PR #507, opened 2014-11-25
Documenting how to deploy a Docker image

dhrp:
I've added a section to the 'using flynn' section about how to deploy a docker container. And I've renamed the current section that was called release to redeploy, because I felt it was a more accurate title, and doesn't confuse with the 'release' cli argument.

Comments welcome. -- I feel looking up the image id using these two curl requests is super clunky, but I wanted to know if it's possible. -- and it is. First downloading the whole thing to be able to do 'docker images --no-trunc' to see the id is also kinda' silly.. 

Ultimately I think this lookup of the id is something that can probably be added in code, but I have not really looked at that yet.

### DIFF
--- a/website/source/docs/using-flynn.html.md
+++ b/website/source/docs/using-flynn.html.md
@@ -285,22 +285,27 @@ $ flynn run bash
 ## Release existing Docker images
 
 Releasing pre-built Docker images is possible by passing the image URL and image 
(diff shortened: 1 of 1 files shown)


lmars:
@dhrp I am going to close this in favour of improving the workflow: #1085

Issue #536, opened 2014-12-01
Deploy Strategy


lmars:
We need to support multiple deploy strategies when creating a new release, with the following being two examples:

* Start and wait for all new release jobs, then kill all old release jobs
* Start and wait for new release jobs one at a time, killing an appropriate old release job each time

We would like to keep the responsibility of both the scheduler and the controller broadly the same, so we plan to add a deployer application.

The deployer application needs the following features:

* HTTP API - POST deploy job (application / release to deploy + strategy)
* Drop deploy jobs in a queue (either etcd / postgres backed)
* Workers work the queue, run jobs based on the strategy and wait for events

The strategy in the deploy job should be generic enough to support custom deploy strategies (e.g. running DB migrations, warming caches, uploading CDN assets etc.)

PR #547, opened 2014-12-03
Add queue package

lmars:
This is based on the [Ruby QueueClassic library](https://github.com/QueueClassic/queue_classic).

Opening now for early feedback.

Prerequisite for #536


titanous:
It's worth looking at https://github.com/bgentry/que-go which claims to use faster queries and appears to solve the locking issues by using a different pg driver which makes more guarantees about connection usage.

lmars:
Closing this as it looks like we are going to use [que-go](https://github.com/bgentry/que-go).

PR #631, opened 2014-12-20
Volume management WIP

heavenlyhash:
First pass on volume management API with an initial ZFS implementation.

Note: this is not at all final and not targeting master.  Some interfaces are not yet complete, and some are in fact just documentation indicating the direction they're expected to go.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -258,6 +258,11 @@
 			"Rev": "7b617783357fec5b5a6d4ecac82e747e4a14d1b5"
 		},
 		{
+			"ImportPath": "github.com/mistifyio/go-zfs",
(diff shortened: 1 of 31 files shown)


heavenlyhash:
@titanous : tentatively ready for another round of review.

Not all features we desire from the final system are present, but the ones that are -- basic volume creation though the API and use of that within job config is covered -- should be working.

heavenlyhash:
Unit tests for zfs stuff requires root.  Travis won't let us do that (nor is it convenient on the command line in dev mode; some of our other unit tests will fail as root (postgres ones, specifically)).  What should we do about that?  We could set up skip flags, but I'd be fairly afraid I wouldn't ever remember to run them again.

heavenlyhash:
tab-enter on github should really *not* close the pr -.-

Issue #170, opened 2014-08-28
Replace rpcplus with HTTP


titanous:
We currently use a custom RPC protocol for a variety of low-level APIs. In the beginning, when the APIs were changing often, this made a lot of sense, but the APIs have not changed very much recently and there isn't a ton of surface area.

For a variety of reasons (introspection, interoperability, etc.) I think it's worth considering switching these APIs from rpcplus to JSON over HTTP.

The main point against HTTP is that we can't easily multiplex multiple streams and requests on a single TCP connection (discoverd would be hit the hardest by this). This problem is already solved by HTTP/2, though I'm not sure if there are any viable packages available currently. I think it makes sense to use server-sent events for streams, as they become trivially available in browsers as well (we already use them in the controller).

Please comment if you have any thoughts about this.

archSeer:
@titanous for multiplexing streams https://github.com/inconshreveable/muxado

archSeer:
Closing as this was resolved with #663

archSeer:
Actually, this is technically not resolved until the discoverd rewrite lands and we completely drop rpcplus from the repository.

PR #663, opened 2014-12-26
Convert host RPC to HTTP API

archSeer:
Opening this up so we can track progress and related discussions here. Prep work should be almost done, now we just need to get this passing on the CI.

Refs #170.


heavenlyhash:
I think this is mostly ready to start review :D  Known caveats follow:

Travis isn't really running our unit tests until we do the rebase&cleanup work and make sure DCO is applied across the board -- but they're looking good on my local runs.

Integration test CI is showing a race somewhere, these three builds are from the same commit:
- failed: https://s3.amazonaws.com/flynn-ci-logs/20150108181454-c6714df5-build-00fa787f83521007ff4850ad3e7bb83a100ad632-2015-01-08-18-21-10.txt
- failed: https://s3.amazonaws.com/flynn-ci-logs/20150108184225-7487daae-build-00fa787f83521007ff4850ad3e7bb83a100ad632-2015-01-08-18-46-01.txt
- passed: https://s3.amazonaws.com/flynn-ci-logs/20150108195326-43c81847-build-00fa787f83521007ff4850ad3e7bb83a100ad632-2015-01-08-20-03-31.txt

The error message is `"cluster: no servers found"`... I think it might have been introduced on this branch, but I'm having trouble reproducing it locally.

We also didn't try to fully convert to using `pkg/stream` since a lot of this work predated that; I'm currently thinking we might find it easiest to pursue that later since this PR has already been in flight quite a while.

host/http.go
```diff
@@ -0,0 +1,129 @@
...
+	w.WriteHeader(200)
+	w.(http.Flusher).Flush()
+	for {
+		select {
+		case data := <-ch:
+			if err := enc.Encode(data); err != nil {
+				return err
+			}
+			w.(http.Flusher).Flush()
+		case <-time.NewTimer(10 * time.Second).C:
```
lmars: `time.After(10 * time.Second)`

host/sampi/http.go
```diff
@@ -0,0 +1,211 @@
...
+	rh := httphelper.NewReponseHelper(w)
+
+	data, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		rh.Error(err)
+		return
+	}
+
+	h := &host.Host{}
+	if err = json.Unmarshal(data, &h); err != nil {
```
lmars: Could use `json.NewDecoder(r.Body).Decode(h)`

host/sampi/http.go
```diff
@@ -0,0 +1,211 @@
...
+	}()
+
+	wr := sse.NewSSEWriter(w)
+	enc := json.NewEncoder(wr)
+	w.Header().Set("Content-Type", "text/event-stream; charset=utf-8")
+	w.WriteHeader(200)
+	wr.Flush()
+	for data := range ch {
+		enc.Encode(data)
+		wr.Flush()
```
lmars: Could use a `flushWriter` here

host/sampi/http.go
```diff
@@ -0,0 +1,211 @@
...
+		rh.Error(err)
+		return
+	}
+	rh.JSON(200, res)
+}
+
+func removeJob(c *Cluster, w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	hostID := ps.ByName("host_id")
+	jobIDs := []string{ps.ByName("job_id")}
+	if err := c.RemoveJobs(hostID, jobIDs); err != nil {
```
lmars: `RemoveJobs` could have arguments `hostID, jobIDs ...string` to avoid the need for the single element slice above

host/sampi/http.go
```diff
@@ -0,0 +1,211 @@
...
+		httphelper.NewReponseHelper(w).Error(err)
```
lmars: This would be more readable as `httphelper.Error(w, err)`

host/sampi/http.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+}
+
+func (cluster *Cluster) ServeHTTP(r *httprouter.Router, sh *shutdown.Handler) error {
+	r.GET("/cluster/hosts", clusterMiddleware(cluster, listHosts))
+	r.PUT("/cluster/hosts/:id", clusterMiddleware(cluster, registerHost))
+	r.POST("/cluster/jobs", clusterMiddleware(cluster, addJobs))
+	r.DELETE("/cluster/hosts/:host_id/jobs/:job_id", clusterMiddleware(cluster, removeJob))
+	r.GET("/cluster/events", clusterMiddleware(cluster, streamHostEvents))
+	return nil
```
lmars: Could define `clusterMiddleware` within `ServeHTTP` and close over `cluster` to avoid passing it as an arg

pkg/cluster/client.go
```diff
@@ -27,25 +32,27 @@ var Attempts = attempt.Strategy{
...
 // a discoverd.ServiceSet.
 type ServiceSetFunc func(name string) (discoverd.ServiceSet, error)
 
-// NewClientWithDial uses the provided dial and services to dial the cluster
-// leader and returns a client. If dial is nil, the default network dialer is
-// used. If services is nil, the default discoverd client is used.
-func NewClientWithDial(dial rpcplus.DialFunc, services ServiceSetFunc) (*Client, error) {
+// NewClientWithServices uses the provided dial and services to dial the cluster
+// leader and returns a client. If services is nil, the default discoverd
+// client is used.
```
lmars: `dial` references need removing from this comment

pkg/cluster/client.go
```diff
@@ -180,86 +151,141 @@ func (c *Client) LeaderID() string {
...
+
 	services := c.service.Select(map[string]string{"id": id})
 	if len(services) == 0 {
 		return nil, ErrNoServers
 	}
-	addr := services[0].Addr
-	rc, err := rpcplus.DialHTTPPath("tcp", addr, rpcplus.DefaultRPCPath, c.dial)
-	return NewHostClient(addr, rc, c.dial), err
+	addr := "http://" + services[0].Addr
+	return NewHostClient(addr, nil), nil
```
lmars: Does this method do any dialling anymore? Should it be renamed to say `HostClient`?

pkg/cluster/client.go
```diff
@@ -180,86 +151,141 @@ func (c *Client) LeaderID() string {
...
+}
+
+// A Stream allows control over a stream sent to a channel.
+type Stream interface {
+	// Close signals the sender to stop sending and then closes the channel.
 	Close() error
+
+	// Err reads the error (if any) that occurred while receiving the
+	// stream. It must only be called after the channel has been closed.
+	Err() error
```
lmars: This can be removed and references to it changed to `stream.Stream`

pkg/cluster/host.go
```diff
@@ -32,53 +35,91 @@ type Host interface {
...
+			close(ch)
+			stream.Close()
+		}()
+
+		r := bufio.NewReader(stream.body)
+		dec := sse.NewDecoder(r)
+		for {
+			event := &host.Event{}
+			if err := dec.Decode(event); err != nil {
+				if err != io.EOF {
```
lmars: Should this same check happen in the other client stream methods? And saying that, could we make something more generic to share between them? They look awfully similar.

pkg/httpclient/json.go
```diff
@@ -12,6 +13,8 @@ import (
 	"github.com/flynn/flynn/pkg/rpcplus"
 )
 
+type DialFunc func(network, addr string) (net.Conn, error)
```
lmars: Is this used anywhere?

pkg/httphelper/httphelper.go
```diff
@@ -0,0 +1,57 @@
...
+const (
+	ContentType = "Content-Type"
+	ContentJSON = "application/json"
+)
+
+type ResponseHelper interface {
+	Error(error)
+	JSON(int, interface{})
+	WriteHeader(int)
+}
```
lmars: We have this same interface [in the controller](https://github.com/flynn/flynn/blob/master/controller/controller.go#L88-L113), can it be shared?

pkg/sse/sse.go
```diff
@@ -26,17 +27,25 @@ type Writer struct {
...
 	defer w.Unlock()
-
-	if _, err := w.Writer.Write([]byte("data: ")); err != nil {
-		return 0, err
-	}
-	if _, err := w.Writer.Write(p); err != nil {
-		return 0, err
+	for _, line := range bytes.Split(p, []byte("\n")) {
+		data := fmt.Sprintf("data: %s\n", line)
+		if _, err := w.Writer.Write([]byte(data)); err != nil {
```
lmars: `fmt.Fprintf(w.Writer, "data: %s\n", line)`

pkg/cluster/host.go
heavenlyhash: I think if we do it, it should be a norm, yeah.  I wanned to confirm though -- When mapping IO to chan, does it seem reasonable that we habitually map EOF to chan close?

I agree this code block seems repetative.  I haven't come with up with a terse way to make it sharable though... Here's a shot: https://github.com/flynn/flynn/commit/b4cf658842574d480916de688df3e2f2b23d06aa , but seems like it ends up requiring a bunch of other interfaces and wrappers to coax channel types into working out.  Thoughts?

pkg/sse/sse.go
heavenlyhash: Should we worry about the return value for bytes written?  (not that the current code does either)

host/sampi/http.go
IceDragon200: We could just be rid of ReponseHelper all together, since it only stores a http.ReponseWriter and no real internal state

pkg/cluster/client.go
IceDragon200: The second argument is a `*http.Client`, but it looks like no one ever sets it anyway, I guess we could remove the second argument, since it defaults to `http.DefaultClient`, wdyt?

pkg/httphelper/httphelper.go
IceDragon200: The pkg is based on that, I copied the code and removed the martini render.Render requirement and made generic, I could try merging the 2 together to make one generic package.

pkg/cluster/host.go
heavenlyhash: Pushed refactor to use reflection based approach as discussed on that tangent :D

Linecount went down nicely and I'm glad the error handling and disconnect handling is now uniform.

controller/jobs.go
```diff
@@ -195,8 +241,8 @@ func jobLog(req *http.Request, app *ct.App, params martini.Params, hc cluster.Ho
 		defer attachClient.Close()
 	}
 
-	sse := strings.Contains(req.Header.Get("Accept"), "text/event-stream")
-	if sse {
+	use_sse := strings.Contains(req.Header.Get("Accept"), "text/event-stream")
```
titanous: Camel case please.

pkg/httpclient/stream.go
```diff
@@ -0,0 +1,63 @@
...
+				{
+					Dir:  reflect.SelectRecv,
+					Chan: stopChanValue,
+				},
+				{
+					Dir:  reflect.SelectSend,
+					Chan: chanValue,
+					Send: reflect.ValueOf(msg),
+				},
+			})
```
lmars: I have never laid eyes on such wizardry :+1:

host/http.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return nil, err
+	}
+	sh.BeforeExit(func() { l.Close() })
+
+	r := httprouter.New()
+	r.POST("/attach", attach.ServeHTTP)
+	r.GET("/host/jobs", hostMiddleware(host, listJobs))
+	r.GET("/host/jobs/:id", hostMiddleware(host, getJob))
+	r.DELETE("/host/jobs/:id", hostMiddleware(host, stopJob))
```
titanous: Instead of `hostMiddleware`, make a struct:

```go
type httpAPI struct {
  Host *Host
}
```

And then just define methods on it and pass method values to the router.

host/sampi/http.go
```diff
@@ -0,0 +1,202 @@
...
+		return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+			handle(cluster, w, r, ps)
+		}
+	}
+
+	r.GET("/cluster/hosts", clusterMiddleware(listHosts))
+	r.PUT("/cluster/hosts/:id", clusterMiddleware(registerHost))
+	r.POST("/cluster/jobs", clusterMiddleware(addJobs))
+	r.DELETE("/cluster/hosts/:host_id/jobs/:job_id", clusterMiddleware(removeJob))
+	r.GET("/cluster/events", clusterMiddleware(streamHostEvents))
```
titanous: Ditto on using a struct instead of a closure.

pkg/httphelper/flushwriter.go
```diff
@@ -0,0 +1,22 @@
+package httphelper
+
+import (
+	"io"
+	"net/http"
+)
+
+type FlushWriter struct {
+	io.Writer
+	OK bool
```
titanous: We should probably name this better.

pkg/httphelper/httphelper.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"encoding/json"
+	"log"
+	"net/http"
+)
+
+// shamelessly taken from martini-contrib/render/render.go
+const (
+	ContentType = "Content-Type"
+	ContentJSON = "application/json"
```
titanous: These constants are overkill.

pkg/sse/sse.go
titanous: No, the return count is the number of bytes read from `p`.

pkg/httpclient/stream.go
heavenlyhash: :) The tests for the feature in the stdlib are really good reading too!  https://github.com/golang/go/blob/083e04c3d963832116deb0d11e1fab5d7d57132a/src/reflect/all_test.go#L1111

pkg/sse/sse.go
```diff
@@ -26,17 +27,25 @@ type Writer struct {
...
+			return 0, err
+		}
 	}
-	_, err := w.Writer.Write([]byte("\n\n"))
+	// add a terminating newline
+	_, err := w.Writer.Write([]byte("\n"))
```
titanous: Is this "terminating newline" in the SSE spec

controller/jobs.go
IceDragon200: Should that be: `useSSE` or `useSse` ?

titanous: The former

pkg/sse/sse.go
archSeer: Yes, since the current message packet might be split into multiple lines, separation between events is done using a double newline. https://developer.mozilla.org/en-US/docs/Server-sent_events/Using_server-sent_events

titanous: Ah, got it, got confused about the context.

controller/jobs.go
```diff
@@ -207,12 +253,12 @@ func jobLog(req *http.Request, app *ct.App, params martini.Params, hc cluster.Ho
 		wf.Flush()
 	}
 
-	fw := flushWriter{w, tail}
-	if sse {
-		ssew := NewSSELogWriter(w)
-		exit, err := attachClient.Receive(flushWriter{ssew.Stream("stdout"), tail}, flushWriter{ssew.Stream("stderr"), tail})
+	fw := httphelper.FlushWriter{w, tail}
```
titanous: Use the field names, otherwise vet will complain.

host/http.go
```diff
@@ -0,0 +1,134 @@
...
+		if err := h.host.streamEvents(id, w); err != nil {
+			httphelper.Error(w, err)
+		}
+		return
+	}
+	job := h.host.state.GetJob(id)
+	httphelper.JSON(w, 200, job)
+}
+
+func (h *httpAPI) stopJob(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
titanous: I like using capitalized ("exported") methods to distinguish from any helper methods.

host/http.go
```diff
@@ -0,0 +1,134 @@
...
+		select {
+		case data := <-ch:
+			if err := enc.Encode(data); err != nil {
+				return err
+			}
+			w.(http.Flusher).Flush()
+		case <-time.After(10 * time.Second):
+			// if the job still doesn't exist after a reasonable timeout, then
+			// break out of this stream (closing the connection will be handled outside)
+			if id != "all" && h.state.GetJob(id) == nil {
```
titanous: Is this timeout necessary? We didn't have it before and I don't think we've fully analyzed the impact. For example I think downloading container images happens before the job exists currently.

host/sampi/http.go
```diff
@@ -0,0 +1,201 @@
...
+	c.cluster.state.Commit()
+	go c.cluster.state.sendEvent(h.ID, "add")
+
+	// "defer" cleanups in a goroutine that waits until the http stream is closed.
+	go func() {
+		<-w.(http.CloseNotifier).CloseNotify()
+		c.cluster.state.Begin()
+		c.cluster.state.RemoveHost(h.ID)
+		c.cluster.state.Commit()
+		go c.cluster.state.sendEvent(h.ID, "remove")
```
titanous: Any reason to run this in a goroutine? We're already detached from the handler goroutine.

pkg/cluster/attach.go
```diff
@@ -33,7 +35,11 @@ func (c *hostClient) Attach(req *host.AttachReq, wait bool) (AttachClient, error
 	if err != nil {
 		return nil, err
 	}
-	conn, err := c.dial("tcp", c.addr)
+	uri, err := url.Parse(c.c.URL)
+	if err != nil {
+		return nil, err
+	}
+	conn, err := net.Dial("tcp", uri.Host)
```
titanous: Is `c.dial` still around? If so, it should be used.

titanous: nvm, looks like it was deleted.

host/sampi/http.go
IceDragon200: Looks like it can be safely changed

host/http.go
heavenlyhash: Hm.  Yeah, we could retract this change.  I haven't seen it do any harm yet, and I'm not super pleased with the alternative behavior (hang), but yeah, I haven't checked for that kind of impacts.

titanous:
I've squashed this down to a more manageable size.

titanous:
LGTM

Issue #34, opened 2014-08-07
Standardize logging


titanous:
Components are using some combination of `fmt`, `log`, `grohl`, and martini request logging. We want to standardize on using [`log15`](https://github.com/inconshreveable/log15) everywhere.

lmars:
We have begun to use `log15` which has been good so far, but I think we need some consensus on the format of the messages and what context is useful to add.

My current thoughts are:

* Components should have a global `logger` with an `app` context:

    `var logger = log15.New("app", "deployer")`

* Loggers should include a `fn` context to help identify where a log message came from:

    `log := logger.New("fn", "HandleJob")`

* Info messages should describe what action is being taken, or what action was taken:

    `log.Info("creating controller client")`
    `log.Info("performing deployment")`
    `log.Info("deployment complete"`)

* Error messages should generally take the form `error ACTION` and include an `err` context:

    `log.Error("error creating controller client", "err", err)`

There is also a convention to include an `at` context which I have not been doing so far, so we should decide whether that provides added benefit when reading logs / finding where logs came from.

titanous:
This is all good, I agree that dropping `at` makes sense.

PR #666, opened 2014-12-27
Deployer

### LINKED ISSUE #63: Zero-downtime deployment
Currently the old formation is terminated before the new one starts, we should have another approach by default that starts new instances before terminating the old ones.

### LINKED ISSUE #602: deployer: Implement Strategies
The deployer app needs to support multiple strategies for performing deployments.

@archSeer started this work but it seems to have been lost from the `deploy-strategy` branch.

Features to implement:

- [x] A `Strategy` interface (e.g. `type Strategy interface { Perform(*Deployment) error }`)
- [x] A `Strategy` object which performs a "one-by-one" deployment
- [x] A `Strategy` object which performs an "all-at-once" deployment
- [x] A lookup function to determine which strategy to use for a given `Deployment` (probably make [Deployment.Strategy](https://github.com/flynn/flynn/blob/bacb1541c97b6750d4ecfa6d920ab3bcc14e0d61/deployer/types/types.go#L10) an enum)
- [x] Update the [handleJob function](https://github.com/flynn/flynn/blob/bacb1541c97b6750d4ecfa6d920ab3bcc14e0d61/deployer/server.go#L70-L79) to call `Perform` on the relevant `Strategy`
- [x] Emit events throughout the deployment (e.g. start / end of deployment, start / end of formation change, job status changes)
- [x] Add integration test for each strategy which does deployment and checks events (partially implemented [here](https://github.com/flynn/flynn/blob/bacb1541c97b6750d4ecfa6d920ab3bcc14e0d61/test/test_deployer.go#L22-L75))
- [x] Restore original formation state if a deployment fails
- [x] Handle crashed formations / processes (depends on a fix for #610)
- [x] Add a releaseID constraint so that only one deployment operating on a single release might be working at once
- [ ] Make sure strategies are idempotent

Part of #536.

### LINKED ISSUE #603: deployer: Reduce risk of workers performing the same deployments
The queue package provides no guarantees that a job will not be locked by two workers at the same time, but two workers performing the same deployment at the same time is problematic.

A possibly approach is to add a `locked_until` column to `deployments` which is bumped every few seconds by the worker performing the deployment, and any other worker happening to lock the same job checks that `locked_until` is in the past before performing the deployment, returning an appropriate `queue.RetryAfterError` if it's in the future.

Part of #536.

archSeer:
Adds a deployer component, which uses the bgentry/que-go to schedule deployments using different strategies (currently one-by-one or all-at-once).

Resolves #602, resolves #603, resolves #63.


archSeer:
@flynn/devs ready for a review!

We track deployment state, as it was used previously with rollbacks, but since the change to que-go, it is unused. Should we make the strategies idempotent and get rid of the state tracking?

archSeer:
Hmm, the integration tests seem to be failing again after the rebase, I'll resolve it in the morning.

controller/scheduler/main.go
```diff
@@ -523,6 +524,13 @@ func (m jobTypeMap) Add(typ, host, id string) *Job {
 
 func (m jobTypeMap) Remove(job *Job) {
 	if jobs, ok := m[job.Type]; ok {
+		j := jobs[jobKey{job.HostID, job.ID}]
+		// cancel job restarts
+		j.timerMtx.Lock()
+		if j.timer != nil {
+			j.timer.Stop()
+		}
+		j.timerMtx.Unlock()
```
lmars: It may be worth breaking this change out into a different pull request

deployer/client/client.go
```diff
@@ -0,0 +1,81 @@
...
+	body   io.ReadCloser
+}
+
+// Close closes the underlying stream.
+func (s *DeploymentEventStream) Close() {
+	s.body.Close()
+}
+
+// StreamDeploymentEvents returns a DeploymentEventStream for an app.
+func (c *Client) StreamDeploymentEvents(deploymentID string, lastID int64) (*DeploymentEventStream, error) {
```
lmars: This should use the [Stream interface](https://github.com/flynn/flynn/blob/master/pkg/stream/stream.go).

deployer/schema.go
```diff
@@ -0,0 +1,58 @@
...
+		`CREATE EXTENSION IF NOT EXISTS "uuid-ossp"`,
+
+		`CREATE TYPE deployment_strategy AS ENUM ('all-at-once', 'one-by-one')`,
+		`CREATE TABLE deployments (
+    deployment_id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+    app_id uuid NOT NULL,
+    old_release_id uuid NOT NULL,
+    new_release_id uuid NOT NULL,
+    strategy deployment_strategy NOT NULL,
+    status int8 NOT NULL DEFAULT 0,
```
lmars: We should drop the status

deployer/schema.go
```diff
@@ -0,0 +1,58 @@
...
+    created_at timestamptz NOT NULL DEFAULT now())`,
+
+		`CREATE SEQUENCE deployment_event_ids`,
+		`CREATE TABLE deployment_events (
+    event_id bigint PRIMARY KEY DEFAULT nextval('deployment_event_ids'),
+    deployment_id uuid NOT NULL REFERENCES deployments (deployment_id),
+    release_id uuid NOT NULL,
```
lmars: We set this to `""` to indicate the end of a deployment, so this needs to be `text` rather than `uuid` (`""` is not a valid uuid)

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/bgentry/que-go"
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-sql"
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/pq"
```
lmars: Do we still need `pq`?

deployer/schema.go
titanous: Wouldn't `NULL` be better? I'm not a huge fan of signalling state via sentinel fields that are used for other things, is there a better way to do it? Columns are cheap.

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+
+type deployID struct {
+	ID string
+}
+
+func main() {
+	var err error
+	client, err = controller.NewClient("", os.Getenv("CONTROLLER_AUTH_KEY"))
+	if err != nil {
+		log.Fatalln("Unable to connect to controller:", err)
```
lmars: `controller.NewClient` doesn't actually connect to the controller, it only returns an error if it can't parse the URL or fails to connect to discoverd (see [here](https://github.com/flynn/flynn/blob/master/controller/client/client.go#L53-L80)), so this error should probably be `Unable to create controller client`

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+		log.Fatal(err)
+	}
+	defer pgxpool.Close()
+
+	q = que.NewClient(pgxpool)
+	wm := que.WorkMap{
+		"Deployment": handleJob,
+	}
+
+	// TODO start new worker on error
```
lmars: We should just exit if the worker stops, making it the scheduler's job to restart the process, possibly with back-off

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+	workers := que.NewWorkerPool(q, wm, 10)
+	go workers.Start()
+	defer workers.Shutdown()
```
lmars: I don't think this deferred method will ever get called

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+
+	router := httprouter.New()
+	router.POST("/deployments", createDeployment)
+	router.GET("/deployments/:deployment_id/events", streamDeploymentEvents)
+
+	log.Println("Listening for HTTP requests on", addr)
```
lmars: We have plans to standardise logging using [log15](https://github.com/inconshreveable/log15) (#34), so we should probably kick that off by using it in the deployer

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+	log.Fatal(http.ListenAndServe(addr, router))
+}
+
+func handleJob(job *que.Job) (e error) {
```
lmars: We should add verbose logging to this method, I suspect it will be involved in a lot of deployment debugging sessions

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+	var args deployID
```
lmars: Can this not just be a string?

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+		return err
+	}
+	// for recovery purposes, fetch old formation
+	f, err := client.GetFormation(deployment.AppID, deployment.OldReleaseID)
+	if err != nil {
+		// TODO: log error
+		return err
+	}
+	setDeploymentStatus(id, deployer.StatusRunning)
+	strategy, err := strategy.GetStrategy(deployment.Strategy, client)
```
lmars: This is a bit of a mouthful, how about `strategy.Get`, or `strategy.Lookup`?

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+	if err != nil {
+		// TODO: log error
+		return err
+	}
+	events := make(chan deployer.DeploymentEvent)
+	defer close(events)
+	go func() {
+		for ev := range events {
+			ev.DeploymentID = deployment.ID
+			sendDeploymentEvent(ev)
```
lmars: We should not ignore the error here (at the very least log it)

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+		}
+	}()
+	defer func() {
+		// rollback failed deploy
+		if e != nil {
```
lmars: I think this should be `deployErr`, or maybe just `err`

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+			events <- deployer.DeploymentEvent{ReleaseID: "fail"}
```
lmars: We should add another column to `deployment_events` rather than using `release_id` like this.

How about a `status` field with values `running`, `complete` or `failed`?

deployer/server.go
```diff
@@ -0,0 +1,406 @@
...
+			err = ErrNotFound
+		}
+		return nil, err
+	}
+	event.DeploymentID = postgres.CleanUUID(event.DeploymentID)
+	event.ReleaseID = postgres.CleanUUID(event.ReleaseID)
+	return event, nil
+}
+
+// TODO: share with controller formatUUID
```
lmars: Since `CleanUUID` is in `pkg/postgres`, we should probably move this there too

deployer/strategies/main.go
```diff
@@ -0,0 +1,89 @@
...
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/deployer/types"
+	"time"
+)
+
+type Performer interface {
+	Perform(*deployer.Deployment, chan<- deployer.DeploymentEvent) error
+}
+
+func GetStrategy(strategy string, client *controller.Client) (Performer, error) {
```
lmars: Since the strategies don't really need any state, this could probably just return a function, e.g.:

```go
package strategy

type PerformFunc func(client *controller.Client, *deployer.Deployment, chan<- deployer.DeploymentEvent) error

var performFuncs = map[string]PerformFunc{
	"all-at-once": allAtOnce,
	"one-by-one": oneByOne,
}

func Get(strategy string) (PerformFunc, error) {
	if f, ok := performFuncs[strategy]; ok {
		return f, nil
	}
	return nil, fmt.Errorf("Unknown strategy '%s'!", strategy)
}

func allAtOnce(client *controller.Client, *deployer.Deployment, chan<- deployer.DeploymentEvent) error {}
func oneByOne(client *controller.Client, *deployer.Deployment, chan<- deployer.DeploymentEvent) error {}
```

deployer/types/types.go
```diff
@@ -0,0 +1,36 @@
...
+	StatusWaiting Status = iota
+	StatusRunning
+	StatusDone
+	StatusCrashed
+	StatusFailed
+)
+
+type Step struct {
+	Cmd []string `json:"cmd,omitempty"`
+}
```
lmars: This should be removed

deployer/server.go
archSeer: No, because the field type is `json` in que's/que-go's schema and it expects a valid JSON object.

lmars: A single string is valid JSON though right?

deployer/server.go
archSeer: `que-go` does not actually return any errors on `WorkerPool.Start()`, so I assume it's foolproof?

deployer/server.go
archSeer: Does changing the name really change anything here?

deployer/server.go
archSeer: Yeah, I'm still using it for the deployment SQL code, as the listener interface from `pgx` is [quite ugly](https://godoc.org/github.com/jackc/pgx#hdr-Listen_and_Notify)

deployer/server.go
lmars: Yeh I think `pgx` will try to make new connections if postgres is unavailable, can probably drop this TODO then

deployer/server.go
archSeer: I've checked, and it indeed needs to be a serialised struct:

```
2015-01-09 03:18:54 UTC ERROR:  invalid input syntax for type json
2015-01-09 03:18:54 UTC DETAIL:  Token "f1e9f1d7543b411f914fc34772e940b8" is invalid.
```

archSeer:
@lmars give it another look, if LGTM, I'll merge and then work on logging in a subsequent PR.

deployer/client/client.go
```diff
@@ -0,0 +1,91 @@
...
+	return &Client{Client: c}, nil
+}
+
+func (c *Client) CreateDeployment(deployment *deployer.Deployment) error {
+	return c.Post("/deployments", deployment, deployment)
+}
+
+// DeploymentEventStream is a wrapper around an Events channel, allowing us to close
+// the stream.
+type DeploymentEventStream struct {
```
lmars: This is unused

deployer/client/client.go
```diff
@@ -0,0 +1,91 @@
...
+	Events chan *deployer.DeploymentEvent
+	body   io.ReadCloser
+}
+
+// Close closes the underlying stream.
+func (s *DeploymentEventStream) Close() {
+	s.body.Close()
+}
+
+// StreamDeploymentEvents returns a DeploymentEventStream for an app.
```
lmars: s/DeploymentEventStream/Stream/

deployer/server.go
```diff
@@ -0,0 +1,401 @@
...
+		// TODO: log error
+		return err
+	}
+	events := make(chan deployer.DeploymentEvent)
+	defer close(events)
+	go func() {
+		for ev := range events {
+			ev.DeploymentID = deployment.ID
+			if err := sendDeploymentEvent(ev); err != nil {
+				log.Fatal(err)
```
lmars: This should just log the error, not stop the process.

deployer/server.go
lmars: Well, I don't think its a real issue, but the code would look something like:

```go
var deployErr error
defer func() {
	if deployErr != nil {
		if err := (...); err != nil {
			return err
		}
	}
}
```

which gets rid of the `e = err` and `e = nil`

deployer/server.go
```diff
@@ -0,0 +1,401 @@
...
+	if err := client.SetAppRelease(deployment.AppID, deployment.NewReleaseID); err != nil {
+		return err
+	}
+	// signal success
+	if err := sendDeploymentEvent(deployer.DeploymentEvent{
+		DeploymentID: deployment.ID,
+		ReleaseID:    deployment.NewReleaseID,
+		Status:       "complete",
+	}); err != nil {
+		log.Fatal(err)
```
lmars: This should just log the error, not stop the process.

lmars:
@archSeer LGTM pending minor comments

controller/scheduler/main.go
```diff
@@ -523,6 +524,13 @@ func (m jobTypeMap) Add(typ, host, id string) *Job {
 
 func (m jobTypeMap) Remove(job *Job) {
 	if jobs, ok := m[job.Type]; ok {
+		j := jobs[jobKey{job.HostID, job.ID}]
+		// cancel job restarts
+		j.timerMtx.Lock()
+		if j.timer != nil {
+			j.timer.Stop()
+		}
```
titanous: This should probably nil the timer too?

controller/scheduler/main.go
```diff
@@ -746,10 +756,19 @@ func (f *Formation) jobType(job *host.Job) string {
...
-func (s sortJobs) Len() int           { return len(s) }
-func (s sortJobs) Less(i, j int) bool { return s[i].startedAt.Sub(s[j].startedAt) > 0 }
-func (s sortJobs) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
-func (s sortJobs) Sort()              { sort.Sort(s) }
+func (s sortJobs) Len() int { return len(s) }
+func (s sortJobs) Less(i, j int) bool {
+	switch {
+	case s[i].timer != nil && s[j].timer == nil:
+		return true
+	case s[i].timer == nil && s[j].timer != nil:
```
titanous: If timer is protected above by a mutex, does it need that protection here?

deployer/server.go
```diff
@@ -0,0 +1,401 @@
...
+	"github.com/flynn/flynn/deployer/types"
+	"github.com/flynn/flynn/discoverd/client"
+	"github.com/flynn/flynn/pkg/postgres"
+	"github.com/flynn/flynn/pkg/random"
+	"github.com/flynn/flynn/pkg/shutdown"
+)
+
+var q *que.Client
+var db *postgres.DB
+var client *controller.Client
```
titanous: I'd rather see these stored in a struct, than global. You can pass `handleJob` to pgx as `worker.HandleJob` if worker is a pointer to a struct with a `HandleJob` method.

deployer/server.go
```diff
@@ -0,0 +1,401 @@
...
+	db, err = postgres.Open("", "")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if err := migrateDB(db.DB); err != nil {
+		log.Fatal(err)
+	}
+
+	pgxcfg, err := pgx.ParseURI(fmt.Sprintf("http://%s:%s@%s/%s", os.Getenv("PGUSER"), os.Getenv("PGPASSWORD"), db.Addr(), os.Getenv("PGDATABASE")))
```
titanous: It's less work just to populate a `pgx.ConnConfig` directly.

deployer/server.go
```diff
@@ -0,0 +1,401 @@
...
+		case pq.ListenerEventConnected:
+			close(connected)
+		case pq.ListenerEventDisconnected:
+			close(done)
+		case pq.ListenerEventConnectionAttemptFailed:
+			err = listenErr
+			close(done)
+		}
+	}
+	listener := pq.NewListener(db.DSN(), 10*time.Second, time.Minute, listenEvent)
```
titanous: Why this over the simpler pgx listener implementation?

deployer/server.go
```diff
@@ -0,0 +1,401 @@
...
+			if err := client.PutFormation(f); err != nil {
+				e = err
+				return
+			}
+			if err := client.DeleteFormation(deployment.AppID, deployment.NewReleaseID); err != nil {
+				e = err
+				return
+			}
+			if err := client.SetAppRelease(deployment.AppID, deployment.NewReleaseID); err != nil {
+				e = err
```
titanous: All of these can be simplified to:

```go
if e = client...
```

deployer/server.go
archSeer: Because this was implemented before the switch to pgx, I plan to merge this PR into the temp branch, then open up additional PRs that add logging and change to use pgx exclusively.

titanous: Okay, sounds good.

controller/scheduler/main.go
archSeer: Hmm, maybe? Lock both mutexes?

deployer/server.go
archSeer: Same length sort of, since I need to parse `db.Addr()` into a `host` and a `port`, but I'll change it since it's clearer.

archSeer: Actually no, I think the original code is much clearer than this:

```
dbaddr := strings.Split(db.Addr(), ":")
dbport, err := strconv.Atoi(dbaddr[1])
if err != nil {
    log.Fatal(err)
}
pgxpool, err := pgx.NewConnPool(pgx.ConnPoolConfig{
    ConnConfig: pgx.ConnConfig{
        Host:     dbaddr[0],
        Port:     uint16(dbport),
        Database: os.Getenv("PGDATABASE"),
        User:     os.Getenv("PGUSER"),
        Password: os.Getenv("PGPASSWORD"),
    },
    AfterConnect: que.PrepareStatements,
})
```

controller/scheduler/main.go
titanous: I don't have enough context currently, it requires  a read-through to figure out if there is the possibility of concurrent writes with any reads.

archSeer:
@titanous @lmars I'm thinking of closing this PR and reopening #665, since I'll be targeting master directly now, and it now includes all of the commits necessary (we opened a temporary branch that included `pkg/queue` so it'd be excluded from the PR diff)

titanous:
@archSeer Just close this one an open a brand new PR.

archSeer:
Closing this and opening a PR to master.

PR #684, opened 2014-12-31
all: Use Stream interface for controller client StreamJobEvents

titanous:
This also fixes some cases that did not properly handle the channel being closed, which resulted in panics in some error paths.

### DIFF
--- a/controller/client/client.go
+++ b/controller/client/client.go
@@ -118,8 +118,7 @@ func (c *Client) StreamFormations(since *time.Time, output chan<- *ct.ExpandedFo
 			res.Body.Close()
 		}()
 
-		r := bufio.NewReader(res.Body)
-		dec := sse.NewDecoder(r)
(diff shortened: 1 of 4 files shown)


controller/client/client.go
```diff
@@ -325,16 +312,26 @@ func (c *Client) StreamJobEvents(appID string, lastID int64) (*JobEventStream, e
...
 	go func() {
-		defer close(stream.Events)
-		dec := sse.NewDecoder(bufio.NewReader(stream.body))
+		defer func() {
+			close(output)
+			res.Body.Close()
+		}()
+
+		r := bufio.NewReader(res.Body)
+		dec := sse.NewDecoder(r)
```
lmars: Any reason for the extra `r` variable?

titanous: Oops, meant to get rid of that.

lmars:
LGTM

PR #685, opened 2015-01-01
Replace blobstore TestLargeAmountOfData with config lookup 

titanous:
`TestLargeAmountOfData` takes upwards of three minutes, so this patch replaces it with a test that reads the ssl renegotiation configuration from postgres by running psql in a container. This takes a second or less to perform.


archSeer:
LGTM

Issue #686, opened 2015-01-01
Scheduler doesn't reliably deliver job events to controller


titanous:
During the last phase of `SchedulerSuite.TestControllerRestart`, scaling the controller down, the update that the controller instance has stopped is sometimes sent to the instance that was just stopped. This results in a timeout waiting for job events and a test failure.

The log message from the controller looks like this:

```text
now=2015-01-01T01:02:47+0000 host.id=94262c37 app=controller-scheduler fn=watchHost at=error job.id=8cd17f80b7894faaa9ba4ac899480a2f event=stop err="Put http://flynn-controller/apps/d58ab68aed2a4fc88bf50de0136b5c24/jobs/94262c37-8cd17f80b7894faaa9ba4ac899480a2f: dial tcp 172.16.31.2:5000: connection refused"
```

While the fix for this specific instance is pretty easy (just retry a few times), it raises a more important question: how should we handle these events?

Currently it is assumed that the controller has an eventually consistent view on the events, but this is not correct. If an event delivery fails from the scheduler to the controller or if the scheduler fails before delivering an event, it will never end up in the `jobs_cache` db table. If this table is not consistent and can have missing events, then the view presented to consumers of the jobs API is incorrect. I don't have any immediate concrete solutions, but I wanted to get this down before I forgot about it.

lmars:
What if the scheduler inserted directly into the `jobs_cache` table?

titanous:
@lmars Hmm, does that solve job changes not being persisted if a scheduler dies? I think we need some kind of anti-entropy?

heavenlyhash:
If we stare at this long enough, does it end up in [Two Generals](https://en.wikipedia.org/wiki/Two_Generals%27_Problem) territory?

lmars:
@titanous so yes there is potential to drop events, and in fact the new scheduler leader will send a whole load of `up` events [as it syncs with the host](https://github.com/flynn/flynn/blob/master/controller/scheduler/main.go#L166-L172)...

titanous:
This should be resolved by the new scheduler (#1914).

PR #687, opened 2015-01-01
discoverd/client/dialer: Try to dial each service before giving up

titanous:
Instead of trying to dial a single random instance once, try to dial each instance in the list in random order before giving up.

This should fix one or two intermittent test failures.

Refs #686


lmars:
LGTM

Issue #145, opened 2014-08-20
Service discovery redux


archSeer:
We've been running into more and more usecases where our current way of going about service discovery doesn't quite fit the model. We've been[ talking with @titanous on IRC about going back to the whiteboard and rethinking the entire thing](https://botbot.me/freenode/flynn/2014-08-20/?msg=20048425&page=1), this issue should serve as a place to discuss this.

I ran into problems during #65, where the idea is to pause the service itself, in order to do updates or maintenance behind the scenes. In our current model, services are hidden away in the implementation, so there are only two options, either pausing the route, or pausing the application itself. The former does not make sense in the case where we want to wait for the backends to be drained, because we might continue receiving new requests trough other routes pointing to that same backend. The latter will not work either, because we might just want to pause a specific service, instead of bringing down the entire application.

I think that instead of hiding away the services, we should expose them, and provide a way for the user to map ip addresses/ports to a service (and then map those services to a route). That will require the user to run a few more commands via the CLI, but I think we can provide a flag that will automatically create the mappings for the user, in simple cases like Heroku-style web apps.

That said, I'm no expert on this topic, so let's get some discussion going on.

Related discussions:
- https://botbot.me/freenode/flynn/2014-08-20/?msg=20048425&page=1
- https://github.com/flynn-archive/slugrunner/pull/11

/cc @lmars @progrium

PR #690, opened 2015-01-04
discoverd2: Add state implementation

titanous:
This implements the core list of services and instances that will be hooked up to a consensus/communication backend like etcd as well as the HTTP and DNS read APIs.

I'm trying to send review-sized PRs instead of just the final result, so this is an intermediate (working, with 84% test coverage) patch for #145.

### DIFF
--- a/discoverd2/server/state.go
+++ b/discoverd2/server/state.go
@@ -1,0 +1,333 @@
+package server
+
+import (
+	"container/list"
+	"errors"
+	"sync"
+
+	"github.com/flynn/flynn/pkg/stream"
+)
+
+type EventKind uint
+
+const (
+	EventKindUp EventKind = 1 << iota
+	EventKindUpdate
+	EventKindDown
+	EventKindLeader
+)
+
+func (k EventKind) String() string {
+	switch k {
+	case EventKindUp:
+		return "up"
+	case EventKindUpdate:
+		return "update"
+	case EventKindDown:
+		return "down"
+	case EventKindLeader:
+		return "leader"
+	default:
+		return "unknown"
+	}
+}
+
+type Event struct {
+	Service string
+	Kind    EventKind
+	*Instance
+}
+
+func eventKindUpdate(existing bool) EventKind {
+	if existing {
+		return EventKindUpdate
+	}
+	return EventKindUp
+}
+
+// Instance is a single running instance of a service.
+type Instance struct {
+	// ID is unique within the service, and is currently defined as
+	// Hex(SHA256(Proto + "-" + Addr)) but this may change in the future.
+	ID string `json:"id"`
+
+	// Addr is the IP/port address that can be used to communicate with the
+	// service. It must be valid to dial this address.
+	Addr string `json:"addr"`
+
+	// Proto is the protocol used to connect to the service, examples include:
+	// tcp, udp, http, https. It must be lowercase alphanumeric.
+	Proto string `json:"proto"`
+
+	// Meta is arbitrary metadata specified when registering the instance.
+	Meta map[string]string `json:"meta,omitempty"`
+
+	// Leader is true if the instance is the leader of the service. Exactly one
+	// instance per service has this set to true at any point in time.
+	Leader bool `json:"leader,omitempty"`
+
+	// Index is the logical epoch of the initial registration of the instance.
+	// It is guaranteed to be unique, not change as long as the instance does
+	// not expire, and sort with other indexes in the order of instance
+	// creation.
+	Index uint `json:"index"`
+}
+
+func (inst *Instance) Equal(other *Instance) bool {
+	return inst.Addr == other.Addr &&
+		inst.Proto == other.Proto &&
+		inst.Index == other.Index &&
+		mapEqual(inst.Meta, other.Meta)
+}
+
+func mapEqual(x, y map[string]string) bool {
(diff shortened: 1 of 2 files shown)


discoverd2/server/state.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+func eventKindUpdate(existing bool) EventKind {
+	if existing {
+		return EventKindUpdate
+	}
+	return EventKindUp
+}
+
+type Instance struct {
```
heavenlyhash: Instance of what?

discoverd2/server/state.go
```diff
@@ -0,0 +1,314 @@
...
+		subscribers: make(map[string]*list.List),
+	}
+}
+
+type State struct {
+	// service name -> instance address -> instance
+	services map[string]map[string]*Instance
+	// TODO: split this mutex up so that it just protects the services map, and
+	// each service has its own mutex as well?
+	mtx sync.RWMutex
```
heavenlyhash: `atomic.Value` has some example code for a "read mostly" datastruct that might be applicable here: http://golang.org/src/sync/atomic/value_test.go#L169

Of course, I'm sure the mutex will stand up fine for quite a while.

discoverd2/server/state.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+func (inst *Instance) Equal(other *Instance) bool {
+	return inst.Addr == other.Addr &&
+		inst.Proto == other.Proto &&
+		inst.Index == other.Index &&
+		mapEqual(inst.Meta, other.Meta)
+}
+
+func mapEqual(x, y map[string]string) bool {
```
heavenlyhash: Could maybe use http://golang.org/pkg/reflect/#DeepEqual if the performance of the refect paths isnt a problem?  But this implementation is also correct.

discoverd2/server/state.go
```diff
@@ -0,0 +1,314 @@
...
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	data, ok := s.services[service]
+	if !ok {
+		data = make(map[string]*Instance)
+		s.services[service] = data
+	}
+	old, existing := data[inst.ID]
+	data[inst.ID] = inst
```
heavenlyhash: looks like this ID is fairly strongly presumed to be unique?  perhaps a doc section about that on the struct layout?

discoverd2/server/state.go
```diff
@@ -0,0 +1,314 @@
...
+func (s *State) getLocked(service string) []*Instance {
+	data, ok := s.services[service]
+	if !ok {
+		return nil
+	}
+
+	res := make([]*Instance, 0, len(data))
+	for _, inst := range data {
+		res = append(res, inst)
+	}
```
heavenlyhash: Can shortly this to use the `copy` builtin.

discoverd2/server/state.go
titanous: Yeah, I think CoW for the top-level services list and a RWMutex for the instances list will probably be the most efficient approach.

discoverd2/server/state.go
titanous: Yeah, I'll add docs.

discoverd2/server/state.go
titanous: data is a map, so we can't.

discoverd2/server/state.go
titanous: This will likely be moved to the client package, named `discoverd`. `ServiceInstance` felt like a mouthful, but if you think it's better I'll gsub.

discoverd2/server/state.go
titanous: I think I'll leave it, there's no way the performance of reflect can match this implementation, and it gets called a lot.

discoverd2/server/state.go
```diff
@@ -0,0 +1,332 @@
...
+	Proto string `json:"proto"`
+
+	// Meta is arbitrary metadata specified when registering the instance.
+	Meta map[string]string `json:"meta,omitempty"`
+
+	// Leader is true if the instance is the leader of the service. Exactly one
+	// instance per service has this set to true at any point in time.
+	Leader bool `json:"leader,omitempty"`
+
+	// Index is the logical epoch of the initial registration of the instance.
```
lmars: Should this read something like:

```
Index is the number of registrations since the initial registration of the service
```

If not, maybe I'm misunderstanding what this is.

titanous: This maps directly to the `CreatedIndex` from etcd, but could map to something else in a different backend. In etcd, the "index" increments with every write to the database, and `CreatedIndex` is the index of the write that created the key.

lmars: Ok, I guess what threw me off was the last comment `increase by at least one with each new registration.`. Is a "registration" a heartbeat of this instance like in the current discoverd? Or is a registration the creation of a new instance?

titanous: Ah, I will clarify.

titanous: Clarified it as sort order instead of incrementing.

titanous:
@heavenlyhash @lmars I think all comments have been addressed.

lmars:
LGTM

heavenlyhash:
LGTM

PR #691, opened 2015-01-04
Add default manifest paths to download and bootstrap

titanous:
This is much less confusing than hanging waiting for STDIN.


heavenlyhash:
LGTM

PR #695, opened 2015-01-05
website: Add sitemap

titanous:
This adds a standard sitemap.xml and sitemap.xml.gz so that crawlers can find everything easily.


lmars:
LGTM

PR #697, opened 2015-01-05
host: Don’t try to set the bridge MTU

titanous:
Setting the bridge MTU errors with EINVAL if the MTU is greater than 1500, which it will be if eth0 supports jumbo frames. The bridge should automatically use the lowest MTU of all interfaces associated with it, so this setting is not necessary.

### DIFF
--- a/host/libvirt_lxc_backend.go
+++ b/host/libvirt_lxc_backend.go
@@ -76,6 +76,7 @@ type LibvirtLXCBackend struct {
 	state     *State
 	pinkerton *pinkerton.Context
 
+	ifaceMTU   int
 	bridgeAddr net.IP
 	bridgeNet  *net.IPNet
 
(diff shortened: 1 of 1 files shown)


host/libvirt_lxc_backend.go
```diff
@@ -76,6 +76,7 @@ type LibvirtLXCBackend struct {
 	state     *State
 	pinkerton *pinkerton.Context
 
+	ifaceMTU   int
```
lmars: Do we need this?

titanous: At some point we should pass it to containerinit, the other option is not parsing the MTU out of the file for now.

lmars: I don't feel strongly about removing it, just looked like an odd diff introducing a field which isn't used

lmars:
LGTM

PR #699, opened 2015-01-06
bootstrap: Log dashboard link and token

lmars:
This adds the dashboard link and token to the log message, and a link to the CLI installation page:

```
12:47:19.635648 log log-complete
12:47:19.635819 log log-complete 

Flynn bootstrapping complete. Install the flynn-cli (see https://cli.flynn.io for instructions) and paste the line below into a terminal window:

flynn cluster add -g dev.localflynn.com:2222 -p tMq6CLHmv2g9ys2Zf/2n4WnC63tdscFsNMflk91538k= default https://controller.dev.localflynn.com aa01ab76e5a68bb33e2678e67813baa4

The built-in dashboard can be accessed at http://dashboard.dev.localflynn.com with login token 3daf12f05c20f88edcf201f386f8818f
```

PR #705, opened 2015-01-07
fix bootstrap success detection in integration test kickoff script

heavenlyhash:
The integration scripts run test checks against output which was changed in the "Log dashboard link and token" commit ( https://github.com/flynn/flynn/pull/699 ), so this is changing to match.


archSeer:
LGTM

titanous:
LGTM

PR #707, opened 2015-01-07
website: Fix swiftype section annotation

titanous:
There was a typo in their docs, it’s ‘sections’, not ‘section’.


jvatic:
LGTM

PR #716, opened 2015-01-12
discoverd2: Implement leader election

titanous:
This is an incremental patch for #145.

### DIFF
--- a/discoverd2/server/state.go
+++ b/discoverd2/server/state.go
@@ -98,21 +98,16 @@ type Instance struct {
 	// Meta is arbitrary metadata specified when registering the instance.
 	Meta map[string]string `json:"meta,omitempty"`
 
-	// Leader is true if the instance is the leader of the service. Exactly one
-	// instance per service has this set to true at any point in time.
-	Leader bool `json:"leader,omitempty"`
-
 	// Index is the logical epoch of the initial registration of the instance.
-	// It is guaranteed to be unique, not change as long as the instance does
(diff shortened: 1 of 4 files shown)


discoverd2/server/state_test.go
```diff
@@ -193,40 +219,237 @@ func (StateSuite) TestSetService(c *C) {
...
+	c.Assert(state.GetLeader("a"), DeepEquals, third)
+
+	// set with same instance and another instance triggers no events
+	fourth := fakeInstance()
+	fourth.Index = 5
+	state.SetService("a", []*Instance{fourth, third})
+	assertNoEvent(c, events)
+	c.Assert(state.GetLeader("a"), DeepEquals, third)
+
+	// set with same instance and lower selects a new leader
```
heavenlyhash: s/lower/lower index/

discoverd2/server/state.go
```diff
@@ -199,93 +196,185 @@ type State struct {
...
+		instances: make(map[string]*Instance),
+	}
+}
+
+type service struct {
+	// instance ID -> instance
+	instances map[string]*Instance
+
+	leaderID     string
+	leaderIndex  uint64
```
heavenlyhash: Perhaps a quick comment that this is usually constrained to being `>0` as per the `Instance.Index` field, with a sentinel value for zero -> unset?

titanous:
@heavenlyhash Comments addressed.

heavenlyhash:
LGTM

PR #717, opened 2015-01-12
cli: Adding --remote flag on "flynn create" command

### LINKED ISSUE #700: Option to rename git remote on "flynn create" command
Hello guys,

I want to create 2 apps from the same repo, one for staging, one for production. All commands accept the "-a" flag to point the app, but the "flynn create" command replaces the "flynn git remote" when I run it twice.

Actually I workaround this way:

```sh
$ git remote -v
flynn	ssh://git@demo.localflynn.com:2222/hares-dumbfound-killaloe.git (fetch) <<< the "stage" app
flynn	ssh://git@demo.localflynn.com:2222/hares-dumbfound-killaloe.git (push)

$ flynn create
Created lemurs-discourage-machin

$ git remote rename flynn flynn-prod

$ git remote add flynn ssh://git@demo.localflynn.com:2222/hares-dumbfound-killaloe.git

$ git remote -v
flynn	ssh://git@demo.localflynn.com:2222/hares-dumbfound-killaloe.git (fetch)
flynn	ssh://git@demo.localflynn.com:2222/hares-dumbfound-killaloe.git (push)
flynn-prod	ssh://git@demo.localflynn.com:2222/lemurs-discourage-machin.git (fetch)
flynn-prod	ssh://git@demo.localflynn.com:2222/lemurs-discourage-machin.git (push)
```

Discussing with @lmars on IRC, we think that could be useful include a flag on the command, giving the name of git remote. Something like this:

```sh
$ flynn create [ -r | --remote ] <name>
```

geovanisouza92:
Hello guys,

Closes #700.

This is my first Go code contribution, so, apologize for any formatting/style/logic issues. I'm open to listen suggestions!

This option allow us to set the local git remote name, or "flynn" if not provided. It checks if the remote name is used and warn user about conflict.

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -99,7 +99,7 @@ func runCreate(args *docopt.Args, client *controller.Client) error {
 		if r == remote {
 			fmt.Println("There is already a git remote called", remote)
 			if !promptYesNo("Are you sure you want to replace it?") {
-				log.Println("Please, declare the desired local git remote name with --remote flag.")
+				log.Println("The app was not created. Please, declare the desired local git remote name with --remote flag.")
 				return nil
 			}
 		}
--- a/cli/app.go
+++ b/cli/app.go
@@ -84,7 +84,10 @@ func promptYesNo(msg string) (result bool) {
 func runCreate(args *docopt.Args, client *controller.Client) error {
 	app := &ct.App{}
 	app.Name = args.String["<name>"]
-	remote := args.String["<remote>"] || "flynn"
+	remote := args.String["<remote>"]
+	if remote == "" {
+		remote = "flynn"
+	}
 
 	// Test if remote name exists and prompt user
 	remotes, err = gitRemoteNames()
--- a/cli/app.go
+++ b/cli/app.go
@@ -111,8 +111,10 @@ func runCreate(args *docopt.Args, client *controller.Client) error {
 	}
 
 	// Register git remote
-	exec.Command("git", "remote", "remove", remote).Run()
-	exec.Command("git", "remote", "add", remote, gitURLPre(clusterConf.GitHost)+app.Name+gitURLSuf).Run()
+	if remote != "" {
+		exec.Command("git", "remote", "remove", remote).Run()
+		exec.Command("git", "remote", "add", remote, gitURLPre(clusterConf.GitHost)+app.Name+gitURLSuf).Run()
+	}
 	log.Printf("Created %s", app.Name)
 	return nil
 }
--- a/cli/app.go
+++ b/cli/app.go
@@ -23,7 +23,7 @@ allows deploying the application via git.
 
 Options:
 	-r, --remote <remote>  Name of git remote to create, empty string for none. [default: flynn]
-	-y, --yes              Skip the confirmation prompt if there is already a "flynn" git remote.
+	-y, --yes              Skip the confirmation prompt if the git remote already exists.
 
 Examples:
 
--- a/cli/app.go
+++ b/cli/app.go
@@ -23,6 +23,7 @@ allows deploying the application via git.
 
 Options:
 	-r, --remote <remote>  Name of git remote to create, empty string for none. [default: flynn]
+	-y, --yes              Skip the confirmation prompt if there is already a "flynn" git remote.
 
 Examples:
 
@@ -92,12 +93,14 @@ func runCreate(args *docopt.Args, client *controller.Client) error {
 		return err
 	}
 
-	for _, r := range remotes {
-		if r == remote {
-			fmt.Println("There is already a git remote called", remote)
-			if !promptYesNo("Are you sure you want to replace it?") {
-				log.Println("The app was not created. Please, declare the desired local git remote name with --remote flag.")
-				return nil
+	if !args.Bool["--yes"] {
+		for _, r := range remotes {
+			if r == remote {
+				fmt.Println("There is already a git remote called", remote)
+				if !promptYesNo("Are you sure you want to replace it?") {
+					log.Println("The app was not created. Please, declare the desired local git remote name with --remote flag.")
+					return nil
+				}
 			}
 		}
 	}
--- a/cli/app.go
+++ b/cli/app.go
@@ -121,19 +121,8 @@ func runDelete(args *docopt.Args, client *controller.Client) error {
 	appName := mustApp()
 
 	if !args.Bool["--yes"] {
-		fmt.Printf("Are you sure you want to delete the app %q? (yes/no): ", appName)
-	loop:
-		for {
-			var answer string
-			fmt.Scanln(&answer)
-			switch answer {
-			case "y", "yes":
-				break loop
-			case "n", "no":
-				return nil
-			default:
-				fmt.Print("Please type 'yes' or 'no': ")
-			}
+		if !promptYesNo(fmt.Sprintf("Are you sure you want to delete the app %q?", appName)) {
+			return nil
 		}
 	}
 
--- a/cli/app.go
+++ b/cli/app.go
@@ -97,7 +97,7 @@ func runCreate(args *docopt.Args, client *controller.Client) error {
 
 	for _, r := range remotes {
 		if r == remote {
-			fmt.Println("There is one git remote called", remote)
+			fmt.Println("There is already a git remote called", remote)
 			if !promptYesNo("Are you sure you want to replace it?") {
 				log.Println("Please, declare the desired local git remote name with --remote flag.")
 				return nil
--- a/cli/app.go
+++ b/cli/app.go
@@ -90,7 +90,7 @@ func runCreate(args *docopt.Args, client *controller.Client) error {
 	}
 
 	// Test if remote name exists and prompt user
-	remotes, err = gitRemoteNames()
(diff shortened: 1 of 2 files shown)


geovanisouza92:
There is another issue. The "flynn delete" command expect the "flynn" remote, so it must be changed too. I'm not sure how to achieve this if the remote have another name. Maybe the same flag?

archSeer:
I wonder if changing remotes is the right solution for this. You're not really creating several different apps from one project repository, you're just creating different deploys of the same app (`staging`, `production`, etc.). This feels more like branches than different remotes. The distinction should be on the server, and changes to be made there to support multiple different deploys (with different environment I guess) of the same app.

geovanisouza92:
@archSeer This sounds logical to me, but it requires more changes on cli and server. How you think that this should work?

titanous:
For now we're going to stick to the Heroku-style 1:1 mapping of git remotes to apps, anything else adds complexity that we don't need right now.

cli/app.go
```diff
@@ -61,16 +64,55 @@ Examples:
 `)
 }
 
+func promptYesNo(msg string) (result bool) {
```
titanous: This should probably replace the implementation here as well: https://github.com/flynn/flynn/blob/40b1a276eec4c1650309cdb438460e0bc7b3eee7/cli/app.go#L83-L96

cli/app.go
```diff
@@ -61,16 +64,55 @@ Examples:
...
+
+	// Test if remote name exists and prompt user
+	remotes, err := gitRemoteNames()
+	if err != nil {
+		return err
+	}
+
+	for _, r := range remotes {
+		if r == remote {
+			fmt.Println("There is one git remote called", remote)
```
titanous: s/one/already a/

cli/app.go
```diff
@@ -61,16 +64,55 @@ Examples:
...
+			if !promptYesNo("Are you sure you want to replace it?") {
+				log.Println("Please, declare the desired local git remote name with --remote flag.")
```
titanous: This should mention that the remote was not created.

cli/app.go
```diff
@@ -61,16 +64,55 @@ Examples:
...
+	// Test if remote name exists and prompt user
+	remotes, err := gitRemoteNames()
+	if err != nil {
+		return err
+	}
+
+	for _, r := range remotes {
+		if r == remote {
+			fmt.Println("There is one git remote called", remote)
+			if !promptYesNo("Are you sure you want to replace it?") {
```
titanous: We also need a `-y` flag now.

cli/git.go
```diff
@@ -34,6 +34,31 @@ type remoteApp struct {
...
+func gitRemoteNames() (results []string, err error) {
+	b, err := exec.Command("git", "remote").Output()
+	if err != nil {
+		return nil, err
+	}
+
+	s := bufio.NewScanner(bytes.NewBuffer(b))
+	s.Split(bufio.ScanWords)
+
+	results = make([]string, 1)
```
titanous: This results in an empty string at the beginning of the slice, is that intentional? If not:

```go
var results []string
```

cli/app.go
```diff
@@ -21,6 +21,9 @@ If a name is not provided, a random name will be generated.
 If run from a git repository, a 'flynn' remote will be created or replaced that
 allows deploying the application via git.
 
+Options:
+	-r, --remote <remote>  Name of git remote on local repo.
```
titanous:
```text
Name of the git remote to create, empty string for none.
```

Also, it should have `[default: flynn]` at the end so that we don't need a special case below.

cli/app.go
geovanisouza92: I code it with this intention in mind :smile:

cli/app.go
geovanisouza92: Sorry for that, I'll fix it

cli/app.go
geovanisouza92: Good point.

cli/app.go
geovanisouza92: I agree.

cli/git.go
geovanisouza92: I don't know. I'll fix that.

cli/app.go
geovanisouza92: Good point.

geovanisouza92: Do you mean skip [this change](https://github.com/flynn/flynn/pull/717/files#diff-333b1ce61f983aec7830fc16c0445bcaR98)?

cli/app.go
```diff
@@ -21,6 +21,10 @@ If a name is not provided, a random name will be generated.
 If run from a git repository, a 'flynn' remote will be created or replaced that
 allows deploying the application via git.
 
+Options:
+	-r, --remote <remote>  Name of git remote to create, empty string for none. [default: flynn]
+	-y, --yes              Skip the confirmation prompt if there is already a "flynn" git remote.
```
titanous:
```text
Skip the confirmation prompt if the git remote already exists.
```

cli/app.go
```diff
@@ -61,16 +65,54 @@ Examples:
...
+	// Create the app
 	if err := client.CreateApp(app); err != nil {
 		return err
 	}
 
-	exec.Command("git", "remote", "remove", "flynn").Run()
-	exec.Command("git", "remote", "add", "flynn", gitURLPre(clusterConf.GitHost)+app.Name+gitURLSuf).Run()
+	// Register git remote
+	exec.Command("git", "remote", "remove", remote).Run()
+	exec.Command("git", "remote", "add", remote, gitURLPre(clusterConf.GitHost)+app.Name+gitURLSuf).Run()
```
titanous: This should check if the remote is an empty string and skip if so.

titanous:
This is looking good! I'd also like to see some integration tests added: https://github.com/flynn/flynn/blob/master/test/test_cli.go

cli/app.go
geovanisouza92: Why we want to create the app and not register the remote?

geovanisouza92:
I'll try add more tests too! :smile:

cli/app.go
titanous: If you are not in a git directory and/or are not going to use git to deploy the app, then a git remote doesn't make sense.

geovanisouza92: Create an orfan app? Without repo/code?

titanous: Yes, git is not the only way to deploy apps to Flynn. We also support specifying a pre-built container image, and will support other methods in the future.

geovanisouza92: Ah, good point! I'm writing it.

geovanisouza92: Ah, good point!

In this case, the command will need more options, to specify container images and other methods in future, right?

cli/app.go
```diff
@@ -61,16 +65,56 @@ Examples:
...
 	if err := client.CreateApp(app); err != nil {
 		return err
 	}
 
-	exec.Command("git", "remote", "remove", "flynn").Run()
-	exec.Command("git", "remote", "add", "flynn", gitURLPre(clusterConf.GitHost)+app.Name+gitURLSuf).Run()
+	// Register git remote
+	if remote != "" {
+		exec.Command("git", "remote", "remove", remote).Run()
+		exec.Command("git", "remote", "add", remote, gitURLPre(clusterConf.GitHost)+app.Name+gitURLSuf).Run()
```
heavenlyhash: We should add an "--" arg before the final two chunks in this `git remote add.  Since the add subcommand takes a variety of other arguments, this will make it absolutely clear to git that these are meant to be the `<name> <url>` chunks.

In this case, the `gitURLPre` function already massages input such that it covers all injections I can currently think of, so the prior state doesn't seem like an active security/stability hole, but if the `remote` value is user input, we definitely want this protection now (and I'd advocate doing it just out of habit for clarity and defense in depth regardless; but now's as good a time as any :) ).

(The adjacent `git remote remove` command supports no other optional arguments, so it doesn't require an end-of-options signifier... looks like it doesn't accept one either.)

geovanisouza92: @heavenlyhash Do you mean

```go
... "add", "--", remote, gitURLPre(clusterConf.GitHost)+app.Name+gitURLSuf).Run()
```

Or

```go
... "add", remote, gitURLPre(clusterConf.GitHost)+app.Name+gitURLSuf, "--").Run()
```

heavenlyhash: The former :)

geovanisouza92: Sorry for my ignorance, but, could be more clear? Tnx

heavenlyhash: The first option you listed is correct.  `... "add", "--", remote, ...`

geovanisouza92: Ok, done. :+1:

titanous:
@geovanisouza92 Are you planning on writing the test for this?

geovanisouza92:
@titanous I'm very busy these days, sorry! :disappointed:

titanous:
Okay, no problem. I'll get someone else to finish it. Thanks!

jvatic:
See #777

titanous:
This got merged, thanks @geovanisouza92!

geovanisouza92:
You're welcome! I hope contribute more in future :smile:

PR #725, opened 2015-01-12
flannel: Use RFC 6598 shared address space

### LINKED ISSUE #724: Flannel network conflicts with EC2 DNS
The flannel network definition of 172.16.0.0/12 currently conflicts with EC2 DNS IP of 172.16.0.23.

This causes any DNS lookups to fail after the creation of the flannel network interface.

titanous:
[RFC 6598](https://tools.ietf.org/html/rfc6598) defines 100.64.0.0/10 as shared address space to be used for carrier-grade NAT, but it also allows its use as private address space.

Using this instead of 172.16.0.0/12 will avoid a ton of conflicts.

Closes #724

### DIFF
--- a/flannel/runner.go
+++ b/flannel/runner.go
@@ -33,7 +33,7 @@ var networkConfigAttempts = attempt.Strategy{
 func main() {
 	var config Config
 	config.Backend.Type = "vxlan"
-	flag.StringVar(&config.Network, "network", "172.16.0.0/12", "container network")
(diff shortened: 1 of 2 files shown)


heavenlyhash:
LGTM.  (Docker also avoids 172.16 for the same reason: https://github.com/docker/docker/blob/3ec695924009421f9b1f79368e9e5b1e3e2ca94f/daemon/networkdriver/bridge/driver.go#L62 )

PR #726, opened 2015-01-12
discoverd2: Move Instance, Event, and EventKind to client package

titanous:
Incremental patch for #145.

There are zero code changes here, just namespace changes. Instead of creating a `types` package, the shared types will live in the client package, the tests will remain in the server package.


lmars:
LGTM

PR #727, opened 2015-01-13
scheduler: Cancel restarts when removing jobs.

### LINKED ISSUE #610: scheduler: Cancel job restart timers
When scaling a formation process type down, jobs which are scheduled to start in the future should have their timers cancelled.

archSeer:
Fixes #610.

### DIFF
--- a/controller/scheduler/main.go
+++ b/controller/scheduler/main.go
@@ -505,6 +505,7 @@ type Job struct {
 
 	restarts  int
 	timer     *time.Timer
+	timerMtx  sync.Mutex
 	startedAt time.Time
 }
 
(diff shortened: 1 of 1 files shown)


controller/scheduler/main.go
```diff
@@ -746,10 +757,19 @@ func (f *Formation) jobType(job *host.Job) string {
...
-func (s sortJobs) Len() int           { return len(s) }
-func (s sortJobs) Less(i, j int) bool { return s[i].startedAt.Sub(s[j].startedAt) > 0 }
-func (s sortJobs) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
-func (s sortJobs) Sort()              { sort.Sort(s) }
+func (s sortJobs) Len() int { return len(s) }
+func (s sortJobs) Less(i, j int) bool {
+	switch {
+	case s[i].timer != nil && s[j].timer == nil:
+		return true
+	case s[i].timer == nil && s[j].timer != nil:
```
titanous: Copying this comment from the other PR: Does this need a mutex? Where is sort called from?

lmars: `sortJobs` is used then removing jobs due to scaling, whereas `Job.timer` is potentially written to when the host job crashes (which will be in another goroutine), so I think it does need a lock here.

titanous:
What do we need to get this merged?

archSeer:
I think adding the locks was the last thing needed.

titanous:
LGTM

PR #728, opened 2015-01-13
simplify httpclient.Stream invocation

heavenlyhash:
This makes creating http client streams significantly terser at the cost of a tiny bit more reflection, by automatically producing message objects based on the output channel's element type.

### DIFF
--- a/pkg/httpclient/stream.go
+++ b/pkg/httpclient/stream.go
@@ -13,18 +13,17 @@ import (
 /*
 	Stream manufactures a `pkg/stream.Stream`, starts a worker pumping events out of decoding, and returns that.
 
-	The 'msgFactory' parameter is invoked to produce a structure that deserialized data is mapped into.
-
-	The 'outputCh' parameter must be a sendable channel.  The channel's content type must be the same as the return type of the factory func.
+	The 'outputCh' parameter must be a sendable channel.  The "zero"-values of channel's content type will be created and used in the deserialization, then sent.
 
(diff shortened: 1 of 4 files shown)


pkg/httpclient/stream.go
```diff
@@ -13,18 +13,17 @@ import (
...
 	The return values from `httpclient.RawReq` are probably a useful starting point for the 'res' parameter.
 
 	Closing the returned `stream.Stream` shuts down the worker.
 */
-func Stream(res *http.Response, msgFactory func() interface{}, outputCh interface{}) stream.Stream {
+func Stream(res *http.Response, outputCh interface{}) stream.Stream {
 	stream := stream.New()
 	chanValue := reflect.ValueOf(outputCh)
 	stopChanValue := reflect.ValueOf(stream.StopCh)
+	msgType := reflect.TypeOf(outputCh).Elem().Elem()
```
titanous: Can just call `chanValue.Type`

pkg/httpclient/stream.go
```diff
@@ -34,7 +33,7 @@ func Stream(res *http.Response, msgFactory func() interface{}, outputCh interfac
 		r := bufio.NewReader(res.Body)
 		dec := sse.NewDecoder(r)
 		for {
-			msg := msgFactory()
+			msg := reflect.New(msgType).Interface()
```
titanous: If you save the value instead of throwing it away, there's no need for the `ValueOf` call below.

titanous:
LGTM

PR #729, opened 2015-01-13
host API cleanup

heavenlyhash:
One tiny change suggested by @archSeer; and a rename that will make sense as the volumes API comes in next to this.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -19,7 +19,6 @@ import (
 	"github.com/flynn/flynn/pkg/attempt"
 	"github.com/flynn/flynn/pkg/cluster"
 	"github.com/flynn/flynn/pkg/shutdown"
-	"github.com/flynn/flynn/pkg/stream"
 )
 
 // discoverdAttempts is the attempt strategy that is used to connect to discoverd.
@@ -175,13 +174,7 @@ func runDaemon(args *docopt.Args) {
 		sh.Fatal(err)
 	}
 
-	var jobStream stream.Stream
-	sh.BeforeExit(func() {
-		if jobStream != nil {
-			jobStream.Close()
-		}
-		backend.Cleanup()
-	})
+	sh.BeforeExit(func() { backend.Cleanup() })
 
 	if force {
 		if err := backend.Cleanup(); err != nil {
@@ -286,7 +279,7 @@ func runDaemon(args *docopt.Args) {
 
 		h.Jobs = state.ClusterJobs()
 		jobs := make(chan *host.Job)
-		jobStream, err = cluster.RegisterHost(h, jobs)
+		jobStream, err := cluster.RegisterHost(h, jobs)
 		if err != nil {
 			sh.Fatal(err)
 		}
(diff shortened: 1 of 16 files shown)


host/host.go
```diff
@@ -252,7 +252,7 @@ func runDaemon(args *docopt.Args) {
 
 	// Check if we are the leader so that we can use the cluster functions directly
 	sampiCluster := sampi.NewCluster(sampi.NewState())
-	sampiAPI := sampi.NewHTTPAPI(sampiCluster)
+	sampiAPI := &sampi.HttpAPI{sampiCluster}
```
titanous: Vet will complain about the missing field name

host/http.go
```diff
@@ -54,11 +54,11 @@ func (h *Host) streamEvents(id string, w http.ResponseWriter) error {
 	return nil
 }
 
-type httpAPI struct {
+type hostAPI struct {
```
heavenlyhash: Or actually -- maybe `jobAPI`?

archSeer: Sounds good

archSeer:
LGTM, pending that rename

archSeer:
We could also get rid of [sampi.NewCluster](https://github.com/flynn/flynn/blob/989341154e25ffde3b84437605f742bd7a04ef78/host/sampi/http.go#L21). This will need to be changed then: https://github.com/flynn/flynn/blob/35e9d7bd89f0d78a0778da92a02c47b886d97884/host/host.go#L254

archSeer:
Ah actually, I'd change `sampi.NewCluster` to be without arguments and create a new state object within the constructor.

host/host.go
```diff
@@ -275,16 +269,18 @@ func runDaemon(args *docopt.Args) {
...
 	for _, s := range metadata {
 		kv := strings.SplitN(s, "=", 2)
 		h.Metadata[kv[0]] = kv[1]
 	}
-	h.ID = hostID
 
+	var jobStream stream.Stream
+	sh.BeforeExit(func() {
+		if jobStream != nil {
+			jobStream.Close()
```
titanous: This seems like unnecessary cleanup?

host/sampi/http.go
```diff
@@ -76,8 +76,8 @@ func (s *Cluster) StreamHostEvents(ch chan host.HostEvent, done chan bool) error
 	return nil
 }
 
-type httpAPI struct {
-	cluster *Cluster
+type HttpAPI struct {
```
titanous: This should be `HTTPAPI`

host/host.go
archSeer: As I said in the commit, I found it easier to follow with the `jobStream` logic being all in one section.

titanous: I mean, closing the stream on exit doesn't really do much, the fds will be reaped by the kernel at exit anyway.

titanous:
LGTM, though a few commit messages need a `pkg/` prefix.

PR #730, opened 2015-01-13
controller,scheduler: Implement job metadata

### LINKED ISSUE #627: Expose job attributes in controller
When creating a one-off job in the controller, I should be able to set job attributes, and when I retrieve the list of jobs from the controller, I should see those attributes. This will allow us to do things like link up taffy deploy logs with apps in the dashboard.

jvatic:
closes #627

### DIFF
--- a/controller/jobs.go
+++ b/controller/jobs.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-sql"
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/pq"
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/pq/hstore"
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-martini/martini"
 	ct "github.com/flynn/flynn/controller/types"
 	"github.com/flynn/flynn/host/types"
@@ -75,7 +76,7 @@ func NewJobRepo(db *postgres.DB) *JobRepo {
 }
 
 func (r *JobRepo) Get(id string) (*ct.Job, error) {
-	row := r.db.QueryRow("SELECT concat(host_id, '-', job_id), app_id, release_id, process_type, state, created_at, updated_at FROM job_cache WHERE concat(host_id, '-', job_id) = $1", id)
+	row := r.db.QueryRow("SELECT concat(host_id, '-', job_id), app_id, release_id, process_type, state, meta, created_at, updated_at FROM job_cache WHERE concat(host_id, '-', job_id) = $1", id)
 	return scanJob(row)
 }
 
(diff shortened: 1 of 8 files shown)


controller/jobs.go
```diff
@@ -87,12 +88,19 @@ func (r *JobRepo) Add(job *ct.Job) error {
...
 	// TODO: actually validate
-	err = r.db.QueryRow("INSERT INTO job_cache (job_id, host_id, app_id, release_id, process_type, state) VALUES ($1, $2, $3, $4, $5, $6) RETURNING created_at, updated_at",
-		jobID, hostID, job.AppID, job.ReleaseID, job.Type, job.State).Scan(&job.CreatedAt, &job.UpdatedAt)
+	err = r.db.QueryRow("INSERT INTO job_cache (job_id, host_id, app_id, release_id, process_type, state, meta) VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING created_at, updated_at",
+		jobID, hostID, job.AppID, job.ReleaseID, job.Type, job.State, meta).Scan(&job.CreatedAt, &job.UpdatedAt)
 	if e, ok := err.(*pq.Error); ok && e.Code.Name() == "unique_violation" {
-		err = r.db.QueryRow("UPDATE job_cache SET state = $3, updated_at = now() WHERE job_id = $1 AND host_id = $2 RETURNING created_at, updated_at",
-			jobID, hostID, job.State).Scan(&job.CreatedAt, &job.UpdatedAt)
+		err = r.db.QueryRow("UPDATE job_cache SET state = $3, meta = $4, updated_at = now() WHERE job_id = $1 AND host_id = $2 RETURNING created_at, updated_at",
+			jobID, hostID, job.State, meta).Scan(&job.CreatedAt, &job.UpdatedAt)
```
titanous: The update doesn't need to set the metadata as the only thing that can change is the state.

controller/jobs.go
```diff
@@ -437,13 +452,16 @@ func runJob(app *ct.App, newJob ct.NewJob, releases *ReleaseRepo, artifacts *Art
 	for k, v := range newJob.Env {
 		env[k] = v
 	}
+	metadata := make(map[string]string)
```
titanous: Add a size hint of `len(newJob.Meta)+3`

controller/scheduler/main.go
```diff
@@ -162,6 +163,14 @@ func (c *context) syncCluster() {
 				f = c.formations.Add(f)
 			}
 
+			jobMeta := make(map[string]string, len(job.Metadata))
+			for k, v := range job.Metadata {
+				if strings.HasPrefix(k, "flynn-controller") {
```
titanous: Add the trailing period as well

titanous:
Neat! This probably deserves an integration test to confirm that the scheduler is working as expected.

controller/jobs.go
```diff
@@ -102,20 +110,27 @@ func (r *JobRepo) Add(job *ct.Job) error {
...
 		if err == sql.ErrNoRows {
 			err = ErrNotFound
 		}
 		return nil, err
 	}
+	if len(meta.Map) > 0 {
+		job.Meta = make(map[string]string, len(meta.Map))
+		for k, v := range meta.Map {
+			job.Meta[k] = v.String
+		}
```
titanous: Do we do this anywhere else? Does a helper function make sense? (ditto for the opposite direction)

controller/scheduler/main.go
titanous: Also, maybe a helper makes sense?

controller/jobs.go
jvatic: I don't think we do as this function is called for both Get and List. The opposite direction could share a helper function with `AppRepo.Add` for creating an `hstore.Hstore` from `map[string]string`.

titanous: Okay, sounds good.

controller/scheduler/main.go
```diff
@@ -169,6 +181,7 @@ func (c *context) syncCluster() {
 				ReleaseID: releaseID,
 				Type:      jobType,
 				State:     "up",
+				Meta:      jobMetaFromMetadata(job.Metadata),
```
lmars: Given the scheduler does not change the metadata, does it need to send it here? I know `PUT` requests are supposed to put the complete resource, maybe change it to `PATCH`?

titanous: I don't think we know whether the job is in the database yet, so we need to include the full details.

jvatic:
Integration test added, some code refactored.

titanous:
LGTM!

PR #732, opened 2015-01-14
pkg/httpclient: Ensure we don’t leak HTTP response bodies

titanous:
`RawReq` does not close the response body if the out parameter is nil. This patch adjusts `Send` to close the body if out is nil, as it does not expose the response to callers.


heavenlyhash:
LGTM

archSeer:
LGTM

PR #733, opened 2015-01-14
pkg/httpclient: Close stream http response body when stopping

titanous:
This unblocks the SSE decoder if the stop request comes in while we are blocked in a read.

### DIFF
--- a/pkg/httpclient/stream.go
+++ b/pkg/httpclient/stream.go
@@ -25,8 +25,17 @@ func Stream(res *http.Response, outputCh interface{}) stream.Stream {
 	stopChanValue := reflect.ValueOf(stream.StopCh)
 	msgType := chanValue.Type().Elem().Elem()
 	go func() {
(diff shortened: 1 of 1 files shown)


heavenlyhash:
And `res.Body.Close()` makes the decoder return immediately?  Awesome!  It was really bugging me that a goroutine could get stuck there (and pass the problem along if you tried to drain the stream after requesting exit!).

I had to read slowly to grok the extra channel and goroutine, but LGTM.

heavenlyhash:
I wonder how this would compare to separating the byte read from the decoding and put it with the other select?  I'm not sure which would be preferable; lifting the read and joining the selects would get rid of a goroutine and channel, but since it's putting things into a reflective select I'm not sure if that's either faster or more readable.

titanous:
I'm not sure how the byte read could go in a select, it's not a channel.

Issue #528, opened 2014-11-30
Data validation


titanous:
We are not currently defensive enough with input. All input to the controller, host, router, and discoverd needs to be fully validated before processing (and tests added for bad input).

In addition we should have standard good error messages and status codes that point at the exact field that is incorrect, this is partially mentioned in #322. We also need to ensure that conflicts with existing objects are handled correctly.

**Controller**
- [x] Release
- [x] Formation
- [x] Job
- [x] Resource
- [x] Provider
- [x] Route
- [x] Artifact
- [x] App
- [x] Key

**Router**
- [ ] Route

Host and discoverd should wait until #170 is addressed.

heavenlyhash:
Error messages stand improved now that #749 has landed.

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

Issue #322, opened 2014-10-16
Creating duplicate apps causes a 500


lmars:
```
$ flynn create example
Created example
$ flynn create example
POST http://dev.localflynn.com:443/apps: controller: unexpected status 500
```

This should print a validation error message instead.

PR #734, opened 2015-01-14
discoverd2: Implement HTTP API server and client

titanous:
Incremental patch for #145. This implements the HTTP API and client along with tests.

### DIFF
--- a/discoverd2/client/client.go
+++ b/discoverd2/client/client.go
@@ -1,0 +1,183 @@
+package discoverd
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/flynn/flynn/pkg/httpclient"
+	"github.com/flynn/flynn/pkg/stream"
+)
+
+type Service interface {
+	Leader() (*Instance, error)
+	Instances() ([]*Instance, error)
+	Addrs() ([]string, error)
+	Leaders(chan *Instance) (stream.Stream, error)
+	Watch(events chan *Event) (stream.Stream, error)
+}
+
+var ErrNotFound = errors.New("discoverd: not found")
+var ErrTimedOut = errors.New("discoverd: timed out waiting for instances")
+
+type Client struct {
+	c *httpclient.Client
+}
+
+func NewClient() *Client {
+	url := os.Getenv("DISCOVERD")
+	if url == "" {
+		url = "http://127.0.0.1:1111"
+	}
+	return NewClientWithURL(url)
+}
+
+func NewClientWithURL(url string) *Client {
+	if !strings.HasPrefix(url, "http") {
+		url = "http://" + url
+	}
+	return &Client{
+		c: &httpclient.Client{
+			ErrPrefix:   "discoverd",
+			ErrNotFound: ErrNotFound,
+			URL:         url,
+			HTTP:        http.DefaultClient,
+		},
+	}
+}
+
+func (c *Client) AddService(name string) error {
+	return c.c.Put("/services/"+name, nil, nil)
+}
+
+func (c *Client) RemoveService(name string) error {
+	return c.c.Delete("/services/" + name)
+}
+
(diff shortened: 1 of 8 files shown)


discoverd2/client/client.go
```diff
@@ -0,0 +1,188 @@
...
+			case <-stream.StopCh:
+				return
+			}
+		}
+	}()
+	return stream, nil
+}
+
+func (s *service) Watch(events chan *Event) (stream.Stream, error) {
+	res, err := s.client.c.RawReq("GET", fmt.Sprintf("/services/%s", s.name), http.Header{"Accept": {"text/event-stream"}}, nil, nil)
```
archSeer: Most of our streaming client code is now really generic. What do you think about a helper method on `httpclient` that would wrap this into

```go
func (s *service) Watch(events chan *Event) (stream.Stream, error) {
    return s.client.c.Stream("GET", fmt.Sprintf("/services/%s", s.name), events)
}
```

Stream itself being something like:
```go
func (c *client) Stream(method, url string, out interface{}) {
    res, err := s.client.c.RawReq("GET", fmt.Sprintf("/services/%s", s.name), http.Header{"Accept": {"text/event-stream"}}, nil, nil)
    if err != nil {
       return nil, err
    }
    return httpclient.Stream(res, out), nil
}
```

archSeer: Wait, I'll implement this in a separate PR.

discoverd2/server/http.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+type httpAPI struct {
+	Store Datastore
+}
+
+func jsonError(w http.ResponseWriter, code int, err error) {
+	httphelper.JSON(w, code, struct {
+		Error string `json:"error"`
+	}{err.Error()})
```
heavenlyhash: Might make sense to lift into httphelper?

Should we consider additional error enum strings that are well defined and exported, so the remote can easily check for them?  Maybe the error messages would be enough, but I always feel a little uncomfortable checking a string that isn't clearly marked as not-for-humans/do-not-refactor-phrasing.

titanous: We should do that, but it should be a project-wide thing, not just a one-off for this API. #322 mentions this and it would make sense to do as a part of #528.

discoverd2/client/client.go
titanous: Sounds good.

discoverd2/client/heartbeat.go
```diff
@@ -0,0 +1,83 @@
...
+		stop:    make(chan struct{}),
+		inst:    inst.Clone(),
+	}
+	go h.run(firstErr)
+	return h, <-firstErr
+}
+
+type heartbeater struct {
+	c *Client
+	sync.Mutex
```
lmars: Whats field is the mutex protecting?

titanous: `inst.Meta`, I'll add a comment.

discoverd2/server/http.go
```diff
@@ -0,0 +1,217 @@
...
+	"encoding/json"
+	"errors"
+	"net/http"
+	"strings"
+
+	"github.com/flynn/flynn/discoverd2/client"
+	"github.com/flynn/flynn/pkg/httphelper"
+	"github.com/flynn/flynn/pkg/sse"
+	"github.com/flynn/flynn/pkg/stream"
+	"github.com/julienschmidt/httprouter"
```
lmars: This should import from `Godeps`

lmars:
LGTM pending minor comments + CI

titanous:
PTAL. Updated to take advantage of the new `Stream` method of the http client.

lmars:
LGTM

discoverd2/server/http.go
heavenlyhash: This can be done after rebasing forward to grab #749 :D

titanous: #764

PR #735, opened 2015-01-14
dashboard: Show deploy logs in UI + minor fixes

jvatic:
screen shot 2015-01-14 at 3 50 52 pm

Note that only apps deployed through the dashboard with this version or later will have deploy logs visible in the dashboard UI.

### DIFF
--- a/dashboard/app/lib/javascripts/dashboard/routers/apps.js
+++ b/dashboard/app/lib/javascripts/dashboard/routers/apps.js
@@ -146,21 +146,25 @@ Dashboard.routers.Apps = Marbles.Router.createClass({
 	},
 
 	appLogs: function (params) {
-		params = params[0];
-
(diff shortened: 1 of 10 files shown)


titanous:
I think we should show at least timestamps in the logs list. Is there any other identifying information that would make sense?

lmars:
I don't think the `DOWN` label makes sense, maybe `Finished`?

jvatic:
We can add any identifying information we want to the job meta (such as the git ref). I think timestamps definitely make sense (for both sets of logs).

Yeah, the `DOWN` label is a bit odd in the context of deploy logs.

jvatic:
screen shot 2015-01-15 at 1 37 17 pm

jvatic:
@titanous, @lmars Comments addressed

titanous:
LGTM

PR #739, opened 2015-01-15
test: Fix vet complaints about formatting directives

titanous:


lmars:
LGTM

test/cluster/cluster.go
```diff
@@ -495,7 +495,7 @@ func (c *Cluster) DumpLogs(w io.Writer) {
 		return err
 	}
 	fallback := func() {
-		fmt.Fprintln(w, "\n*** Error getting job logs via flynn-host, falling back to tail log dump\n")
+		fmt.Fprintf(w, "\n*** Error getting job logs via flynn-host, falling back to tail log dump\n\n")
```
heavenlyhash: ...does `vet` just dislike the `Fprintln` function in general?  There's no format directives.

titanous: It got confused by the trailing newline. Instead of leaving the warning their I got rid of it by changing the call.

heavenlyhash:
LGTM

PR #746, opened 2015-01-16
discoverd2: Implement DNS server

titanous:
Incremental patch for #145.

This implements a DNS server backed by the discoverd database.

Queries that are not for the `discoverd.` domain will be proxied to a recursive nameservers if configured.

Each registered service exposes a number of domains:

#### Service

`SERVICE.discoverd.` returns randomly sorted A, AAAA, and SRV records for each instance.

#### RFC 2782

`_SERVICE._PROTO.discoverd.` returns randomly sorted A, AAAA, and SRV records for each instance, filtered by protocol.

#### Leader

`leader.SERVICE.discoverd.` returns single A or AAAA and SRV records pointing at the leader of the service.

#### Instance

`INSTANCE_ID.SERVICE._i.discoverd.` returns single A or AAAA and SRV records for a single instance. This domain pattern is typically only used as the target for SRV records (which include the port but not the IP address).

Lookups over UDP are limited to up to three instances and the response does not have the truncation bit set (which would trigger a subsequent lookup over TCP). TCP lookups return all instances. The weight and priority of SRV records is always 1. The response to SRV questions over TCP include extra A/AAAA records matching each target in the answer. The TTL is always 0.

### DIFF
--- a/discoverd2/server/dns.go
+++ b/discoverd2/server/dns.go
@@ -1,0 +1,388 @@
+package server
+
+import (
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/miekg/dns"
+	"github.com/flynn/flynn/discoverd2/client"
+	"github.com/flynn/flynn/pkg/random"
+)
+
+type DNSStore interface {
+	Get(string) []*discoverd.Instance
+	GetLeader(string) *discoverd.Instance
+}
+
+type DNSServer struct {
+	UDPAddr   string
+	TCPAddr   string
+	Store     DNSStore
+	Domain    string
+	Recursors []string
+
+	servers []*dns.Server
+}
+
+const maxUDPRecords = 3
+const dnsDomain = "discoverd."
+
+func (srv *DNSServer) ListenAndServe() error {
+	if srv.Store == nil {
+		panic("missing Store")
+	}
+	if srv.Domain == "" {
+		srv.Domain = dnsDomain
+	}
+	if err := srv.validateRecursors(); err != nil {
+		return err
+	}
+
+	api := dnsAPI{srv}
+	mux := dns.NewServeMux()
+	mux.HandleFunc(srv.Domain, api.ServiceLookup)
+	if len(srv.Recursors) > 0 {
+		mux.HandleFunc(".", api.Recurse)
+	}
+
+	errors := make(chan error, 4)
+	done := func() { errors <- nil }
+
+	if srv.UDPAddr != "" {
+		server := &dns.Server{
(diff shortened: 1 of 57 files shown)


discoverd2/server/dns.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	api := dnsAPI{srv}
+	mux := dns.NewServeMux()
+	mux.HandleFunc(srv.Domain, api.ServiceLookup)
+	if len(srv.Recursors) > 0 {
+		mux.HandleFunc(".", api.Recurse)
+	}
+
+	errors := make(chan error, 4)
```
heavenlyhash: 4?

titanous: Two `nil`s from `done()` plus two errors from `ListenAndServe`, probably overkill but I didn't want to risk leaking goroutines.

heavenlyhash: I guess this has something to do with 2 servers x 2 statuses, but I think this would make more sense as a channel per server, then selecting over them.

titanous: A channel per server would require more code, as there would be a different `NotifyStartedFunc` for each one, and a for loop wrapping a select, is it really that hard to follow the current code?

heavenlyhash: I didn't follow what sending the nils into the errors channel does?  It looks like the range just eats them, and isn't waiting on them to trigger anything?

titanous: There are two places where a send can occur for each server:

- From the `NotifyStartedFunc` (`nil`) - this indicates that the server has started correctly
- From the goroutine running `server.ListenAndServe()` - this indicates that the server failed to start or is stopping

If the server fails to start, `NotifyStartedFunc` will not be called, and we'll get an error sent on `errors`. The range receives `len(srv.servers)` times from `errors` and returns an error if one is received.

discoverd2/server/dns_test.go
```diff
@@ -0,0 +1,552 @@
...
+			domain: "leader.a.discoverd.",
+			data:   simpleData,
+			qs:     leaderQs,
+		},
+		{
+			name:   "v6 leader",
+			domain: "leader.a.discoverd.",
+			data:   v6v4Data[:1],
+			qs:     v6LeaderQs,
+		},
```
heavenlyhash: Do any of these test datasets hit the recursors?

(I guess that's a hard area to reach... if we can't swallow the network dependency, maybe mocked recursors?)

titanous: https://github.com/flynn/flynn/pull/746/files#diff-7b854941b569c25c17fd931d5e47c443R52

And they hit real servers as I wanted to be sure that it works.

heavenlyhash:
LGTM

Issue #744, opened 2015-01-15
Controller refactor


titanous:
- [x] Move from martini to httprouter (use [Context](https://godoc.org/golang.org/x/net/context) to pass around request-scoped data)
- [x] Validate requests using JSON schemas (#471)
- [x] Full input validation (#528)
- [ ] List pagination (#533)
- [x] Standardize error returns (status code + body) (discussed in #322)

titanous:
@lmars @archSeer If you have any suggestions about things that could be improved in the controller, please propose them here.

archSeer:
I think it'd be good if we got to use concrete types on repos (for instance `AppRepo.Add`), as that would do away with a lot of type casts. I don't know if it's possible though, due to the way the CRUD mapper works.

titanous:
I might be possible to use reflection with the CRUD mapper, I can take a look.

archSeer:
It's already using reflection in most of the calls, the real problem is going around the the interface validation we do [here](https://github.com/flynn/flynn/blob/24e4c1bce037ec88d3b8abe7059ddce7470cdada/controller/crud.go#L11-L15), since then every repository will have different function signatures.

archSeer:
One thing I was also looking into is https://github.com/flynn/flynn/commit/e2508323722173ebc3427a303d14fea13ef3c1b8, I'd like to do away with this and only have one common way to specify the controller credentials; less things to go wrong that way.

titanous:
I don't think the JS eventsource API allows setting headers.

archSeer:
Ah I see, figured there had to be a good reason for that.

Another thing I'd love is a more specialised `pkg/sse` method that allows setting the other event fields (`event`, `id`), so we can simplify cases like [these](https://github.com/flynn/flynn/blob/48b1a0d379056b810c4818a4601a583da54e3073/controller/jobs.go#L307) into something that's more similar to our other streams. Maybe something in the form of `sse.Send(&sse.Event{id: '1234', event: 'name', data: someInterface})`.

Would also be lovely if we can find more generalisations on the backend, for instance a `Stream(chan in, out http.ResponseWriter)` would greatly simplify all of our streaming.
```go
func (h *Host) streamEvents(id string, w http.ResponseWriter) error {
	ch := h.state.AddListener(id)
	go func() {
		<-w.(http.CloseNotifier).CloseNotify()
		h.state.RemoveListener(id, ch)
	}()
	enc := json.NewEncoder(sse.NewWriter(w))
[...]

PR #748, opened 2015-01-16
controller: Change CRUD helper to use httprouter instead of martini

jvatic:
refs #744

Still returns martini handler func for lookup
Moving towards using httphelper pkg for Error and JSON responses

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-martini/martini"
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/julienschmidt/httprouter"
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/martini-contrib/binding"
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/martini-contrib/render"
 	"github.com/flynn/flynn/controller/name"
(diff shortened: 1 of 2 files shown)


titanous:
Typo in the commit message `s/marini/martini`.

controller/controller.go
```diff
@@ -112,6 +114,31 @@ func (r *responseHelper) Error(err error) {
...
+		if err == ErrNotFound {
+			rw.WriteHeader(404)
+			return
+		}
+		log.Println(err)
+		respondWithJSON(rw, 500, struct{}{})
+	}
+}
+
+func respondWithJSON(rw http.ResponseWriter, status int, thing interface{}) {
```
titanous: We have very similar helpers in `pkg/httphelper`, let's standardize on those.

jvatic:
fixed typo

controller/controller.go
```diff
@@ -145,11 +172,17 @@ func appHandler(c handlerConfig) (http.Handler, *martini.Martini) {
...
+	httpRouter := httprouter.New()
+	httpRouter.NotFound = http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		m.ServeHTTP(res, req)
+	})
+
+	getAppMiddleware := crud("apps", ct.App{}, appRepo, httpRouter)
+	getReleaseMiddleware := crud("releases", ct.Release{}, releaseRepo, httpRouter)
+	getProviderMiddleware := crud("providers", ct.Provider{}, providerRepo, httpRouter)
+	crud("artifacts", ct.Artifact{}, artifactRepo, httpRouter)
+	crud("keys", ct.Key{}, keyRepo, httpRouter)
```
titanous: It feels like it makes more sense to pass `httpRouter` as the first parameter.

jvatic:
@titanous comments addressed*

*`httphelper.Error` doesn't currently handle `ErrNotFound` or `ct.ValidationError` so we'll need to standardize on something for those.

controller/controller.go
```diff
@@ -112,6 +115,19 @@ func (r *responseHelper) Error(err error) {
 	}
 }
 
+func respondWithError(w http.ResponseWriter, err error) {
```
titanous: Add a comment noting this is a temporary measure.

titanous:
LGTM pending minor comment

PR #749, opened 2015-01-16
httphelper: Standardize response JSON for errors

jvatic:
refs #744, #322

All errors respond with `code`, `message`, and optionally `details`

### DIFF
--- a/pkg/httphelper/httphelper.go
+++ b/pkg/httphelper/httphelper.go
@@ -2,18 +2,66 @@ package httphelper
 
 import (
 	"encoding/json"
+	"fmt"
 	"log"
 	"net/http"
 )
 
+type ErrorCode string
+
+const (
+	NotFoundError       ErrorCode = "not_found"
+	ObjectNotFoundError           = "object_not_found"
+	ObjectExistsError             = "object_exists"
+	SyntaxError                   = "syntax_error"
+	ValidationError               = "validation_error"
+	UnknownError                  = "unknown_error"
+)
+
+var errorResponseCodes = map[ErrorCode]int{
+	NotFoundError:       404,
+	ObjectNotFoundError: 404,
+	ObjectExistsError:   409,
+	SyntaxError:         400,
+	ValidationError:     400,
+	UnknownError:        500,
+}
+
+type JSONError struct {
+	Code    ErrorCode       `json:"code"`
+	Message string          `json:"message"`
+	Detail  json.RawMessage `json:"detail,omitempty"`
+}
+
+func (jsonError JSONError) Error() string {
+	return fmt.Sprintf("%s: %s", jsonError.Code, jsonError.Message)
+}
+
 func Error(w http.ResponseWriter, err error) {
+	var jsonError JSONError
 	switch err.(type) {
 	case *json.SyntaxError, *json.UnmarshalTypeError:
-		JSON(w, 400, "The provided JSON input is invalid")
+		jsonError = JSONError{
+			Code:    SyntaxError,
+			Message: "The provided JSON input is invalid",
+		}
+	case JSONError:
+		jsonError = err.(JSONError)
+	case *JSONError:
+		jsonError = *err.(*JSONError)
 	default:
 		log.Println(err)
-		JSON(w, 500, struct{}{})
+		jsonError = JSONError{
+			Code:    UnknownError,
+			Message: "Something went wrong",
+		}
 	}
+
+	responseCode, ok := errorResponseCodes[jsonError.Code]
+	if !ok {
+		responseCode = 500
+	}
+	JSON(w, responseCode, jsonError)
 }
 
 func JSON(w http.ResponseWriter, status int, v interface{}) {


pkg/httphelper/httphelper.go
```diff
@@ -6,14 +6,30 @@ import (
 	"net/http"
 )
 
-func Error(w http.ResponseWriter, err error) {
+type JSONError struct {
+	Code    int         `json:"code"`
```
titanous: I was thinking that `code` should be a string that can be used as a static identifier to compare against when error handling. Examples:

```text
syntax_error
object_exists
not_found
```

What I'm not sure about is whether we need to namespace them based on the component and then have generic codes as well.

jvatic: Makes sense. We could add an additional field for component if that's needed.

heavenlyhash: Agree with having a string, enumerated somewhere.

Perhaps we try to keep them generic APIerrors as long as we can and revist component namespaces later if we find overwhelming need?  So far the most major itches I've all seem to be general ones like "API not found" vs "ID not found".  "object exists" is also a good example (I think @archSeer metioned in another thread that that maps nicely on to http 409, but it a json chuck clarifying that the "conflict" is because "exists" makes total sense).

pkg/httphelper/httphelper.go
```diff
@@ -6,14 +6,30 @@ import (
...
+	Message string      `json:"message"`
+	Detail  interface{} `json:"detail,omitempty"`
```
heavenlyhash: How do we expect to use this `Detail` field?  Is it just a stringifiable, and the reciever is expected to be able to display it but probably not parse it?

I'm not sure if this will apply to the intended use pattern, but are you familiar with http://golang.org/pkg/encoding/json/#RawMessage ?  It can be useful when you have a "generic" field with a type hint field next to it; you can defer deserializing it on the first pass, then do a switch on the type hint to another deserialize call that expects a concrete type.

jvatic:
@titanous comment addressed

pkg/httphelper/httphelper.go
```diff
@@ -6,14 +6,69 @@ import (
...
+const (
+	NotFoundError ErrorCode = iota
+	ObjectExistsError
+	SyntaxError
+	ValidationError
+	ServerError
+)
+
+var errorStrings = map[ErrorCode]string{
+	NotFoundError:     "not_found",
```
heavenlyhash: Does it makes sense to add two types of not found?  i.e. to address this problem: https://github.com/flynn/flynn/pull/631/files#diff-e088955583d3e8787905eb3bdace3b49R88 .  Doing the host-rpc-to-http branch had a lot of head-on-desk debuging moments where I mistook a 404 for invalid route format when it was actually a misparsed ID, or vice versa.

Perhaps
```
NotFoundError
ObjectNotExistsError
```
(to parallel the `ObjectExistsError` name)

pkg/httphelper/httphelper.go
jvatic: The `Detail` field is expected to be any json serializable struct providing the API consumer with relevant information.

I don't think the `RawMessage` type is useful here as it would require consumers of `httphelper.Error` to first serialize their details object.

pkg/httphelper/httphelper.go
jvatic: I agree, distinguishing between "object not found" and "route not found" is important.

pkg/httphelper/httphelper.go
heavenlyhash: Yeah, you're right.  We'd end up having to write a factory method and so forth; I'm not a fan scattering map[] work around but in this case for presumably shallow structures, multi-tier serialization won't necessarily come out simplier.

pkg/httphelper/httphelper.go
```diff
@@ -6,14 +6,69 @@ import (
...
 
-func Error(w http.ResponseWriter, err error) {
+type ErrorCode int
+
+const (
+	NotFoundError ErrorCode = iota
+	ObjectExistsError
+	SyntaxError
+	ValidationError
+	ServerError
```
titanous: These constant should be strings, not untyped integers.

pkg/httphelper/httphelper.go
titanous: Yeah, it should be `json.RawMessage` otherwise we'll get a `map[string]interface{}` when we decode which sucks.

jvatic: Okay, yeah, given that this will be used on both client and server end of things, it's best to keep the complexity at the narrow end and use `json.RawMessage`.

jvatic:
@titanous @heavenlyhash comments addressed

pkg/httphelper/httphelper.go
```diff
@@ -6,14 +6,52 @@ import (
...
-func Error(w http.ResponseWriter, err error) {
+type ErrorCode string
+
+const (
+	NotFoundError       ErrorCode = "not_found"
+	ObjectNotFoundError           = "object_not_found"
+	ObjectExistsError             = "object_exists"
+	SyntaxError                   = "syntax_error"
+	ValidationError               = "validation_error"
+	ServerError                   = "server_error"
```
titanous: This should be `UnknownError`

pkg/httphelper/httphelper.go
```diff
@@ -6,14 +6,52 @@ import (
...
+	ServerError:         500,
+}
+
+type JSONError struct {
+	Code    ErrorCode       `json:"code"`
+	Message string          `json:"message"`
+	Detail  json.RawMessage `json:"detail,omitempty"`
+}
+
+func Error(w http.ResponseWriter, err interface{}) {
```
titanous: Implement `Error() string` on `JSONError` and we can return this to `err error`

pkg/httphelper/httphelper.go
```diff
@@ -6,14 +6,52 @@ import (
...
+	var jsonError JSONError
 	switch err.(type) {
 	case *json.SyntaxError, *json.UnmarshalTypeError:
-		JSON(w, 400, "The provided JSON input is invalid")
+		jsonError = JSONError{
+			Code:    SyntaxError,
+			Message: "The provided JSON input is invalid",
+		}
+	case JSONError:
```
titanous: Probably should handle `*JSONError` too

heavenlyhash:
LGTM pending that last rename suggestion.

pkg/httphelper/httphelper.go
```diff
@@ -6,14 +6,52 @@ import (
...
 	default:
 		log.Println(err)
-		JSON(w, 500, struct{}{})
+		jsonError = JSONError{
+			Code:    ServerError,
+			Message: "Server error",
+		}
 	}
+
+	responseCode := errorResponseCodes[jsonError.Code]
```
titanous: This should default to 500 if unknown.

jvatic:
comments addressed

titanous:
LGTM

PR #750, opened 2015-01-16
Volume management

heavenlyhash:
Add 'host/volume' package.

This volume system creates shared persistent storage areas, exposing this ability through an API on the host, and allows job creation to specify volumes for mounting into the job container.  The prototype implementation is backed by ZFS.  Not all features we desire from the final system are present, but the ones that are -- basic volume creation though the API and use of that within job config -- should be working.


host/volume/api/http.go
```diff
@@ -0,0 +1,111 @@
...
+
+	httphelper.JSON(w, 200, vol.Info())
+}
+
+func (api *HttpAPI) Inspect(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	vol := api.vman.GetVolume(ps.ByName("id"))
+	if vol == nil {
+		// TODO: produce a message that hints this is an id-not-found rather than youre-barking-up-the-wrong-api?
+		// Currently wouldn't matter; httpclient.RawReq helpers just return a predefined error val for 404's.
+		httphelper.JSON(w, 404, nil)
```
titanous: You can use the new error stuff that was just merged.

heavenlyhash: Yayy :3

host/libvirt_lxc_backend.go
```diff
@@ -423,6 +426,33 @@ func (l *LibvirtLXCBackend) Run(job *host.Job) (err error) {
...
+		}
+		if err := os.MkdirAll(filepath.Join(rootPath, v.Target), 0755); err != nil {
+			g.Log(grohl.Data{"at": "volume_mkdir", "dir": v.Target, "status": "error", "err": err})
+			return err
+		}
+		// REVIEW: strange though it may seem, it's actually looking like it'll make more sense to let
+		// the volume system tell the backend has produced our filesystem already, and since the backend
+		// already knows how to deal with the binds.
+		// Neither system wants to admit internals to the other, something's gotta give ¯\_(ツ)_/¯
+		volumePath, err := vol.Mount(job.ID, v.Target)
```
titanous: This looks fine, drop or rewrite the comment?

host/volume/api/http.go
```diff
@@ -0,0 +1,114 @@
...
+	"net/http"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/julienschmidt/httprouter"
+	"github.com/flynn/flynn/host/volume"
+	"github.com/flynn/flynn/host/volume/zfs"
+	"github.com/flynn/flynn/pkg/httphelper"
+	"github.com/flynn/flynn/pkg/random"
+)
+
+type HttpAPI struct {
```
titanous: s/HttpAPI/HTTPAPI/

host/volume/api/http.go
```diff
@@ -0,0 +1,114 @@
...
+
+func RegisterRoutes(api *HttpAPI, r *httprouter.Router) {
+	r.POST("/volume/provider", api.CreateProvider)
+	r.POST("/volume/provider/:provider_id/newVolume", api.Create)
+	r.PUT("/volume/instances/:id/snapshot", api.Snapshot)
+	r.GET("/volume/instances/:id/inspect", api.Inspect)
+}
+
+func (api *HttpAPI) CreateProvider(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	data, err := ioutil.ReadAll(r.Body)
```
titanous: Do this instead: `json.NewDecoder(r.Body).Decode...`

host/volume/api/http.go
```diff
@@ -0,0 +1,114 @@
...
+		var err error
+		if provider, err = zfs.NewProvider(config); err != nil {
+			httphelper.JSON(w, 500, err)
+			return
+		}
+	case "":
+		httphelper.JSON(w, 400, errors.New("host: volume provider kind must not be blank"))
+		return
+	default:
+		httphelper.JSON(w, 400, fmt.Errorf("host: volume provider kind '%s' is not known"))
```
titanous: These errors should use the `JSONError` system we are currently working towards standardizing.

host/volume/api/http.go
```diff
@@ -0,0 +1,114 @@
...
+type HttpAPI struct {
+	vman *volume.Manager
+}
+
+func NewHttpAPI(vman *volume.Manager) *HttpAPI {
+	return &HttpAPI{vman: vman}
+}
+
+func RegisterRoutes(api *HttpAPI, r *httprouter.Router) {
+	r.POST("/volume/provider", api.CreateProvider)
```
titanous: Do we need/want this endpoint?

host/volume/api/http.go
```diff
@@ -0,0 +1,114 @@
...
+	r.POST("/volume/provider/:provider_id/newVolume", api.Create)
+	r.PUT("/volume/instances/:id/snapshot", api.Snapshot)
+	r.GET("/volume/instances/:id/inspect", api.Inspect)
```
titanous: I think this API should be something along these lines (not totally sold on `volume_providers`)

```text
GET  /volume_providers - list volume providers
GET  /volume_providers/:provider_id/volumes - list volumes
POST /volume_providers/:provider_id/volumes - create volume
GET  /volume_providers/:provider_id/volumes/:volume_id - get volume details
POST /volume_providers/:provider_id/volumes/:volume_id/snapshots - create snapshot
```

host/volume/backend.go
```diff
@@ -0,0 +1,29 @@
+package volume
+
+import (
+	"encoding/json"
+)
+
+/*
+	Provider defines the interface of a, well, a provider of volumes.
```
titanous: Tone down the conversationality.

host/volume/manager.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// `map[providerName]provider`
+	//
+	// It's possible to configure multiple volume providers for a flynn-host daemon.
+	// This can be used to create volumes using providers backed by different storage resources,
+	// or different volume backends entirely.
+	providers map[string]Provider
+
+	// `map[volume.Id]volume`
+	volumes map[string]Volume
```
titanous: Do we need some mutexes here?

host/volume/manager.go
```diff
@@ -0,0 +1,89 @@
...
+func (m *Manager) GetVolume(id string) Volume {
+	return m.volumes[id]
+}
+
+/*
+	Proxies `volume.Provider` while making sure the manager remains
+	apprised of all volume lifecycle events.
+
+	@implements Provider
+*/
```
titanous: Take a look at these comments in the godoc web interface to confirm that they render sanely

host/volume/volume_test.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type S struct{}
+
+var _ = Suite(&S{})
+
+func (S) TestThings(c *C) {
+}
```
titanous: Remove this?

host/volume/zfs/helpers_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return false, err.Error()
+	}
+	defer dir.Close()
+	fileinfos, err := dir.Readdir(-1)
+	if err != nil {
+		return false, err.Error()
+	}
+
+	actualFilenames := make([]string, 0)
```
titanous: `make([]string, len(fileinfos))` then use `actualFilenames[i]` to assign below.

host/volume/zfs/integrations.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+	"strings"
+
+	gzfs "github.com/flynn/flynn/Godeps/_workspace/src/github.com/mistifyio/go-zfs"
+)
+
+type Logger struct{}
+
+func (*Logger) Log(msg []string) {
+	fmt.Printf("[zfs] %s\n", strings.Join(msg, " ")) // TODO replace with log15
```
titanous: `log15` is in the tree now

host/volume/zfs/zfs.go
```diff
@@ -0,0 +1,181 @@
...
+
+func NewProvider(config *ProviderConfig) (volume.Provider, error) {
+	if _, err := exec.LookPath("zfs"); err != nil {
+		return nil, fmt.Errorf("zfs command is not available")
+	}
+	dataset, err := zfs.GetDataset(config.DatasetName)
+	if err != nil {
+		err = eunwrap(err)
+		switch err.Error() {
+		case fmt.Sprintf("cannot open '%s': dataset does not exist\n", config.DatasetName):
```
titanous: Can we just match against `dataset does not exist` instead of the exact string? This seems like it could break across versions of the zfs userspace tools pretty easily.

host/volume/zfs/integrations.go
heavenlyhash: How should we tag output and at what level?  This gets called for every zfs command go-zfs invokes, so probably trace... I'd like to have a way to tag it with which provider invoked it but I don't see a ready way to set a logger per provider without doing some very janky mutexes and "context" objects.

Alternatively, we may not have much need for this at all in the final code.

host/volume/api/http.go
heavenlyhash: Unsure.  It's fairly inexpensive in code complexity to provide it since it's effectively the same code as needed in host bootstrap, so I tilted towards inclusion.

host/volume/api/http.go
heavenlyhash: So the big question here is, do we want the volume to visibly maintain its binding to the provider?  I'd argue no: the IDs are unique on their own, and so fit in a top level grouping of their own; and we should be able to provide the same operations regardless of backend, so it should matter.  Thoughts?

archSeer: I agree with having volumes under the top level, however I'd still follow the shape of the rest of the API and make the operations more RESTful:

```
GET /volumes/providers - list providers
POST /volumes/providers - create provider
GET  /volumes/providers/:provider_id/volumes - list volumes for provider
POST /volumes/providers/:provider_id/volumes - create volume
GET  /volumes/:volume_id - get volume details
PUT  /volumes/:volume_id/snapshots - create snapshot
```

Namespacing providers under `/volumes` directly might end up a bit ugly, so we could top-level them and namespace the entire API under `/data` or `/storage` (`/volume/volumes/:id` seems a bit ugly) -- `GET /storage/providers`:

```
GET  /storage/providers - list providers
POST /storage/providers - create provider
GET  /storage/providers/:provider_id/volumes - list volumes for provider
POST /storage/volumes/providers/:provider_id/volumes - create volume
GET  /storage/volumes/:volume_id - get volume details
PUT  /storage/volumes/:volume_id/snapshots - create snapshot
```

archSeer: Now that I think about it, I think a `/storage` namespace makes a lot of sense, since we already namespace other calls under `/host` and `/cluster`.

host/volume/zfs/integrations.go
titanous: I'd just swap this out for context-less Debug-level calls with a `package=zfs` or something annotation.

host/volume/api/http.go
titanous: Yeah, I like the `/storage` prefix. I'm fine with not namespacing volumes under providers, the layout proposed by @archSeer looks good.

heavenlyhash: Very much like that storage layout.  Adopted.

heavenlyhash:
@titanous @archSeer I believe all comments addressed... though the approach to testing needs still some consideration before we can merge this, due to use of root.

host/types/types.go
```diff
@@ -86,6 +87,14 @@ type Mount struct {
 	Writeable bool   `json:"writeable,omitempty"`
 }
 
+type VolumeBinding struct {
+	// Target defines the filesystem path inside the container where the volume will be mounted.
+	Target string
+	// VolumeID can be thought of as the source path if this were a simple bind-mount.  It is resolved by a VolumeManager.
+	VolumeID  string
+	Writeable bool
```
titanous: These fields need json tags.

titanous:
This will also need installation docs changes that explain how to install ZFS.

archSeer:
LGTM, pending the JSON comment.

heavenlyhash:
Rebased to latest master, yanked commits working on cross-host-reboot persistence for a future branch.

Since host reboots cause zfs to forget about file vdevs as written, detection of existing zpools will be required before this feature is really usable; the behavior here is just to stomp over it to make sure flynn-host startup doesn't get hung up.

host/volume/api/http.go
```diff
@@ -0,0 +1,127 @@
...
+	case "":
+		httphelper.Error(w, httphelper.JSONError{
+			Code:    httphelper.ValidationError,
+			Message: fmt.Sprintf("volume provider 'kind' field must not be blank"),
+		})
+		return
+	default:
+		httphelper.Error(w, httphelper.JSONError{
+			Code:    httphelper.ValidationError,
+			Message: fmt.Sprintf("volume provider kind %q is not known"),
```
titanous: missing sprintf param, probably worth a `go vet`

host/volume/api/http.go
```diff
@@ -0,0 +1,127 @@
...
+	httphelper.JSON(w, 200, vol.Info())
+}
+
+func (api *HTTPAPI) List(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	vols := api.vman.Volumes()
+	volList := make([]*volume.Info, len(vols))
+	i := 0
+	for _, v := range vols {
+		volList[i] = v.Info()
+		i++
```
titanous: I prefer `append`, it's clearer and has no alloc cost when you allocate enough capacity ahead of time.

host/volume/api/http.go
```diff
@@ -0,0 +1,127 @@
...
+
+type HTTPAPI struct {
+	vman *volume.Manager
+}
+
+func NewHTTPAPI(vman *volume.Manager) *HTTPAPI {
+	return &HTTPAPI{vman: vman}
+}
+
+func RegisterRoutes(api *HTTPAPI, r *httprouter.Router) {
```
archSeer: I'd make this a `func (api *HTTPAPI) RegisterRoutes(r *httprouter.Router)`, like in Sampi.

host/volume/zfs/zfs_basics_test.go
```diff
@@ -0,0 +1,142 @@
...
+var _ = Suite(&S{})
+
+func (S) SetUpSuite(c *C) {
+	// Skip all tests in this suite if not running as root.
+	// Many zfs operations require root priviledges.
+	skipIfNotRoot(c)
+}
+
+func (S) TestSnapshotShouldCarryFiles(c *C) {
+	err := WithTmpfileZpool("testpool", func() error {
```
titanous: Since we do this in every test in this file, make a suite and do the setup in `SetUpTest` and the teardown in `TearDownTest`.

host/volume/zfs/zpool.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"io/ioutil"
+	"math"
+	"os"
+
+	zfs "github.com/flynn/flynn/Godeps/_workspace/src/github.com/mistifyio/go-zfs"
+)
+
+func WithTmpfileZpool(poolName string, fn func() error) error {
```
titanous: Should this be in the test code or do we use it somewhere else?

host/volume/api/http.go
heavenlyhash: Fixed instance; what's the invocation I should use to find more?  `go vet ./host/...` didn't emit any warnings for me.  Checked with `-x` and it was covering this (warn-worth) file.

titanous: Use `go tool vet`, `go vet` is kinda sketchy.

host/volume/zfs/zfs.go
```diff
@@ -0,0 +1,178 @@
...
+			if _, err = zfs.CreateZpool(
+				config.DatasetName,
+				nil,
+				"-f",     // force.  TODO: as part of reboot persistence, this will be removed
+				"-mnone", // do not mount the root dataset.  (we'll mount our own datasets as necessary.)
+				config.Make.BackingFilename,
+			); err != nil {
+				return nil, err
+			}
+			// get the datazet again... `zfs.Zpool` isn't a `zfs.Dataset`
```
titanous: s/datazet/dataset/

test/test_host.go
```diff
@@ -96,3 +96,105 @@ func (s *HostSuite) TestNetworkedPersistentJob(t *c.C) {
...
+	// most of the volume tests (snapshotting, quotas, etc) are unit tests under their own package.
+	// these tests exist to cover the last mile where volumes are bind-mounted into containers.
+
+	cluster := s.clusterClient(t)
+
+	// create a volume!
+	vol, err := s.anyHostClient(t).CreateVolume("default")
+	t.Assert(err, c.IsNil)
+
+	// FIXME: yes, there's massive code duplication here.  boil it down later.
```
titanous: How much work to fix this now?

titanous:
LGTM pending minor comments.

test/test_host.go
heavenlyhash: Extracted the repeated test job creation components and the "api" call to the test shell into functions.  Much better now.

host/types/types.go
```diff
@@ -68,12 +68,39 @@ type ContainerConfig struct {
...
+	}
+	if y.Cmd != nil {
+		x.Cmd = y.Cmd
+	}
+	for k, v := range y.Env {
+		x.Env[k] = v
+	}
+	x.Mounts = append(x.Mounts, y.Mounts...)
+	x.Volumes = append(x.Volumes, y.Volumes...)
+	x.Ports = append(x.Ports, y.Ports...)
```
titanous: These slices/maps should probably be copied.

test/test_host.go
```diff
@@ -73,26 +78,114 @@ func (s *HostSuite) TestNetworkedPersistentJob(t *c.C) {
...
+
+func (s *HostSuite) TestVolumeCreation(t *c.C) {
+	h := s.anyHostClient(t)
+
+	vol, err := h.CreateVolume("default")
+	t.Assert(err, c.IsNil)
+	t.Assert(vol.ID, c.Not(c.Equals), "")
+}
+
+func (s *HostSuite) TestVolumeCreationFailsForNonexistantProvider(t *c.C) {
```
titanous: s/Nonexistant/Nonexistent/

test/test_host.go
```diff
@@ -73,26 +78,114 @@ func (s *HostSuite) TestNetworkedPersistentJob(t *c.C) {
...
+	h := s.anyHostClient(t)
+
+	vol, err := h.CreateVolume("non-existent")
+	t.Assert(err, c.NotNil)
+	t.Assert(vol.ID, c.Equals, "")
```
titanous: This assert shouldn't happen, as you're not guaranteed that the first param is not nil if there is a non-nil err return.

host/types/types.go
heavenlyhash: I don't think that needs any extra effort here since we're handling the config by value instead of reference.  I always second guess myself about how function calls by value isolate things in golang, so I made a playground example doing similar... http://play.golang.org/p/h2iRBxw8z2

heavenlyhash: I stand corrected... that example only works because the append call happens to be allocating a new copy.  Here's an example with capacity preallocated which behaves differently: http://play.golang.org/p/5KyiyTKSTK

titanous:
LGTM

PR #754, opened 2015-01-17
discoverd2: Add healthcheck package

titanous:
Incremental patch for #145.

This adds a simple package that can check the health of HTTP(S) and TCP endpoints, it will be used to automatically register/unregister services (probably from containerinit). The tests cover 100% of the statements.

### DIFF
--- a/discoverd2/health/check.go
+++ b/discoverd2/health/check.go
@@ -1,0 +1,103 @@
+package health
+
+import (
+	"bytes"
+	"crypto/tls"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"time"
+)
+
+const defaultTimeout = 2 * time.Second
+
+type Check interface {
+	Check() error
+}
+
+var _ Check = &TCPCheck{}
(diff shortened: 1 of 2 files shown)


archSeer:
LGTM

discoverd2/health/check.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if res.StatusCode != expectedStatus {
+		return fmt.Errorf("healthcheck: expected HTTP status %d, got %d", expectedStatus, res.StatusCode)
+	}
+	if len(c.MatchBytes) == 0 {
+		return nil
+	}
+
+	var buf bytes.Buffer
+	if _, err := buf.ReadFrom(io.LimitReader(res.Body, 5120)); err != nil {
```
heavenlyhash: Add a quick comment to the struct that mentioned any match bytes need to be found in the first few k of the page?

titanous: Sure, will add.

PR #757, opened 2015-01-17
Fix unit test env in devbox

### LINKED ISSUE #689: Fix unit tests in dev box
The unit tests depend on the `discoverd` and `etcd` binaries. Currently they are not in the `PATH` even after compiling Flynn. The following steps should be taken to solve this:

- [ ] Download/extract etcd outside of the docker build. [Here's a commit](https://gist.github.com/titanous/2da2140d0b5e745e490e) I made a while ago that does this.
- [ ] Add the `discoverd`/`etcd` build directories to the `PATH` from the `Vagrantfile`, this will allow the unit tests to find the binaries
- [ ] Add make targets for running the integration tests, unit tests, and both

heavenlyhash:
Fixes #689

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,4 +15,13 @@ release:
 clean:
 	git clean -Xdf -e '!.tup' -e '!.vagrant' -e '!script/custom-vagrant'
 
-.PHONY: all clean dev release
+test: test-unit test-integration
+
+test-unit:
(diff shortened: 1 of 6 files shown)


titanous:
LGTM

archSeer:
LGTM

PR #759, opened 2015-01-18
discoverd2: Implement health monitor

titanous:
Incremental patch for #145.

This builds on the health check capabilities to provide a monitor that runs checks on a regular basis and signals a channel when a check changes state.

### DIFF
--- a/discoverd2/health/monitor.go
+++ b/discoverd2/health/monitor.go
@@ -1,0 +1,114 @@
+package health
+
+import (
+	"time"
+
+	"github.com/flynn/flynn/pkg/stream"
+)
+
+type MonitorConfig struct {
+	// StartInterval is the check interval to use when waiting for the service
+	// to transition from created -> up. It defaults to 100ms.
+	StartInterval time.Duration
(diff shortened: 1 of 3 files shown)


archSeer:
Very neat!

discoverd2/health/monitor.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"time"
+
+	"github.com/flynn/flynn/pkg/stream"
+)
+
+type MonitorConfig struct {
+	// StartInterval is the check interval to use when waiting for the service
+	// to transition from created -> up. It defaults to 100ms.
+	StartInterval time.Duration
```
archSeer: So if an app requires longer to start than 100ms, it will be marked as down until the next check 2 seconds later? Seems a bit penalising, but I can't think of a better solutio n.

archSeer: Ah nevermind, I misread, just checked the implementation.

archSeer:
LGTM.

PR #760, opened 2015-01-19
host: Pass config to containerinit via JSON

titanous:
Instead of passing the configuration via arguments, we write it to a JSON file. We were already using JSON for environment variables, and putting the arguments in a struct, so this patch mostly just renames fields.

### DIFF
--- a/host/containerinit/init.go
+++ b/host/containerinit/init.go
@@ -17,7 +17,6 @@ package containerinit
 import (
 	"encoding/json"
 	"errors"
-	"flag"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -26,7 +25,6 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
-	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


host/containerinit/init.go
```diff
@@ -349,37 +347,37 @@ func setupNetworking(args *ContainerInitArgs) error {
...
+	if err := setupNetworking(c); err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func getEnv(args *ContainerInitArgs, key string) string {
-	for _, kv := range args.env {
+func getEnv(c *Config, key string) string {
```
archSeer: We could now change the `ENV` field to be a `map[string]string`, which would then get rid of `getEnv` and [`writeContainerEnv` on the backend](https://github.com/flynn/flynn/blob/75d6dcf8883eb1ae049483353b1013a408585a6e/host/libvirt_lxc_backend.go#L146).

host/libvirt_lxc_backend.go
```diff
@@ -446,58 +446,53 @@ func (l *LibvirtLXCBackend) Run(job *host.Job) (err error) {
...
 		if len(job.Config.Cmd) > 0 {
-			args = append(args, job.Config.Cmd...)
+			config.Args = append(config.Args, job.Config.Cmd...)
 		} else {
-			args = append(args, imageConfig.Cmd...)
+			config.Args = append(config.Args, imageConfig.Cmd...)
 		}
 	}
 
+	g.Log(grohl.Data{"at": "write_env"})
```
archSeer: This should now be `write_config`

titanous: Fixed.

host/containerinit/init.go
titanous: We need it to be a `[]string` when we pass it to `os/exec` so merging here is fine.

archSeer:
LGTM

heavenlyhash:
LGTM

PR #761, opened 2015-01-19
pkg/httpclient: Consolidate HTTP hijacking.

archSeer:
Since we're using an uniform client for doing http requests, we can provide a
convenience method for hijacking. This also standardises hijacking -- the attach
endpoint used status 200, but http.SwitchingProtocols seems more appropriate.

Whereas Stream is defined as a SSE stream of JSON objects, Hijack is more
general and unspecified, allowing for custom protocols where SSE would not work
well.

### DIFF
--- a/pkg/httpclient/json.go
+++ b/pkg/httpclient/json.go
@@ -1,12 +1,14 @@
 package httpclient
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net"
 	"net/http"
+	"net/http/httputil"
 	"net/url"
 	"strings"
 
@@ -15,6 +17,16 @@ import (
 )
 
 type DialFunc func(network, addr string) (net.Conn, error)
+
+type writeCloser interface {
+	io.WriteCloser
+	CloseWrite() error
+}
+
+type ReadWriteCloser interface {
+	io.ReadWriteCloser
+	CloseWrite() error
+}
 
 type Client struct {
 	ErrNotFound error
@@ -38,7 +50,7 @@ func ToJSON(v interface{}) (io.Reader, error) {
 	return bytes.NewBuffer(data), err
 }
 
-func (c *Client) RawReq(method, path string, header http.Header, in, out interface{}) (*http.Response, error) {
+func (c *Client) prepareReq(method, path string, header http.Header, in interface{}) (*http.Request, error) {
 	var payload io.Reader
 	switch v := in.(type) {
 	case io.Reader:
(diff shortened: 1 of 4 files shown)


archSeer:
Note that I wasn't able to get rid of `utils.HijackRequest` yet, as a [controller test references it](https://github.com/flynn/flynn/blob/0f87379eb4e0fe2146ef774b92a6c8ff53feb01e/controller/jobs_test.go#L297-L303). I think the controller tests should have the [utility methods](https://github.com/flynn/flynn/blob/0f87379eb4e0fe2146ef774b92a6c8ff53feb01e/controller/controller_test.go#L77-L137) replaced with `httpclient`, then we can replace that piece of code with a call to `httpclient.Hijack`. That would also unify all of our API calls to use `httpclient`, one drawback however is that the tests expect `http.Response` objects; in most cases we use them to simply check the status code, which could be replaced with checking for a specific error (`t.Assert(err, c.IsNil)` for 200, `t.Assert(err, Equals, client.NotFoundError)` for 404).

archSeer:
@flynn/devs needs a review.

pkg/httpclient/json.go
```diff
@@ -15,6 +17,16 @@ import (
...
 
+type writeCloser interface {
+	io.WriteCloser
+	CloseWrite() error
+}
+
+type ReadWriteCloser interface {
+	io.ReadWriteCloser
+	CloseWrite() error
+}
```
titanous: Is this a duplicate of `utils.ReadWriteCloser`?

pkg/httpclient/json.go
```diff
@@ -97,6 +117,51 @@ func (c *Client) RawReq(method, path string, header http.Header, in, out interfa
...
+	res, err := clientconn.Do(req)
+	if err != nil && err != httputil.ErrPersistEOF {
+		return nil, err
+	}
+	if res.StatusCode != http.StatusSwitchingProtocols {
+		res.Body.Close()
+		return nil, &url.Error{
+			Op:  req.Method,
+			URL: req.URL.String(),
+			Err: fmt.Errorf(c.ErrPrefix+": unexpected status %d", res.StatusCode),
```
titanous: `ErrPrefix` has been removed in master.

pkg/httpclient/json.go
archSeer: Yes, since I am essentially trying to get rid of `utils/hijack.go`.

pkg/httpclient/json.go
archSeer: Fixed.

pkg/httpclient/json.go
titanous: Ah, I see.

titanous:
LGTM

PR #764, opened 2015-01-19
discoverd2: Use new standardized error types for HTTP responses

titanous:


discoverd2/server/http.go
```diff
@@ -75,23 +75,21 @@ type httpAPI struct {
 	Store Datastore
 }
 
-func jsonError(w http.ResponseWriter, code int, err error) {
-	httphelper.JSON(w, code, struct {
-		Error string `json:"error"`
-	}{err.Error()})
+func jsonError(w http.ResponseWriter, code hh.ErrorCode, err error) {
+	hh.Error(w, hh.JSONError{Code: code, Message: err.Error()})
```
heavenlyhash: Should we maybe just fold this error message unwrapping behavior into the `hh.Error` function instead of the "Something went wrong" default text?

titanous: For unknown errors we can't expose the error text to clients as it could contain sensitive information. I'm not passing any unknown errors here. Having default text for the error codes would be useful, as would having a method that didn't require building a `JSONError` first.

heavenlyhash:
LGTM

PR #765, opened 2015-01-19
pkg/httpclient: Decode JSON errors

titanous:
If a non-200 response has an `application/json` Content-Type and decodes successfully into a `JSONError`, return it. If not, fallback to the old behavior of returning generic errors.


lmars:
LGTM pending CI

PR #768, opened 2015-01-20
appliance: Delete unused mongodb appliance

titanous:
This is completely untested and does not currently ship as part of the standard distribution. It will be resurrected later.


lmars:
LGTM

Issue #267, opened 2014-09-28
Allow specifying wildcard certificates in router


titanous:
I should be able to load the router with wildcard certificates (and perhaps certs with SANs) and have them automatically applied to matching HTTP routes.

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

PR #769, opened 2015-01-20
router: add default tls keypair to server

benburkert:
This is an initial attempt at implementing #267. I'm still getting up to speed on the code base so I'd love to get some feedback on this approach & any style issues.

The keypair passed into the new `HTTPListener` is used as the default `tls.Certificate` for SNI & non-SNI clients alike. This is handled by `tls.Config`'s `NameToCertificate` map, which is populated in the `tls.Server` call.

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -47,13 +47,14 @@ type HTTPListener struct {
 	tlsListener net.Listener
 	closed      bool
 	cookieKey   *[32]byte
+	keypair     tls.Certificate
 }
 
 type DiscoverdClient interface {
 	NewServiceSet(string) (discoverd.ServiceSet, error)
 }
 
-func NewHTTPListener(addr, tlsAddr string, cookieKey *[32]byte, ds DataStore, discoverdc DiscoverdClient) *HTTPListener {
+func NewHTTPListener(addr, tlsAddr string, cookieKey *[32]byte, keypair tls.Certificate, ds DataStore, discoverdc DiscoverdClient) *HTTPListener {
 	l := &HTTPListener{
 		Addr:      addr,
 		TLSAddr:   tlsAddr,
@@ -64,6 +65,7 @@ func NewHTTPListener(addr, tlsAddr string, cookieKey *[32]byte, ds DataStore, di
 		services:  make(map[string]*httpService),
 		wm:        NewWatchManager(),
 		cookieKey: cookieKey,
+		keypair:   keypair,
 	}
 	if cookieKey == nil {
 		var k [32]byte
@@ -288,7 +290,7 @@ func fail(sc *httputil.ServerConn, req *http.Request, code int, msg string) {
 var errMissingTLS = errors.New("router: route not found or TLS not configured")
 
 func (s *HTTPListener) handle(conn net.Conn, isTLS bool) {
-	defer conn.Close()
+	defer conn.Close() // TODO(benburkert): for TLS, check that closing this conn sends the TLS "close notify"
 
 	var r *httpRoute
 
@@ -298,14 +300,11 @@ func (s *HTTPListener) handle(conn net.Conn, isTLS bool) {
 			if r == nil {
 				return nil, errMissingTLS
 			}
-			if r.keypair == nil {
-				return nil, errMissingTLS
-			}
 			return r.keypair, nil
 		}
 		conn = tls.Server(conn, tlsconfig.SecureCiphers(&tls.Config{
 			GetCertificate: certForHandshake,
-			Certificates:   []tls.Certificate{{}},
+			Certificates:   []tls.Certificate{s.keypair},
 		}))
 	}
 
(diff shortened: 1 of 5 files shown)


router/server.go
```diff
@@ -67,9 +68,19 @@ func main() {
 	tcpIP := flag.String("tcpip", "", "tcp router listen ip")
 	tcpRangeStart := flag.Int("tcp-range-start", 3000, "tcp port range start")
 	tcpRangeEnd := flag.Int("tcp-range-end", 3500, "tcp port range end")
+	certFile := flag.String("tlscert", "", "TLS (SSL) cert file in pem format")
+	keyFile := flag.String("tlskey", "", "TLS (SSL) key file in pem format")
```
titanous: This is good for anyone running the router standalone, but won't work inside of Flynn as there's no good way to inject config files into a container currently. We should support environment variables as well (which is what we're currently using for secrets in lieu of a more formal/secure system).

router/http.go
```diff
@@ -47,13 +47,14 @@ type HTTPListener struct {
...
 	cookieKey   *[32]byte
+	keypair     tls.Certificate
 }
 
 type DiscoverdClient interface {
 	NewServiceSet(string) (discoverd.ServiceSet, error)
 }
 
-func NewHTTPListener(addr, tlsAddr string, cookieKey *[32]byte, ds DataStore, discoverdc DiscoverdClient) *HTTPListener {
+func NewHTTPListener(addr, tlsAddr string, cookieKey *[32]byte, keypair tls.Certificate, ds DataStore, discoverdc DiscoverdClient) *HTTPListener {
```
titanous: This function is getting pretty overloaded. What do you think about moving to setting fields on the struct directly and then do a lazy init in `Start`?

titanous:
This looks good!

titanous:
I think a second component of #267 is allowing an arbitrary set of global wildcard and SAN certs to be used as a fallback for SNI connections, but that's much lower priority.

router/server.go
benburkert: gotcha, any concerns with abusing `flag.String` to fallback to the environment variable?

```go
flag.String("tlscert", os.Getenv("TLSCERT"), "...")
```

benburkert: also, `tlscert`/`tlskey` was the first name that popped into my head, lemme know if there's a better term for these.

router/http.go
benburkert: sounds good. :+1:

router/server.go
titanous: Will that fallback work? The environment variable will contain the data, but the flag is supposed to be a file path. The flag names are fine, but the description should note that they are for the default certificate presented by the HTTP listener (not the API, which is also HTTP).

benburkert: ahh, the environment variables should the PEM encoded data, not the file path, gotcha.

titanous:
FYI each commit needs a [DCO sign-off](https://flynn.io/docs/contributing#developer’s-certificate-of-origin) to make Travis happy.

titanous:
Oh, one other note that hasn't gotten added to the CONTRIBUTING document yet is that we are trying to add a bit of context to each commit message, as described here: https://groups.google.com/d/msg/golang-dev/6M4dmZWpFaI/SyU5Sl4zZLYJ

benburkert:
I added a test for a SAN domain (in the default keypair, not a route's keypair) and changed the listeners to use lazy init in the `Start` methods. The environment variable stuff still needs to be done, but this is ready for another :eyes:.

router/http.go
```diff
@@ -88,6 +68,27 @@ func (s *HTTPListener) Close() error {
...
+	if s.Watcher != nil {
+		return errors.New("http listener already started")
+	}
+	if s.wm == nil {
+		s.wm = NewWatchManager()
+	}
+	s.Watcher = s.wm
+
+	if s.ds == nil {
+		return errors.New("http listener missing data store")
```
titanous: Per convention, these errors should start with `router: `.

router/server.go
```diff
@@ -106,9 +117,23 @@ func main() {
...
+			ds:        NewEtcdDataStore(etcdc, path.Join(prefix, "tcp/")),
+			discoverd: d,
+		},
+		HTTP: &HTTPListener{
+			Addr:      *httpAddr,
+			TLSAddr:   *httpsAddr,
+			cookieKey: cookieKey,
+			keypair:   keypair,
+			ds:        NewEtcdDataStore(etcdc, path.Join(prefix, "http/")),
+			discoverd: d,
```
titanous: What do you think about 'exporting' these fields to distinguish them from the internal bookkeeping fields?

benburkert: I agree that it's awkward, but i'd rather save this for a future refactor.

benburkert:
This is ready for another :eyes:

router/server.go
titanous: K, that's fine.

titanous:
LGTM.

Please squash c13579302141aecbac7f6fa8f2f94236baeed7a0, and it looks like this needs a rebase to make travis happy.

titanous:
:sunrise:

PR #772, opened 2015-01-20
appliance/postgresql: Temporarily disable follower mode

titanous:
This appliance will be rewritten very shortly, and we don't currently use follower mode, as it is prone to split-brain.

Disabling follower mode makes landing discoverd2 easier, so this patch changes the server to exit if there is already a leader.


lmars:
LGTM

PR #775, opened 2015-01-20
router: Remove fake etcd/discoverd from tests

### LINKED ISSUE #771: router: remove fake test mode
The router test suite currently has a fake version of etcd/discoverd to optionally use instead of starting them before each test. While this is way faster, it probably doesn't make sense to maintain the code. Unless we have a compelling reason to keep it, the 'fake' code should be removed.

/cc @benburkert @bgentry

titanous:
This code is brittle as it makes many assumptions about the behavior of etcd/discoverd, which may not be correct. It was originally added to make the tests faster, but inaccurate tests are useless, no matter how fast.

Closes #771

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,6 @@ install:
 
 script:
   - go test -race -cover ./...
-  - go test -race ./router -fake=false
 
 notifications:
   irc:
--- a/router/api_test.go
+++ b/router/api_test.go
@@ -25,9 +25,7 @@ func newTestAPIServer(t etcdrunner.TestingT) *testAPIServer {
 		cleanup:   []func(){killDiscoverd, killEtcd},
 	}
 
-	discoverd := newFakeDiscoverd()
-	discoverd.Register("router-api", ts.Listener.Addr().String())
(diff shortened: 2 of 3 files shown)


lmars:
LGTM

bgentry:
FYI this broke tests locally for me until I manually did a `go install` on discoverd. Saw the following in my router test logs with `DEBUG=1`:

```
/Users/bgentry/go/src/github.com/flynn/flynn/discoverd/testutil/testutil.go:34:
    t.Fatal("discoverd start failed: ", err)
... Error: discoverd start failed: exec: "discoverd": executable file not found in $PATH
```

Should I be running the router tests via `make` instead of `go test` ?

titanous:
The new `make test-unit` target should work inside the dev VM, but will run all of the tests, not just the router tests. As long as you have discoverd/etcd in your `PATH`, `go test` is fine (and works on OS X as well as Linux).

bgentry:
Running the tests with `go test` does "work", although it seems to be execing etcd and discoverd for each individual test, which means the router tests now take ~26s locally :(

PR #778, opened 2015-01-20
router: make domain name lookup case insensitive

benburkert:
Some http clients do not down-case the hostname in the Host header of the HTTP request and the servername in the TLS client-hello. Matching route domains in a case-insensitive way prevents 404 and TLS handshake errors for such clients. The request's Host header is not altered and keeps the original casing.


titanous:
LGTM.

PR #762, opened 2015-01-19
controller: Refactor to use httprouter instead of martini

jvatic:

Issue #755, opened 2015-01-17
HTTP request logging


titanous:
We should have a standard HTTP request logging middleware that logs at the start and end of every request, and adds a `log15` logging context to a [`Context`](https://godoc.org/golang.org/x/net/context) that is then passed through to every method called out of the handler. This way every log line will have context about the request (probably a random request ID).

Calls to `httphelper.Error` should use the same context and log the error.

See review comments at the end of #745 for more context.

titanous:
This was done a while ago.

PR #780, opened 2015-01-20
httphelper: Add handler for request logging

jvatic:


jvatic:
This currently exposes the same functionality as the martini logging middleware, and should be expanded to satisfy #755

titanous:
This should use log15 to do key=value logging

pkg/httphelper/request_logger.go
```diff
@@ -0,0 +1,23 @@
...
+
+	log "github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
+	"github.com/flynn/flynn/pkg/random"
+)
+
+func NewRequestLogger(ctx log.Ctx, handler http.Handler) http.Handler {
+	l := log.New(ctx)
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		start := time.Now()
+		logger := l.New(log.Ctx{"req_id": random.UUID()})
```
titanous: This should be set to the `X-Request-ID` header if it exists.

pkg/httphelper/request_logger.go
```diff
@@ -0,0 +1,23 @@
...
+		logger.Info("request started", "method", req.Method, "path", req.URL.Path)
```
titanous: Might be useful to include the source IP/port here too

jvatic:
@titanous comments addressed

jvatic:
moved into #762

Issue #67, opened 2014-08-11
Robust queuing/buffering/timeout behavior in router


titanous:
See [Heroku's docs](https://devcenter.heroku.com/articles/http-routing#request-queueing) for inspiration.

bgentry:
In my upcoming PR, we've added some timeouts on the TCP side. However, there are not yet any timeouts on the request (corresponding to `http.Client.Timeout`). Just noting that here so we don't forget to add it.

bgentry:
#955 includes a 120 second timeout for receiving headers on responses.

titanous:
Note that the upgraded connection path does not have the same timeout.

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

PR #781, opened 2015-01-22
router: switch to from net/http/httputil to net/http

### LINKED ISSUE #61: Persistent backend connections in router
Due to flynn-archive/strowger#26, we changed the router to close the HTTP backend connection after each request in flynn-archive/strowger#27. This is not optimal, and we need a long-term solution that allows us to have a 1:1 mapping of client connections to backend connections. The main thing that needs to be solved is detection of connections closed by the backend before we send the next request.

**Relevant IRC logs**

- https://botbot.me/freenode/flynn/2014-06-19/?msg=16568761&page=1
- https://botbot.me/freenode/flynn/2014-06-20/?msg=16619353&page=1
- https://botbot.me/freenode/flynn/2014-06-20/?msg=16657413&page=1

### LINKED ISSUE #64: HTTP/1.0 support in router
Currently the router only supports HTTP/1.1, but clients exist that use HTTP/1.0 (mostly telnet, probably). See flynn-archive/flynn-demo#31 for discussion.

bgentry:
This PR switches Flynn's router away from using the somewhat-deprecated `httputil.ClientConn` and `httputil.ServerConn` to using `net/http` HTTP parsers and handlers.

Because there are subtle differences between handling protocol upgrades (i.e. for websockets) and regular request proxying, there's a lot of duplicated logic between the functions that handle these paths. I think we'll be able to clean that up a lot, but would rather this doesn't spend too much more time in a side branch :)

We pretty quickly had to move away from `httputil.ReverseProxy` as it makes it basically impossible to do some of the stuff we want to do with trying conns to multiple backends, and it also wouldn't work for the websocket upgrade. In doing so, we've currently lost some of the nice parts of that package like periodic flushing of buffers (see [ReverseProxy.FlushInterval](http://golang.org/pkg/net/http/httputil/#ReverseProxy)). I've left a couple of TODOs for that in the code and intend to bring it in in a future PR. I also think that we can better package up some of the stuff we've done into a nice encapsulation similar to `httputil.ReverseProxy`, but again, didn't want to branch off for too long.

Finally, the way we pass down the sticky session setting from the route to the service is pretty ugly (using a temporary internal header). I didn't see an obvious way to access this bool (`r.Sticky`) from the part of the code where I needed to use it, so this was a compromise for expediency. I suspect we can find a much better way to do this, either by refactoring to make the setting available where we need it or by storing that setting inside of [a request context](https://blog.golang.org/context).

Fixes #61. Fixes #64. Makes some progress on #67 by setting some sane defaults for TCP timeouts.

### DIFF
--- a/router/handlers_test.go
+++ b/router/handlers_test.go
@@ -14,8 +14,12 @@ func TestFwdProtoHandler(t *testing.T) {
 
 	rec := httptest.NewRecorder()
 	request, _ := http.NewRequest("GET", "http://test.com", nil)
+	request.RemoteAddr = "1.2.3.4:5678"
 	h := fwdProtoHandler{Handler: nopHandler, Proto: "https", Port: "443"}
 	h.ServeHTTP(rec, request)
+	if v := request.Header.Get("X-Forwarded-For"); v != "1.2.3.4" {
+		t.Errorf("want X-Forwarded-For %s, got %s", "1.2.3.4", v)
+	}
 	if v := request.Header.Get("X-Forwarded-Proto"); v != "https" {
 		t.Errorf("want X-Forwarded-Proto %s, got %s", "https", v)
 	}
--- a/router/handlers_test.go
+++ b/router/handlers_test.go
@@ -3,12 +3,13 @@ package main
 import (
 	"net/http"
 	"net/http/httptest"
-	"testing"
+
+	. "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
 )
 
 var nopHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
 
-func TestFwdProtoHandler(t *testing.T) {
+func (s *S) TestFwdProtoHandler(c *C) {
 	prevForwardedProto := "fakeproto"
 	prevForwardedPort := "8080"
 
@@ -17,15 +18,9 @@ func TestFwdProtoHandler(t *testing.T) {
 	request.RemoteAddr = "1.2.3.4:5678"
 	h := fwdProtoHandler{Handler: nopHandler, Proto: "https", Port: "443"}
 	h.ServeHTTP(rec, request)
-	if v := request.Header.Get("X-Forwarded-For"); v != "1.2.3.4" {
-		t.Errorf("want X-Forwarded-For %s, got %s", "1.2.3.4", v)
-	}
-	if v := request.Header.Get("X-Forwarded-Proto"); v != "https" {
-		t.Errorf("want X-Forwarded-Proto %s, got %s", "https", v)
-	}
-	if v := request.Header.Get("X-Forwarded-Port"); v != "443" {
-		t.Errorf("want X-Forwarded-Port %s, got %s", "443", v)
-	}
+	c.Assert(request.Header.Get("X-Forwarded-For"), Equals, "1.2.3.4")
+	c.Assert(request.Header.Get("X-Forwarded-Proto"), Equals, "https")
+	c.Assert(request.Header.Get("X-Forwarded-Port"), Equals, "443")
 
 	// test with headers already set, make sure we append correctly
 	rec = httptest.NewRecorder()
@@ -34,14 +29,6 @@ func TestFwdProtoHandler(t *testing.T) {
 	request.Header.Set("X-Forwarded-Port", prevForwardedPort)
 
 	h.ServeHTTP(rec, request)
-	want := prevForwardedProto + ", https"
-	got := request.Header.Get("X-Forwarded-Proto")
-	if want != got {
-		t.Errorf("want X-Forwarded-Proto %q, got %q", want, got)
-	}
-	want = prevForwardedPort + ", 443"
-	got = request.Header.Get("X-Forwarded-Port")
-	if want != got {
-		t.Errorf("want X-Forwarded-Port %q, got %q", want, got)
-	}
+	c.Assert(request.Header.Get("X-Forwarded-Proto"), Equals, prevForwardedProto+", https")
+	c.Assert(request.Header.Get("X-Forwarded-Port"), Equals, prevForwardedPort+", 443")
 }
--- a/router/handlers_test.go
+++ b/router/handlers_test.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+var nopHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
+
+func TestFwdProtoHandler(t *testing.T) {
+	prevForwardedProto := "fakeproto"
+	prevForwardedPort := "8080"
+
+	rec := httptest.NewRecorder()
+	request, _ := http.NewRequest("GET", "http://test.com", nil)
+	h := fwdProtoHandler{Handler: nopHandler, Proto: "https", Port: "443"}
+	h.ServeHTTP(rec, request)
+	if v := request.Header.Get("X-Forwarded-Proto"); v != "https" {
+		t.Errorf("want X-Forwarded-Proto %s, got %s", "https", v)
+	}
+	if v := request.Header.Get("X-Forwarded-Port"); v != "443" {
+		t.Errorf("want X-Forwarded-Port %s, got %s", "443", v)
+	}
+
+	// test with headers already set, make sure we append correctly
+	rec = httptest.NewRecorder()
+	request, _ = http.NewRequest("GET", "http://test.com", nil)
+	request.Header.Set("X-Forwarded-Proto", prevForwardedProto)
+	request.Header.Set("X-Forwarded-Port", prevForwardedPort)
+
+	h.ServeHTTP(rec, request)
+	want := prevForwardedProto + ", https"
+	got := request.Header.Get("X-Forwarded-Proto")
+	if want != got {
+		t.Errorf("want X-Forwarded-Proto %q, got %q", want, got)
+	}
+	want = prevForwardedPort + ", 443"
+	got = request.Header.Get("X-Forwarded-Port")
+	if want != got {
+		t.Errorf("want X-Forwarded-Port %q, got %q", want, got)
+	}
+}
--- a/router/http.go
+++ b/router/http.go
@@ -224,20 +224,28 @@ func (h *httpSyncHandler) Remove(id string) error {
 }
 
 func (s *HTTPListener) listenAndServe(started chan<- error) {
-	var err error
+	_, port, err := net.SplitHostPort(s.Addr)
+	if err != nil {
+		started <- err
+		return
+	}
 	s.listener, err = reuseport.NewReusablePortListener("tcp4", s.Addr)
 	started <- err
 	if err != nil {
 		return
 	}
-	for {
-		conn, err := s.listener.Accept()
-		if err != nil {
-			// TODO: log error
-			break
-		}
-		go s.handle(conn, false)
+
+	server := &http.Server{
+		Addr: s.Addr,
+		Handler: fwdProtoHandler{
+			Handler: s,
+			Proto:   "http",
+			Port:    port,
+		},
 	}
+
+	// TODO: log error
+	_ = server.Serve(s.listener)
 }
 
 func (s *HTTPListener) listenAndServeTLS(started chan<- error) {
--- a/router/http.go
+++ b/router/http.go
@@ -580,6 +580,14 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	outreq.ProtoMinor = 1
 	outreq.Close = false
 
+	// TODO: Proxy HTTP CONNECT? (example: Go RPC over HTTP)
+
+	// Directly bridge `Connection: Upgrade` requests
+	if strings.ToLower(outreq.Header.Get("Connection")) == "upgrade" {
+		s.forwardAndProxyTCP(w, outreq)
+		return
+	}
+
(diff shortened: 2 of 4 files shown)


bgentry:
one fail so far in the Flynn CI:

```
FAIL: test_git_deploy.go:72: GitDeploySuite.TestRubyBuildpack

++ 18:47:44.769 /usr/bin/git clone https://github.com/flynn-examples/ruby-flynn-example /tmp/check-8674665223082153551/10/repo
Cloning into '/tmp/check-8674665223082153551/10/repo'...

++ 18:47:45.999 /home/ubuntu/go/src/github.com/flynn/flynn/test/../cli/flynn-cli create ruby-flynn-example
Created ruby-flynn-example

++ 18:47:46.581 /usr/bin/git push flynn master
remote: [1G[K-----> Building ruby-flynn-example...        
remote: Build failed: cluster: no servers found        
To ssh://git@flynn-1bd56847017dad2c.local:2222/ruby-flynn-example.git
 ! [remote rejected] master -> master (pre-receive hook declined)
error: failed to push some refs to 'ssh://git@flynn-1bd56847017dad2c.local:2222/ruby-flynn-example.git'

test_git_deploy.go:73:
    s.runBuildpackTest(t, "ruby-flynn-example", nil)
test_git_deploy.go:106:
    t.Assert(push, Succeeds)
... result *main.CmdResult = &main.CmdResult{Cmd:[]string{"git", "push", "flynn", "master"}, Output:"remote: \x1b[1G\x1b[K-----> Building ruby-flynn-example...        \nremote: Build failed: cluster: no servers found        \nTo ssh://git@flynn-1bd56847017dad2c.local:2222/ruby-flynn-example.git\n ! [remote rejected] master -> master (pre-receive hook declined)\nerror: failed to push some refs to 'ssh://git@flynn-1bd56847017dad2c.local:2222/ruby-flynn-example.git'\n", Err:(*exec.ExitError)(0xc2083f8400)}
```

router/handlers_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+var nopHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
+
+func TestFwdProtoHandler(t *testing.T) {
```
benburkert: are we :ok_hand: with mixing the use of `testing.T` & `check.C` in the router?

bgentry: yeah not sure. We had these tests already from one of our own project so I just imported them as-is (with additions for `X-Forwarded-For`). Let me know if you want to switch this to using gocheck instead.

router/http.go
```diff
@@ -275,62 +303,28 @@ func (s *HTTPListener) findRouteForHost(host string) *httpRoute {
...
-		StatusCode:    code,
-		ProtoMajor:    1,
-		ProtoMinor:    0,
-		Request:       req,
-		Body:          ioutil.NopCloser(bytes.NewBufferString(msg)),
-		ContentLength: int64(len(msg)),
-	}
-	sc.Write(req, resp)
+func fail(w http.ResponseWriter, code int, msg string) {
+	w.Header().Set("Content-Length", strconv.Itoa(len(msg)))
```
benburkert: Is this to force a fixed length response? Do we want to `Connection: close` on an error?

router/http.go
```diff
@@ -422,109 +376,269 @@ func (s *httpService) getBackendSticky(req *http.Request) (*httputil.ClientConn,
...
-		if prior, ok := req.Header["X-Forwarded-For"]; ok {
-			clientIP = strings.Join(prior, ", ") + ", " + clientIP
+	isSticky := false
+	stickyAddr := ""
+	if req.Header.Get(hdrUseStickySessions) == "true" {
+		// TODO(bgentry): switch to better way to check sticky setting
+		isSticky = true
+		req.Header.Del(hdrUseStickySessions)
+		if stickyAddr = s.stickyCookieAddr(req); stickyAddr != "" {
+			sortStringFirst(addrs, stickyAddr)
```
benburkert: :+1: I like this approach.

router/http.go
bgentry: It does force a fixed-length (non-chunked) response. As far as whether we want to `Connection: close`, after an error, that's generally a no, but there may be cases (specifically most `fail()` calls inside the TCP proxying path) where we do want to close the conn.

We can either add a param to `fail()` to indicate that, or just call `w.Header().Set("Connection", "close")` in each specific place where we want to do that. Extra param sounds better to me.

bgentry: also, that means more subtle differences between the two paths :(

router/http_test.go
```diff
@@ -401,6 +409,65 @@ func (s *S) TestStickyHTTPRoute(c *C) {
...
+	defer l.Close()
+
+	addStickyHTTPRoute(c, l)
+
+	discoverdRegisterHTTP(c, l, srv1.Listener.Addr().String())
+	defer discoverd.UnregisterAll()
+
+	// triggers the TCP-mode conn upgrade path, but does not hit the websocket
+	// handler and so the upgrade is not successful.
+	assertGetCookieUpgradeClosed := func(c *C, url, host, expected string, cookie *http.Cookie) *http.Cookie {
```
benburkert: this is really funky. It could probably be rewritten as table driven tests.

bgentry: can it really work as a table? The ordering of events matters here.

router/http_test.go
```diff
@@ -401,6 +409,65 @@ func (s *S) TestStickyHTTPRoute(c *C) {
 	}
 }
 
+func (s *S) TestStickyHTTPRouteWebsocket(c *C) {
+	wshandler := func(w http.ResponseWriter, req *http.Request) {
+		w.Write([]byte("ok " + strings.TrimPrefix(req.URL.Path, "/")))
```
benburkert: because the request path is used for the response body, it's possible for a request to `srv2` with path `/1` to respond with `"ok 1"`.

benburkert:
I think this is a big step forward, glad to see this coming together quickly. My two concerns are that the `TestStickyHTTPRouteWebsocket` test doesn't actually perform the websocket handshake, and that it's hard to follow along with how the vendored code from `reverseproxy.go` has been patched. For the second concern, I'm gonna look into refactoring the proxy code into a `router/proxy` package.

router/http_test.go
bgentry: ugh, I knew this ended up being too simple after I cleaned it up. Good catch.

benburkert:
> Finally, the way we pass down the sticky session setting from the route to the service is pretty ugly (using a temporary internal header). I didn't see an obvious way to access this bool (r.Sticky) from the part of the code where I needed to use it, so this was a compromise for expediency. I suspect we can find a much better way to do this, either by refactoring to make the setting available where we need it or by storing that setting inside of a request context.

I agree 100%, and I suspect that adding a context to the request path will also be useful for addressing #67.

router/handlers_test.go
titanous: I'd prefer changing this to gocheck for consistency.

router/http.go
```diff
@@ -275,62 +303,28 @@ func (s *HTTPListener) findRouteForHost(host string) *httpRoute {
...
-		}
-
-		req.RemoteAddr = conn.RemoteAddr().String()
-		if r.service.handle(req, sc, isTLS, r.Sticky) {
-			return
-		}
+	// TODO(bgentry): find a better way to access this setting in the service
+	// where it's needed.
+	if r.Sticky {
+		req.Header.Set(hdrUseStickySessions, "true")
```
titanous: +1 for using a context here. In the mean time, always set this header so that it can't be injected.

router/http.go
```diff
@@ -422,109 +376,269 @@ func (s *httpService) getBackendSticky(req *http.Request) (*httputil.ClientConn,
...
 	}
-	if tls {
-		req.Header.Set("X-Forwarded-Proto", "https")
-	} else {
-		req.Header.Set("X-Forwarded-Proto", "http")
-	}
-	// TODO: Set X-Forwarded-Port
+
+	// Most of this is borrowed from httputil.ReverseProxy
+	outreq := new(http.Request)
```
titanous: We've standardized on not using `new` to initialize structs, use `&http.Request{}`.

router/http.go
bgentry: *facepalm* great catch.

titanous:
> one fail so far in the Flynn CI:

This is an unrelated failure.

router/http.go
bgentry: opted for a `failAndClose()` that just calls `fail()` after setting the `Connection: close` header. See cfff35ce2f3123fc6703d9bfab9d7847615915d3

router/http.go
bgentry: fixed in fc0efc7622c2ac8a609897d557208f85aa34b029

router/http.go
```diff
@@ -422,109 +383,271 @@ func (s *httpService) getBackendSticky(req *http.Request) (*httputil.ClientConn,
...
+	}
+
+	if res.StatusCode != 101 {
+		return
+	}
+	res.Body.Close() // close this now since we've copied everything
+
+	hj, ok := w.(http.Hijacker)
+	if !ok {
+		log.Println("not a hijacker")
```
titanous: Is it reasonable to panic here? There should be no situation where we hit this code path with the current code.

router/handlers_test.go
bgentry: Done. c6fa8810ca8be812adca3659bb9abe25b0ea1a91

router/http.go
bgentry: yeah, I think so on this particular error. What do you think about the line down below (logged with `hijack failed:`)?

titanous: Currently we would never hit that path either: https://github.com/golang/go/blob/4b96409aacfb569ac2564240d8aadd668b7c25fa/src/net/http/server.go#L142-L147

But there's no guarantee that other more dynamic errors won't be added in the future.

bgentry: yeah, ok, I vote to keep that other scenario as a non-panic.

router/http_test.go
bgentry: FYI @benburkert has reorganized this test case so that it's much clearer what's going on: https://github.com/bgentry/flynn/pull/1

Will bring that in after the last couple of tweaks.

router/http.go
```diff
@@ -422,109 +383,271 @@ func (s *httpService) getBackendSticky(req *http.Request) (*httputil.ClientConn,
...
+		outreq.URL.Host = backend
+		res, err = transport.RoundTrip(outreq)
+		if err != nil {
+			if _, ok := err.(*dialErr); ok {
+				// retry, maybe log a message about it
+				continue
 			}
-			return true
+			log.Println("http: proxy error:", err)
+			fail(w, 503, http.StatusText(503))
```
titanous: Seems like `fail` could be modified to do the status text lookup itself? `fail(w, 503)`

bgentry: yeah, as long as you're ok with always using the default text for that status code. Looks like that's pretty much what's already happening in the other places that were using it.

titanous: Yeah, I expect we'll have nice error pages at some point, but for now it looks like we're just using variants of the status text.

router/http.go
bgentry: Fixed in 2fd76aeed2ed65ca08316cbe4c194b4e28c88594

router/http.go
bgentry: Fixed in 7d7dd9d8996968d862d786722b78f2d38b85edee

router/http_test.go
bgentry: Cherry-picked that PR as 89e66509b7a59097d4e443c5fc5baa161bf08a8b

router/http.go
bgentry: Fixed in 4245d38.

bgentry:
I think I've resolved everything mentioned so far. :beers:

router/http.go
```diff
@@ -275,62 +303,36 @@ func (s *HTTPListener) findRouteForHost(host string) *httpRoute {
...
 
-var errMissingTLS = errors.New("router: route not found or TLS not configured")
-
-func (s *HTTPListener) handle(conn net.Conn, isTLS bool) {
-	defer conn.Close() // TODO(benburkert): for TLS, check that closing this conn sends the TLS "close notify"
+func fail(w http.ResponseWriter, code int) {
+	msg := []byte(http.StatusText(code))
+	w.Header().Set("Content-Length", strconv.Itoa(len(msg)))
+	w.WriteHeader(code)
+	w.Write(msg)
```
titanous: Add a trailing newline so that when we curl it looks better?

router/http.go
```diff
@@ -422,109 +384,269 @@ func (s *httpService) getBackendSticky(req *http.Request) (*httputil.ClientConn,
...
+		return
+	}
+
+	if res.StatusCode != 101 {
+		return
+	}
+	res.Body.Close() // close this now since we've copied everything
+
+	hj, ok := w.(http.Hijacker)
+	if !ok {
```
titanous: Can just type assert without checking: `w.(http.Hijacker).Hijack()` which will have an even better error message in the panic.

titanous:
This looks great! Please squash any minor fixup commits, after the last two comments are addressed we should be good to merge. :dancer:

bgentry:
I think I've squashed all the ones that make sense. Lmk if you see any others worth squashing down.

titanous:
LGTM, will merge when CI looks good.

titanous:
:tada:

bgentry:
:v:

PR #784, opened 2015-01-24
Deployer

### LINKED ISSUE #63: Zero-downtime deployment

### LINKED ISSUE #536: Deploy Strategy

### LINKED ISSUE #602: deployer: Implement Strategies

### LINKED ISSUE #603: deployer: Reduce risk of workers performing the same deployments

archSeer:
Adds a deployer component, which uses the bgentry/que-go to schedule deployments using different strategies (currently one-by-one or all-at-once).

Resolves #602, resolves #603, resolves #63, resolves #536.


archSeer:
@flynn/devs This is now integrated with the controller and ready for another review pass

titanous:
Please move 3a094cc422b66901db004fcac92dfba1fafa328b and eae30dd28b65f268fd383f671ad767ec4d8dcae6 into a separate PR.

bootstrap/manifest_template.json
```diff
@@ -77,6 +77,10 @@
         "scheduler": {
           "cmd": ["scheduler"],
           "omni": true
+        },
+        "deployer": {
+          "cmd": ["deployer"],
+          "omni": true
```
titanous: Should this be an omni job?

archSeer: If the host that contains the deployer dies, we do want to continue to be able to deploy (same reasoning as for the scheduler being omni).

titanous: The scheduler will handle resurrecting it.

archSeer: True, removed.

controller/client/client.go
```diff
@@ -268,6 +268,50 @@ func (c *Client) GetApp(appID string) (*ct.App, error) {
...
+
+func (c *Client) DeployApp(appID, releaseID string) (initial bool, err error) {
+	d, err := c.CreateDeployment(appID, releaseID)
+	if err != nil {
+		return true, err
+	}
+
+	// if initial deploy, just stop here
+	if d.ID == "" {
+		return true, nil
```
lmars: It looks like no callers of `DeployApp` use the `initial` return value

controller/client/client.go
```diff
@@ -268,6 +268,50 @@ func (c *Client) GetApp(appID string) (*ct.App, error) {
...
+	defer stream.Close()
+	select {
+	case e := <-events:
+		if e.Status == "complete" {
+			break
+		}
+	case <-time.After(10 * time.Second):
+		return false, fmt.Errorf("Timed out waiting for deployment completion!")
+
+	}
```
lmars: This should be in a `for` loop

controller/deployer/main.go
```diff
@@ -0,0 +1,149 @@
...
+	client *controller.Client
+}
+
+func main() {
+	client, err := controller.NewClient("", os.Getenv("AUTH_KEY"))
+	if err != nil {
+		log.Fatalln("Unable to create controller client:", err)
+	}
+
+	if err := discoverd.Register("flynn-deployer", ":"+os.Getenv("PORT")); err != nil {
```
lmars: Do we need to register with discoverd?

controller/deployer/main.go
```diff
@@ -0,0 +1,149 @@
...
+	defer pgxpool.Close()
+
+	q := que.NewClient(pgxpool)
+	wm := que.WorkMap{"Deployment": cxt.handleJob}
+
+	workers := que.NewWorkerPool(q, wm, 10)
+	go workers.Start()
+	shutdown.BeforeExit(func() { workers.Shutdown() })
+
+	<-make(chan bool) // block and keep running
```
lmars: This should probably just call `workers.Start` directly rather than in a goroutine

controller/client/client.go
archSeer: Yeah, we could use it in [the receiver](https://github.com/flynn/flynn/blob/deploy-strategy/receiver/flynn-receive.go#L136), or we can remove that parameter, which would make it neater to use, since we can just `return client.DeployApp`

controller/deployer/main.go
```diff
@@ -0,0 +1,149 @@
...
+		if e != nil {
+			events <- ct.DeploymentEvent{
+				ReleaseID: deployment.NewReleaseID,
+				Status:    "failed",
+			}
+			if e = c.rollback(deployment, f); e != nil {
+				return
+			}
+		}
+		e = nil
```
lmars: This can be dropped and just assign `e = c.rollback(...)` without the `nil` check above

controller/deployer/main.go
```diff
@@ -0,0 +1,149 @@
...
+	}()
+	if err := strategyFunc(c.client, deployment, events); err != nil {
+		// TODO: log/handle error
+		return err
+	}
+	if err := c.setDeploymentDone(deployment.ID); err != nil {
+		log.Print(err)
+	}
+	if err := c.client.SetAppRelease(deployment.AppID, deployment.NewReleaseID); err != nil {
+		log.Print(err)
```
lmars: Should we not rollback if this fails?

controller/deployer/main.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (c *context) rollback(deployment *ct.Deployment, original *ct.Formation) error {
+	if err := c.client.PutFormation(original); err != nil {
+		return err
+	}
+	if err := c.client.DeleteFormation(deployment.AppID, deployment.NewReleaseID); err != nil {
+		return err
+	}
+	return nil
```
lmars: `return c.client.DeleteFormation(...)`

controller/deployer/main.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (c *context) setDeploymentDone(id string) error {
+	return c.db.Exec("UPDATE deployments SET finished_at = now() WHERE deployment_id = $1", id)
+}
+
+func (c *context) sendDeploymentEvent(e ct.DeploymentEvent) error {
```
lmars: As this doesn't so any "sending", I think it would be better named `createDeploymentEvent`

controller/deployer/strategies/all-at-once.go
```diff
@@ -0,0 +1,64 @@
...
+				JobState:  "stopping",
+				JobType:   typ,
+			}
+		}
+		expect[d.OldReleaseID] = map[string]map[string]int{typ: {"down": n}}
+	}
+	if err := waitForJobEvents(jobStream, events, expect); err != nil {
+		return err
+	}
+	return nil
```
lmars: `return waitForJobEvents(...)`

controller/schema.go
```diff
@@ -134,6 +137,53 @@ $$ LANGUAGE plpgsql`,
...
+		run_at      timestamptz NOT NULL DEFAULT now(),
+		job_id      bigserial   NOT NULL,
+		job_class   text        NOT NULL,
+		args        json        NOT NULL DEFAULT '[]'::json,
+		error_count integer     NOT NULL DEFAULT 0,
+		last_error  text,
+		queue       text        NOT NULL DEFAULT '',
+
+		CONSTRAINT que_jobs_pkey PRIMARY KEY (queue, priority, run_at, job_id))`,
+		`COMMENT ON TABLE que_jobs IS '3'`,
```
lmars: Any reason not to use the same white space convention as the first migration?

receiver/flynn-receive.go
```diff
@@ -125,8 +125,8 @@ func main() {
 	if err := client.CreateRelease(release); err != nil {
 		log.Fatalln("Error creating release:", err)
 	}
-	if err := client.SetAppRelease(app.Name, release.ID); err != nil {
-		log.Fatalln("Error setting app release:", err)
+	if _, err := client.CreateDeployment(app.Name, release.ID); err != nil {
```
lmars: Should this call `DeployApp`?

test/test_deployer.go
```diff
@@ -0,0 +1,117 @@
...
+			events = append(events, e)
+			if e.Status == "complete" {
+				break loop
+			}
+		case <-time.After(5 * time.Second):
+			t.Fatal("timed out waiting for deployment event")
+		}
+	}
+	compare := func(t *c.C, i *ct.DeploymentEvent, j *ct.DeploymentEvent) {
+		fmt.Println("Comparing", i, j)
```
lmars: This should call `debug` so that it's only outputted on failure and includes timestamps

controller/deployer/main.go
archSeer: Ah true, I don't think so.

controller/deployer/main.go
archSeer: `workers.Start` does not block indefinitely however: https://github.com/bgentry/que-go/blob/4ad3cf5c7e0aa6382d5a437deed9951efbe5d2fa/worker.go#L156

archSeer:
Comments addressed.

archSeer:
@lmars @titanous Rollback test is added, I think this should be enough to merge this PR, and then follow up with more patches.

titanous:
Okay, will review in a bit.

bootstrap/deploy_app_action.go
```diff
@@ -112,9 +112,8 @@ func (a *DeployAppAction) Run(s *State) error {
 	}
 	as.Formation = formation
 
-	if err := client.SetAppRelease(a.App.ID, a.Release.ID); err != nil {
+	if _, err := client.DeployApp(a.App.ID, a.Release.ID); err != nil {
```
titanous: This should probably be called `DeployAppRelease`

controller/deployer/main.go
```diff
@@ -0,0 +1,142 @@
...
+		ConnConfig:   pgxcfg,
+		AfterConnect: que.PrepareStatements,
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer pgxpool.Close()
+
+	q := que.NewClient(pgxpool)
+	wm := que.WorkMap{"Deployment": cxt.handleJob}
```
titanous: I'd prefer lowercase `deployment`

controller/deployer/main.go
```diff
@@ -0,0 +1,142 @@
...
+
+	workers := que.NewWorkerPool(q, wm, 10)
+	go workers.Start()
+	shutdown.BeforeExit(func() { workers.Shutdown() })
+
+	<-make(chan bool) // block and keep running
+}
+
+func (c *context) handleJob(job *que.Job) (e error) {
```
titanous: Uppercase: `HandleJob`

controller/deployer/strategies/one-by-one.go
```diff
@@ -0,0 +1,63 @@
...
+			if err := client.PutFormation(&ct.Formation{
+				AppID:     d.AppID,
+				ReleaseID: d.OldReleaseID,
+				Processes: oldFormation,
+			}); err != nil {
+				return err
+			}
+			events <- ct.DeploymentEvent{
+				ReleaseID: d.OldReleaseID,
+				JobState:  "stopping",
```
titanous: I'd like to see constants used for these states.

controller/deployment.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	deployment.ID = postgres.CleanUUID(deployment.ID)
+	deployment.OldReleaseID = postgres.CleanUUID(deployment.OldReleaseID)
+	deployment.NewReleaseID = postgres.CleanUUID(deployment.NewReleaseID)
+
+	args, err := json.Marshal(ct.DeployID{ID: deployment.ID})
+	if err != nil {
+		return err
+	}
+	if err := r.q.Enqueue(&que.Job{
```
titanous: This should use the same transaction as the insert above.

controller/deployment.go
```diff
@@ -0,0 +1,272 @@
...
+		respondWithError(w, err)
+		return
+	}
+	deployment := &ct.Deployment{
+		AppID:        app.ID,
+		OldReleaseID: oldRelease.ID,
+		NewReleaseID: release.ID,
+		Strategy:     app.Strategy,
+	}
+	c.deploymentRepo.Add(deployment)
```
titanous: Missing error check?

controller/deployment.go
```diff
@@ -0,0 +1,272 @@
...
+	app := c.getApp(ctx)
+
+	fs, err := c.formationRepo.List(app.ID)
+	if err != nil {
+		respondWithError(w, err)
+		return
+	}
+	if len(fs) == 0 || (len(fs) == 1 && fs[0].ReleaseID == release.ID) {
+		// immediately set app release
+		c.appRepo.SetRelease(app.ID, release.ID)
```
titanous: Missing error check?

controller/deployment.go
```diff
@@ -0,0 +1,272 @@
...
+				Message: fmt.Sprintf("could not find release with ID %s", rid.ID),
+			}
+		}
+		respondWithError(w, err)
+		return
+	}
+	release := rel.(*ct.Release)
+	app := c.getApp(ctx)
+
+	fs, err := c.formationRepo.List(app.ID)
```
titanous: Add a TODO comment that all of this should probably be in a transaction

controller/schema.go
```diff
@@ -134,6 +137,53 @@ $$ LANGUAGE plpgsql`,
     FOR EACH ROW EXECUTE PROCEDURE notify_job_event()`,
 
 		`CREATE SEQUENCE name_ids MAXVALUE 4294967295`,
+		`CREATE TABLE deployments (
+    deployment_id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+    app_id uuid NOT NULL,
+    old_release_id uuid NOT NULL,
+    new_release_id uuid NOT NULL,
```
titanous: These should have `REFERENCES` foreign keys

controller/schema.go
```diff
@@ -134,6 +137,53 @@ $$ LANGUAGE plpgsql`,
...
+
+		`CREATE UNIQUE INDEX isolate_deploys ON deployments (old_release_id, new_release_id)
+    WHERE finished_at is NULL`,
+
+		`CREATE SEQUENCE deployment_event_ids`,
+		`CREATE TYPE deployment_status AS ENUM ('running', 'complete', 'failed')`,
+		`CREATE TABLE deployment_events (
+    event_id bigint PRIMARY KEY DEFAULT nextval('deployment_event_ids'),
+    deployment_id uuid NOT NULL REFERENCES deployments (deployment_id),
+    release_id uuid NOT NULL,
```
titanous: needs a foreign key as well

test/test_deployer.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	waitForDeploymentEvents(t, events, expected)
+
+	// check that we're running the old release
+	rel, err := s.controllerClient(t).GetAppRelease(deployment.AppID)
+	t.Assert(err, c.IsNil)
+	t.Assert(rel.ID, c.Equals, oldReleaseID)
+
+	// check that the old formation is the same and there's no new formation
+}
```
titanous: Is there supposed to be more code here? The comment implies more assertions?

archSeer: Forgot to push after doing a fixup.

controller/deployment.go
archSeer: The above call uses `pq`, while this call uses `pgx`.

controller/deployer/strategies/one-by-one.go
archSeer: This field is an aggregate of deployer events, job events and deploy status (`running`, `failed`, `complete`), the latter two being strings because StreamJobEvents provides those states as strings, and deploy status being a postgres enum.

controller/deployment.go
titanous: Okay, add a comment.

controller/deployer/main.go
lmars: Ok I didn't realise, no need to call `workers.Start()` in a goroutine then

controller/deployer/main.go
```diff
@@ -0,0 +1,142 @@
...
+	defer func() {
+		// rollback failed deploy
+		if e != nil {
+			events <- ct.DeploymentEvent{
+				ReleaseID: deployment.NewReleaseID,
+				Status:    "failed",
+			}
+			if err = c.rollback(deployment, f); err != nil {
+				e = err
+			}
```
lmars: This could just be:

```
e = c.rollback(deployment, f)
```

archSeer: I simplified it to that, but then I noticed that if rollback succeeds and returns `nil`, then we will return nil from the worker, and the job will be marked as `done` instead of `failed` (and the worker then won't retry the job)

controller/client/client.go
lmars: I would remove it since it is not being used.

controller/deployer/main.go
lmars: Ok gotcha

controller/client/client.go
archSeer: Done.

controller/controller_test.go
```diff
@@ -48,8 +51,20 @@ func (s *S) SetUpSuite(c *C) {
...
+			Host:     "/var/run/postgresql",
+			Database: dbname,
+		},
+		AfterConnect: que.PrepareStatements,
+	})
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	shutdown.BeforeExit(func() { pgxpool.Close() })
```
titanous: No need for this shutdown call.

controller/deployer/main.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	pgxpool, err := pgx.NewConnPool(pgx.ConnPoolConfig{
+		ConnConfig:   pgxcfg,
+		AfterConnect: que.PrepareStatements,
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer pgxpool.Close()
```
titanous: This should use `shutdown.BeforeExit`

titanous:
LGTM

titanous:
Please pull in the latest que-go and pgx before merging.

PR #785, opened 2015-01-24
host: Create/configure bridge directly instead of via libvirt

titanous:
This prevents libvirt from enabling some iptables rules and starting dnsmasq which conflicts with discoverd2.


lmars:
LGTM

PR #787, opened 2015-01-24
host: Fix libvirt network XML

lmars:
The empty fields were causing the following error:

```
[Code-27] [Domain-19] XML error: Missing required address attribute in network 'flynn'
```


titanous:
LGTM

PR #792, opened 2015-01-24
Run unit tests in flynn CI

heavenlyhash:
Move unit tests into flynn CI.  Will deprecate travis.

Adds dependencies of unit tests to the rootfs used in CI.

Adds a script 'test/scripts/test-unit.sh' which can drive unit tests.  In the future, this script can be modified without requiring updates to the main CI binary.

If unit tests fail, the entire run will be halted; cluster bootstrapping and integration tests will not be run.

### DIFF
--- a/test/cluster/cluster.go
+++ b/test/cluster/cluster.go
@@ -122,6 +122,11 @@ func (c *Cluster) BuildFlynn(rootFS, commit string, merge bool) (string, error) 
 		return build.Drive("hda").FS, fmt.Errorf("error running build script: %s", err)
 	}
 
+	if err := runUnitTests(build, c.out); err != nil {
+		build.Kill()
+		return build.Drive("hda").FS, fmt.Errorf("unit tests failed: %s", err)
+	}
+
 	if err := build.Shutdown(); err != nil {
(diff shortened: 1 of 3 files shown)


test/cluster/cluster.go
```diff
@@ -122,6 +122,13 @@ func (c *Cluster) BuildFlynn(rootFS, commit string, merge bool) (string, error)
 		return build.Drive("hda").FS, fmt.Errorf("error running build script: %s", err)
 	}
 
+	// REVIEW: do we have any desire to log to a separate stream?
+	// REVIEW: we could split out the DCO and other checks while we're here; desire?
```
titanous: We don't need to do either of these.

test/scripts/test-unit.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -eo pipefail
```
titanous: Can't hurt to `set -x` here

titanous:
Could probably squash all of these commits together, as they are pretty useless standalone.

titanous:
LGTM

PR #794, opened 2015-01-24
router: Strip port from Host header when looking up route

### LINKED ISSUE #789: controller: resource not found
Integration tests sometimes fail in bulk with messages of "controller: resource not found".

The host log contains the following line: "Error creating default route for controller: Post http://router-api/routes: dialer: no services found"

titanous:
It is totally valid for a Host header to look like:

    Host: example.com:80

or

    Host: example.com:443

This will cause a 404, as the route doesn't contain the port. So when doing the backend lookup, we need to strip the port off if it is included.

This fixes the "controller: resource not found" error that slipped through broken CI.

Closes #789.

/cc @bgentry @benburkert

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -287,6 +287,9 @@ func (s *HTTPListener) listenAndServeTLS(started chan<- error) {
 
 func (s *HTTPListener) findRouteForHost(host string) *httpRoute {
 	host = strings.ToLower(host)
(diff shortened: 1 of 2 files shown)


heavenlyhash:
LGTM

bgentry:
ahhh, the `Host` sometimes being `example.com:port` gets me almost every time. :+1:

PR #795, opened 2015-01-24
host: Check error from reading flannel config

titanous:
If we don’t do this, and there is an error, we’ll end up panicking later.


heavenlyhash:
LGTM

lmars:
LGTM

PR #798, opened 2015-01-24
router: Flush HTTP response headers and body writes

titanous:
Without this, some of our streaming clients would hang forever waiting for SSE events that would just hang out in the response body buffer. It may make sense to switch to an interval instead of flushing every write, but I want to get this in immediately, as it is breaking a bunch of stuff.


bgentry:
yeah, makes sense. @benburkert's refactor includes flushing. I'll work on getting some tests added on Monday to make sure this behavior is covered. The built-in ReverseProxy includes some tests for this: http://golang.org/src/net/http/httputil/reverseproxy_test.go

archSeer:
LGTM

PR #800, opened 2015-01-24
all: Fix attach HTTP upgrade protocol

titanous:
The previous implementation was not compliant with the HTTP spec, and broke when the router was refactored. The new protocol dance looks like:

    POST /attach
    Connection: upgrade
    Upgrade: flynn-attach/0

    ...

    101 Switching Protocols
    Connection: upgrade
    Upgrade: flynn-attach/0

    <attach frames>


lmars:
LGTM

PR #803, opened 2015-01-25
shutdown handler should be a singleton.

heavenlyhash:
Use of multiple handlers as written will result in the first one of them to finish exiting the whole process without waiting for other handler instances, which may result in incomplete cleanup.  There is now only one allowed handler instance.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -150,7 +150,6 @@ func runDaemon(args *docopt.Args) {
 		"udp": ports.NewAllocator(55000, 65535),
 	}
 
-	sh := shutdown.NewHandler()
 	state := NewState(hostID, stateFile)
 	var backend Backend
 	var err error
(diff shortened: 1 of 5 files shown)


lmars:
LGTM

archSeer:
LGTM

host/host.go
```diff
@@ -298,9 +297,9 @@ func runDaemon(args *docopt.Args) {
 		g.Log(grohl.Data{"at": "sampi_disconnected", "err": jobStream.Err})
 
 		// if the process is shutting down, just block
-		if sh.Active {
-			<-make(chan struct{})
-		}
+		//if shutdown.Active { // Wot?
```
heavenlyhash: Sanity check, can we remove this?  I'm almost certain the other loop behavior continuing on here is fine.

heavenlyhash: Belay that, by the light of the morning I understand what this is for.

pkg/shutdown/handler.go
```diff
@@ -7,33 +7,41 @@ import (
...
 	"syscall"
 )
 
-type Handler struct {
-	Active bool
+var h = newHandler()
 
-	mtx  sync.RWMutex
-	done chan struct{}
+type Handler struct {
```
titanous: I think we can unexport the `Handler` type?

PR #804, opened 2015-01-25
discoverd: Replace with discoverd2

titanous:
Incremental patch for #145.

No code changes, this just deletes the old discoverd code and replaces it with discoverd2.


heavenlyhash:
Some test applications are still broken by the API change.

PR #805, opened 2015-01-25
discoverd2 API changes

titanous:
Incremental patch for #145.

This changes all of Flynn to use the new discoverd2 APIs.


appliance/postgresql/runner.go
```diff
@@ -83,14 +91,14 @@ func startLeader() (*exec.Cmd, <-chan struct{}) {
 }
 
 func register(attrs map[string]string) {
-	err := discoverd.RegisterWithAttributes(*serviceName, addr, attrs)
+	err := heartbeater.SetMeta(attrs)
 	if err != nil {
 		log.Fatalln("discoverd registration error:", err)
```
archSeer: The error and the function name could be changed to be a bit more accurate (since by this point, we've already registered on discoverd)

titanous: This code will be thrown out in the next week, I didn't change anything extra to avoid increasing the diff size.

archSeer:
LGTM, but let's wait for more people to review this.

heavenlyhash:
LGTM.

PR #807, opened 2015-01-25
discoverd2

### LINKED ISSUE #69: Consolidate watches in discoverd
Each service should only use one watch (instead of one per stream).

### LINKED ISSUE #145: Service discovery redux

### LINKED ISSUE #170: Replace rpcplus with HTTP

titanous:
Closes #145
Closes #69 
Closes #170

There is no unreviewed code here. The only change since #805 is that buildpack web processes now bind to port 8080 instead of 80 (which they are not privileged to do).

There is one non-flaky test failing (the Go example app has an old version of discoverd). This will be fixed as soon as this PR is merged.


heavenlyhash:
LGTM

archSeer:
Ship it!

PR #808, opened 2015-01-25
test: Bump service timeout in TestNetworkedPersistentJob

titanous:
This was failing intermittently, 15s should be enough.


archSeer:
LGTM

PR #809, opened 2015-01-25
discoverd/health: Add more logging to tests

titanous:
`CheckSuite.TestHTTPConnectRefused` is failing intermittently, this will log the actual error that is being received.


heavenlyhash:
LGTM

PR #812, opened 2015-01-25
bootstrap: Try multiple times to connect to discoverd

lmars:
It can take a few seconds for discoverd to come up as it first waits for etcd + flannel + network setup.


titanous:
LGTM

PR #813, opened 2015-01-25
test: Output stderr of `flynn-host ps` when dumping logs

lmars:
The command sometimes hangs in CI, so we want to be able to see the output when sending SIGQUIT.


titanous:
LGTM

PR #814, opened 2015-01-25
discoverd/server: Wait a second before retrying after errors

titanous:
If we don't do this, when etcd goes down we'll go into a tight loop spamming the logs with errors.


lmars:
LGTM

PR #815, opened 2015-01-25
httpclient: Remove Dial and DialClose.

archSeer:
Primarily used for the way old discoverd managed service discovery.

Because of this simplification, we can now get rid of Close().

### DIFF
--- a/pkg/httpclient/json.go
+++ b/pkg/httpclient/json.go
@@ -33,16 +33,6 @@ type Client struct {
 	URL         string
 	Key         string
 	HTTP        *http.Client
-	Dial        DialFunc
-	DialClose   io.Closer
-}
-
-// Close closes the underlying transport connection.
-func (c *Client) Close() error {
-	if c.DialClose != nil {
-		c.DialClose.Close()
-	}
-	return nil
 }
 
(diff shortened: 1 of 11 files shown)


pkg/httpclient/json.go
```diff
@@ -119,7 +109,7 @@ func (c *Client) Hijack(method, path string, header http.Header, in interface{})
 	if err != nil {
 		return nil, err
 	}
-	dial := c.Dial
+	dial := c.HTTP.Transport.(*http.Transport).Dial
```
archSeer: I'm not completely sure if this will work in all cases?

titanous: Just use net.Dial

archSeer: Done.

archSeer: I was a bit careful because of the `NewClientWithPin`, that specifies it's own dial in the http client:
https://github.com/flynn/flynn/blob/eea338421d545727fb6eb179d5e1bfd634f74676/controller/client/client.go#L68

titanous:
LGTM, but needs a rebase.

PR #819, opened 2015-01-26
Graceful node shutdowns

heavenlyhash:


appliance/postgresql/runner.go
```diff
@@ -198,7 +202,7 @@ func runCmd(cmd *exec.Cmd) {
 				os.Exit(status.ExitStatus())
```
titanous: This looks important.

sdutil/sdutil.go
```diff
@@ -1,6 +1,11 @@
 package main
 
+import (
+	"github.com/flynn/flynn/pkg/shutdown"
+)
+
 func main() {
+	defer shutdown.Exit()
```
titanous: We already do signal handling in sdutil and it will be deleted soon, so no need.

test/apps/echoer/main.go
```diff
@@ -15,12 +15,15 @@ func main() {
 	port := os.Getenv("PORT")
 	addr := ":" + port
 
-	if _, err := discoverd.AddServiceAndRegister(service, addr); err != nil {
+	hb, err := discoverd.AddServiceAndRegister(service, addr)
+	if err != nil {
 		log.Fatal(err)
 	}
+	defer hb.Close()
```
titanous: Any reason not to use shutdown for these test apps?

archSeer: In one of the commit messages: "Use the pkg/shutdown system in all flynn services.  Small test/demo/example apps just use plain defers and idiomatic go error handling, to make it perfectly clear there's nothing special about the flynn shutdown convenience package."

heavenlyhash: They're short enough there's no compelling need, and I figured they might be used as examples.  We could convert them; wanna?

titanous: Yeah, the example apps are fine without it but use it for the test apps.

titanous:
LGTM

heavenlyhash:
There's a couple more races to be addressed in the cli 'run' subcommand, then this should finally be good.

heavenlyhash:
This should be good now.

titanous:
LGTM

PR #820, opened 2015-01-26
httpclient: Fix Dial for cases where we use pkg/pinned.

archSeer:


pkg/httpclient/json.go
```diff
@@ -33,6 +33,7 @@ type Client struct {
 	URL         string
 	Key         string
 	HTTP        *http.Client
+	PinDial     DialFunc
```
titanous: Let's call this HijackDial, since that's where it is used.

titanous:
LGTM

PR #825, opened 2015-01-26
util/release: Drop minimum commit statuses to one

titanous:
We are no longer using Travis, so the only commit status will be from Flynn CI.


archSeer:
LGTM

PR #826, opened 2015-01-26
router: RFC7230-compatible Connection header handling, more tests for header behaviors

bgentry:
Add a test to make sure that hop-by-hop headers, such as `Connection`, are
removed before the request is proxied to the backend.

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -431,6 +431,11 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 	req.Header.Del(hdrUseStickySessions) // delete this no matter what
 
+	// A proxy or gateway MUST parse a received Connection header field before a
+	// message is forwarded: https://tools.ietf.org/html/rfc7230#section-6.1
+	reqConnHdrOpts := parseConnHeader(req.Header.Get("Connection"))
+	isRequestConnUpgrade := isConnUpgrade(reqConnHdrOpts)
+
 	// Most of this is borrowed from httputil.ReverseProxy
 	outreq := &http.Request{}
 	*outreq = *req // includes shallow copies of maps, but okay
@@ -445,27 +450,36 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 
 	// TODO: Proxy HTTP CONNECT? (example: Go RPC over HTTP)
 
+	// Remove hop-by-hop headers to the backend.  This is modifying the same
+	// underlying map from req (shallow copied above) so we only copy it if
+	// necessary.
+	outreq.Header = make(http.Header, len(req.Header))
+	copyHeader(outreq.Header, req.Header)
+	for _, h := range alwaysHopHeaders {
+		outreq.Header.Del(h)
+	}
+
+	// remove the Upgrade header if HTTP < 1.1 or if Connection header didn't
+	// contain "upgrade": https://tools.ietf.org/html/rfc7230#section-6.7
+	if outreq.Header.Get("Upgrade") != "" && (!req.ProtoAtLeast(1, 1) || !isRequestConnUpgrade) {
+		outreq.Header.Del("Upgrade")
+	}
+
 	// Directly bridge `Connection: Upgrade` requests
-	if strings.ToLower(outreq.Header.Get("Connection")) == "upgrade" {
+	if isRequestConnUpgrade {
 		s.forwardAndProxyTCP(w, outreq, addrs, stickyAddr, isSticky)
 		return
 	}
 
-	// Remove hop-by-hop headers to the backend.  Especially
-	// important is "Connection" because we want a persistent
-	// connection, regardless of what the client sent to us.  This
-	// is modifying the same underlying map from req (shallow
-	// copied above) so we only copy it if necessary.
-	copiedHeaders := false
-	for _, h := range hopHeaders {
-		if outreq.Header.Get(h) != "" {
-			if !copiedHeaders {
-				outreq.Header = make(http.Header)
-				copyHeader(outreq.Header, req.Header)
-				copiedHeaders = true
-			}
-			outreq.Header.Del(h)
-		}
+	// A proxy or gateway MUST parse a received Connection header field before a
+	// message is forwarded and, for each connection-option in this field, remove
+	// any header field(s) from the message with the same name as the
+	// connection-option, and then remove the Connection header field itself (or
+	// replace it with the intermediary's own connection options for the
+	// forwarded message): https://tools.ietf.org/html/rfc7230#section-6.1
+	outreq.Header.Del("Connection")
+	for _, opt := range reqConnHdrOpts {
+		outreq.Header.Del(opt)
 	}
 
 	var (
@@ -503,8 +517,21 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		http.SetCookie(w, s.newStickyCookie(backend))
 	}
 
-	for _, h := range hopHeaders {
+	// A proxy or gateway MUST parse a received Connection header field before a
+	// message is forwarded: https://tools.ietf.org/html/rfc7230#section-6.1
+	respConnHdrOpts := parseConnHeader(res.Header.Get("Connection"))
+	res.Header.Del("Connection")
+	for _, h := range alwaysHopHeaders {
 		res.Header.Del(h)
+	}
+	// remove hop-by-hop headers as specified in connection options
+	for _, opt := range respConnHdrOpts {
+		res.Header.Del(opt)
+	}
+	// remove the Upgrade header if HTTP < 1.1 or if Connection header didn't
+	// contain "upgrade": https://tools.ietf.org/html/rfc7230#section-6.7
+	if res.Header.Get("Upgrade") != "" && (!req.ProtoAtLeast(1, 1) || !isConnUpgrade(respConnHdrOpts)) {
+		res.Header.Del("Upgrade")
 	}
 
 	copyHeader(w.Header(), res.Header)
@@ -588,6 +615,14 @@ func (s *httpService) forwardAndProxyTCP(w http.ResponseWriter, req *http.Reques
 	}
 	defer res.Body.Close()
 
+	respConnHdrOpts := parseConnHeader(res.Header.Get("Connection"))
+	for _, h := range alwaysHopHeaders {
+		res.Header.Del(h)
+	}
+	if res.Header.Get("Upgrade") != "" && !isConnUpgrade(respConnHdrOpts) {
+		res.Header.Del("Upgrade")
+	}
+
 	// Copy the response headers and body over to the downstream ResponseWriter,
 	// the same as done in the non-TCP path.
 	copyHeader(w.Header(), res.Header)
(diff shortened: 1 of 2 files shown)


router/http_test.go
```diff
@@ -339,6 +339,34 @@ func (s *S) TestHTTPHeadersFromClient(c *C) {
 	c.Assert(res.StatusCode, Equals, 200)
 }
 
+func (s *S) TestConnectionCloseHeaderFromClient(c *C) {
+	h := http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		// Connection: close header should be stripped by the reverse proxy so it
+		// always does keep-alive with backends.
+		c.Assert(req.Close, Equals, false)
+		c.Assert(req.Header.Get("Proxy-Authenticate"), Equals, "")
+		c.Assert(req.Header.Get("Proxy-Authorization"), Equals, "")
```
titanous: I'm not convinced that stripping these headers is correct. Per [RFC 7235](https://tools.ietf.org/html/rfc7235#section-4.4):

> the Proxy-Authorization header field applies only to the next inbound proxy that demanded authentication using the Proxy-Authenticate field.  When multiple proxies are used in a chain, the Proxy-Authorization header field is consumed by the first inbound proxy that was expecting to receive credentials.

Since we are not expecting credentials, do we need to "consume" it?

bgentry: yeah, that's a great question. I didn't do any research on the `Proxy-` headers, just left them in the list because they were in `httputil.ReverseProxy`. It certainly sounds like those shouldn't be removed by that proxy either.

bgentry:
The full list of headers that are being dropped by our proxy (inherited from `httputil.ReverseProxy`):

```go
// Hop-by-hop headers. These are removed when sent to the backend.
// http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html
var hopHeaders = []string{
	"Connection",
	"Keep-Alive",
	"Proxy-Authenticate",
	"Proxy-Authorization",
	"Te", // canonicalized version of "TE"
	"Trailers",
	"Transfer-Encoding",
	"Upgrade",
}
```

That's the list for non-upgrade requests. For the current TCP upgrade / websocket path, none of these is removed as of now. @benburkert's upcoming PR, however, does remove this same list of headers from upgrade requests, with the exception of `Connection` and `Upgrade` (required for the websocket upgrade handshake).

I *think* that's correct.

Do we need to drop anything from this list other than `Proxy-Authenticate` and `Proxy-Authorization`?

router/http_test.go
titanous: It looks like "hop-by-hop" got redefined in [RFC 7230](https://tools.ietf.org/html/rfc7230#section-6.1):

> When a header field aside from Connection is used to supply control information for or about the current connection, the sender MUST list the corresponding field-name within the Connection header field.  A proxy or gateway MUST parse a received Connection header field before a message is forwarded and, for each connection-option in this field, remove any header field(s) from the message with the same name as the connection-option, and then remove the Connection header field itself (or replace it with the intermediary's own connection options for the forwarded message).
>
> Hence, the Connection header field provides a declarative way of distinguishing header fields that are only intended for the immediate recipient ("hop-by-hop") from those fields that are intended for all recipients on the chain ("end-to-end"), enabling the message to be self-descriptive and allowing future connection-specific extensions to be deployed without fear that they will be blindly forwarded by older intermediaries.

titanous: And in the "Changes from RFC2616" section:

> Also, "hop-by-hop" header fields are required to appear in the Connection header field; just because they're defined as hop-by-hop in this specification doesn't exempt them.

titanous:
Based on the excerpts/links posted above and the wording in the new HTTP/1.1 RFCs, I think that the "correct" behavior is to do more parsing of the `Connection` header and strip only headers listed in that header or headers for transfer options that we are not capable of proxying.

bgentry:
OK, so after a very thorough reading of [RFC7230 Section 6.1](https://tools.ietf.org/html/rfc7230#section-6.1), here's my understanding of the situation:

Any value (or series of values) may be supplied for the `Connection` header., so long as each is a valid `token` (i.e. header format, such as `keep-alive` or `whatever-i-want`). As a proxy, we MUST parse this `Connection` header, and either remove it altogether or replace it with our own connection options (for conns to the backend) before forwarding the request.

Additionally, before forwarding, we MUST remove from the request any headers corresponding to the values listed in the `Connection` header. So in the example above, we would remove `Keep-Alive` and `Whatever-I-Want` headers from the request before forwarding it along, because these are specific to the current connection.

As @titanous said, a close reading of [RFC7235 Section 4.3](https://tools.ietf.org/html/rfc7235#section-4.3) shows that `Proxy-Authenticate` and `Proxy-Authorization` headers are only supposed to be consumed by the first proxy that was expecting them. Which is not us, so we shouldn't touch those.

The case of `Upgrade` is a little bit different. That turns out to be very well documented in [RFC720 Section 6.7](https://tools.ietf.org/html/rfc7230#section-6.7). We should never be removing `Upgrade` from responses, as a server may use that field in any response to advertise its support for other protocols that clients may upgrade to.

One detail that sticks out is that we MUST ignore an Upgrade header sent in an HTTP/1.0 request. To me that suggests that we should not use the TCP upgrade / websocket route if this is not at least an HTTP/1.1 request, and should delete that header before sending the request to the backend (if HTTP/1.0).

The RFC says that any `Connection` headers are supposed to apply to only a single hop, but I don't think that's the correct behavior for the `upgrade` case with a transparent reverse proxy like ours. Otherwise websockets would have to be negotiated one leg at a time.

Regarding `Transfer-Encoding` and `TE`, my understanding is that those are still single-hop headers and that any encoding is handled by Go's net/http client or server, which is also responsible for specifying which encodings it supports. So we should still strip those headers out.

Going to work on tests for these behaviors.

bgentry:
@titanous I think this is ready for another look.

Obviously it will take some work to deal with the conflicts that will come up between this and #842. But at least this gives us some better coverage on what the behaviors should be.

LMK if any of it doesn't make sense. I've done my best to stay true to the RFC723X specs, though there's a lot to dig through there.

router/http.go
```diff
@@ -639,17 +671,33 @@ func copyHeader(dst, src http.Header) {
...
-// Hop-by-hop headers. These are removed when sent to the backend.
-// http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html
-var hopHeaders = []string{
-	"Connection",
-	"Keep-Alive",
-	"Proxy-Authenticate",
-	"Proxy-Authorization",
+func isConnUpgrade(connOpts []string) bool {
+	for i := range connOpts {
+		if connOpts[i] == "upgrade" {
```
titanous: Can we just do:

```go
for _, opt := range connOpts {
    if opt == "upgrade" {
```

router/http.go
```diff
@@ -639,17 +671,33 @@ func copyHeader(dst, src http.Header) {
...
 	"Transfer-Encoding",
-	"Upgrade",
+}
+
+func parseConnHeader(value string) []string {
+	splitOpts := strings.Split(value, ",")
+	headerOpts := make([]string, 0, len(splitOpts))
+	for i := range splitOpts {
+		// remove empty values
+		if opt := strings.ToLower(strings.TrimSpace(splitOpts[i])); opt != "" {
```
titanous: Ditto.

router/http_test.go
```diff
@@ -339,6 +339,202 @@ func (s *S) TestHTTPHeadersFromClient(c *C) {
...
+	req.Header.Set("Keep-Alive", "test-keep-alive")
+	req.Header.Set("Transfer-Encoding", "test-transfer-encoding")
+	req.Header.Set("Upgrade", "test-upgrade")
+	res, err := httpClient.Do(req)
+	c.Assert(err, IsNil)
+	defer res.Body.Close()
+	c.Assert(res.StatusCode, Equals, 200)
+}
+
+func testConnectionHeadersFromServer(
```
titanous: Any reason these are separate functions? Seems like they could just go inline in the for loop.

router/http_test.go
```diff
@@ -339,6 +339,202 @@ func (s *S) TestHTTPHeadersFromClient(c *C) {
...
+	srv *httptest.Server,
+	conn string,
+	emptyHeaders,
+	presentHeaders []string,
+) {
+	c.Logf("testing client with Connection: %q", conn)
+	srv.Config.Handler = http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		for _, k := range emptyHeaders {
+			c.Logf("checking if %q header is empty", k)
+			c.Assert(req.Header.Get(k), Equals, "")
```
titanous: Could do this instead of the `Logf`:

```go
c.Assert(req.Header.Get(k), Equals, "", Commentf("header = %s", k))
```

router/http_test.go
bgentry: the use of `defer` was the reason I pulled them out into their own functions. Could still inline them as separate funcs, or find a way to drop the defer if you'd like.

titanous: It looks like the defer isn't necessary as we don't read the body? I think we can just drop the `defer` prefix and close right away.

router/http_test.go
bgentry: oh nice, wasn't aware of that usage. :v:

router/http.go
bgentry: yeah, good call. Fixed this and the other similar one.

router/http.go
```diff
@@ -445,27 +450,38 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 
 	// TODO: Proxy HTTP CONNECT? (example: Go RPC over HTTP)
 
+	// Remove hop-by-hop headers to the backend.  This is modifying the same
+	// underlying map from req (shallow copied above) so we only copy it if
+	// necessary.
+	outreq.Header = make(http.Header)
```
titanous: Could add a capacity hint of `len(req.Header)` here.

router/http.go
```diff
@@ -445,27 +450,38 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-		}
+	// A proxy or gateway MUST parse a received Connection header field before a
+	// message is forwarded and, for each connection-option in this field, remove
+	// any header field(s) from the message with the same name as the
+	// connection-option, and then remove the Connection header field itself (or
+	// replace it with the intermediary's own connection options for the
+	// forwarded message): https://tools.ietf.org/html/rfc7230#section-6.1
+	outreq.Header.Del("Connection")
+	for i := range reqConnHdrOpts {
+		outreq.Header.Del(reqConnHdrOpts[i])
```
titanous: Another range where we can use the second param

router/http.go
```diff
@@ -503,9 +519,22 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	// A proxy or gateway MUST parse a received Connection header field before a
+	// message is forwarded: https://tools.ietf.org/html/rfc7230#section-6.1
+	respConnHdrOpts := parseConnHeader(res.Header.Get("Connection"))
+	res.Header.Del("Connection")
+	for _, h := range alwaysHopHeaders {
 		res.Header.Del(h)
 	}
+	// remove hop-by-hop headers as specified in connection options
+	for i := range respConnHdrOpts {
+		res.Header.Del(respConnHdrOpts[i])
```
titanous: Ditto

router/http.go
```diff
@@ -445,27 +450,38 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 	// TODO: Proxy HTTP CONNECT? (example: Go RPC over HTTP)
 
+	// Remove hop-by-hop headers to the backend.  This is modifying the same
+	// underlying map from req (shallow copied above) so we only copy it if
+	// necessary.
+	outreq.Header = make(http.Header)
+	copyHeader(outreq.Header, req.Header)
+	for _, h := range alwaysHopHeaders {
+		if outreq.Header.Get(h) != "" {
+			outreq.Header.Del(h)
```
titanous: Any reason to do this check? `Del` shouldn't fail if it's not there.

bgentry: No. It's a relic from how this was done in ReverseProxy. Fixed.

bgentry:
I think I fixed everything you brought up so far. Thanks for the review!

titanous:
LGTM

titanous:
:metal:

PR #831, opened 2015-01-26
discoverd: Use current etcd index for watch after full sync

titanous:
Previously we were using the `ModifiedIndex` of the returned node, which is incorrect, as we can enter an infinite loop if that index is old. Instead we use the `EtcdIndex` which is the current index of the server and will be the most recent index.


archSeer:
LGTM

Issue #617, opened 2014-12-17
updater: Rolling out new images


lmars:
First we need to make sure images have associated tarsums, see #612.

Then:

- [ ] Use an S3 backed [tuf](https://github.com/theupdateframework/tuf) repository for version manifests (initial Go implementation [here](https://github.com/flynn/go-tuf))
- [ ] Add the generated image manifest (including tarsums) to the tuf repo (possibly embedded in the [release manifest](https://github.com/flynn/flynn/blob/master/script/release-flynn#L129-L137))
- [ ] Create an `updater` binary which takes a tuf root key, downloads the image manifest and passes it off to `flynn-host download` (or downloads images itself)

Part of #276.

PR #834, opened 2015-01-26
util/release: Support exporting images

lmars:
Incremental patch for #617


titanous:
Does this dump just the version we want of each image? Or all images in the "repo"?

lmars:
@titanous previously it would dump the "latest" tag of each of the images in the manifest, but I have now updated it to explicitly tag the images for the export (like the upload currently does).

titanous:
LGTM

PR #840, opened 2015-01-27
Testable host restarts

heavenlyhash:
Add features to the test cluter API to support the rebooting of both flynn-host processes, and whole test VMs.

Some of this is refactor to extract existing api code from tests that were previously the only consumers of it.

Still WIP.

### DIFF
--- a/test/cluster/cluster.go
+++ b/test/cluster/cluster.go
@@ -241,6 +241,26 @@ loop:
 	return nil
 }
 
+func (c *Cluster) BounceHost(id string) error {
+	inst, err := c.Instances.Get(id)
+	if err != nil {
+		return err
+	}
+
+	// send reboot to the host, and wait for new ssh
+	c.log("bouncing host", id)
+	if err := inst.Reboot(); err != nil {
+		return err
+	}
+
+	// restart the flynn-host
+	c.log("restarting flynn-host process", id)
+	if err := c.startHostProcess(inst); err != nil {
+		return err
+	}
+	return nil
+}
+
 func (c *Cluster) Size() int {
 	return len(c.Instances)
 }
--- a/test/cluster/cluster.go
+++ b/test/cluster/cluster.go
@@ -222,12 +222,7 @@ func (c *Cluster) RemoveHost(id string) error {
 	defer stream.Close()
 
 	// ssh into the host and tell the flynn-host daemon to stop
-	var cmd string
-	switch c.bc.Backend {
-	case "libvirt-lxc":
-		cmd = "sudo start-stop-daemon --stop --pidfile /var/run/flynn-host.pid --retry 15"
-	}
-	if err := inst.Run(cmd, nil); err != nil {
+	if err := c.stopHostProcess(inst); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


test/cluster/cluster.go
```diff
@@ -36,6 +37,9 @@ type Cluster struct {
 	ControllerKey string        `json:"controller_key"`
 	RouterIP      string        `json:"router_ip"`
 
+	discMtx sync.Mutex
+	disc    *discoverd.Client
```
lmars: Is the mutex necessary?

test/cluster/cluster.go
```diff
@@ -185,20 +198,99 @@ func (c *Cluster) RemoveHost(id string) error {
...
+	defer stream.Close()
+
+	// ssh into the host and tell the flynn-host daemon to stop
+	c.stopHostProcess(inst)
+
+loop:
+	for {
+		select {
+		case event := <-events:
+			if event.Kind == discoverd.EventKindDown {
```
lmars: Is it worth explicitly checking that the event came from the removed host?

test/cluster/cluster.go
```diff
@@ -185,20 +198,99 @@ func (c *Cluster) RemoveHost(id string) error {
...
+	if !ok {
+		return fmt.Errorf("unknown host backend: %s", c.bc.Backend)
+	}
+	return inst.Run(cmd, nil)
+}
+
 func (c *Cluster) startVMs(rootFS string, initial bool, count int) ([]*Instance, error) {
-	tmpl, ok := flynnHostScripts[c.bc.Backend]
+	_, ok := flynnHostStartScripts[c.bc.Backend]
 	if !ok {
```
lmars: `if _, ok := ...; !ok`

lmars:
@heavenlyhash I have rebased this on master since #901 got merged.

I think `TestBounceHostVM` should be expanded to check that flynn-host is running after rebooting the VM.

heavenlyhash:
Hm... I think we already have a lot of it covered, the question is just if it's direct and race-free enough.

 `RemoveHosts` will error at the if the flynn-host process doesn't start and clusterize enough to get a router-api job heartbeating.  Though that's way late in the game -- a huge amount of code covered and a pretty indirect way to detect a specific failure, admittedly.

`startHostProcess` will error and that will bubble out of `BounceHost` if the flynn-host daemon process fails immediately.  So then the question is just how much resolution we want to add between those two steps.

We could add a discoverd watch for flynn-host up events to the test to make it a more direct observation?

If I put that watch in right before the `testCluster.BounceHost`, is that going to be a race with the earlier `AddHosts`, or does the start script wait long enough for that to be clear?

lmars:
> startHostProcess will error and that will bubble out of BounceHost if the flynn-host daemon process fails immediately

I don't think this is correct. We start flynn-host using start-stop-daemon which exits straight after fork/exec. We should be relying on host / discoverd events to determine if things are running.

TBH I don't think the test is really testing anything which isn't covered elsewhere. From a users perspective, they would probably expect flynn-host to recover on it's own after a reboot, whereas the test is rebooting and then explicitly starting flynn-host again, which isn't all that different from how we start flynn-host in the first place.

titanous:
Closing as this is a WIP, feel free to open a new PR later.

PR #841, opened 2015-01-27
controller: Remove "_id" suffixes from deployment json

jvatic:
This makes the JSON consistent with other controller resources


titanous:
LGTM

PR #842, opened 2015-01-27
router: add proxy package

benburkert:
The `router/proxy` package contains the (reverse)proxy functionality extracted from `router/http.go` & `router/tcp.go`. The package is broken up into two main types: `ReverseProxy` & `transporter`. They are broken up this way to keep features (like sticky-sessions) orthogonal to protocols (like HTTP & TCP) and out of `ReverseProxy`.

The `ReverseProxy` is a modified version of `httputil.ReverseProxy` from stdlib. There are a few changes to how it behaves internally that differ from the stdlib implementation. Additionally, support for generic TCP proxying has also been added via `ServeConn`. `ReverseProxy` is able to handle HTTP upgrades internally by using the same dialing logic as `ServeConn`. `transport` is based off `http.Transport` and implements the `http.RoundTripper` interface, but is also modified to support raw `net.Conn` connections via `Connect` and HTTP upgrades via `UpgradeHTTP`.

#### TODO

- [x] godoc all exported types & funcs.
- [x] rewrite tests to not use `StickyCookie`, then stop exporting it.
- [x] fix backend list fetching/updating mechanism in `ReverseProxy`.
- [x] cleanup, squash & reorder commits.

#### For Later PRs

- move `router/*_test.go` tests that assert internal proxy behavior into the `router/proxy` package.
- toggle stickiness on existing services.
- context & request logging.
- request/response timeouts.

/cc @bgentry

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -1,15 +1,10 @@
 package main
 
 import (
-	"bufio"
 	"crypto/md5"
-	"crypto/rand"
 	"crypto/tls"
-	"encoding/base64"
 	"encoding/hex"
 	"errors"
-	"io"
-	"log"
 	"net"
 	"net/http"
 	"strconv"
@@ -18,11 +13,10 @@ import (
 	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/kavu/go_reuseport"
-	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/crypto/nacl/secretbox"
 	"github.com/flynn/flynn/discoverd/client"
-	"github.com/flynn/flynn/pkg/httphelper"
 	"github.com/flynn/flynn/pkg/random"
 	"github.com/flynn/flynn/pkg/tlsconfig"
+	"github.com/flynn/flynn/router/proxy"
 	"github.com/flynn/flynn/router/types"
 )
 
@@ -187,7 +181,11 @@ func (h *httpSyncHandler) Set(data *router.Route) error {
 		if err != nil {
 			return err
 		}
-		service = &httpService{name: r.Service, sc: sc, cookieKey: h.l.cookieKey}
+		service = &httpService{
+			name: r.Service,
+			sc:   sc,
+			rp:   proxy.NewReverseProxy(sc.Addrs, h.l.cookieKey, r.Sticky),
+		}
 		h.l.services[r.Service] = service
 	}
 	service.refs++
@@ -312,22 +310,12 @@ func fail(w http.ResponseWriter, code int) {
 	w.Write(msg)
 }
 
-const hdrUseStickySessions = "Flynn-Use-Sticky-Sessions"
-
 func (s *HTTPListener) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	r := s.findRouteForHost(req.Host)
 	if r == nil {
 		fail(w, 404)
 		return
 	}
-
-	// TODO(bgentry): find a better way to access this setting in the service
-	// where it's needed.
-	stickyValue := "false"
-	if r.Sticky {
-		stickyValue = "true"
-	}
-	req.Header.Set(hdrUseStickySessions, stickyValue)
 
 	r.service.ServeHTTP(w, req)
 }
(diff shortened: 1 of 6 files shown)


router/http.go
```diff
@@ -356,310 +345,27 @@ type httpService struct {
...
-
-func shuffle(s []string) []string {
-	for i := len(s) - 1; i > 0; i-- {
-		j := random.Math.Intn(i + 1)
-		s[i], s[j] = s[j], s[i]
-	}
-	return s
+func isWebSocketUpgrade(req *http.Request) bool {
+	return strings.ToLower(req.Header.Get("Connection")) == "upgrade" &&
+		strings.ToLower(req.Header.Get("Upgrade")) == "websocket"
```
bgentry: I don't think there's any reason we need to restrict our upgrade path to being websocket-specific. The Upgrade header semantics are more general than that: https://tools.ietf.org/html/rfc7230#section-6.7

AFAICT we should only care about the `Connection: upgrade` header, and whether the response was a 101 or not. Then we should be able to proceed into TCP proxying mode.

router/http_test.go
```diff
@@ -478,23 +489,25 @@ func (s *S) TestStickyHTTPRouteWebsocket(c *C) {
...
 
-	assertGet(c, "http://"+l.Addr, "example.com", "")
+	req := newReq(url, "example.com")
+	req.Header.Set("Connection", "Upgrade")
+	req.Header.Set("Upgrade", "websocket")
+	res, err := httpClient.Do(req)
+	defer res.Body.Close()
+
+	c.Assert(err, IsNil)
+	c.Assert(res.Close, Equals, true)
```
bgentry: Turns out this behavior wasn't correct after all. See 616d112a9fd1e6d195239705cdf98699c5792393 for an explanation and my suggested fix.

router/http_test.go
```diff
@@ -514,7 +527,7 @@ func (s *S) TestNoBackends(c *C) {
 	c.Assert(res.StatusCode, Equals, 503)
 	data, err := ioutil.ReadAll(res.Body)
 	c.Assert(err, IsNil)
-	c.Assert(string(data), Equals, "Service Unavailable\n")
+	c.Assert(string(data), Equals, http.StatusText(http.StatusServiceUnavailable))
```
bgentry: It was intentional that I didn't use the constants in the tests. I felt it was a better guarantee against regressions, though that might not really be true.

Regardless, it looks like you've lost the curl-friendly `\n` that was in these responses before.

router/http_test.go
```diff
@@ -646,6 +662,7 @@ func runTestErrorAfterConnOnlyHitsOneBackend(c *C, upgrade bool) {
 	req := newReq("http://"+l.Addr, "example.com")
 	if upgrade {
 		req.Header.Set("Connection", "Upgrade")
+		req.Header.Set("Upgrade", "websocket")
```
bgentry: same comment re: websocket-specific upgrade path

router/http_test.go
```diff
@@ -654,7 +671,7 @@ func runTestErrorAfterConnOnlyHitsOneBackend(c *C, upgrade bool) {
 	c.Assert(res.StatusCode, Equals, 503)
 	data, err := ioutil.ReadAll(res.Body)
 	c.Assert(err, IsNil)
-	c.Assert(string(data), Equals, "Service Unavailable\n")
+	c.Assert(string(data), Equals, http.StatusText(http.StatusServiceUnavailable))
```
bgentry: missing newline

router/proxy/export.go
```diff
@@ -0,0 +1,3 @@
+package proxy
+
+var StickyCookie = stickyCookie
```
bgentry: Should this be a constant instead of a var? Was this intentionally left in a separate file like this?

Seems weird that this can be globally overridden.

router/http.go
```diff
@@ -356,310 +345,27 @@ type httpService struct {
...
-	copyHeader(w.Header(), res.Header)
-
-	w.WriteHeader(res.StatusCode)
-	w.(http.Flusher).Flush()
-	_, err = io.Copy(httphelper.FlushWriter{Writer: w, Enabled: true}, res.Body) // TODO(bgentry): consider using a flush interval
-	if err != nil {
-		log.Println("reverse proxy copy err:", err)
-		return
+	if isWebSocketUpgrade(req) {
+		s.proxy.ServeWebSocket(w, req)
```
bgentry: I still don't think that this interface should be exposed to users of the `Proxy` type. These should be internal implementation details for that type.

Why not stick with the `http.Handler` interface and let the implementation of `Proxy` choose to be smart enough to handle conn upgrades?

router/http.go
```diff
@@ -356,310 +345,27 @@ type httpService struct {
...
-		backend string
-		res     *http.Response
-		err     error
-	)
-
-	for _, backend = range addrs {
-		// TODO: limit number of backends tried
-		// TODO: temporarily quarantine failing backends
+	//TODO(benburkert): only call this when the ss set is updated, instead of every request
+	s.proxy.UpdateBackends(s.sc.Addrs())
```
bgentry: I'm guessing it's going to be quite expensive to pass around the full list of backends every time one changes. But even more importantly, we already have that information in memory on another struct. It'd be better if we didn't have to duplicate the lists of backends on the proxy, and instead could look them up directly on the thing that's already storing them.

Ultimately that'd give us fewer copies of data to have to keep in sync, and fewer locks to deal with.

bgentry:
I'm not able to load all of this up into my head again right now, but at a glance I do still have most of the same concerns I did in the earlier revision ([comments here](https://github.com/benburkert/flynn/compare/flynn:master...benburkert:router-proxy-package#commit_comments_bucket)). I'm concerned about the multiple added layers of abstraction, new interfaces, and big increase in lines of code.

I also haven't spent any time understanding the TCP routing stuff yet, so I can't review those changes.

I'll try to give it another full look tomorrow to see if I can better understand why you've chosen these particular interfaces.

router/http.go
titanous: +1

We use an upgraded connection that is not a websocket to do bidirectional attachment to the stdin/stderr/stdout of containers currently.

router/http.go
benburkert: Keeping the "protocol detection" step outside of the proxy package allows each implementation of `Proxy` to pick and choose which protocols it supports. If the websocket handshake check is done in the router, then any transport that supported HTTP would be obligated to support WebSockets. This allows for proxy implementations that only apply to a small number of protos (like `StickyProxy`), and generic ones that work for all protos (like `DynamicProxy`).

I played around with adding an interface type per proto, but that lead to an explosion of additional types, and it's too soon to add them in IMO, but that may be the way to go in the future. For right now, this means: 1) a single `Proxy` interface, 2) allowing implementations to pick and choose protos, 3) the burden is on the user of the package to pick the right instance of `Proxy`.

router/http.go
benburkert: I agree with the concern for duplicating this list, but would like to see the list of backends kept in the proxy for two reasons:

Keeping the `discoverd` client (or a callback func to discoverd) outside of the proxy simplifies testing. Updates to the backend list would not have to be pushed through `discoverd`.

The proxy code that accesses the backend list is much more sensitive to latency than the discoverd code that updates the list. One of these two components must traverse a few structs to access that list, IMO it should be the thing updating the list.

router/proxy/export.go
benburkert: yeah, it's on my TODO list to alter the tests so that this exported var/const is not necessary.

router/http.go
benburkert: Ahh, ok. I'll switch it to detect generic upgrade request here and in `ReverseProxy.ServeWebSocket`.

bgentry: also maybe rename that to `ServeUpgrade`?

benburkert: :+1:

bgentry:
just FYI, @benburkert and I spent a bunch of time going over this PR this morning and are both working on refactoring a bunch of things. Consider this a WIP for now.

benburkert:
@bgentry please have a look at the last commit WRT #826.

bgentry:
@benburkert well, the header tests pass, so that's a :+1: as far as I'm concerned right now :) I updated my one commit ( f00f89e71c1e7a61289039c9892cd0476ab06fdf ) to be cherry-picked on top of your latest: https://github.com/bgentry/flynn/compare/benburkert:router-proxy-package...benburkert-router-proxy-package-transport-refactor

I'm guessing you rebased on master and pulled in all of the tests that were added in the last few days?

bgentry:
@titanous we just finished another big pass pairing on this and I think it's in a much better place. There are still a few more minor items left (see the TODOs list), but structurally I think this is pretty close to where we want it to end up.

Feel free to take a look through the diff and make any notes you want. We'll be picking back up on this right after lunch and should have it finished pretty soon.

router/proxy/reverseproxy.go
```diff
@@ -0,0 +1,335 @@
...
+// backends, a stickyKey for encrypting sticky session cookies, and a flag
+// sticky to enable sticky sessions.
+func NewReverseProxy(backends []string, stickyKey [32]byte, sticky bool) *ReverseProxy {
+	return &ReverseProxy{
+		transport: &transport{
+			backends:          backends,
+			stickyCookieKey:   stickyKey,
+			useStickySessions: sticky,
+		},
+		FlushInterval: 50 * time.Millisecond,
```
benburkert: 50ms is probably too high, any recommendation here?

benburkert:
@titanous this PR is functionally complete, we're just working on squashing the number of commits down, but it's ready for :eyes:.

router/http_test.go
```diff
@@ -596,24 +591,37 @@ func (s *S) TestStickyHTTPRoute(c *C) {
...
 	unregister := discoverdRegisterHTTP(c, l, srv1.Listener.Addr().String())
 
-	cookie := assertGet(c, "http://"+l.Addr, "example.com", "1")
+	cookies := assertGet(c, "http://"+l.Addr, "example.com", "1")
 	discoverdRegisterHTTP(c, l, srv2.Listener.Addr().String())
 	for i := 0; i < 10; i++ {
-		resCookie := assertGetCookie(c, "http://"+l.Addr, "example.com", "1", cookie)
-		c.Assert(resCookie, IsNil)
+		resCookies := assertGetCookies(c, "http://"+l.Addr, "example.com", "1", cookies)
+		c.Assert(len(resCookies), Equals, 0)
```
titanous:
```go
c.Assert(resCookies, HasLen, 0)
```

router/http_test.go
```diff
@@ -648,32 +656,33 @@ func (s *S) TestStickyHTTPRouteWebsocket(c *C) {
...
-			for _, c := range res.Cookies() {
-				if c.Name == stickyCookie {
-					cookieSet = true
-					sessionCookie = c
-				}
+			if len(res.Cookies()) > 0 {
+				//TODO(benburkert): instead of assuming that a session cookie
+				//is set if a response has cookies, switch back to checking for
+				//the session cookie once this test can access
+				//proxy.stickyCookie
```
titanous: nit: Add spaces between the `//` and the text.

router/http_test.go
titanous: Do we have a test for a non-websocket upgrade? (ie "Hijack")

router/proxy/reverseproxy.go
```diff
@@ -0,0 +1,332 @@
...
+const (
+	stickyCookie = "_backend"
+)
+
+// onExitFlushLoop is a callback set by tests to detect the state of the
+// flushLoop() goroutine.
+var onExitFlushLoop func()
+
+// Hop-by-hop headers. These are removed when sent to the backend.
+// http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html
```
titanous: This RFC is deprecated, please update the comment/link.

router/proxy/reverseproxy.go
```diff
@@ -0,0 +1,332 @@
...
+// backends, a stickyKey for encrypting sticky session cookies, and a flag
+// sticky to enable sticky sessions.
+func NewReverseProxy(bf BackendListFunc, stickyKey [32]byte, sticky bool) *ReverseProxy {
+	return &ReverseProxy{
+		transport: &transport{
+			getBackends:       bf,
+			stickyCookieKey:   stickyKey,
+			useStickySessions: sticky,
+		},
+		FlushInterval: 50 * time.Millisecond,
```
titanous: Maybe drop this to 10ms? I'm not certain what it should be.

router/proxy/reverseproxy.go
```diff
@@ -0,0 +1,332 @@
...
+	if transport == nil {
+		panic("router: nil transport for proxy")
+	}
+	defer conn.Close()
+
+	dconn, err := transport.Connect(conn.RemoteAddr())
+	if err != nil {
+		panic(err)
+		p.logf("router: proxy error: %v", err)
+		return
```
titanous: Should this panic?

router/proxy/reverseproxy.go
```diff
@@ -0,0 +1,332 @@
...
+	if res.StatusCode != 101 {
+		return
+	}
+
+	hj, ok := rw.(http.Hijacker)
+	if !ok {
+		p.logf("router: hijack failed")
+		return
+	}
+	dconn, bufrw, err := hj.Hijack()
```
titanous: Drop the type check and do an unchecked assertion here, the panic will have detail and it should never happen at runtime if the use is correct.

router/proxy/transport.go
```diff
@@ -0,0 +1,201 @@
...
+
+func decrypt(data []byte, key [32]byte) []byte {
+	var nonce [24]byte
+	if len(data) < len(nonce) {
+		return []byte{}
+	}
+	copy(nonce[:], data)
+	res, ok := secretbox.Open(nil, data[len(nonce):], &nonce, &key)
+	if !ok {
+		return []byte{}
```
titanous: `return nil` for both of these cases?

router/tcp.go
```diff
@@ -170,9 +171,11 @@ func (h *tcpSyncHandler) Set(data *router.Route) error {
 		if err != nil {
 			return err
 		}
+		var stickyKey [32]byte // TODO: change constructor to not need this
```
titanous: Please fix this TODO before merging.

router/proxy/transport.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		if _, ok := err.(dialErr); !ok {
+			return nil, err
+		}
+		// retry, maybe log a message about it
+	}
+	return nil, errNoBackends
+}
+
+func (t *transport) Connect(_ net.Addr) (net.Conn, error) {
```
titanous: Name the method argument `clientAddr` or similar to avoid confusion, I wasn't sure what it was until I looked at usage.

router/proxy/transport.go
```diff
@@ -0,0 +1,201 @@
...
+	return res, conn, nil
+}
+
+func dialTCP(addrs []string) (*net.TCPConn, string, error) {
+	for _, addr := range addrs {
+		if conn, err := dialer.Dial("tcp", addr); err == nil {
+			if tcpConn, ok := conn.(*net.TCPConn); ok {
+				return tcpConn, addr, nil
+			}
+			panic("proxy: tcp dialer provided non-TCP connection")
```
titanous: Can just type assert without the ok check, the panic will have more detail.

router/proxy/reverseproxy.go
```diff
@@ -0,0 +1,332 @@
...
+	closeWrite(dconn)
+	<-done
+}
+
+func prepareRequest(req *http.Request) *http.Request {
+	outreq := new(http.Request)
+	*outreq = *req // includes shallow copies of maps, but okay
+
+	// Pass the Request-URI verbatim without any modifications
+	outreq.URL.Opaque = strings.Split(strings.TrimPrefix(req.RequestURI, req.URL.Scheme+":"), "?")[0]
```
titanous: Do we have a test that exercises query params? What about full URL in the Request URI?

router/proxy/reverseproxy.go
```diff
@@ -0,0 +1,332 @@
...
+type closeWriter interface {
+	CloseWrite() error
+}
+
+func closeWrite(conn net.Conn) {
+	if cw, ok := conn.(closeWriter); ok {
+		cw.CloseWrite()
+	} else {
+		conn.Close()
+	}
```
titanous: Is there a situation where we don't have `CloseWrite`?

router/proxy/reverseproxy.go
```diff
@@ -0,0 +1,332 @@
...
+	if err != nil {
+		p.logf("router: hijack failed: %v", err)
+		return
+	}
+	joinConns(uconn, &streamConn{bufrw.Reader, dconn})
+}
+
+func (p *ReverseProxy) writeResponse(rw http.ResponseWriter, res *http.Response) {
+	// remove hop-by-hop headers as specified in connection options.
+	for _, h := range hopHeaders {
```
titanous: These are the global hop-by-hop headers, not the headers from the connection options, comment is wrong?

titanous:
First pass review done, overall this looks great!

router/http_test.go
bgentry: Fixed.

router/http_test.go
bgentry: Fixed.

router/proxy/reverseproxy.go
bgentry: yeah, this crept back in during the branch. Fixed.

bgentry:
@benburkert FYI there are still a number of commits that are somewhat broken from rebasing here (where the code won't even build at that commit due to missing types & stuff). May want to run the tests at each step and check those out.

Also the commit message in my one commit (0e73656 as of this writing) makes mention of removing stuff like the `StreamConn` interface which no longer exist.

I'll let you take care of the rest of these.

router/proxy/reverseproxy.go
bgentry: yeah, probably should do that in the commit where we switch to RFC7230-style connection header handling.

router/proxy/reverseproxy.go
bgentry: We were also thinking that 10ms might be a good default.

bgentry:
This PR loses 3 of TODOs that are present [in master](https://github.com/flynn/flynn/blob/3e3ea48ac7c6da0898d44e109d2fe0b368046260/router/http.go):

> * `// TODO: Proxy HTTP CONNECT? (example: Go RPC over HTTP)`
> * `// TODO: limit number of backends tried`
> * `// TODO: temporarily quarantine failing backends`

@titanous do you want these to be restored where appropriate, or would you rather track these in new Github issues?

titanous:
@bgentry GitHub issues are fine.

bgentry:
Gave this a fresh rebase to include #875.

router/proxy/reverseproxy.go
bgentry: There is already a test for the request URI (assuming this is what you mean): https://github.com/flynn/flynn/blob/2a0f1e7ac1f95fdd5585dd62e659210378f918c8/router/http_test.go#L921-L946

I've also just opened #876 with a test to guarantee query params.

router/http_test.go
benburkert: I've add `TestHTTPHijackUpgrade` that uses `httpclient.Hijack`.

router/proxy/reverseproxy.go
benburkert: yes, `streamConn` is no longer a `closeWriter`. Previously, it had a shim `CloseWrite()` that just called `Close()`.

benburkert:
ok, i finished squashing down a reasonable 18 :stuck_out_tongue:

I also addressed (hopefully) all of @titanous` feedback, I think this is ready!

router/http_test.go
```diff
@@ -1059,3 +1054,42 @@ func (s *S) TestHTTPResponseStreaming(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(string(buf), Equals, "a")
 }
+
+func pongHijackHandler(rw http.ResponseWriter, req *http.Request) {
+}
```
titanous: This doesn't appear to be used?

router/http_test.go
```diff
@@ -1059,3 +1054,42 @@ func (s *S) TestHTTPResponseStreaming(c *C) {
...
+
+	rwc, err := client.Hijack("GET", "/", nil, nil)
+	c.Assert(err, IsNil)
+
+	rwc.Write([]byte("ping!\n"))
+
+	pong, err := ioutil.ReadAll(rwc)
+	c.Assert(err, IsNil)
+	c.Assert(string(pong), Equals, "pong!\n")
+}
```
titanous: This doesn't appear to test the router?

benburkert: lol, yup, on it.

benburkert: should be fixed now.

titanous:
LGTM! :tropical_drink:

benburkert:
:metal: :beers:

bgentry:
:beers: and :tropical_drink::tropical_drink:  are deserved for this one

bgentry:
@benburkert did you make sure there are issues filed for the TODOs that were removed?

benburkert:
@bgentry opened as #886, #887, & #888.

PR #844, opened 2015-01-27
CI fixes

lmars:


test/cluster/cluster.go
```diff
@@ -500,6 +502,19 @@ func lookupUser(name string) (int, int, error) {
...
 
 func (c *Cluster) DumpLogs(w io.Writer) {
+	done := make(chan struct{})
+	go func() {
+		c.dumpLogs(w)
+		close(done)
+	}()
+	select {
+	case <-done:
+	case <-time.After(60 * time.Second):
```
heavenlyhash: This timeout covers all the log gather attempts at once... do we need the big one, if the timeout per ssh connection is now low?

titanous: The ssh timeout only covers the connect time, this covers execution time too.

heavenlyhash: OH.... so if the dialSSH call times out or errors once, it doesn't stop future runs from trying and timing out again.  Maybe that's not the greatest behavior, maybe we should just set an "error and stop trying" flag for that?  WDYT?

titanous: I think this is fine for now.

heavenlyhash:
LGTM

titanous:
LGTM

archSeer:
LGTM

PR #846, opened 2015-01-27
controller: Validate input through json schemas

jvatic:


jvatic:
refs #744, #528

archSeer:
Very neat!

controller/controller.go
```diff
@@ -121,6 +121,8 @@ func respondWithError(w http.ResponseWriter, err error) {
 }
 
 func appHandler(c handlerConfig) http.Handler {
+	loadSchemas()
```
archSeer: What do you think about renaming the package to `schema` or `schemas`? Then we can do `schema.Load()`

controller/jobs.go
```diff
@@ -237,6 +237,13 @@ func (c *controllerAPI) PutJob(ctx context.Context, w http.ResponseWriter, req *
 	}
 
 	job.AppID = app.ID
+
+	schema := schemaForType(ct.Job{})
```
archSeer: What do you think about a `schema.Validate(thing interface{})` that would join the two calls?

```go
func Validate(thing interface{}) err {
    schema := schemaForType(thing)
    if err := validateInput(schema, thing); err != nil {
        return
    }
}
```

controller/controller.go
jvatic: Ideally this will end up in `pkg/schema`, but I don't think it makes sense to separate it form the controller in it's current state.

jvatic:
- [ ] Copy schemas into the container

controller/controller.go
archSeer: Well I'd keep it in the same location, just change the package declaration at the top of the file to `schema`

jvatic:
Screenshot of invalid app example in docs:
screen shot 2015-01-29 at 10 57 27 am

jvatic:
@archSeer comments addressed

controller/schema/schema.go
```diff
@@ -0,0 +1,98 @@
...
+		return schemaCache["https://flynn.io/schema/router/route"]
+	}
+	cacheKey := "https://flynn.io/schema/controller/" + name
+	return schemaCache[cacheKey]
+}
+
+func Validate(thing interface{}) error {
+	schema := schemaForType(thing)
+	if schema == nil {
+		return errors.New(fmt.Sprintf("Unknown resource: %T %v", thing, thing))
```
titanous: Needs a `schema: ` prefix.

jvatic:
@titanous comment addressed

titanous:
LGTM :tropical_drink:

PR #848, opened 2015-01-27
controller/deployer: Test transitioning

archSeer:


test/test_deployer.go
```diff
@@ -144,7 +165,55 @@ func (s *DeployerSuite) TestRollback(t *c.C) {
 	// check that the old formation is the same and there's no new formation
 	f, err := s.controllerClient(t).GetFormation(deployment.AppID, oldReleaseID)
 	t.Assert(err, c.IsNil)
-	t.Assert(f.Processes, c.DeepEquals, map[string]int{"crasher": 2})
+	t.Assert(f.Processes, c.DeepEquals, map[string]int{"failer": 2})
 	_, err = s.controllerClient(t).GetFormation(deployment.AppID, releaseID)
 	t.Assert(err, c.NotNil)
 }
+
+func (s *DeployerSuite) TestTransitioning(t *c.C) {
```
lmars: I think `TestNoDowntime` is a bit more descriptive.

test/test_deployer.go
```diff
@@ -144,7 +165,55 @@ func (s *DeployerSuite) TestRollback(t *c.C) {
...
+	app, _ := s.createApp(t)
+	route := "ping-transitioning.dev"
+	s.createFormation(t, app, map[string]int{"ping": 1})
+	t.Assert(flynn(t, "/", "-a", app.Name, "route", "add", "http", "-s", "ping-service", route), Succeeds)
+	<-time.After(1 * time.Second) // give the job a bit of time to start
```
lmars: This should wait for the service to appear in service discovery.

test/test_deployer.go
```diff
@@ -144,7 +165,55 @@ func (s *DeployerSuite) TestRollback(t *c.C) {
...
+		expected := []*ct.DeploymentEvent{
+			{ReleaseID: releaseID, JobType: "ping", JobState: "starting", Status: "running"},
+			{ReleaseID: releaseID, JobType: "ping", JobState: "up", Status: "running"},
+			{ReleaseID: oldReleaseID, JobType: "ping", JobState: "stopping", Status: "running"},
+			{ReleaseID: oldReleaseID, JobType: "ping", JobState: "down", Status: "running"},
+			{ReleaseID: releaseID, JobType: "", JobState: "", Status: "complete"},
+		}
+
+		defer close(done)
+		waitForDeploymentEvents(t, events, expected)
```
lmars: As discussed, test assertions cannot be made inside a goroutine due to the way go-check works, so this will need to send errors via a channel to the main test goroutine.

test/test_deployer.go
```diff
@@ -144,7 +165,55 @@ func (s *DeployerSuite) TestRollback(t *c.C) {
...
+			client := &http.Client{}
+			req, err := http.NewRequest("GET", "http://"+routerIP, nil)
+			t.Assert(err, c.IsNil)
+			req.Host = route
+			res, err := client.Do(req)
+			t.Assert(err, c.IsNil)
+			res.Body.Close()
+			t.Assert(res.StatusCode, c.Equals, 200)
+		}
+	}
```
lmars: This should be wrapped in a timeout

archSeer:
@lmars Comments addressed.

test/test_deployer.go
```diff
@@ -144,7 +173,69 @@ func (s *DeployerSuite) TestRollback(t *c.C) {
...
+					fail <- err
+				}
+				req.Host = route
+				res, err := client.Do(req)
+				if err != nil {
+					fail <- err
+				}
+				res.Body.Close()
+				if res.StatusCode != 200 {
+					fail <- fmt.Errorf("Expected res.StatusCode to equal 200, got %s", res.StatusCode)
```
lmars: These failures should break the loop

test/test_deployer.go
```diff
@@ -116,22 +136,31 @@ func (s *DeployerSuite) TestAllAtOnceStrategy(t *c.C) {
...
+	   	for {
+	   		select {
+	   		case e := <-events:
+	   			if e.Status == "failed" {
+	   				break outer
+	   			}
+	   		case <-time.After(10 * time.Second):
+	   			t.Fatal("Timed out waiting for deployment event")
+	   		}
+	   	}*/
```
lmars: This comment needs removing

archSeer:
@lmars Fixed.

test/test_deployer.go
```diff
@@ -144,7 +166,71 @@ func (s *DeployerSuite) TestRollback(t *c.C) {
...
+				req.Host = route
+				res, err := client.Do(req)
+				if err != nil {
+					fail <- err
+					break
+				}
+				res.Body.Close()
+				if res.StatusCode != 200 {
+					fail <- fmt.Errorf("Expected res.StatusCode to equal 200, got %s", res.StatusCode)
+					break
```
lmars: These should break the outer loop

test/test_deployer.go
```diff
@@ -144,7 +162,71 @@ func (s *DeployerSuite) TestRollback(t *c.C) {
...
+		{ReleaseID: oldReleaseID, JobType: "ping", JobState: "down", Status: "running"},
+		{ReleaseID: releaseID, JobType: "", JobState: "", Status: "complete"},
+	}
+
+	waitForDeploymentEvents(t, events, expected)
+	close(done)
+	err = <-fail
+	if err != nil {
+		t.Error(err)
+	}
```
lmars:
```go
if err := <-fail; err != nil {
	t.Error(err)
}
```

lmars:
LGTM pending minor comment

lmars:
@archSeer as discussed offline there is a possibility that this test will fail if the HTTP request is in flight to the old job when it is stopped by the deployer, leading to the test receiving an unexpected EOF. 

One solution is to update the deployer to pause new requests for the service and wait for the job to finish processing current requests before killing it (e.g. via some router API endpoint), but given the requests could be arbitrarily long running, maybe dropping the connection and forcing the clients to reconnect is the correct approach for now?

titanous:
@lmars I think we can resolve this by not exiting the old job until all pending HTTP requests have been served. This is called "graceful shutdown" and the net/http library supports it.

lmars:
@titanous that works for standard HTTP requests, but what about persistent connections? And would a deployment wait for the requests to finish before moving on (e.g. in a one-by-one deployment)? If so, how would it detect that condition? We also need to make sure new requests don't go to the job we are waiting on too, is request pausing still possible with the router?

titanous:
> that works for standard HTTP requests, but what about persistent connections?

For some services it makes sense to break those connections, for others we need to wait for them to drain.

> And would a deployment wait for the requests to finish before moving on (e.g. in a one-by-one deployment)?

In the one-by-one case, absolutely, the user needs to be aware of this and plan accordingly.

> We also need to make sure new requests don't go to the job we are waiting on too, is request pausing still possible with the router?

No need to do request pausing, we just need to unregister from service discovery.

lmars:
@archSeer I think it makes sense to close this PR and open an issue to track making the changes to make this test pass.

archSeer:
@lmars Indeed, #966

PR #850, opened 2015-01-27
pass panics through shutdown

heavenlyhash:
The newly added consistent shutdown sequence defers shouldn't hide panics.  This change recovers them, does the usual shutdown sequences, and then re-raises the original panic (meaning we get all relevant stacks and the usual exit code for a panic).


archSeer:
LGTM

PR #851, opened 2015-01-27
controller/deployer: Initial release creates placeholder deployment

### LINKED ISSUE #845: deployer: Return value for CreateDeployment
Currently `CreateDeployment` returns an empty deployment in the case where we have our first release (since we always transition between two releases, and simply setting the initial formation will basically deploy the release). This was so that from the code, you can call `CreateDeployment`, and it will simply call the SetAppRelease if this was the initial release, or schedule a deployment for any other release -- you do not need to care whether this was an initial deployment or not.

After discussing with @jvatic, it is a bit weird from the API side (since we return `{}` in the first case), so we're considering the alternative of throwing an error, which would then need additional checks in the code:
```
err := client.CreateDeployment(appID, releaseID)
if err != nil && err != InitialRelease { // ignore InitialRelease error
    return err
}
```

archSeer:
Closes #845.

### DIFF
--- a/controller/deployment.go
+++ b/controller/deployment.go
@@ -32,19 +32,40 @@ func NewDeploymentRepo(db *postgres.DB, pgxpool *pgx.ConnPool) *DeploymentRepo {
 }
 
 func (r *DeploymentRepo) Add(data interface{}) error {
-	deployment := data.(*ct.Deployment)
-	if deployment.ID == "" {
-		deployment.ID = random.UUID()
+	d := data.(*ct.Deployment)
+	if d.ID == "" {
+		d.ID = random.UUID()
 	}
-	query := "INSERT INTO deployments (deployment_id, app_id, old_release_id, new_release_id, strategy) VALUES ($1, $2, $3, $4, $5) RETURNING created_at"
-	if err := r.db.QueryRow(query, deployment.ID, deployment.AppID, deployment.OldReleaseID, deployment.NewReleaseID, deployment.Strategy).Scan(&deployment.CreatedAt); err != nil {
+	var oldReleaseID *string
+	if d.OldReleaseID != "" {
+		oldReleaseID = &d.OldReleaseID
+	}
+	tx, err := r.db.Begin()
+	if err != nil {
 		return err
 	}
-	deployment.ID = postgres.CleanUUID(deployment.ID)
-	deployment.OldReleaseID = postgres.CleanUUID(deployment.OldReleaseID)
-	deployment.NewReleaseID = postgres.CleanUUID(deployment.NewReleaseID)
+	query := "INSERT INTO deployments (deployment_id, app_id, old_release_id, new_release_id, strategy) VALUES ($1, $2, $3, $4, $5) RETURNING created_at"
+	if err := tx.QueryRow(query, d.ID, d.AppID, oldReleaseID, d.NewReleaseID, d.Strategy).Scan(&d.CreatedAt); err != nil {
+		tx.Rollback()
+		return err
+	}
+	d.ID = postgres.CleanUUID(d.ID)
+	d.OldReleaseID = postgres.CleanUUID(d.OldReleaseID)
+	d.NewReleaseID = postgres.CleanUUID(d.NewReleaseID)
 
-	args, err := json.Marshal(ct.DeployID{ID: deployment.ID})
+	// fake initial deployment
+	if d.FinishedAt != nil {
+		if _, err := tx.Exec("UPDATE deployments SET finished_at = $2 WHERE deployment_id = $1", d.ID, d.FinishedAt); err != nil {
+			tx.Rollback()
+			return err
+		}
+		return tx.Commit()
+	}
+	if err := tx.Commit(); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


lmars:
LGTM

controller/deployment.go
```diff
@@ -122,8 +122,11 @@ func (c *controllerAPI) CreateDeployment(ctx context.Context, w http.ResponseWri
 			respondWithError(w, err)
 			return
 		}
-		// empty ID means initial deploy
-		httphelper.JSON(w, 200, &ct.Deployment{})
+		// initial deploy
+		httphelper.Error(w, httphelper.JSONError{
+			Code:    httphelper.UnknownError,
```
titanous: UnknownError is incorrect. It should be a distinct error.

archSeer:
@titanous Comments addressed.

controller/deployment.go
```diff
@@ -30,19 +30,34 @@ func NewDeploymentRepo(db *postgres.DB, pgxpool *pgx.ConnPool) *DeploymentRepo {
...
-	query := "INSERT INTO deployments (deployment_id, app_id, old_release_id, new_release_id, strategy) VALUES ($1, $2, $3, $4, $5) RETURNING created_at"
-	if err := r.db.QueryRow(query, deployment.ID, deployment.AppID, deployment.OldReleaseID, deployment.NewReleaseID, deployment.Strategy).Scan(&deployment.CreatedAt); err != nil {
-		return err
+	if d.OldReleaseID == "" {
+		query := "INSERT INTO deployments (deployment_id, app_id, new_release_id, strategy) VALUES ($1, $2, $3, $4) RETURNING created_at"
+		if err := r.db.QueryRow(query, d.ID, d.AppID, d.NewReleaseID, d.Strategy).Scan(&d.CreatedAt); err != nil {
+			return err
+		}
+	} else {
+		query := "INSERT INTO deployments (deployment_id, app_id, old_release_id, new_release_id, strategy) VALUES ($1, $2, $3, $4, $5) RETURNING created_at"
```
titanous: You can avoid this by passing a **string which will insert NULL if the inner pointer is nil. We do this for artifact_ids on releases.

titanous:
This needs tests.

archSeer:
@titanous Tests are there, I just forgot to update them. Fixed.

controller/deployment.go
```diff
@@ -30,19 +30,31 @@ func NewDeploymentRepo(db *postgres.DB, pgxpool *pgx.ConnPool) *DeploymentRepo {
...
-	deployment.OldReleaseID = postgres.CleanUUID(deployment.OldReleaseID)
-	deployment.NewReleaseID = postgres.CleanUUID(deployment.NewReleaseID)
+	d.ID = postgres.CleanUUID(d.ID)
+	d.OldReleaseID = postgres.CleanUUID(d.OldReleaseID)
+	d.NewReleaseID = postgres.CleanUUID(d.NewReleaseID)
 
-	args, err := json.Marshal(ct.DeployID{ID: deployment.ID})
+	// fake initial deployment
+	if d.FinishedAt != nil {
+		if err := r.db.Exec("UPDATE deployments SET finished_at = $2 WHERE deployment_id = $1", d.ID, d.FinishedAt); err != nil {
```
titanous: This should be done in a transaction.

archSeer:
@titanous Changed to a transaction, but now everything stalls on the second call for some reason.

controller/deployment.go
```diff
@@ -32,19 +32,40 @@ func NewDeploymentRepo(db *postgres.DB, pgxpool *pgx.ConnPool) *DeploymentRepo {
...
+	d.OldReleaseID = postgres.CleanUUID(d.OldReleaseID)
+	d.NewReleaseID = postgres.CleanUUID(d.NewReleaseID)
+
+	// fake initial deployment
+	if d.FinishedAt != nil {
+		if _, err := tx.Exec("UPDATE deployments SET finished_at = $2 WHERE deployment_id = $1", d.ID, d.FinishedAt); err != nil {
+			tx.Rollback()
+			return err
+		}
+		return nil
```
titanous: This return is before the transaction has been commited.

archSeer:
@titanous Fixed.

controller/deployment.go
```diff
@@ -32,19 +32,43 @@ func NewDeploymentRepo(db *postgres.DB, pgxpool *pgx.ConnPool) *DeploymentRepo {
...
+	d.NewReleaseID = postgres.CleanUUID(d.NewReleaseID)
 
-	args, err := json.Marshal(ct.DeployID{ID: deployment.ID})
+	// fake initial deployment
+	if d.FinishedAt != nil {
+		if _, err := tx.Exec("UPDATE deployments SET finished_at = $2 WHERE deployment_id = $1", d.ID, d.FinishedAt); err != nil {
+			tx.Rollback()
+			return err
+		}
+		if err := tx.Commit(); err != nil {
```
titanous: Can just `return tx.Commit()`

archSeer:
@titanous ^

titanous:
LGTM, please rebase before merging.

PR #853, opened 2015-01-28
controller: Simplify resource testing.

archSeer:
Creating another appHandler was used back when we reconfigured a few things
using the old discoverd version.


heavenlyhash:
LGTM

Issue #753, opened 2015-01-17
Flynn CI failures


lmars:
Most Flynn CI builds are currently failing with errors like `cluster: no servers found`.

The errors happen in a variety of tests, for example:

* [b15bd36b](https://s3.amazonaws.com/flynn-ci-logs/20150117011022-b15bd36b-build-9ebc30ffdbd45dd0ce90e61ad257679f6e898bd4-2015-01-17-01-18-45.txt) - 2 `SchedulerSuite` tests
* [65d30a41](https://s3.amazonaws.com/flynn-ci-logs/20150117010508-65d30a41-build-e96ce72124b8bf13e74e6c0ff85443f2efd8cd32-2015-01-17-01-07-19.txt) - all of the `GitDeploySuite` and `SchedulerSuite` tests
* [5141df62](https://s3.amazonaws.com/flynn-ci-logs/20150117005500-5141df62-build-9a27063fbaf38730fd6678ccf50f5026ec2d96cb-2015-01-17-01-05-25.txt) - `CLISuite.TestLogFollow` and `SchedulerSuite.TestOmniJobs`

The changes which triggered these builds are unrelated.

archSeer:
`HostSuite.TestNetworkedPersistentJob` failed intermittently locally:

```
----------------------------------------------------------------------
FAIL: test_host.go:62: HostSuite.TestNetworkedPersistentJob

test_host.go:85:
    t.Assert(err, c.IsNil)
... value *errors.errorString = &errors.errorString{s:"discover: timeout exceeded"} ("discover: timeout exceeded")
```

lmars:
Sometimes `TestOmniJobs` fails to start new hosts and this is because the flynn-host daemon exits with `cluster: resource not found` when [trying to register itself](https://github.com/flynn/flynn/blob/master/host/host.go#L313), so the test never receives a "host added" event.

titanous:
@lmars that is fixed in #896

titanous:
Closing in favor of individual issues with the "ci-flake" tag. If you see a unrelated failure on CI please look for an issue and add a link or create a new one if you can't find a relevant one.

PR #857, opened 2015-01-28
test/cluster: Use virtio network interface

titanous:
Refs #753

This resolves a long-standing networking instability issue on CI. The default qemu NIC driver is apparently broken and was dropping/mangling packets right and left. This patch changes the NIC driver to virtio which is used by everybody and works correctly.

The problem and fix were extensively verified by running a program that exercised the network by making many connections and transferring many bytes.

Unfortunately this does not appear to resolve all of the intermittent CI failures, but it's a step forward.


archSeer:
LGTM

lmars:
LGTM

PR #858, opened 2015-01-28
util/packer: Don't attempt to copy AMI to eu-central-1

titanous:
The eu-central-1 region requires v4 signatures, which are unsupported by packer and its fork of goamz. Trying to copy to this region results in an AuthError and the release does not continue.

See https://github.com/mitchellh/packer/issues/1646 for the upstream issue tracking this.


archSeer:
LGTM

Issue #190, opened 2014-09-03
cli: make dealing with multiple clusters better


titanous:
A few things I'd like to be able to do:

- [ ] Use a flag to specify the cluster I'm talking to (currently this is only available in the `FLYNN_CLUSTER` env var)
- [ ] Add a force flag so that I can replace a cluster without removing it first
- [ ] Change which cluster is the default

titanous:
Closed by #859.

PR #859, opened 2015-01-28
Cli multiple clusters

geovanisouza92:
Solves #190

### DIFF
--- a/cli/cluster.go
+++ b/cli/cluster.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"errors"
+	"fmt"
 	"log"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
@@ -10,20 +12,24 @@ import (
 func init() {
 	register("cluster", runCluster, `
 usage: flynn cluster
-       flynn cluster add [-g <githost>] [-p <tlspin>] <cluster-name> <url> <key>
+       flynn cluster add [-f] [-d] [-g <githost>] [-p <tlspin>] <cluster-name> <url> <key>
        flynn cluster remove <cluster-name>
+       flynn cluster default [<cluster-name>]
 
 Manage clusters in the ~/.flynnrc configuration file.
 
 Options:
+	-f, --force               force add cluster
+	-d, --default             set as default cluster
 	-g, --git-host <githost>  git host (if host differs from api URL host)
 	-p, --tls-pin <tlspin>    SHA256 of the cluster's TLS cert (useful if it is self-signed)
 
 Commands:
 	With no arguments, shows a list of clusters.
 
-	add     adds a cluster to the ~/.flynnrc configuration file
-	remove  removes a cluster from the ~/.flynnrc configuration file
+	add      adds a cluster to the ~/.flynnrc configuration file
+	remove   removes a cluster from the ~/.flynnrc configuration file
+	default  set or print the default cluster
 
 Examples:
 
@@ -41,6 +47,8 @@ func runCluster(args *docopt.Args) error {
 		return runClusterAdd(args)
 	} else if args.Bool["remove"] {
 		return runClusterRemove(args)
+	} else if args.Bool["default"] {
+		return runClusterDefault(args)
 	}
 
 	w := tabWriter()
@@ -48,7 +56,11 @@ func runCluster(args *docopt.Args) error {
 
 	listRec(w, "NAME", "URL")
 	for _, s := range config.Clusters {
-		listRec(w, s.Name, s.URL)
+		if s.Name == config.Default {
+			listRec(w, s.Name, s.URL, "(default)")
+		} else {
+			listRec(w, s.Name, s.URL)
+		}
 	}
 	return nil
 }
@@ -61,14 +73,26 @@ func runClusterAdd(args *docopt.Args) error {
 		GitHost: args.String["--git-host"],
 		TLSPin:  args.String["--tls-pin"],
 	}
-	if err := config.Add(s); err != nil {
+
+	if err := config.Add(s, args.Bool["--force"]); err != nil {
 		return err
 	}
+
+	setDefault := args.Bool["--default"] || len(config.Clusters) == 1
+
+	if setDefault && !config.SetDefault(s.Name) {
+		return errors.New(fmt.Sprintf("Cluster %q does not exist and cannot be set as default.", s.Name))
+	}
+
 	if err := config.SaveTo(configPath()); err != nil {
 		return err
 	}
 
-	log.Printf("Cluster %q added.", s.Name)
+	if setDefault {
+		log.Printf("Cluster %q added and set as default.", s.Name)
+	} else {
+		log.Printf("Cluster %q added.", s.Name)
+	}
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


cli/cluster.go
```diff
@@ -85,3 +93,17 @@ func runClusterRemove(args *docopt.Args) error {
 
 	return nil
 }
+
+func runClusterDefault(args *docopt.Args) error {
+	name := args.String["<cluster-name>"]
+
+	if config.SetDefault(name) {
```
archSeer: Flip the conditional, so that we avoid indenting code:

```go
if !config.SetDefault(name) {
   return nil
}

if err := config.SaveTo(configPath()); err != nil {
   return err
}
...
```

cli/config/config.go
```diff
@@ -39,7 +40,7 @@ func (c *Config) Marshal() []byte {
 	return buf.Bytes()
 }
 
-func (c *Config) Add(s *Cluster) error {
+func (c *Config) Add(s *Cluster, f bool, d bool) error {
```
archSeer: Even though Golang usually prefers short variable names, I'd use `force` and `default` here, so it's a bit more obvious what's going on (`s` is a bit more straightforward since the type tells you what it is).

archSeer:
Nice work! I'd say we need a few integration tests (see `test/test_cli.go`) for the new behaviour.

cli/config/config.go
```diff
@@ -39,7 +40,7 @@ func (c *Config) Marshal() []byte {
 	return buf.Bytes()
 }
 
-func (c *Config) Add(s *Cluster) error {
+func (c *Config) Add(s *Cluster, force bool, set_as_default bool) error {
```
archSeer: `default`

test/test_cli.go
```diff
@@ -478,14 +480,22 @@ func (s *CLISuite) TestCluster(t *c.C) {
...
 	cfg, err := config.ReadFile(file.Name())
 	t.Assert(err, c.IsNil)
 	t.Assert(cfg.Clusters, c.HasLen, 1)
 	t.Assert(cfg.Clusters[0].Name, c.Equals, "test")
-	// overwriting should not work
+	// overwriting (without --force) should not work
 	t.Assert(flynn("cluster", "add", "test", "foo", "bar"), c.Not(Succeeds))
 	t.Assert(flynn("cluster"), OutputContains, "test")
+	// change default cluster
+	t.Assert(flynn("custer", "default", "test"), OutputContains, "Cluster test is default now.")
```
archSeer: custer -> cluster

geovanisouza92:
@archSeer This seems weird. The recent build reports that ```Config.Default``` is not set. Maybe you can give a look and see what I don't see?

archSeer:
The test failure is a bug in the TOML decoding library, the `Default` key is printed right after the `[[cluster]]`, and then parsed incorrectly. Switching [these two lines](https://github.com/geovanisouza92/flynn/blob/d08f41933687f08033503192581179c98b6990e5/cli/config/config.go#L22-L23), so that `Default` is first fixes it, I'll also try updating the library itself.

cli/cluster.go
```diff
@@ -61,7 +67,9 @@ func runClusterAdd(args *docopt.Args) error {
 		GitHost: args.String["--git-host"],
 		TLSPin:  args.String["--tls-pin"],
 	}
-	if err := config.Add(s); err != nil {
+	f := args.Bool["--force"]
+	d := args.Bool["--default"]
+	if err := config.Add(s, f, d); err != nil {
```
archSeer: Instead of passing the default flag into `config.Add`, I'd call `config.SetDefault` right after calling `config.Add`:

```go
if err := config.Add(s, args.Bool["--force"]); err != nil {
  return err
}
if args.Bool["--default"] {
  if err := config.SetDefault(s.Name); err != nil {
    return err
  }
}
```

cli/cluster.go
```diff
@@ -85,3 +96,19 @@ func runClusterRemove(args *docopt.Args) error {
...
+	if !config.SetDefault(name) {
+		log.Printf("Cluster %q not found.", name)
+		return nil
+	}
+
+	if err := config.SaveTo(configPath()); err != nil {
+		return err
+	}
+
+	log.Printf("Cluster %q is default now.", name)
```
archSeer: `%q is now the default cluster.`

cli/cluster.go
```diff
@@ -69,6 +77,9 @@ func runClusterAdd(args *docopt.Args) error {
 	}
 
 	log.Printf("Cluster %q added.", s.Name)
+	if d {
+		log.Printf("Cluster %q is default now.", s.Name)
```
archSeer: I'd make it a single line, saying `Cluster %q added and set as default.`

archSeer:
Yes, the TOML bug is fixed on the latest version, so wait until #957 is merged, then rebase on master.

archSeer:
Also, it'd be good to know which cluster is the default, so `flynn cluster default` should print out the current default, and maybe `flynn cluster` could print a `(default)` at the end of the row for the default cluster. Other than that, seems to work fine :)

geovanisouza92:
Finally green! :smile:

archSeer:
@geovanisouza92 Great, can you squash these up into a single commit? :)

/cc @flynn/devs for a review

cli/cluster.go
```diff
@@ -10,20 +11,24 @@ import (
...
 	-p, --tls-pin <tlspin>    SHA256 of the cluster's TLS cert (useful if it is self-signed)
 
 Commands:
 	With no arguments, shows a list of clusters.
 
-	add     adds a cluster to the ~/.flynnrc configuration file
-	remove  removes a cluster from the ~/.flynnrc configuration file
+	add      adds a cluster to the ~/.flynnrc configuration file
+	remove   removes a cluster from the ~/.flynnrc configuration file
+	default  set the default cluster
```
lmars: `set or print the default cluster`

cli/cluster.go
```diff
@@ -61,14 +72,26 @@ func runClusterAdd(args *docopt.Args) error {
...
-	if err := config.Add(s); err != nil {
+
+	if err := config.Add(s, args.Bool["--force"]); err != nil {
 		return err
 	}
+
+	setDefault := args.Bool["--default"] || len(config.Clusters) == 1
+
+	if setDefault && !config.SetDefault(s.Name) {
+		return errors.New("Cluster cannot be set as default.")
```
titanous: This error is only shown if the cluster doesn't exist, right? If so, the message should reflect that.

cli/config/config.go
```diff
@@ -52,19 +53,34 @@ func (c *Config) Add(s *Cluster) error {
...
+			// Remove existing match
+			c.Clusters = append(c.Clusters[:i], c.Clusters[i+1:]...)
+			msg = ""
 		}
 	}
 
-	c.Clusters = append(c.Clusters, s)
+	// Did not match
+	if msg == "" {
+		c.Clusters = append(c.Clusters, s)
```
titanous: Is there any reason for the if check here?

cli/config/config.go
```diff
@@ -52,19 +53,34 @@ func (c *Config) Add(s *Cluster) error {
 		}
 	}
 
-	for _, existing := range c.Clusters {
-		if existing.Name == s.Name {
-			return fmt.Errorf("Cluster %q already exists in ~/.flynnrc", s.Name)
+	msg := ""
+	for i, existing := range c.Clusters {
```
titanous: Looks like `msg` can be scoped inside this loop?

titanous:
This looks good! It can be merged after my minor comments are addressed and rebased on top of master.

geovanisouza92:
@titanous RFC :sparkles:

cli/config/config.go
```diff
@@ -52,19 +53,33 @@ func (c *Config) Add(s *Cluster) error {
...
+
+		// The new cluster config match with existing one
+		if msg != "" {
+			if !force {
+				return fmt.Errorf(msg)
+			}
+
+			// Remove existing match
+			c.Clusters = append(c.Clusters[:i], c.Clusters[i+1:]...)
+			msg = ""
```
archSeer: No need to do this anymore.

archSeer:
@geovanisouza92 One last change, then we merge! :)

geovanisouza92:
Code changed, rebased on master, but fails on ```test_deployer.go:175```, apparently nothing related to PR.

Anyway, considering that this is my 2nd Go code, I think that's awesome, thanks for you patience and guidance! :smile:

titanous:
LGTM, thanks! (I restarted the CI build)

PR #861, opened 2015-01-28
util/packer,test/rootfs: Install kernel 3.16.0

titanous:
The default trusty kernel is 3.13, but the backported “LTS Enablement” package is already available and will be the default in 14.04.2


archSeer:
LGTM

lmars:
LGTM

PR #862, opened 2015-01-28
router: properly determine fwd port from listener

bgentry:
Rather than just pulling the forwarding port out of the input address,
determine it based off of the listener's actual address. This makes sure
that the value isn't 0, even in tests where that was previously the
case.

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -223,23 +223,18 @@ func (h *httpSyncHandler) Remove(id string) error {
 }
 
 func (s *HTTPListener) listenAndServe(started chan<- error) {
-	_, port, err := net.SplitHostPort(s.Addr)
-	if err != nil {
-		started <- err
-		return
-	}
+	var err error
 	s.listener, err = reuseport.NewReusablePortListener("tcp4", s.Addr)
 	started <- err
 	if err != nil {
 		return
 	}
-
 	server := &http.Server{
 		Addr: s.listener.Addr().String(),
 		Handler: fwdProtoHandler{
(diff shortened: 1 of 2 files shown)


router/http.go
```diff
@@ -223,14 +223,15 @@ func (h *httpSyncHandler) Remove(id string) error {
...
+	var err error
+	s.listener, err = reuseport.NewReusablePortListener("tcp4", s.Addr)
+	started <- err
 	if err != nil {
-		started <- err
 		return
 	}
-	s.listener, err = reuseport.NewReusablePortListener("tcp4", s.Addr)
-	started <- err
+	_, port, err := net.SplitHostPort(s.listener.Addr().String())
```
bgentry: Maybe we don't need to handle this error here and can just get it from `portFromAddr()` as we're doing in the tests. Then in that func, we can just panic if we hit an error.

titanous: Yeah, that seems reasonable.

bgentry: updated.

router/http_test.go
```diff
@@ -55,6 +55,8 @@ func httpTestHandler(id string) http.Handler {
 	})
 }
 
+var httpNilHandler = http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {})
```
titanous: Where is this used?

bgentry: ugh, it's not, should have been deleted. Gone now.

titanous:
LGTM.

PR #865, opened 2015-01-29
Manifests and named volumes

heavenlyhash:
Create persistent volumes when evaluating manifests.  Add simple name->volume mapping so if the manifest is run again, services get the same volumes again if were created previously.

Using etcd appliance as canary.

Once this and other `volume.Manager` state is persisted via boltdb, this will allow etcd appliances to keep their disk state across flynn-host daemon reboots.

### DIFF
--- a/host/manifest_template.json
+++ b/host/manifest_template.json
@@ -4,7 +4,7 @@
     "image": "$image_url_prefix/etcd?id=$image_id[etcd]",
     "expose_env": ["ETCD_INITIAL_CLUSTER", "ETCD_INITIAL_CLUSTER_STATE", "ETCD_NAME", "ETCD_DISCOVERY", "ETCD_PROXY"],
     "args": [
-      "-data-dir={{ .Volume \"/data\" }}",
+      "-data-dir={{ .Volume \"flynn-etcd-data\" \"/data\" }}",
       "-advertise-client-urls=http://{{ .ExternalIP }}:{{ .TCPPort 0 }}",
       "-listen-client-urls=http://0.0.0.0:{{ .TCPPort 0 }}",
       "-initial-advertise-peer-urls=http://{{ .ExternalIP }}:{{ .TCPPort 1 }}",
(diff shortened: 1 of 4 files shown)


host/manifest_template.json
```diff
@@ -4,7 +4,7 @@
     "image": "$image_url_prefix/etcd?id=$image_id[etcd]",
     "expose_env": ["ETCD_INITIAL_CLUSTER", "ETCD_INITIAL_CLUSTER_STATE", "ETCD_NAME", "ETCD_DISCOVERY", "ETCD_PROXY"],
     "args": [
-      "-data-dir={{ .Volume \"/data\" }}",
+      "-data-dir={{ .Volume \"etcd-vol\" \"/data\" }}",
```
titanous: maybe call this `etcd-data`?

titanous:
This looks like a good approach. The only thing I wonder about is if we should carve out a `flynn/` or `flynn-` name prefix by convention?

heavenlyhash:
Since it's an internal service?  Yeah, that makes sense.  I'll apply "flynn/".

host/volume/manager.go
```diff
@@ -104,3 +115,23 @@ func (p managerProviderProxy) NewVolume() (Volume, error) {
...
+	Gets a reference to a volume by name if that exists; if no volume is so named,
+	it is created using the named provider (the zero string can be used to invoke
+	the default provider).
+*/
+func (m *Manager) CreateOrGetNamedVolume(name string, providerID string) (Volume, error) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	if v, ok := m.namedVolumes[name]; ok {
+		return m.volumes[v], nil
+	} else {
```
titanous: Drop the else and outdent.

heavenlyhash:
In the event of cluster-wide names in the future, I'm thinking we may end up with another manifest function that creates the volume name with a host-id {pre/suf}fix or something like that... but we're not there yet.

titanous:
Hmm, how does a slash in the name interact with the HTTP API?

titanous:
LGTM

heavenlyhash:
Slash in the name: so far, it's fine, because the context that's evaluating the manifest has direct access to the volume manager, so there's no interesting API trips.  I changed it to a dash anyway.  We should do an id sanitation filter in general.

archSeer:
LGTM

PR #866, opened 2015-01-29
Zfs file vdev recovery

heavenlyhash:
ZFS volume providers should re-import file vdevs if already existing.

This makes it possible for the system to easily reassemble itself after a host reboot in which zfs has forgotten about the zpools based on file vdevs.


titanous:
LGTM

PR #868, opened 2015-01-29
Set deployer worker interval

titanous:
If this is not set, it defaults to zero which will peg the CPU querying for jobs from postgres.

### DIFF
--- a/controller/deployer/main.go
+++ b/controller/deployer/main.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/bgentry/que-go"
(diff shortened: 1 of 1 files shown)


titanous:
Upstream fix at bgentry/que-go#4.

controller/deployer/main.go
```diff
@@ -46,8 +47,9 @@ func main() {
 	}
 
 	pgxpool, err := pgx.NewConnPool(pgx.ConnPoolConfig{
-		ConnConfig:   pgxcfg,
-		AfterConnect: que.PrepareStatements,
+		ConnConfig:     pgxcfg,
+		AfterConnect:   que.PrepareStatements,
+		MaxConnections: 10,
```
lmars: Is it worth putting this int in a constant and also using it when creating the worker pool?

titanous: Sure.

titanous: Done.

lmars:
LGTM

PR #870, opened 2015-01-29
test/scripts: Disable network debug logging

titanous:
We’ve eliminated most of the network instability, so this output is not useful.


heavenlyhash:
LGTM

PR #874, opened 2015-01-29
httphelper: Fix marshalling of JSONError

jvatic:
Calling `json.Marshal` on `JSONError` instead of `*JSONError` causes the
detail field to be base64 encoded.

https://play.golang.org/p/xzPIPOrfoO
https://play.golang.org/p/55jQt2QD_e


titanous:
LGTM

Issue #872, opened 2015-01-29
router can't retry request with body


titanous:
After looking at a few intermittent failures on CI ([1](https://s3.amazonaws.com/flynn-ci-logs/20150129052251-0fdd772b-build-d377f3932992bf601169ed11fe4c62063dcf30bb-2015-01-29-05-27-28.txt), [2](https://s3.amazonaws.com/flynn-ci-logs/20150129051901-2b18f75c-build-f990e4871b2ddfde4d0065788f67dbfb998f6e9f-2015-01-29-05-25-08.txt), [3](https://s3.amazonaws.com/flynn-ci-logs/20150129053909-00ef6c24-build-d14b2002006768b2de0368224ab35a1e7c7d7674-2015-01-29-05-43-42.txt)) I noticed a pattern. They are all failing POST requests right after some service instance churn.

It turns out they all have corresponding log lines that look like:

```text
2015/01/29 05:43:28 http: proxy error: http: invalid Read on closed Body
```

After looking at the code a bit, I think I know what is going on:

1. We hit a dial error when trying to contact a backend that just went down ([router code](https://github.com/flynn/flynn/blob/master/router/http.go#L487-L492))
1. As soon as `Transport.RoundTrip` encounters an error it closes `req.Body` (somewhere [around here](https://github.com/golang/go/blob/go1.4.1/src/net/http/transport.go#L220))
1. We try to `RoundTrip` to another backend, but `req.Body.Read` errors because it was closed by the first `RoundTrip` that failed to dial.
1. The request fails with a 503 and logs the above error.

I think we need to wrap the request body so that `Close` is a no-op and then manually close the body after the request is complete. We also need a few tests around this behavior (I don't think we have _any_ tests that make requests with inbound bodies currently).

/cc @benburkert @bgentry

bgentry:
Damn. Thanks for doing the digging on this, I think your analysis is spot on.

At one point I noticed that we were only testing simple `GET` requests and meant to fix that, but forgot about it when I was actually adding tests. Will get some good tests for this first thing in the morning.

PR #875, opened 2015-01-29
router: fix request retries with bodies

bgentry:
As reported in #872, http.Transport will call Close() on any Request.Body after it encounters any error. We are retrying the same request under certain circumstances (dial errors in particular), so we need to keep the body open until we're done retrying.

Adds some test coverage for the error case reported in #872 by also testing `POST` requests with bodies, and fixes the underlying issue as well.

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -438,6 +438,7 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	outreq.ProtoMajor = 1
 	outreq.ProtoMinor = 1
 	outreq.Close = false
+	outreq.Body = &fakeCloseReadCloser{outreq.Body}
 
 	// TODO: Proxy HTTP CONNECT? (example: Go RPC over HTTP)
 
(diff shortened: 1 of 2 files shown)


router/http.go
```diff
@@ -491,9 +492,11 @@ func (s *httpService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 				continue
 			}
 			log.Println("http: proxy error:", err)
+			outreq.Body.(*fakeCloseReadCloser).RealClose()
 			fail(w, 503)
 			return
 		}
+		outreq.Body.(*fakeCloseReadCloser).RealClose()
```
bgentry: I'm not super happy with the way this ends up being duplicated here, but didn't think it was worth the effort to try and factor out this retry loop in light of our other pending PR (#842). Maybe we should just make sure it's done more cleanly there?

titanous: Yeah, fixing this in the refactor is fine. Thanks for the quick bugfix!

titanous:
LGTM.

bgentry:
oh /cc @benburkert

bamarni:
I'm coming from a linked issue which is similar in traefik (https://github.com/containous/traefik/issues/1008) but I'm not sure this patch is enough.

It seems like even though subsequent reads wouldn't trigger errors, they wouldn't read anything either, or do I miss something?

titanous:
@bamarni If I remember correctly, the problem this was working around was the body being closed by `net/http` during a failed dial, before anything was read, so turning the close into a no-op is a valid solution.

bamarni:
@titanous : makes sense, thanks for the feedback.

PR #876, opened 2015-01-29
router: add a test for query params

bgentry:
Make sure that query params are kept intact as the request is proxied to the backend.


titanous:
LGTM

PR #878, opened 2015-01-29
Fix stack depth in shutdown log

heavenlyhash:
Just remove the remaining indirections of methods on the singleton struct.  Stack depth is now uniform again (the '3' offset for the log output is thus again correct).


titanous:
LGTM

PR #890, opened 2015-01-30
test: Add instance to local state when adding new hosts

titanous:
If we don't do this and need to dump the logs, they will not be dumped from the pending host.


lmars:
LGTM

PR #891, opened 2015-01-30
router: Add wait between retries if connection to etcd fails

titanous:
Without this patch, if/when etcd goes down it will fill up the logs with errors due to a tight retry loop.


lmars:
LGTM

PR #896, opened 2015-01-30
Fix discoverd leader election

titanous:
In the rewrite we dropped the workaround for coreos/etcd#407, which is still not fixed in etcd.

Original fix: f7a99cbe10cf3408c02ee7d9f016a4687e974fdf

This fixes issues adding hosts to an existing cluster, which was presenting intermittently on CI.


titanous:
ping @flynn/devs please take a look at this.

lmars:
LGTM

PR #897, opened 2015-01-30
test/runner: Redirect local log requests to log URL

### LINKED ISSUE #877: CI build log links should redirect to S3
If a build log is not available locally the request should 301 to the log on S3.

titanous:
If we've uploaded the log, the local log file will not be available, but we can redirect to the copy on S3.

Closes #877


lmars:
LGTM

PR #899, opened 2015-01-30
SSE abstraction

jvatic:


pkg/sse/sse.go
```diff
@@ -19,6 +21,18 @@ type Writer struct {
 	mtx sync.Mutex
 }
 
+func (w *Writer) writeInterface(data interface{}) error {
+	w.mtx.Lock()
+	defer w.mtx.Unlock()
+
+	wr := hh.FlushWriter{Writer: NewWriter(w.w), Enabled: true}
+	enc := json.NewEncoder(wr)
```
archSeer: I think all of this should be moved to `sse/stream.go`, so that we don't need to recreate the encoder on every write.

pkg/sse/stream.go
```diff
@@ -0,0 +1,52 @@
...
+	"net/http"
+	"reflect"
+)
+
+type Stream struct {
+	w    *Writer
+	done chan struct{}
+}
+
+func OpenStream(w http.ResponseWriter, ch interface{}) *Stream {
```
titanous: Call this `ServeStream`, it should probably handle `CloseNotify` too.

pkg/sse/stream.go
```diff
@@ -0,0 +1,52 @@
...
+			} else {
+				break
+			}
+		}
+	}()
+
+	return s
+}
+
+func (s *Stream) Send(data interface{}) error {
```
titanous: This method does not appear to be externally useful without a constructor that doesn't start the send loop?

pkg/sse/stream.go
```diff
@@ -0,0 +1,52 @@
...
+	return s.w.writeInterface(data)
+}
+
+func (s *Stream) Error(err error) {
+	s.w.Error(err)
```
titanous: This should stop the stream from sending first.

host/http.go
```diff
@@ -42,16 +41,8 @@ func (h *Host) streamEvents(id string, w http.ResponseWriter) error {
 		<-w.(http.CloseNotifier).CloseNotify()
```
archSeer: I'd also abstract the `CloseNotify` away and specify a cleanup function on the stream, like in my proposal [here](https://github.com/flynn/flynn/issues/744#issuecomment-70214292). I'd also probably simplify the implementation to just be a single blocking function -- since the channel we pass in is of type `chan interface{}`, to log out errors, we could simply detect the error type using reflection, and then write them out as an error.

pkg/sse/stream.go
```diff
@@ -0,0 +1,52 @@
...
+	w.WriteHeader(200)
+	sw.Flush()
+
+	s := &Stream{w: sw, done: make(chan struct{})}
+
+	chValue := reflect.ValueOf(ch)
+	go func() {
+		defer s.Close()
+		for {
+			if e, ok := chValue.Recv(); ok {
```
titanous: This should select on done.

host/http.go
titanous: I think an explicit `Error` or `CloseWithError` on the stream is better than using a `chan interface{}`.

jvatic:
@archSeer @titanous comments addressed

jvatic:
- [ ] Abstract SSE in [controller/jobs.go](https://github.com/flynn/flynn/blob/419b1702d881f8f18e895592015b6ccb9c176630/controller/jobs.go)

pkg/sse/stream.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+type Stream struct {
+	w       *Writer
+	done    chan struct{}
+	cleanup func()
+	closed  bool
+}
+
+func ServeStream(w http.ResponseWriter, ch interface{}, cleanup func()) *Stream {
```
titanous: Why the cleanup function? It'd be better if `Close` blocked until the stream sender is done.

jvatic:
- [ ] Wrap events: Full event details (id, name, data) should be included in the JSON for data events

pkg/sse/stream.go
jvatic: Yup, no need for it, `ServeStream` blocks so I moved the contents of all the cleanup funcs to after it returns.

pkg/sse/stream.go
```diff
@@ -0,0 +1,83 @@
...
+func (s *Stream) wait() {
+	<-s.done
+}
+
+func (s *Stream) Close() {
+	if !s.closed {
+		s.closed = true
+		close(s.done)
+	}
+}
```
titanous: `Close` should wait until the streamer exits.

archSeer:
I'd also hide away the [keep-alive calls](https://github.com/flynn/flynn/blob/ab546c9e574e6b358aca9a6953f766b08d2ca453/controller/jobs.go#L338). Every SSE stream should send that every 30s or so, and it won't interfere with anything since that is just a blank comment.

host/http.go
```diff
@@ -38,20 +37,8 @@ func (h *Host) StopJob(id string) error {
...
-	w.WriteHeader(200)
-	w.(http.Flusher).Flush()
-	for data := range ch {
-		if err := enc.Encode(data); err != nil {
-			return err
-		}
-		w.(http.Flusher).Flush()
-	}
+	sse.ServeStream(w, ch)
+	h.state.RemoveListener(id, ch)
```
archSeer: I'd put this in a defer before `ServeStream`

jvatic:
@archSeer @titanous ready for review again

controller/jobs.go
```diff
@@ -287,32 +266,30 @@ func (c *controllerAPI) JobLog(ctx context.Context, w http.ResponseWriter, req *
...
-		exit, err := attachClient.Receive(ssew.Stream("stdout"), ssew.Stream("stderr"))
+	if strings.Contains(req.Header.Get("Accept"), "text/event-stream") {
+		ch := make(chan *sseLogChunk)
+		s := sse.NewStream(w, ch)
+		defer s.Close()
+		s.Serve()
+		exit, err := attachClient.Receive(&sseLogStream{Name: "stdout", Chan: ch}, &sseLogStream{Name: "stderr", Chan: ch})
 		if err != nil {
-			fmt.Fprintf(fw, "event: error\ndata: %s\n\n", err)
+			ch <- &sseLogChunk{Event: "error", Data: []byte(fmt.Sprintf(`"%s"`, err.Error()))}
```
titanous: This method of serializing the error is unsafe, there could be special characters.

controller/jobs.go
```diff
@@ -27,45 +27,24 @@ import (
...
-	return len(p), err
+	s.Chan <- &sseLogChunk{Stream: s.Name, Data: data}
+	return len(p), nil
 }
 
-func (w *sseLogStreamWriter) Flush() {
-	w.w.Writer.Flush()
+type sseLogChunk struct {
+	Event  string          `json:"event,omitempty"`
+	Stream string          `json:"stream,omitempty"`
```
titanous: Can we just use `Event` to hold the data that currently goes in `Stream`?

controller/types/types.go
```diff
@@ -123,6 +124,10 @@ type DeploymentEvent struct {
 	CreatedAt    *time.Time `json:"created_at"`
 }
 
+func (de *DeploymentEvent) GetID() string {
+	return strconv.FormatInt(de.ID, 10)
```
titanous: `strconv.Itoa`

host/sampi/http.go
```diff
@@ -206,16 +194,10 @@ func (c *HTTPAPI) StreamHostEvents(w http.ResponseWriter, r *http.Request, ps ht
...
-		close(done)
-	}()
-	enc := startJSONEventStreaming(w)
 	l.Debug("streaming host events")
-	for data := range ch {
-		enc.Encode(data)
-	}
+	sse.ServeStream(w, ch)
+	l.Debug("http stream closed")
+	done <- true
```
titanous: Any reason this was changed from `close(done)`?

pkg/sse/stream.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (s *Stream) Serve() {
+	s.rw.Header().Set("Content-Type", "text/event-stream; charset=utf-8")
+	s.rw.WriteHeader(200)
+	s.w.Flush()
+
+	s.fw = hh.FlushWriter{Writer: newWriter(s.rw), Enabled: true}
+
+	go func() {
+		if cw, ok := s.rw.(http.CloseNotifier); ok {
```
titanous: Could move this condition outside the goroutine.

pkg/sse/stream.go
```diff
@@ -0,0 +1,134 @@
...
+			<-cw.CloseNotify()
+			s.Close()
+		}
+	}()
+
+	closeChanValue := reflect.ValueOf(s.closeChan)
+	chValue := reflect.ValueOf(s.ch)
```
titanous: These could be done inline in the select.

pkg/sse/stream.go
```diff
@@ -0,0 +1,134 @@
...
+			})
+			switch chosen {
+			case 0:
+				return
+			case 1:
+				s.sendKeepAlive()
+			default:
+				if ok {
+					if err := s.send(v.Interface()); err != nil {
+						fmt.Println("sse.(*Stream).Serve error: ", err)
```
titanous: probably shouldn't print here.

pkg/sse/stream.go
```diff
@@ -0,0 +1,134 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"reflect"
+	"time"
+
+	hh "github.com/flynn/flynn/pkg/httphelper"
+)
+
+type identifiable interface {
```
titanous: `identifier`

pkg/sse/stream.go
```diff
@@ -0,0 +1,134 @@
...
+	GetID() string
```
titanous: s/GetID/ID`

pkg/sse/stream.go
```diff
@@ -0,0 +1,134 @@
...
+func NewStream(w http.ResponseWriter, ch interface{}) *Stream {
+	sw := newWriter(w)
+	return &Stream{rw: w, w: sw, ch: ch, closeChan: make(chan struct{}), doneChan: make(chan struct{}), Done: make(chan struct{})}
+}
+
+func ServeStream(w http.ResponseWriter, ch interface{}) *Stream {
+	s := NewStream(w, ch)
+	s.Serve()
+	s.Wait()
+	return s
```
titanous: Any reason to return `s` since we are calling `Wait` inline?

pkg/sse/stream.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = s.fw.Write(data)
+	return err
+}
+
+func (s *Stream) sendKeepAlive() error {
+	_, err := s.w.w.Write([]byte(":\n"))
+	if err != nil {
```
titanous:
```go
if _, err := ...
```

controller/types/types.go
jvatic: `strconv.Itoa` doesn't appear to support `int64`

pkg/sse/stream.go
jvatic: `ID` conflicts with property names

titanous: Ah, `EventID`, then?

controller/types/types.go
titanous: Ah, missed that. nvm

pkg/sse/stream.go
jvatic: Is the overhead of doing so acceptable?

titanous: Ah, forgot that it's in a loop. You're right, that would slow things down.

jvatic:
@titanous comments addressed

controller/deployment.go
```diff
@@ -105,7 +106,7 @@ func (c *controllerAPI) GetDeployment(ctx context.Context, w http.ResponseWriter
 		return
 	}
 	if strings.Contains(req.Header.Get("Accept"), "text/event-stream") {
-		if err := streamDeploymentEvents(deployment.ID, w, c.deploymentRepo); err != nil {
+		if err := streamDeploymentEvents(deployment.ID, w, c.deploymentRepo, ctx); err != nil {
```
titanous: ctx should always be the first argument

host/sampi/http.go
```diff
@@ -132,6 +119,7 @@ func (c *HTTPAPI) RegisterHost(w http.ResponseWriter, r *http.Request, ps httpro
 	}
 
 	ch := make(chan *host.Job)
+	ch1 := make(chan *host.Job)
```
titanous: maybe `inCh` and `outCh`?

pkg/sse/sse.go
```diff
@@ -10,16 +10,23 @@ import (
...
 }
 
-type Writer struct {
+type writer struct {
 	w   io.Writer
 	mtx sync.Mutex
 }
 
-func (w *Writer) Write(p []byte) (int, error) {
+func (w *writer) writeID(id string) error {
```
titanous: Capital W

pkg/sse/stream.go
```diff
@@ -0,0 +1,149 @@
...
+	closeChan chan struct{}
+	doneChan  chan struct{}
+	closed    bool
+	logger    log.Logger
+	Done      chan struct{}
+}
+
+func NewStream(w http.ResponseWriter, ch interface{}, l log.Logger) *Stream {
+	sw := newWriter(w)
+	return &Stream{rw: w, w: sw, ch: ch, closeChan: make(chan struct{}), doneChan: make(chan struct{}), Done: make(chan struct{}), logger: l}
```
titanous: Reformat this so that each field is on its own line.

host/sampi/http.go
archSeer: Or just `in`, `out`.

pkg/sse/stream.go
```diff
@@ -0,0 +1,157 @@
...
+					Chan: chValue,
+				},
+			})
+			switch chosen {
+			case 0:
+				return
+			case 1:
+				s.sendKeepAlive()
+			default:
+				if ok {
```
titanous: Change this to `if !ok { return }`

titanous:
LGTM!

PR #901, opened 2015-01-30
test: Improve cluster HTTP API

lmars:
Use `httprouter` in the test runner and create a client for making HTTP requests from the test builds.

This is an extraction of changes in #840, and that PR will be rebased on top of these changes once merged.


heavenlyhash:
LGTM

PR #903, opened 2015-01-30
test: Remove unnecessary setup from TestExampleOutput

titanous:
There's no need to use the controller since we are using pkg/exec.


jvatic:
LGTM

PR #904, opened 2015-01-30
controller/examples: Send log output to STDERR by default

titanous:
The previous default was to discard any messages, which will suppress errors.


lmars:
LGTM

PR #909, opened 2015-01-30
router: enable pprof handler via build flag

benburkert:
Profiling the router with the `net/http/pprof` handler is useful for debugging the router, but it has a lot of baggage. Pulling in the `runtime/pprof` dependency should not be done by default. This adds a `pkg/pprof` package that only pulls in the `runtime/pprof`/`net/http/pprof` packages when the `pprof` build tag is set (`go build -tags=pprof`).

### DIFF
--- a/pkg/pprof/nopprof.go
+++ b/pkg/pprof/nopprof.go
@@ -1,0 +1,11 @@
+// +build !pprof
+
+package pprof
+
+import (
+	"net/http"
+)
+
+const Enabled = false
+
+var Handler = http.HandlerFunc(http.NotFound)
--- a/pkg/pprof/pprof.go
+++ b/pkg/pprof/pprof.go
@@ -1,0 +1,19 @@
+// +build pprof
+
+package pprof
+
+import (
+	"net/http"
+	hpprof "net/http/pprof"
+)
+
+const Enabled = true
+
+var Handler = http.NewServeMux()
+
+func init() {
+	Handler.Handle("/debug/pprof/", http.HandlerFunc(hpprof.Index))
+	Handler.Handle("/debug/pprof/cmdline", http.HandlerFunc(hpprof.Cmdline))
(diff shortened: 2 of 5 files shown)


router/server.go
```diff
@@ -72,6 +73,7 @@ func main() {
 	certFile := flag.String("tlscert", "", "TLS (SSL) cert file in pem format")
 	keyFile := flag.String("tlskey", "", "TLS (SSL) key file in pem format")
 	apiAddr := flag.String("apiaddr", ":"+apiPort, "api listen address")
+	pprofAddr := flag.String("pprofaddr", ":6060", "pprof handler listen address")
```
titanous: What do you think about mounting this on the API HTTP server?

pkg/pprof/pprof.go
```diff
@@ -0,0 +1,21 @@
...
+
+package pprof
+
+import (
+	"net/http"
+	hpprof "net/http/pprof"
+)
+
+var (
+	Enabled = true
```
titanous: This could be a const

router/server.go
benburkert: sounds good. :+1:

benburkert:
ok, the pprof handler is now mounted on the API listener instead of staring it's own listener.

pkg/pprof/nopprof.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"net/http"
+)
+
+const Enabled = false
+
+var Handler = http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+	rw.WriteHeader(404)
+	rw.Write([]byte(`pprof disabled, go build with -tags=pprof to enable\n`))
```
titanous: The `\n` will be two literal bytes

benburkert:
turns out pprof only prints the status code on a bad URL, so the message body wasn't useful anyways.

titanous:
LGTM

PR #911, opened 2015-01-31
run CI itself with race detection.

heavenlyhash:


titanous:
Not sure how much of a performance hit this will have, have you tried it on a staging CI instance?

heavenlyhash:
Nope!  Will do.  My theory is "not much" since the runner itself doesn't do much lifting aside from driving around qemu.

pkg/iotool/timeoutwriter.go
```diff
@@ -0,0 +1,56 @@
...
+	done bool
+}
+
+var TimeoutErr = fmt.Errorf("writer closed due to timeout")
+
+func NewTimeoutWriter(w io.Writer, patience time.Duration) *TimeoutWriter {
+	return NewTimeoutWriterFromChan(w, time.After(patience))
+}
+
+func NewTimeoutWriterFromChan(w io.Writer, done <-chan time.Time) *TimeoutWriter {
```
titanous: This and the above function should return `io.Writer`

titanous: nvm, just noticed the `Finished` method.

pkg/iotool/timeoutwriter.go
```diff
@@ -0,0 +1,56 @@
...
+	t := &TimeoutWriter{w: w}
+	go func() {
+		<-done
+		t.mtx.Lock()
+		defer t.mtx.Unlock()
+		t.done = true
+		if w != nil {
+			fmt.Fprintf(w, "%s\n", TimeoutErr)
```
titanous: Could just do `Fprintln`. Do we want some tag to identify the writer that timed out?

heavenlyhash: Could, though without parametric polymorphism the constructor names are going to get hilarious.

In the current use case, it doesn't really matter, because the uses are actually serial anyway.  The only reason it raced was in that when the timeout caused the main thread to move on, nothing stopped the slow writers.

titanous: Okay, no need.

titanous:
LGTM

heavenlyhash:
Note: Confirmed no discernible impact on test durations, so can just leave this on forever..  Data races detected in the runner can't be easily separated from data races in the tests afaik since the logs are combined, but changing that doesn't seem important at the moment (and it can be discerned from the actual stack content).

This will quietly go into effect the next time we reboot CI.

Issue #905, opened 2015-01-30
panic from netlink in networking init


titanous:
```text
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x603385]

goroutine 1 [running]:
github.com/flynn/flynn/pkg/shutdown.(*handler).exit(0xc20803c4b0, 0x0, 0x0, 0x0, 0xa0b960, 0xc20802a050)
	/home/ubuntu/go/src/github.com/flynn/flynn/pkg/shutdown/handler.go:65 +0x12e
github.com/flynn/flynn/pkg/shutdown.Exit()
	/home/ubuntu/go/src/github.com/flynn/flynn/pkg/shutdown/handler.go:40 +0x73
github.com/flynn/flynn/Godeps/_workspace/src/github.com/docker/libcontainer/netlink.networkLinkIpAction(0x14, 0x604, 0xc2080db2c0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/ubuntu/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/docker/libcontainer/netlink/netlink_linux.go:883 +0xaa5
github.com/flynn/flynn/Godeps/_workspace/src/github.com/docker/libcontainer/netlink.NetworkLinkAddIp(0xc2080db2c0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/ubuntu/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/docker/libcontainer/netlink/netlink_linux.go:924 +0xa4
main.(*LibvirtLXCBackend).ConfigureNetworking(0xc20800a420, 0x0, 0xc2080eef00, 0x2e, 0x1, 0x0, 0x0)
	/home/ubuntu/go/src/github.com/flynn/flynn/host/libvirt_lxc_backend.go:261 +0x11e3
main.(*manifestRunner).runManifest(0xc2080dc140, 0x7f48c686fd58, 0xc2080502a0, 0x0, 0x0, 0x0)
	/home/ubuntu/go/src/github.com/flynn/flynn/host/manifest.go:247 +0x196d
main.runDaemon(0xc2080a01c0)
	/home/ubuntu/go/src/github.com/flynn/flynn/host/host.go:228 +0x1a20
[...]

PR #912, opened 2015-01-31
Flannel fixes

titanous:
The combination of these two fixes should prevent the panic in #905.

Flannel patch: https://github.com/flynn/flannel/commit/04ff05d566d705715882b20a566be6113ca50679


heavenlyhash:
LGTM

PR #913, opened 2015-01-31
discoverd: Register

archSeer:
Incremental patch towards getting rid of sdutil.

/cc @titanous

### DIFF
--- a/discoverd/health/register_test.go
+++ b/discoverd/health/register_test.go
@@ -1,0 +1,298 @@
+package health
+
+import (
+	"errors"
+	"time"
+
+	. "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
+	"github.com/flynn/flynn/discoverd/client"
+	"github.com/flynn/flynn/pkg/random"
+	"github.com/flynn/flynn/pkg/stream"
+)
+
+type RegisterSuite struct{}
+
+var _ = Suite(&RegisterSuite{})
+
+type RegistrarFunc func(service string, inst *discoverd.Instance) (discoverd.Heartbeater, error)
+
+func (f RegistrarFunc) RegisterInstance(service string, inst *discoverd.Instance) (discoverd.Heartbeater, error) {
+	return f(service, inst)
+}
+
+type FakeHeartbeat struct {
+	closeFn   func() error
+	setMetaFn func(map[string]string) error
+	addrFn    func() string
+}
+
+func (f FakeHeartbeat) SetMeta(meta map[string]string) error { return f.setMetaFn(meta) }
+func (f FakeHeartbeat) Close() error                         { return f.closeFn() }
+func (f FakeHeartbeat) Addr() string                         { return f.addrFn() }
+
+func init() {
+	registerErrWait = time.Millisecond
+}
+
+func (RegisterSuite) TestRegister(c *C) {
+	type step struct {
+		event      bool // send an event
+		up         bool // event type
+		register   bool // event should trigger register
+		unregister bool // event should unregister service
+		setMeta    bool // attempt SetMeta
+		success    bool // true if SetMeta or Register should succeed
+	}
+
+	type called struct {
+		args      map[string]interface{}
+		returnVal chan bool
+	}
+
+	run := func(c *C, steps []step) {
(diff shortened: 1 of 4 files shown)


archSeer:
All statements are covered except `Addr()`, which I added to satisfy the heartbeater interface.

discoverd/health/register_test.go
```diff
@@ -0,0 +1,255 @@
...
+			steps: []step{
+				{event: true, up: true, register: true, success: false},
+				{register: true, success: false},
+				{event: true},
+				{}, // make sure register does not run
+				{},
+			},
+		},
+	} {
+		fmt.Println("--- TEST:", t.name, "---")
```
titanous: Use `c.Log`

discoverd/health/register_test.go
```diff
@@ -0,0 +1,255 @@
...
+			steps: []step{
+				{setMeta: true, success: false},
+			},
+		},
+		{
+			name: "setmeta while erroring registration",
+			steps: []step{
+				{event: true, up: true, register: true, success: false},
+				{register: true, success: false},
+				{setMeta: true, success: false},
```
titanous: This and the above check should ensure that when register succeeds the new meta is reflected.

discoverd/health/register_test.go
```diff
@@ -0,0 +1,255 @@
...
+type RegistrarFunc func(service string, inst *discoverd.Instance) (discoverd.Heartbeater, error)
+
+func (f RegistrarFunc) RegisterInstance(service string, inst *discoverd.Instance) (discoverd.Heartbeater, error) {
+	return f(service, inst)
+}
+
+type HeartbeatFunc func(map[string]string) error
+
+func (f HeartbeatFunc) SetMeta(meta map[string]string) error { return f(meta) }
+func (f HeartbeatFunc) Close() error                         { return nil }
```
titanous: Do we have any tests that the registration is closed when the service goes down?

archSeer:
@titanous Everything addressed, also added a test for `Addr()`, so we now have 100% coverage.

discoverd/health/register_test.go
titanous: Was this addressed?

archSeer: [Yes.](https://github.com/flynn/flynn/pull/913/files#diff-2e1f378428df428e27bf8eabe9dc247eR167)

titanous: It doesn't look like the necessary register step is called after the failing setmeta?

archSeer: Ah true, nice catch. Fixed.

discoverd/health/register_test.go
```diff
@@ -0,0 +1,295 @@
...
+			steps: []step{
+				{event: true, up: true, register: true, success: true},
+				{setMeta: true, success: true},
+			},
+		},
+		{
+			name: "setmeta while offline",
+			steps: []step{
+				{setMeta: true, success: false},
+			},
```
titanous: It think this also needs an extra step or two. `setMeta` while offline should just save locally, it should not call the underlying `SetMeta` func

discoverd/health/register_test.go
```diff
@@ -0,0 +1,298 @@
...
+				// Addr should be nil now
+				c.Assert(hb.Addr(), Equals, "")
+			} else {
+				wait++
+				errCheck(unregisterChan, stop)
+			}
+			if step.setMeta {
+				go func() {
+					call := <-metaChan
+					call.returnVal <- step.success
```
titanous: Does this ever happen if the heartbeater is "offline" (unregistered)?

discoverd/health/register_test.go
archSeer: Fixed as well, we now call register afterwards to confirm that the correct meta is transmitted.

discoverd/health/register_test.go
archSeer: Oh, good point... The logic for this is getting fairly complicated though, I think it should be obvious from the implementation that SetMeta will work correctly.

titanous:
Okay, LGTM. Make sure you've rebased on master before merging.

PR #914, opened 2015-01-31
host: Fix exit status attach return with no stdin

### LINKED ISSUE #626: event-stream of job log gives incorrect exit status
```
GET /apps/:app_id/jobs/:job_id/log?tail=true
Accept: text/event-stream
```

Exit status reported in event-stream doesn't match the one reported in the build output.

See https://gist.github.com/jvatic/82fe06a6637b63ac3d3c for output.

Edit: flynn-host logs: https://gist.github.com/2af253ef6759d74b6b9f

### LINKED ISSUE #810: pkg/exec: No error returned when non-zero exit status
For example:

```go
// ...
cmd := exec.Command(exec.DockerImage(...), "cmd-that-will-crash")
out, err := cmd.Output()
// err is nil
// ...
```

titanous:
Instead of only looking up the containerinit client when we need it, always look it up, and only error when we need it. This allows us to get retrieve the exit status in cases where stdin is provided and the job is not in TTY mode.

Closes #810
Closes #626

### DIFF
--- a/host/libvirt_lxc_backend.go
+++ b/host/libvirt_lxc_backend.go
@@ -798,12 +798,9 @@ func (l *LibvirtLXCBackend) Signal(id string, sig int) error {
 }
 
 func (l *LibvirtLXCBackend) Attach(req *AttachRequest) (err error) {
-	var client *libvirtContainer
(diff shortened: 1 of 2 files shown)


archSeer:
LGTM!

PR #916, opened 2015-01-31
pinkerton: Move LayerPullInfo into package

lmars:
This is so that components don't need to import pinkerton, which introduces a dependency on libdevmapper.


titanous:
LGTM

PR #920, opened 2015-02-01
discoverd/server: Run a single etcd server for all tests

titanous:
Instead of booting a new server, we can just use a different prefix for each test case.

This shaves ~20s off of the test runtime.


archSeer:
LGTM

PR #923, opened 2015-02-01
util/packer: Remove pre-installed guest additions

titanous:
These will typically fail to install when we install the latest kernel, and we install the latest version of the additions later.


lmars:
LGTM

Issue #709, opened 2015-01-07
Delete routes when deleting app


temujin9:
Currently, stale routes are retained when deleting an app. This causes issues with duplicate routes, if an app is deleted and recreated.

titanous:
One approach is to do #549 and just use a table in the controller database and then delete in the same transaction as the app delete. This breaks some boundaries but will always work correctly. I haven't though much about other solutions, I'd trade my kingdom for [HAT](http://www.bailis.org/blog/hat-not-cap-introducing-highly-available-transactions/).

titanous:
Fixed by #1533

PR #924, opened 2015-02-01
volume persistence database

heavenlyhash:
This adds use of boltdb to the volume management system.  Volume lifecycle events and metadata will be persisted so that volumes can be recalled and reconnected to jobs across flynn-host daemon reboots.

Also, a `flynn-host destroy-volumes` command has been added to remove state from previous hosts if a clean slate is desired.


host/volume/zfs/persistence_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+var _ = Suite(&PersistenceTests{})
+
+func (PersistenceTests) SetUpSuite(c *C) {
+	// Skip all tests in this suite if not running as root.
+	// Many zfs operations require root priviledges.
+	testutils.SkipIfNotRoot(c)
+}
+
+func (s *PersistenceTests) TestApplePie(c *C) {
```
titanous: What does this test again?

host/volume/zfs/zfs.go
```diff
@@ -110,77 +115,108 @@ func NewProvider(config *ProviderConfig) (volume.Provider, error) {
...
-		Location: path,
+func (b *Provider) MarshalGlobalState() (json.RawMessage, error) {
+	return json.Marshal(b.config)
+}
+
+func (b *Provider) MarshalVolumeState(volumeID string) (json.RawMessage, error) {
+	vol := b.volumes[volumeID]
+	c := make(map[string]string)
+	c["basemount"] = vol.basemount
+	c["dataset"] = vol.dataset.Name
```
titanous: Probably should make a struct for this.

host/volume/zfs/persistence_test.go
heavenlyhash: Creating the universe from scratch :)

Kidding, placeholder needs rename badly.

heavenlyhash:
Comments addressed and WIP commit replaced with actual description.

I think this is broadly ready for merge, with one major exception: currently, retaining etcd's state is causing the router application to crash on start.  As a result, though CI passes, running the integration tests twice in a row on the dev vm results in bootstrap failure.  I think this is the same root issue as #709 and will be fixed by #959.

titanous:
Okay, can you add a quick command to flynn-host that nukes all of the volumes/state?

host/volume/manager/persistence_test.go
```diff
@@ -0,0 +1,126 @@
...
+	volProv, err := zfs.NewProvider(&zfs.ProviderConfig{
+		DatasetName: zfsDatasetName,
+		Make: &zfs.MakeDev{
+			BackingFilename: zfsVdevFilePath,
+			Size:            int64(math.Pow(2, float64(30))),
+		},
+	})
+	c.Assert(err, IsNil)
+
+	// new volume manager with that shiney new backing zfs vdev file and a new boltdb
```
titanous: s/shiney/shiny/

archSeer:
Very nice! Still reviewing, but it seems fine. I think we do need to address the issue of cleaning up behind the app, namely resources and routes.

titanous:
LGTM

PR #926, opened 2015-02-02
router: context, cancelation of requests when clients go away

bgentry:
Introduce contexts into the router stack. Use them to carry request metadata (currently just start time) down the stack, and to coordinate cancelation of requests to backends when clients disappear. That behavior currently only applies to the non-upgrade path.

Also, add a 120 second timeout for response headers from backends. There's no test for this yet, and I'm not sure if/how I should add one. Perhaps we could just test that the value is indeed set on the `httpTransport` since we can already trust that to use the setting correctly?

### DIFF
--- a/router/http_test.go
+++ b/router/http_test.go
@@ -869,6 +869,50 @@ func (s *S) runTestErrorAfterConnOnlyHitsOneBackend(c *C, upgrade bool) {
 	c.Assert(string(data), Equals, "Service Unavailable\n")
 }
 
+func (s *S) TestClientClosesConnectionBeforeResponse(c *C) {
+	gotReq := make(chan struct{})
+	backendDone := make(chan struct{})
+
+	h := http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		close(gotReq)
+		defer close(backendDone)
+
+		cn := w.(http.CloseNotifier).CloseNotify()
+		select {
+		case <-cn:
+		case <-time.After(time.Second):
+			c.Fatal("expected reverse proxy to close the connection quickly")
+		}
+	})
+	srv := httptest.NewServer(h)
+	defer srv.Close()
+
+	l := s.newHTTPListener(c)
+	defer l.Close()
+
+	addHTTPRoute(c, l)
+	discoverdRegisterHTTP(c, l, srv.Listener.Addr().String())
+
+	req := newReq("http://"+l.Addr, "example.com")
+
+	go func() {
+		httpClient.Do(req)
+	}()
+
+	select {
+	case <-gotReq: // wait for backend to get the request
+	case <-time.After(2 * time.Second):
+		c.Fatal("never received on gotReq")
+	}
+	httpClient.Transport.(*http.Transport).CancelRequest(req)
+
+	select {
+	case <-backendDone:
+	case <-time.After(2 * time.Second):
+		c.Fatal("never received on backendDone")
+	}
+}
+
 // issue #152
 func (s *S) TestKeepaliveHostname(c *C) {
 	srv1 := httptest.NewServer(httpTestHandler("1"))
--- a/router/http_test.go
+++ b/router/http_test.go
@@ -869,6 +869,75 @@ func (s *S) runTestErrorAfterConnOnlyHitsOneBackend(c *C, upgrade bool) {
 	c.Assert(string(data), Equals, "Service Unavailable\n")
 }
 
+func (s *S) TestUpgradeAfterKeepAlive(c *C) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		// empty 200
+	}))
+
+	defer srv.Close()
+
+	l := s.newHTTPListener(c)
+	defer l.Close()
+
+	addHTTPRoute(c, l)
+	discoverdRegisterHTTP(c, l, srv.Listener.Addr().String())
+
+	conn, err := net.Dial("tcp", l.listener.Addr().String())
+	c.Assert(err, IsNil)
+	defer conn.Close()
+
+	req := newReq("http://"+l.Addr, "example.com")
+	err = req.Write(conn)
+	c.Assert(err, IsNil)
+
+	connReader := bufio.NewReader(conn)
+	res, err := http.ReadResponse(connReader, req)
+	c.Assert(err, IsNil)
+	c.Assert(res.StatusCode, Equals, 200)
+
+	data, err := ioutil.ReadAll(res.Body)
+	c.Assert(err, IsNil)
+	res.Body.Close()
+	c.Assert(string(data), Equals, "")
+
+	srv.Config.Handler = http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		w.Header().Set("Connection", "upgrade")
+		w.WriteHeader(101)
+		conn, bufrw, err := w.(http.Hijacker).Hijack()
+		c.Assert(err, IsNil)
+		defer conn.Close()
+
+		bufrw.Write([]byte("data after upgrade"))
+	})
+
+	req = newReq("http://"+l.Addr, "example.com")
+	req.Header.Set("Connection", "upgrade")
+	err = req.Write(conn)
+	c.Assert(err, IsNil)
+
+	res, err = http.ReadResponse(connReader, req)
+	c.Assert(err, IsNil)
+
+	data, err = ioutil.ReadAll(res.Body)
+	c.Assert(err, IsNil)
+	res.Body.Close()
+	c.Assert(res.StatusCode, Equals, 101)
+	c.Assert(string(data), Equals, "")
+
+	datac := make(chan []byte)
+	go func() {
+		data, err = ioutil.ReadAll(connReader)
+		c.Assert(err, IsNil)
+		datac <- data
+	}()
+
+	select {
+	case <-datac:
+	case <-time.After(2 * time.Second):
+		c.Fatal("never received from datac")
+	}
+}
+
 func (s *S) TestClientClosesConnectionBeforeResponse(c *C) {
 	gotReq := make(chan struct{})
 	backendDone := make(chan struct{})
--- a/router/proxy/reverseproxy.go
+++ b/router/proxy/reverseproxy.go
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/net/context"
 )
 
 const (
@@ -71,7 +73,7 @@ func NewReverseProxy(bf BackendListFunc, stickyKey *[32]byte, sticky bool) *Reve
 }
 
 // ServeHTTP implements http.Handler.
-func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+func (p *ReverseProxy) ServeHTTP(ctx context.Context, rw http.ResponseWriter, req *http.Request) {
 	transport := p.transport
 	if transport == nil {
 		panic("router: nil transport for proxy")
@@ -84,8 +86,24 @@ func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	res, err := transport.RoundTrip(outreq)
+	clientGone := rw.(http.CloseNotifier).CloseNotify()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel() // tells cancellation goroutine to exit
+
+	go func() {
+		select {
+		case <-clientGone:
+			cancel() // client went away, cancel request
+		case <-ctx.Done():
+		}
+	}()
+
+	res, err := transport.RoundTrip(ctx, outreq)
 	if err != nil {
+		if err == errRequestCanceled {
+			// TODO(bgentry): anything special to log here?
+			return
+		}
 		p.logf("router: proxy error: %v", err)
 		rw.WriteHeader(http.StatusServiceUnavailable)
 		rw.Write(serviceUnavailable)
--- a/router/proxy/reverseproxy.go
+++ b/router/proxy/reverseproxy.go
@@ -127,6 +127,7 @@ func (p *ReverseProxy) ServeConn(conn net.Conn) {
 		p.logf("router: proxy error: %v", err)
 		return
 	}
+	defer dconn.Close()
 
 	joinConns(conn, dconn)
 }
@@ -158,6 +159,7 @@ func (p *ReverseProxy) serveUpgrade(rw http.ResponseWriter, req *http.Request) {
 		p.logf("router: hijack failed: %v", err)
 		return
 	}
+	defer dconn.Close()
 	joinConns(uconn, &streamConn{bufrw.Reader, dconn})
 }
 
--- a/router/proxy/reverseproxy.go
+++ b/router/proxy/reverseproxy.go
@@ -115,21 +115,21 @@ func (p *ReverseProxy) ServeHTTP(ctx context.Context, rw http.ResponseWriter, re
 }
 
 // ServeConn takes an inbound conn and proxies it to a backend.
-func (p *ReverseProxy) ServeConn(conn net.Conn) {
+func (p *ReverseProxy) ServeConn(dconn net.Conn) {
 	transport := p.transport
 	if transport == nil {
 		panic("router: nil transport for proxy")
 	}
-	defer conn.Close()
+	defer dconn.Close()
 
-	dconn, err := transport.Connect(conn.RemoteAddr())
+	uconn, err := transport.Connect()
 	if err != nil {
 		p.logf("router: proxy error: %v", err)
 		return
 	}
-	defer dconn.Close()
+	defer uconn.Close()
 
-	joinConns(conn, dconn)
+	joinConns(uconn, dconn)
 }
 
 func (p *ReverseProxy) serveUpgrade(rw http.ResponseWriter, req *http.Request) {
--- a/router/proxy/transport.go
+++ b/router/proxy/transport.go
@@ -11,11 +11,13 @@ import (
 	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/crypto/nacl/secretbox"
+	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/net/context"
 	"github.com/flynn/flynn/pkg/random"
 )
 
 var (
-	errNoBackends = errors.New("router: no backends available")
+	errNoBackends      = errors.New("router: no backends available")
+	errRequestCanceled = errors.New("router: request canceled")
 
 	httpTransport = &http.Transport{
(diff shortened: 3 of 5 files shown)


bgentry:
In order to protect the upgrade path against the client going away, I think we'll end up needing to do something similar to what Go's server does to enable `CloseNotify()`: https://github.com/golang/go/blob/928c83ff2cd13e5973827b2abde43b6dc901ec85/src/net/http/server.go#L173-L179

router/proxy/transport.go
```diff
@@ -64,13 +66,33 @@ func (t *transport) setStickyBackend(res *http.Response, originalStickyBackend s
...
+
+	go func() {
+		defer close(resultc)
+		res, err := t.roundTrip(req)
+		resultc <- httpResult{res, err}
+	}()
+
+	select {
+	case <-ctx.Done():
+		httpTransport.CancelRequest(req)
```
bgentry: Damn, I think this might have a small race. If `roundTrip()` gets a `dialErr` trying to hit one backend, then it might already be moving on to another request when the old request gets canceled.

I originally had some more coordination that trickled down into `roundTrip()` but thought it might be unnecessary. Now I'm not sure if there's a way to avoid that.

router/proxy/transport.go
```diff
@@ -64,13 +66,33 @@ func (t *transport) setStickyBackend(res *http.Response, originalStickyBackend s
...
+		err := ctx.Err()
+		if err == context.Canceled {
+			return nil, nil
```
benburkert: IMO it would be safer to return an `errRequestCanceled` here. It's correct right now, but in the future if the context could be canceled by something other than the `ReverseProxy`, then the `defer result.res.Body.Close()` in `ReverseProxy.ServeHTTP` would panic b/c `result.res` is nil.

bgentry: yeah, I think that definitely fits better with the [definition of RoundTripper](http://golang.org/pkg/net/http/#RoundTripper):

> A non-nil err should be reserved for failure to obtain a response.

I'll change it, thanks.

router/proxy/transport.go
benburkert: yeah, it looks like we want to pass the context all the way down to `transport.roundTrip`. `http.Transporter` doesn't scope dials to requests; if a cancel happens during a dial, `net.Conn` (wrapped in `dialRes`) is moved to `idleConn` (via `handlePendingDial`). So I think it's optimal to check for cancelations before dialing at the beginning of the for loop of `roundTrip`:
```go
for _, backend := range backends {
	select {
	case <-ctx.Done():
		return nil, errRequestCanceled
	default:
	}

	req.URL.Host = backend
	// ...
}
```

bgentry: yeah, that's exactly what I had originally :)

Gonna reintroduce that.

benburkert: cool. i suspect it will be easier to follow if the `for` loop is moved into `RoundTrip` and the `resultc/go/select` for `httpTransport.RoundTrip` stuff is moved to `roundTrip`.

bgentry: one problem with that... there's still another subtle race. After the `select` in `RoundTrip()` enters the `case <-ctx.Done()`, we could get a response from `roundTrip()` and send that on the `resultc` before the request is actually canceled. That would result in leaving that response and its backend conn open forever.

Gonna take a fresh look at this after lunch to see if something like the they have in [`httputil.ReverseProxy`](https://github.com/golang/go/blob/fa851b17a2fd2350043b3e56dce398a606692c23/src/net/http/httputil/reverseproxy.go#L130-L156) might actually simplify this.

bgentry: yeah, your suggestion might help as well. Will play around with a few things when I've got fresher eyes.

bgentry: Another subtle case to handle is covered in the review for that `httputil.ReverseProxy` change: https://go-review.googlesource.com/#/c/2320/5/src/net/http/httputil/reverseproxy.go@120

> If the CloseNotifier triggers very quickly (like if the client sends a request and then immediately closes their connection), it seems like this call could happen before the call to transport.RoundTrip begins.

So that's why they don't spawn their cancellation goroutine until after the Transport starts reading the request body.

benburkert:
> Also, add a 120 second timeout for response headers from backends. There's no test for this yet, and I'm not sure if/how I should add one. Perhaps we could just test that the value is indeed set on the httpTransport since we can already trust that to use the setting correctly?

If the test were in the proxy package, we can do something evil like mutating `httpTransport` in the test file. :smiling_imp:

bgentry:
@benburkert yeah, but then we'd still be testing the underlying behavior of the `http.Transport`, which is already well-tested by the net/http tests. To me it feels sufficient to add a test just making sure it's being set properly. Though I guess that still doesn't ensure that you're actually using the transport you're testing against... =/

router/proxy/transport.go
bgentry: I just pushed a new version that I think resolves all these subtle race conditions.

router/proxy/reverseproxy.go
```diff
@@ -84,16 +87,47 @@ func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		return
+	case <-clientGone:
+		cancel() // client went away, cancel request
+		go throwawayResponse(resultc)
+		// TODO(bgentry): anything special to log here?
 		return
+	case result := <-resultc:
+		if result.err != nil {
+			if result.err == errRequestCanceled {
+				// TODO(bgentry): anything special to log here?
```
bgentry: Is there anything I should log for these scenarios, or should I just drop the TODOs?

router/proxy/transport.go
benburkert: > You might end up trying to cancel the request before it's begun, and then the request to the backend will still happen.

I think checking `ctx.Done()` (the `select`/`default`) prevents a quickly canceled request from being sent to a backend. But we aren't doing anything (e.g. reading the request body) to trigger that `CloseNotifier`. We've already got the request body wrapped in `fakeCloseReadCloser`, perhaps we should `Peek` right before starting the `for` loop? i think that would trigger the `CloseNotifier` and trickle down to `ctx.Done()` (but across different goroutines, so it's not guaranteed).

bgentry:
So I'm having a tough time deciding how I should handle disappearing clients from the `Upgrade` path. We arranged `serveUpgrade` so that it wouldn't hijack until the last possible moment, thus preserving persistent inbound conns from clients until we've already upgraded them.

However, because [you can't `Hijack()` after you've invoked `CloseNotify`](https://github.com/golang/go/blob/928c83ff2cd13e5973827b2abde43b6dc901ec85/src/net/http/server.go#L146), this leaves us vulnerable to disappearing clients that send `Upgrade` requests.

I'm not sure how to handle this short of immediately Hijacking in `serveUpgrade` and giving up on reusing any conn that hits that path. Then we'd be able to detect when those clients' conns go away [just as the `http.Server` does](https://github.com/golang/go/blob/928c83ff2cd13e5973827b2abde43b6dc901ec85/src/net/http/server.go#L173-L179). It wouldn't be *that* bad if we did that... Users would just incur a reconnect after a failed upgrade. Probably not a scenario we need to optimize for. I'd be much more concerned about leaving that path as-is and not handling disappearing clients at all.

router/proxy/transport.go
bgentry: > But we aren't doing anything (e.g. reading the request body) to trigger that `CloseNotifier`.

Not sure I follow. We don't have to "trigger" the `CloseNotifier`; that could happen anytime after we call `CloseNotify()`: https://github.com/golang/go/blob/928c83ff2cd13e5973827b2abde43b6dc901ec85/src/net/http/server.go#L173-L180

Anyway, the current iteration of this does seem the most correct of the ones we've had so far. Admittedly, though, it does take a lot of digging to really understand why it works.

router/proxy/transport.go
```diff
@@ -64,26 +67,55 @@ func (t *transport) setStickyBackend(res *http.Response, originalStickyBackend s
...
+		// https://go-review.googlesource.com/#/c/2320/5/src/net/http/httputil/reverseproxy.go@120
+		Reader: &runOnFirstRead{
+			Reader: req.Body,
+			fn: func() {
+				go func() {
+					select {
+					case <-cancelc:
+						httpTransport.CancelRequest(req)
+					case <-reqDone:
+					}
```
benburkert: the only place `close(reqDone)` is called is in the `defer` above, which indicates that the first read on the request body won't finish until the response from the backend has been returned, or the request is canceled.

bgentry: no, this is being run in a separate goroutine that is spawned from within `runOnFirstRead.Read()`. It doesn't block `Read()`.

benburkert: ahh, ok, I overlooked the the `go` on 89.

This still creates a goroutine to watch for `cancelc` and call then call `httpTransport.CancelRequest(req)`. It doesn't reduce the number of goroutines, but does make it harder to follow along than the previous version.

Why does this need the extra complexity of `runOnFirstRead`? Won't the `select` catch clients that go away quickly? It depends on the `CloseNotifier` triggering in a separate goroutine [here](https://github.com/golang/go/blob/master/src/net/http/server.go#L179), so it's not deterministic, but neither is `runOnFirstRead` due to the goroutine there.

router/proxy/reverseproxy.go
```diff
@@ -84,16 +87,47 @@ func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		}
+		defer result.res.Body.Close()
+
+		p.writeResponse(rw, result.res)
 	}
-	defer res.Body.Close()
+}
 
-	p.writeResponse(rw, res)
+func throwawayResponse(resultc <-chan httpResult) {
```
benburkert: `throwawayResult`? Also, calling it everywhere with `(<-resultc)` seems simpler than passing the `<-chan httpResult`.

bgentry: if I called it with `defer throwawayResult(<-resultc)`, that line would block until the `<-resultc` returned (`defer` args are evaluated immediately, though the func isn't run until later. It'd have to be:

```go
defer func() {
    throwawayResult(<-resultc)
}()
```

Which didn't seem better than just changing the cleanup func.

router/proxy/reverseproxy.go
```diff
@@ -84,16 +87,47 @@ func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	go func() {
+		res, err := transport.RoundTrip(ctx, outreq)
+		defer close(resultc)
+		resultc <- httpResult{res, err}
+	}()
+
+	select {
+	case <-ctx.Done():
+		go throwawayResponse(resultc)
+		return
```
benburkert: this is an undefined situation at this point, should it write a 503 response here? checking `ctx.Done()` here seems unnecessary right now. It's already going to be handled in `transporter`, so it could throw away the result right after the request is canceled?

router/proxy/reverseproxy.go
benburkert: ahh, agreed.

router/proxy/transport.go
bgentry: > Why does this need the extra complexity of `runOnFirstRead`? Won't the `select` on 104 catch clients that go away quickly?

No, not always. It will only catch the case where they've gone away before the `select` statement began. It will not handle the case where client goes away after the `select` falls through to the `default` case, but before `httpTransport.RoundTrip()` is called on 109. Even after `RoundTrip()` is called, you still have a window before the in-flight request has been internally registered by the `http.Transport`. If `CancelRequest()` is called during that window, it's a no-op (because the `Transport` doesn't know about the request). This scenario is discussed in [Golang Change 2320](https://go-review.googlesource.com/#/c/2320/) on patch set 5.

> it's not deterministic, but neither is `runOnFirstRead` due to the goroutine there.

Can you walk me through any racy behavior that still remains with the current version of this PR? I'm not seeing any problems.

bgentry:
just found a way to get rid of the `throwawayResponse()` cleanup thing altogether and simplify `ReverseProxy.ServeHTTP()` a bit. Please take another look at the changes to reverse_proxy.go.

router/http_test.go
```diff
@@ -869,6 +869,42 @@ func (s *S) runTestErrorAfterConnOnlyHitsOneBackend(c *C, upgrade bool) {
...
+	req := newReq("http://"+l.Addr, "example.com")
+
+	go func() {
+		httpClient.Do(req)
+	}()
+
+	<-gotReq // wait for backend to get the request
+	httpClient.Transport.(*http.Transport).CancelRequest(req)
+
+	<-backendDone
```
titanous: Should add selects with timeouts for both of these so that the tests don't hang if they break.

router/proxy/transport.go
```diff
@@ -138,19 +170,30 @@ type dialErr struct {
...
+
+type runOnFirstRead struct {
+	io.Reader
+
+	fn func() // Run before first Read, then set to nil
+}
+
+func (c *runOnFirstRead) Read(bs []byte) (int, error) {
+	if c.fn != nil {
+		c.fn()
```
titanous: Could change to run this in a goroutine here, which would reduce the indentation above. not sure if `runOnFirstRead` is useful in it's current form elsewher?

bgentry: Not useful elsewhere, it's left as it was in https://github.com/golang/go/commit/ececbe89d4d5ab151333cbadae725619c0ad9dd8

Want it changed?

router/proxy/transport.go
```diff
@@ -64,26 +67,55 @@ func (t *transport) setStickyBackend(res *http.Response, originalStickyBackend s
...
-		req.URL.Host = backend
-		res, err := httpTransport.RoundTrip(req)
-		if err == nil {
-			t.setStickyBackend(res, stickyBackend)
-			return res, nil
+		select {
+		case <-cancelc:
+			return nil, errRequestCanceled
+		default:
+			req.URL.Host = backend
```
titanous: I normally use an empty default case in this type of scenario to avoid the indentation (can just move all this code down and outdent.

router/http_test.go
bgentry: fixed.

router/proxy/transport.go
titanous: I don't have a strong preference, but it seems like if it's a single-purpose type, we can make it specific to the purpose.

router/proxy/transport.go
benburkert: > Even after RoundTrip() is called, you still have a window before the in-flight request has been internally registered by the http.Transport. If CancelRequest() is called during that window, it's a no-op (because the Transport doesn't know about the request).

gotcha, i was failing to understand this window.

router/proxy/transport.go
bgentry: :+1: will change that

router/proxy/transport.go
bgentry: yeah, I think it cleans this block up substantially to move the `go ...` into `runOnFirstRead.Read()`. Will change it :v:

router/proxy/transport.go
bgentry: yeah, also took me awhile to understand why they did what they did in that change, and why it works :)

benburkert:
It looks like `runOnFirstRead` could be folded into `fakeCloseReadCloser` (as `backendBody`, with a `sync.Once` too). I think that would make it easier to follow.

> So I'm having a tough time deciding how I should handle disappearing clients from the Upgrade path. We arranged serveUpgrade so that it wouldn't hijack until the last possible moment, thus preserving persistent inbound conns from clients until we've already upgraded them.

Sounds like the consensus in IRC is to duplicate the `CloseNotify` behavior in `http/server.go`, which might also work as a new body type used by `transport.UpgradeHTTP`.

router/proxy/transport.go
```diff
@@ -88,7 +118,7 @@ func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	return nil, errNoBackends
 }
 
-func (t *transport) Connect(remoteAddr net.Addr) (net.Conn, error) {
+func (t *transport) Connect() (net.Conn, error) {
```
titanous: I assumed that this argument was there to support the PROXY protocol in the future, but I think removing it for now is fine.

router/proxy/reverseproxy.go
```diff
@@ -286,6 +306,11 @@ func prepareRequest(req *http.Request) *http.Request {
 	return outreq
 }
 
+type httpResult struct {
+	res *http.Response
+	err error
+}
```
titanous: This looks unused?

bgentry: yeah, meant to nuke it.

bgentry:
oh, fuck. It looks like it's not just that `CloseNotify()` and `Hijack()` can't be used together in the same request... they can never be used together on the same HTTP conn: https://github.com/golang/go/blob/928c83ff2cd13e5973827b2abde43b6dc901ec85/src/net/http/server.go#L160-L165

You'll see that `closeNotifyc` gets set with `c.closeNotifyc = make(chan bool, 1)`, but it never gets unset. Combined with the fact that an error is thrown any time you try to `Hijack()` a conn that has `closeNotifyc != nil`, and that means that we can't use `CloseNotify()` at all in the same server as we're also doing `Hijack()`.

Ideas?

titanous:
LGTM. Anything left on this before merging? Seems like we can sort out the upgrade path nonsense in a future PR?

bgentry:
@titanous if I'm understanding this correctly, we can't actually use both `CloseNotifier` and `Hijacker` safely in the same `http.Server` if we're also using keep-alive. It would be possible/likely for a request to come in on the non-upgrade path, then reuse that conn for an upgrade request. The attempted `Hijack()` call in the upgrade request would fail due to the fact that `closeNotifyc` had already been allocated on that server conn.

titanous:
Ugh. Do we have a test for that? (upgrade from keep-alive conn)

bgentry:
@titanous I pushed up my best attempt at a test for now...You can see the error message, but because of the order of events in `serveUpgrade` it doesn't actually cause the upgrade to fail, and the 101 is simply returned to the client, making the client think the upgrade was successful.

That's a subtle bug with the current behavior :(

benburkert:
Damn, that really sucks. Perhaps we can send the `101` response after we attempt to hijack, and if the hijack fails, send a `302` to the same URL with `Connection: close` instead.

bgentry:
So the CloseNotify stuff is blocked due to https://github.com/golang/go/issues/9763 . I'm hoping we can resolve that with the Go team in the near future, but for now it looks like we'll probably need to leave out handling of disconnecting clients.

Once we come up with a solution in net/http, we can discuss whether it makes sense to vendor that.

In the mean time, I opened #955 to pull in all of the other bits of this PR. We may want to just open a separate issue to track the CloseNotify stuff and close this PR altogether.

titanous:
Yes, please go ahead and open a tracking issue for CloseNotify.

bgentry:
Opened #958 to track it. Closing this PR for now. Hopefully it can be a good reference when the underlying net/http limitation is gone.

PR #931, opened 2015-02-02
pinkerton: Fix fetching image ancestors

lmars:
Previously the ancestor layers would not have ParentID set because we did not fetch the layer JSON, so the layer would not get the correct JSON when extracted into the graph driver.

Incremental patch for #617.


titanous:
LGTM

PR #932, opened 2015-02-02
split volume manager into separate package

heavenlyhash:
Upcoming changes were running aground on import cycles: the manager needs to refer into the provider implementation packages as it gains responsibility for initializing them during restores from prior state; and the providers of course need to refer to the volume basic types.

The new import graph is host|api -> manager -> {providers} -> volume.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -17,7 +17,7 @@ import (
 	"github.com/flynn/flynn/host/ports"
 	"github.com/flynn/flynn/host/sampi"
 	"github.com/flynn/flynn/host/types"
-	"github.com/flynn/flynn/host/volume"
+	"github.com/flynn/flynn/host/volume/manager"
 	zfsVolume "github.com/flynn/flynn/host/volume/zfs"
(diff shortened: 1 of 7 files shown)


archSeer:
Couldn't you do the same thing we do in the router/controller/host, and move the type definitions into a subpackage? `vt github.com/flynn/flynn/host/volume/types`

heavenlyhash:
Yeah, it's about six of one, half dozen of the other.  I think in this case it makes sense for the types to be the central package and the manager to be off to the side; the manager is more a bunch of glue than the focal point, and all the providers can technically be used without any reference to one.

host/host.go
```diff
@@ -173,7 +173,7 @@ func runDaemon(args *docopt.Args) {
 	}
 
 	// create volume manager
-	vman := volume.NewManager(volProv)
+	vman := volumemanager.NewManager(volProv)
```
titanous: Can we change this signature to `volumemanager.New` to avoid the stutter?

titanous:
LGTM

PR #933, opened 2015-02-02
discoverd: Die sdutil!

archSeer:
This is ready for merge. I'm writing a quick prototype of how the CLI-side interactions with these services would go.


receiver/flynn-receive.go
```diff
@@ -103,11 +104,15 @@ func main() {
...
-				proc.Env = make(map[string]string)
-			}
-			proc.Env["SD_NAME"] = app.Name + "-web"
+			proc.Ports = []ct.Port{{
+				Port:  8080,
+				Proto: "tcp",
+				Service: &host.PortService{
+					Name:   app.Name + "-web",
+					Create: true,
+					Check:  &host.PortServiceCheck{Type: "http"},
```
archSeer: HTTP check by default. Works for majority of cases, but might not fit where the user has a resource that is not available under `/`, however I think we can just note that in the usage manual, and then have instructions on how to change the service checker using the CLI.

host/containerinit/init.go
```diff
@@ -395,6 +401,71 @@ func getCmdPath(c *Config) (string, error) {
...
+		}
+	}
+	portStr := strconv.Itoa(port.Port)
+	if portStr == "0" {
+		portStr = env["PORT"]
+	}
+	localAddr := fmt.Sprintf("127.0.0.1:%s", portStr)
+	addr := fmt.Sprintf("%s:%s", env["EXTERNAL_IP"], portStr)
+	var check health.Check
+	switch config.Check.Type {
```
archSeer: Should probably add support for registering a service without a checker? Would skip the `health.Registration` and instead just return a heartbeater which gets closed when the process exits.

archSeer:
@titanous I'm thinking of renaming the types `PortService` -> `Service`, `PortServiceChecker` -> `Checker/HealthChecker`. WDYT?

receiver/flynn-receive.go
titanous: We could add a loose check that accepts any status code (maybe status code `-1`), what do you think?

host/containerinit/init.go
titanous: Yes, absolutely.

host/containerinit/init.go
```diff
@@ -535,9 +619,6 @@ func Main() {
 		log.Fatalf("Unable to unmarshal config: %v", err)
 	}
 
-	// Propagate the plugin-specific container env variable
-	config.Env["container"] = os.Getenv("container")
-
```
titanous: Why was this removed?

titanous:
Changing the type names is fine, `HealthChecker` sounds good.

host/containerinit/init.go
```diff
@@ -395,6 +401,71 @@ func getCmdPath(c *Config) (string, error) {
...
+		// TODO: maybe reuse maybeAddService() from the client
+		if err := client.AddService(config.Name); err != nil {
+			if je, ok := err.(hh.JSONError); !ok || je.Code != hh.ObjectExistsError {
+				return nil, fmt.Errorf("something went wrong with discoverd: %s", err)
+			}
+		}
+	}
+	portStr := strconv.Itoa(port.Port)
+	if portStr == "0" {
+		portStr = env["PORT"]
```
titanous: Do this before converting to a string.

host/containerinit/init.go
```diff
@@ -395,6 +401,71 @@ func getCmdPath(c *Config) (string, error) {
...
+	}
+	localAddr := fmt.Sprintf("127.0.0.1:%s", portStr)
+	addr := fmt.Sprintf("%s:%s", env["EXTERNAL_IP"], portStr)
```
titanous: No need to convert  port to a string if you are using `fmt`.

host/containerinit/init.go
```diff
@@ -395,6 +401,71 @@ func getCmdPath(c *Config) (string, error) {
...
+	if config.Check.KillDown {
+		// ignore events for the first StartTimeout interval
+		events := make(chan health.MonitorEvent)
+		reg.Events = events
+		go func() {
+			for e := range events {
+				if e.Status != health.MonitorStatusDown {
+					continue
+				}
+				// kill process
```
titanous: Is this a TODO?

host/containerinit/init.go
```diff
@@ -513,6 +584,19 @@ func containerInitApp(c *Config) error {
...
 
 	init.mtx.Unlock() // Allow calls
+	// monitor services
+	for _, port := range c.Ports {
+		if port.Service == nil {
+			continue
+		}
+		hb, err := monitor(port, c.Env)
+		if err != nil {
+			// TODO: handle err
```
titanous: TODO needs handling?

titanous:
This should have a bunch of integration tests covering each option of the service check, especially stuff like `KillDown` and thresholds.

host/containerinit/init.go
archSeer: Is it used anywhere? I think it's a leftover from the original code.

host/containerinit/init.go
archSeer: `env["PORT"]` is a string, so I need to convert that to an integer then.

host/containerinit/init.go
archSeer: Ah, yes.

host/containerinit/init.go
archSeer: I am unsure how to handle this. Should the entire application fail, if the Register fails to spin up for some reason, or should we simply skip it (like we do now) and let the application run?

titanous: Entire app should fail.

host/containerinit/init.go
titanous: It sets an environment variable that should be passed on. It is 'used' in the sense that that env var should be available.

host/containerinit/init.go
titanous: Ah, misunderstood this. Can't we just serialize the info after port assignment?

host/containerinit/init.go
archSeer: Yes, I figured that `os.Getenv("container") == ""`, since I couldn't find anything setting `container`

titanous: libvirt-lxc sets it. Check out the environment of a container.

host/containerinit/init.go
archSeer: What do you mean by that?

receiver/flynn-receive.go
archSeer: I think that could work. We could also just use the TCP checker?

titanous: Yeah, probably easier to just do TCP and won't show up in the logs, we can always change it later.

archSeer:
@titanous Do we need to test all cases though? I'd test the `KillDown + StartTimeout`, but all of the `Monitor` and `Checker` options are already tested in the discoverd unit tests.

titanous:
A sanity check or two wouldn't hurt, but you don't need to go overboard.

archSeer:
@titanous Tests pass.

host/containerinit/init.go
```diff
@@ -395,6 +400,89 @@ func getCmdPath(c *Config) (string, error) {
...
+
+	if config.Create {
+		// TODO: maybe reuse maybeAddService() from the client
+		if err := client.AddService(config.Name); err != nil {
+			if je, ok := err.(hh.JSONError); !ok || je.Code != hh.ObjectExistsError {
+				return nil, fmt.Errorf("something went wrong with discoverd: %s", err)
+			}
+		}
+	}
+	localAddr := fmt.Sprintf("127.0.0.1:%v", port.Port)
```
titanous: Do we need this? Can't we just use `addr`?

host/containerinit/init.go
```diff
@@ -395,6 +400,89 @@ func getCmdPath(c *Config) (string, error) {
...
+			Threshold: config.Check.Threshold,
+		}.Run,
+		Check: check,
+	}
+
+	if config.Check.KillDown {
+		reg.Events = make(chan health.MonitorEvent)
+		go func() {
+			if config.Check.StartTimeout == 0 {
+				config.Check.StartTimeout = 10
```
titanous: This would be 10 nanoseconds, which is probably not what we want.

host/containerinit/init.go
```diff
@@ -395,6 +400,89 @@ func getCmdPath(c *Config) (string, error) {
...
+			}
+
+			start := false
+			lastStatus := health.MonitorStatusDown
+
+			maybeKill := func() {
+				if lastStatus == health.MonitorStatusDown {
```
titanous: This check is racy

host/containerinit/init.go
```diff
@@ -395,6 +400,89 @@ func getCmdPath(c *Config) (string, error) {
...
+					container.Signal(int(syscall.SIGKILL), &struct{}{})
+				}
+			}
+			go func() {
+				// ignore events for the first StartTimeout interval
+				<-time.After(config.Check.StartTimeout * time.Second)
+				maybeKill() // check if the app is down
+				start = true
```
titanous: This set is racy

host/containerinit/init.go
archSeer: Should I just wrap all of it with a `sync.RWMutex`, or do you have a better approach in mind?

host/containerinit/init.go
archSeer: Hmm, I'll change it for clarity, but the test definitely takes 10 seconds to timeout (passing tests take 2s, failing test takes 12s)

host/containerinit/init.go
archSeer: I ran into problems originally, but I think that might have happened because of other problems. Now that we have tests, I'll give it a try.

host/containerinit/init.go
titanous: yeah, a `sync.Mutex` is fine, probably overkill to use `RWMutex`

archSeer:
@titanous Comments addressed.

host/containerinit/init.go
```diff
@@ -395,6 +400,94 @@ func getCmdPath(c *Config) (string, error) {
...
+	if config.Check.KillDown {
+		reg.Events = make(chan health.MonitorEvent)
+		go func() {
+			if config.Check.StartTimeout == 0 {
+				config.Check.StartTimeout = 10 * time.Second
+			}
+
+			start := false
+			lastStatus := health.MonitorStatusDown
+			mtx := &sync.Mutex{}
```
titanous: Can just do `var mtx sync.Mutex`

archSeer: Addressed.

test/apps/ping/main.go
```diff
@@ -0,0 +1,29 @@
...
+
+const service = "ping-service"
+
+func main() {
+	defer shutdown.Exit()
+
+	port := os.Getenv("PORT")
+	addr := ":" + port
+
+	hb, err := discoverd.AddServiceAndRegister(service, addr)
```
titanous: Why does this service register with discoverd?

archSeer: Ouch, totally missed that, I think I copied the boilerplate code from the TCP app...

test/test_healthcheck.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (s *HealthcheckSuite) TestWithoutChecker(t *c.C) {
+	// start app with a service but no checker
+	app, _ := s.createAppWithService(t, "ping", &host.Service{
+		Name:   "ping-without-checker",
+		Create: true,
+	})
+	t.Assert(flynn(t, "/", "-a", app.Name, "scale", "ping=1"), Succeeds)
+	// make sure app is registered and unregistered when the process terminates
+	_, err := s.discoverdClient(t).Instances("ping-without-checker", 10*time.Second)
```
titanous: This doesn't do what it says on the tin. Where is the verification that it is unregistered?

test/test_healthcheck.go
```diff
@@ -0,0 +1,86 @@
...
+	c "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/host/types"
+)
+
+type HealthcheckSuite struct {
+	Helper
+}
+
+var _ = c.Suite(&HealthcheckSuite{})
```
titanous: This should be a ConcurrentSuite?

test/test_healthcheck.go
```diff
@@ -0,0 +1,86 @@
...
+		Create: true,
+		Check:  &host.HealthCheck{Type: "tcp"},
+	})
+	t.Assert(flynn(t, "/", "-a", app.Name, "scale", "printer=1"), Succeeds)
+	// confirm that it's never registered
+	_, err := s.discoverdClient(t).Instances("healthcheck-failure", 10*time.Second)
+	t.Assert(err, c.NotNil)
+}
+
+func (s *HealthcheckSuite) TestKilldown(t *c.C) {
```
titanous: s/Killdown/KillDown/

test/test_healthcheck.go
```diff
@@ -0,0 +1,86 @@
...
+	// start an app that is failing checks /w killdown
+	app, _ := s.createAppWithService(t, "printer", &host.Service{
+		Name:   "healthcheck-killdown",
+		Create: true,
+		Check:  &host.HealthCheck{Type: "tcp", KillDown: true},
```
titanous: This should drop the `StartTimeout`

test/test_healthcheck.go
```diff
@@ -0,0 +1,86 @@
...
+	// start an app that is failing checks
+	app, _ := s.createAppWithService(t, "printer", &host.Service{
+		Name:   "healthcheck-failure",
+		Create: true,
+		Check:  &host.HealthCheck{Type: "tcp"},
+	})
+	t.Assert(flynn(t, "/", "-a", app.Name, "scale", "printer=1"), Succeeds)
+	// confirm that it's never registered
+	_, err := s.discoverdClient(t).Instances("healthcheck-failure", 10*time.Second)
+	t.Assert(err, c.NotNil)
```
titanous: Drop the StartTimeout so we don't have to wait ten seconds for this test to fail.

archSeer:
@titanous Addressed again. Note that TestFailure takes 11s for some reason, the `current == nil` event is sent straight away, but the down event is delayed. Might have something to do with `Threshold` on the monitor?

titanous:
@archSeer Yes, sounds like a threshold/timeout issue.

archSeer:
@titanous Ready for another pass, should pass CI after the server is restarted.

discoverd/health/register_test.go
```diff
@@ -136,7 +136,11 @@ func (RegisterSuite) TestRegister(c *C) {
 			Events:    make(chan MonitorEvent),
 		}
 		hb := reg.Register()
-		defer hb.Close()
+		defer func() {
+			go func() { <-unregisterChan }()
```
titanous: What does this goroutine do?

archSeer: 1559e5fc620367c9e1c1c1d47c03420598feb40f calls the wrapped heartbeater's close method, which then [triggers the mocked `Close()` method](https://github.com/flynn/flynn/blob/1559e5fc620367c9e1c1c1d47c03420598feb40f/discoverd/health/register_test.go#L61), that sends a notification to the `unregisterChan`. I'm making sure the channel gets drained here, so that the function does not block indefinitely.

titanous:
LGTM

PR #935, opened 2015-02-03
gather test utils together

heavenlyhash:
Gather test utils.

Shuffle some postgres stuff; the general package needn't carry pg references.

(This is triggered by more import cycle avoidance for volume persistence work.)


titanous:
LGTM

PR #939, opened 2015-02-03
cli: Output flag parse errors plainly.

### LINKED ISSUE #938: shutdown log nonsense in `flynn` error output
```text
$ flynn box
22:27:15.485829 handler.go:52: [box is not a flynn command. See 'flynn help']
```

@heavenlyhash please fix this.

heavenlyhash:
`shutdown.Fatal`'s behavior is to format output as a log line, which isn't relevant here.  Replace it with a plain log statement and explicit exit code.

Fixes #938.


lmars:
LGTM

Issue #942, opened 2015-02-03
controller: Successive deployments fail after timeout


jvatic:
```
Error deploying app release: Timed out waiting for deployment completion!
```
```
Error deploying app release: validation_error: Cannot create deploy, there is already one in progress for this app.
```
Logs: https://gist.github.com/anonymous/986405cfdad32a3c95d9

diclophis:
I am also running into this problem I think, when I go to `flynn set env` each time it returns:

    10:37:56.107555 handler.go:52: [Timed out waiting for deployment completion!]
    10:38:07.137258 handler.go:52: [Timed out waiting for deployment completion!]
    10:38:18.494601 handler.go:52: [Timed out waiting for deployment completion!]
    10:38:29.964765 handler.go:52: [Timed out waiting for deployment completion!]
    10:38:41.139897 handler.go:52: [Timed out waiting for deployment completion!]

In my debug info log I also have similar lines:

	2015-02-04 18:31:33 UTC ERROR:  duplicate key value violates unique constraint "formations_pkey"
	2015-02-04 18:31:33 UTC DETAIL:  Key (app_id, release_id)=(e958be71-b3d3-471f-8a43-f84bb802054a, e2e653e4-9638-4211-95b7-0bb27a97822c) already exists.
	2015-02-04 18:31:33 UTC STATEMENT:  INSERT INTO formations (app_id, release_id, processes) VALUES ($1, $2, $3) RETURNING created_at, updated_at
	2015-02-04 18:31:34 UTC LOG:  loaded library "$libdir/plugins/pgextwlist"
	2015-02-04 18:31:34 UTC LOG:  loaded library "$libdir/plugins/pgextwlist"
	2015-02-04 18:31:34 UTC LOG:  loaded library "$libdir/plugins/pgextwlist"
	2015-02-04 18:31:34 UTC LOG:  loaded library "$libdir/plugins/pgextwlist"
	2015-02-04 18:31:34 UTC ERROR:  duplicate key value violates unique constraint "job_cache_pkey"
	2015-02-04 18:31:34 UTC DETAIL:  Key (job_id, host_id)=(3cf10ac294ca4d898a261763aecca56e, XXX.com) already exists.
	2015-02-04 18:31:34 UTC STATEMENT:  INSERT INTO job_cache (job_id, host_id, app_id, release_id, process_type, state, meta) VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING created_at, updated_at
[...]

PR #946, opened 2015-02-03
test: Add taffy integration test

jvatic:
refs #942

### DIFF
--- a/test/test_taffy_deploy.go
+++ b/test/test_taffy_deploy.go
@@ -1,0 +1,104 @@
+package main
+
+import (
+	"bytes"
+
+	c "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/pkg/cluster"
+)
+
+type TaffyDeploySuite struct {
+	Helper
+}
+
+var _ = c.ConcurrentSuite(&TaffyDeploySuite{})
+
+func (s *TaffyDeploySuite) deployWithTaffy(t *c.C, app *ct.App, github map[string]string) {
+	client := s.controllerClient(t)
+
+	taffyRelease, err := client.GetAppRelease("taffy")
+	t.Assert(err, c.IsNil)
+
+	rwc, err := client.RunJobAttached("taffy", &ct.NewJob{
+		ReleaseID: taffyRelease.ID,
+		Cmd: []string{
+			app.Name,
+			github["clone_url"],
+			github["ref"],
+			github["sha"],
+		},
+		Meta: map[string]string{
+			"type":       "github",
+			"user_login": github["user_login"],
+			"repo_name":  github["repo_name"],
+			"ref":        github["ref"],
+			"sha":        github["sha"],
+			"clone_url":  github["clone_url"],
+			"app":        app.ID,
+		},
+	})
+	t.Assert(err, c.IsNil)
+	attachClient := cluster.NewAttachClient(rwc)
+	var outBuf bytes.Buffer
+	exit, err := attachClient.Receive(&outBuf, &outBuf)
+	t.Log(outBuf.String())
+	t.Assert(exit, c.Equals, 0)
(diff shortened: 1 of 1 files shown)


test/test_taffy_deploy.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+	})
+	t.Assert(err, c.IsNil)
+	attachClient := cluster.NewAttachClient(rwc)
+	var outBuf bytes.Buffer
+	exit, err := attachClient.Receive(&outBuf, &outBuf)
+	t.Log(outBuf.String())
+	outBuf.Reset()
+	t.Assert(exit, c.Equals, 0)
+	t.Assert(err, c.IsNil)
```
lmars: Could use `pkg/exec.Cmd`, unless there is a specific reason to go via the controller?

test/test_taffy_deploy.go
```diff
@@ -0,0 +1,119 @@
...
+	release, err := client.GetAppRelease(app.ID)
+	t.Assert(err, c.IsNil)
+
+	release = &ct.Release{
+		ArtifactID: artifact.ID,
+		Env:        release.Env,
+		Processes:  release.Processes,
+	}
+	t.Assert(client.CreateRelease(release), c.IsNil)
+	t.Assert(client.SetAppRelease(app.ID, release.ID), c.IsNil)
```
lmars: What is the purpose of changing the release to use the `test-apps` image? I think a comment would be good.

test/test_taffy_deploy.go
titanous: This test emulates dashboard deploys (there should be a comment), hence the use of the controller.

test/test_taffy_deploy.go
```diff
@@ -0,0 +1,119 @@
...
+			"app":        app.ID,
+		},
+	})
+	t.Assert(err, c.IsNil)
+	attachClient := cluster.NewAttachClient(rwc)
+	var outBuf bytes.Buffer
+	exit, err := attachClient.Receive(&outBuf, &outBuf)
+	t.Log(outBuf.String())
+	outBuf.Reset()
+	t.Assert(exit, c.Equals, 0)
```
lmars: Should we also test the actual outcome of running taffy? i.e. that the app got a new release?

lmars:
It may be worth adding helper methods to make the test more readable (e.g. `deployWithTaffy(app, url)`)

archSeer:
@jvatic @titanous Pushed a fix to the same branch.

titanous:
@jvatic please rebase this and address the comments.

test/test_taffy_deploy.go
jvatic: This is emulating [the dashboard](https://github.com/flynn/flynn/blob/ab3478079d1ae3418f3c97a5581e43bed0076d09/dashboard/app/lib/javascripts/dashboard/stores/app.js#L422-L449). As [releases no longer require an artifact](https://github.com/flynn/flynn/commit/6be3d4376bb4d2405b622f9b7f2444cd4c0743de), this step can be removed from both the test and dashboard app.

jvatic:
@lmars @titanous comments addressed, rebased on latest master

(see 03af0bc42d8ff2ca48f9caccc0ce89b554849e21 and 488860aed9149aff8f3b245115904ea9545794b9 for changes squashed into original commit)

test/test_taffy_deploy.go
```diff
@@ -0,0 +1,104 @@
...
+	c "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/pkg/cluster"
+)
+
+type TaffyDeploySuite struct {
+	Helper
+}
+
+var _ = c.Suite(&TaffyDeploySuite{})
```
titanous: I think this can be marked as a `ConcurrentSuite`?

titanous:
LGTM pending squash

PR #950, opened 2015-02-04
router: hijack conn before writing response

bgentry:
If we don't attempt to hijack the client conn before we write a response
to it, the hijack may fail, but we will have no way of telling the
client or backend about that.

Couldn't think of a way to write a test for this. Any ideas?

Regardless, this should fix the issue we discussed towards the end of #926. /cc @benburkert

### DIFF
--- a/router/proxy/reverseproxy.go
+++ b/router/proxy/reverseproxy.go
@@ -128,22 +128,31 @@ func (p *ReverseProxy) serveUpgrade(rw http.ResponseWriter, req *http.Request) {
 	}
 	defer uconn.Close()
 
-	p.writeResponse(rw, res)
-	res.Body.Close()
+	dconn, bufrw, err := rw.(http.Hijacker).Hijack()
+	if err != nil {
+		p.logf("router: hijack failed: %v", err)
+		rw.WriteHeader(http.StatusServiceUnavailable)
+		rw.Write(serviceUnavailable)
+		return
+	}
+	defer dconn.Close()
+
+	prepareResponseHeaders(res)
+	if res.StatusCode != 101 {
+		res.Header.Set("Connection", "close")
+	}
+	if err := res.Write(dconn); err != nil {
+		return
+	}
 
 	if res.StatusCode != 101 {
 		return
 	}
 
-	dconn, bufrw, err := rw.(http.Hijacker).Hijack()
-	if err != nil {
-		p.logf("router: hijack failed: %v", err)
-		return
-	}
 	joinConns(uconn, &streamConn{bufrw.Reader, dconn})
 }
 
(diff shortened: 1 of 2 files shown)


router/proxy/reverseproxy.go
```diff
@@ -128,22 +128,32 @@ func (p *ReverseProxy) serveUpgrade(rw http.ResponseWriter, req *http.Request) {
 	}
 	defer uconn.Close()
 
-	p.writeResponse(rw, res)
-	res.Body.Close()
+	dconn, bufrw, err := rw.(http.Hijacker).Hijack()
```
benburkert: I think the `prepareResponseHeaders` call and the status code check can happen moved before call to hijack. That way, if `res.StatusCode != 101`, it can use `rw` to send back the response. And I believe it's OK to always call `prepareResponseHeaders`.

benburkert:
> Couldn't think of a way to write a test for this. Any ideas?

the only way I can think of testing a failed hijack at this point is to build a request in the test, call `CloseNotify` on it, create a `ReverseProxy`, and call `ServeHTTP` directly with the request.

router/proxy/reverseproxy.go
benburkert: ohh, that might end up being more steps than what's below.

router/proxy/reverseproxy.go
```diff
@@ -128,22 +128,32 @@ func (p *ReverseProxy) serveUpgrade(rw http.ResponseWriter, req *http.Request) {
...
+	defer dconn.Close()
 
+	prepareResponseHeaders(res)
 	if res.StatusCode != 101 {
+		res.Close = true
+	}
+	if err := res.Write(dconn); err != nil {
 		return
 	}
+	res.Body.Close()
```
benburkert: isn't the body closed after `res.Write`?

benburkert:
> the only way I can think of testing a failed hijack at this point is to build a request in the test, call CloseNotify on it, create a ReverseProxy, and call ServeHTTP directly with the request.

this would probably have to be done in a test handler where the request is a real request that the handler received.

router/proxy/reverseproxy.go
bgentry: yep, missed that, good call :v:

router/proxy/reverseproxy.go
bgentry: yeah, I think it would, because it would need 2 blocks for sending back the response.

bgentry:
I adapted the existing websocket test to test 2 different scenarios: an upgrade on a new connection, and an upgrade on a reused connection. I think that's really the behavior we want to ensure.

titanous:
LGTM

PR #952, opened 2015-02-04
pkg/cluster,controller,dashboard: Return -1 from attach client if exit code unknown

jvatic:


jvatic:
ping @titanous

controller/jobs.go
```diff
@@ -279,7 +279,12 @@ func (c *controllerAPI) JobLog(ctx context.Context, w http.ResponseWriter, req *
 			return
 		}
 		if tail {
-			ch <- &sseLogChunk{Event: "exit", Data: []byte(fmt.Sprintf(`{"status": %d}`, exit))}
+			if exit == -1 {
+				// Don't send exit status if we don't know it
+				ch <- &sseLogChunk{Event: "exit"}
```
titanous: No need for this event, we already send eof

jvatic: `exit` is sent instead of `eof` (see `return` below), should this change?

titanous: Yeah, if we don't know the exit status, we should send eof

titanous:
LGTM

PR #955, opened 2015-02-04
router: context

bgentry:
This includes everything from #926 except for the CloseNotify stuff (which became an issue due to conflicts w/ Hijack).

Introduce contexts into the router stack. Use them to carry request metadata (currently just start time) down the stack.

Also, add a 120 second timeout for response headers from backends. There's no test for this yet, and I'm not sure if/how I should add one. Perhaps we could just test that the value is indeed set on the `httpTransport` since we can already trust that to use the setting correctly?

### DIFF
--- a/pkg/ctxhelper/ctxhelper.go
+++ b/pkg/ctxhelper/ctxhelper.go
@@ -1,6 +1,8 @@
 package ctxhelper
 
 import (
+	"time"
+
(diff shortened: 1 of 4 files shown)


bgentry:
Notably, the 120 second timeout only applies to non-upgrade requests. If we want to add it to the upgrade path too, that shouldn't be too difficult to do. We'll definitely want tests for that, though.

titanous:
LGTM. I'm fine with not adding a test.

PR #959, opened 2015-02-05
router: Postgres datastore

### LINKED ISSUE #549: Postgres datastore for the router
We've standardized on postgres for all persistent data storage, so the router should have a `DataStore` implementation that persists to PostgreSQL.

benburkert:
Store routing state in a PostgreSQL database. Closes #549.

#### TODO

- [x] fix broken tests.
- [x] remove etcd datastore.
- [x] investigate slow tests (14s on CI).

/cc @bgentry

### DIFF
--- a/router/client/client.go
+++ b/router/client/client.go
@@ -50,9 +50,9 @@ func NewWithAddr(addr string) Client {
 type Client interface {
 	// CreateRoute creates a new route.
 	CreateRoute(*router.Route) error
-	// SetRoute updates an existing route. If the route does not exist, it
-	// creates a new one.
-	SetRoute(*router.Route) error
+	// UpdateRoute updates an existing route by overwriting all fields on the route
+	// except ID and Domain.
+	UpdateRoute(*router.Route) error
 	// DeleteRoute deletes the route with the specified routeType and id.
 	DeleteRoute(routeType, id string) error
 	// GetRoute returns a route with the specified routeType and id.
@@ -66,8 +66,8 @@ func (c *client) CreateRoute(r *router.Route) error {
 	return c.Post("/routes", r, r)
 }
 
-func (c *client) SetRoute(r *router.Route) error {
-	return c.Put("/routes", r, r)
+func (c *client) UpdateRoute(r *router.Route) error {
+	return c.Put("/routes/"+r.Type+"/"+r.ID, r, r)
 }
 
 func (c *client) DeleteRoute(routeType, id string) error {
--- a/router/client/client.go
+++ b/router/client/client.go
@@ -53,10 +53,10 @@ type Client interface {
 	// SetRoute updates an existing route. If the route does not exist, it
 	// creates a new one.
 	SetRoute(*router.Route) error
-	// DeleteRoute deletes the route with the specified id.
-	DeleteRoute(id string) error
-	// GetRoute returns a route with the specified id.
-	GetRoute(id string) (*router.Route, error)
+	// DeleteRoute deletes the route with the specified routeType and id.
+	DeleteRoute(routeType, id string) error
+	// GetRoute returns a route with the specified routeType and id.
+	GetRoute(routeType, id string) (*router.Route, error)
 	// ListRoutes returns a list of routes. If parentRef is not empty, routes
 	// are filtered by the reference (ex: "controller/apps/myapp").
 	ListRoutes(parentRef string) ([]*router.Route, error)
@@ -70,13 +70,13 @@ func (c *client) SetRoute(r *router.Route) error {
 	return c.Put("/routes", r, r)
 }
 
-func (c *client) DeleteRoute(id string) error {
-	return c.Delete("/routes/" + id)
+func (c *client) DeleteRoute(routeType, id string) error {
+	return c.Delete("/routes/" + routeType + "/" + id)
 }
 
-func (c *client) GetRoute(id string) (*router.Route, error) {
+func (c *client) GetRoute(routeType, id string) (*router.Route, error) {
 	res := &router.Route{}
-	err := c.Get("/routes/"+id, res)
+	err := c.Get(fmt.Sprintf("/routes/%s/%s", routeType, id), res)
 	return res, err
 }
 
--- a/router/data_store.go
+++ b/router/data_store.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"strings"
 	"sync"
 	"time"
 
@@ -37,62 +36,136 @@ type SyncHandler interface {
 type pgDataStore struct {
 	pgx *pgx.ConnPool
 
-	rt string
+	routeType string
+	tableName string
 
 	doneo sync.Once
 	donec chan struct{}
 }
 
+const (
+	routeTypeHTTP = "http"
+	routeTypeTCP  = "tcp"
+	tableNameHTTP = "http_routes"
+	tableNameTCP  = "tcp_routes"
+)
+
 // NewPostgresDataStore returns a DataStore that stores route information in a
 // Postgres database. It uses pg_notify and a listener connection to watch for
 // route changes.
 func NewPostgresDataStore(routeType string, pgx *pgx.ConnPool) *pgDataStore {
+	tableName := ""
+	switch routeType {
+	case routeTypeHTTP:
+		tableName = tableNameHTTP
+	case routeTypeTCP:
+		tableName = tableNameTCP
+	default:
+		panic(fmt.Sprintf("unknown routeType: %q", routeType))
+	}
 	return &pgDataStore{
-		pgx:   pgx,
-		rt:    routeType,
-		donec: make(chan struct{}),
+		pgx:       pgx,
+		routeType: routeType,
+		tableName: tableName,
+		donec:     make(chan struct{}),
 	}
 }
 
-const sqlAddRoute = `INSERT INTO routes (parent_ref, type, config) VALUES ($1, $2, $3::json) RETURNING route_id, created_at, updated_at`
+const sqlAddRouteHTTP = `
+INSERT INTO ` + tableNameHTTP + ` (parent_ref, service, domain, tls_cert, tls_key, sticky)
+	VALUES ($1, $2, $3, $4, $5, $6)
+	RETURNING id, created_at, updated_at`
 
-func (d *pgDataStore) Add(r *router.Route) error {
-	if r.CreatedAt == nil {
-		r.CreatedAt = &time.Time{}
+const sqlAddRouteTCP = `
+INSERT INTO ` + tableNameTCP + ` (parent_ref, service, port)
+	VALUES ($1, $2, $3)
+	RETURNING id, created_at, updated_at`
+
+func (d *pgDataStore) Add(r *router.Route) (err error) {
+	switch d.tableName {
+	case tableNameHTTP:
+		err = d.pgx.QueryRow(
+			sqlAddRouteHTTP,
+			r.ParentRef,
+			r.Service,
+			r.Domain,
+			r.TLSCert,
+			r.TLSKey,
+			r.Sticky,
+		).Scan(&r.ID, &r.CreatedAt, &r.UpdatedAt)
+	case tableNameTCP:
+		err = d.pgx.QueryRow(
+			sqlAddRouteTCP,
+			r.ParentRef,
+			r.Service,
+			r.Port,
+		).Scan(&r.ID, &r.CreatedAt, &r.UpdatedAt)
 	}
-	if r.UpdatedAt == nil {
-		r.UpdatedAt = &time.Time{}
-	}
-	return d.pgx.QueryRow(sqlAddRoute, r.ParentRef, r.Type, r.Config).Scan(&r.ID, r.CreatedAt, r.UpdatedAt)
+	r.Type = d.routeType
+	return err
 }
 
-const sqlSetRoute = `UPDATE routes SET parent_ref = $1, type = $2, config = $3::json WHERE route_id = $4 RETURNING updated_at`
+const sqlSetRouteHTTP = `
+UPDATE %s SET parent_ref = $1, service = $2, domain = $3, tls_cert = $4, tls_key = $5, sticky = $6
+	WHERE id = $7 AND deleted_at IS NULL
+	RETURNING updated_at`
+
+const sqlSetRouteTCP = `
+UPDATE %s SET parent_ref = $1, service = $2, port = $3
+	WHERE id = $4 AND deleted_at IS NULL
+	RETURNING updated_at`
 
 func (d *pgDataStore) Set(r *router.Route) error {
-	if r.UpdatedAt == nil {
-		r.UpdatedAt = &time.Time{}
+	var row *pgx.Row
+
+	switch d.tableName {
+	case tableNameHTTP:
+		row = d.pgx.QueryRow(
+			fmt.Sprintf(sqlSetRouteHTTP, d.tableName),
+			r.ParentRef,
+			r.Service,
+			r.Domain,
+			r.TLSCert,
+			r.TLSKey,
+			r.Sticky,
+			r.ID,
+		)
+	case tableNameTCP:
+		row = d.pgx.QueryRow(
+			fmt.Sprintf(sqlSetRouteTCP, d.tableName),
+			r.ParentRef,
+			r.Service,
+			r.Port,
+			r.ID,
+		)
 	}
-	err := d.pgx.QueryRow(sqlSetRoute, r.ParentRef, r.Type, r.Config, r.ID).Scan(r.UpdatedAt)
+	err := row.Scan(&r.UpdatedAt)
 	if err == pgx.ErrNoRows {
 		return ErrNotFound
 	}
+	r.Type = d.routeType
 	return err
 }
 
-const sqlRemoveRoute = `UPDATE routes SET deleted_at = now() WHERE route_id = $1`
+const sqlRemoveRoute = `UPDATE %s SET deleted_at = now() WHERE id = $1`
 
 func (d *pgDataStore) Remove(id string) error {
-	_, err := d.pgx.Exec(sqlRemoveRoute, id)
+	_, err := d.pgx.Exec(fmt.Sprintf(sqlRemoveRoute, d.tableName), id)
 	return err
 }
 
-const sqlGetRoute = `SELECT parent_ref, type, config, created_at, updated_at FROM routes WHERE route_id = $1 AND type = $2 AND deleted_at IS NULL`
+const sqlGetRoute = `SELECT %s FROM %s WHERE id = $1 AND deleted_at IS NULL`
 
 func (d *pgDataStore) Get(id string) (*router.Route, error) {
-	r := newRoute()
-	row := d.pgx.QueryRow(sqlGetRoute, id, d.rt)
+	if id == "" {
+		return nil, ErrNotFound
+	}
 
-	err := row.Scan(&r.ParentRef, &r.Type, r.Config, r.CreatedAt, r.UpdatedAt)
+	query := fmt.Sprintf(sqlGetRoute, d.columnNames(), d.tableName)
+	row := d.pgx.QueryRow(query, id)
+
+	r := &router.Route{}
+	err := d.scanRoute(r, row)
 	if err == pgx.ErrNoRows {
 		return nil, ErrNotFound
 	}
@@ -100,14 +173,14 @@ func (d *pgDataStore) Get(id string) (*router.Route, error) {
 		return nil, err
 	}
 
-	r.ID = id
 	return r, nil
 }
 
-const sqlListRoutes = `SELECT route_id, parent_ref, type, config, created_at, updated_at FROM routes WHERE type = $1 AND deleted_at IS NULL`
+const sqlListRoutes = `SELECT %s FROM %s WHERE deleted_at IS NULL`
 
 func (d *pgDataStore) List() ([]*router.Route, error) {
-	rows, err := d.pgx.Query(sqlListRoutes, d.rt)
+	query := fmt.Sprintf(sqlListRoutes, d.columnNames(), d.tableName)
+	rows, err := d.pgx.Query(query)
 	if err != nil {
 		return nil, err
 	}
@@ -115,12 +188,13 @@ func (d *pgDataStore) List() ([]*router.Route, error) {
 
 	routes := []*router.Route{}
 	for rows.Next() {
-		r := newRoute()
-		err := rows.Scan(&r.ID, &r.ParentRef, &r.Type, r.Config, r.CreatedAt, r.UpdatedAt)
+		r := &router.Route{}
+		err := d.scanRoute(r, rows)
 		if err != nil {
 			return nil, err
 		}
 
+		r.Type = d.routeType
 		routes = append(routes, r)
 	}
 	return routes, rows.Err()
@@ -189,13 +263,13 @@ func (d *pgDataStore) startListener(idc chan<- string) error {
 	if err != nil {
 		return err
 	}
-	if err = conn.Listen("routes"); err != nil {
+	if err = conn.Listen(d.tableName); err != nil {
 		d.pgx.Release(conn)
 		return err
 	}
 
 	go func() {
-		defer unlistenAndRelease(d.pgx, conn, "routes")
+		defer unlistenAndRelease(d.pgx, conn, d.tableName)
 		defer close(idc)
 
 		for {
@@ -214,21 +288,59 @@ func (d *pgDataStore) startListener(idc chan<- string) error {
 				return
 			}
 
-			data := strings.Split(notification.Payload, ":")
-			if len(data) != 2 {
-				log.Printf("router: invalid route notification: %s", notification.Payload)
-				defer d.StopSync()
-				return
-			}
-
-			routeType, id := data[0], data[1]
-			if routeType == d.rt {
-				idc <- id
-			}
+			idc <- notification.Payload
 		}
 	}()
 
 	return nil
+}
+
+const (
+	selectColumnsHTTP = "id, parent_ref, service, domain, sticky, tls_cert, tls_key, created_at, updated_at"
+	selectColumnsTCP  = "id, parent_ref, service, port, created_at, updated_at"
+)
+
+func (d *pgDataStore) columnNames() string {
+	switch d.routeType {
+	case routeTypeHTTP:
+		return selectColumnsHTTP
+	case routeTypeTCP:
+		return selectColumnsTCP
+	default:
+		panic(fmt.Sprintf("unknown routeType: %q", d.routeType))
+	}
+}
+
+type scannable interface {
+	Scan(dest ...interface{}) (err error)
+}
+
+func (d *pgDataStore) scanRoute(route *router.Route, s scannable) error {
+	route.Type = d.routeType
+	switch d.tableName {
+	case tableNameHTTP:
+		return s.Scan(
+			&route.ID,
+			&route.ParentRef,
+			&route.Service,
+			&route.Domain,
+			&route.Sticky,
+			&route.TLSCert,
+			&route.TLSKey,
+			&route.CreatedAt,
+			&route.UpdatedAt,
+		)
+	case tableNameTCP:
+		return s.Scan(
+			&route.ID,
+			&route.ParentRef,
+			&route.Service,
+			&route.Port,
+			&route.CreatedAt,
+			&route.UpdatedAt,
+		)
+	}
+	panic("unknown tableName: " + d.tableName)
 }
 
 const sqlUnlisten = `UNLISTEN %s`
@@ -240,12 +352,4 @@ func unlistenAndRelease(pool *pgx.ConnPool, conn *pgx.Conn, channel string) {
 		return
 	}
 	pool.Release(conn)
-}
-
-func newRoute() *router.Route {
-	return &router.Route{
-		Config:    &router.Config{},
-		CreatedAt: &time.Time{},
-		UpdatedAt: &time.Time{},
-	}
 }
--- a/router/data_store.go
+++ b/router/data_store.go
@@ -15,7 +15,7 @@ var ErrNotFound = errors.New("router: route not found")
 
 type DataStore interface {
 	Add(route *router.Route) error
-	Set(route *router.Route) error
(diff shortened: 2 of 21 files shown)


router/schema.go
```diff
@@ -0,0 +1,46 @@
...
+		NEW.updated_at = CURRENT_TIMESTAMP AT TIME ZONE 'UTC';
+		RETURN NEW;
+	END;
+$$ language 'plpgsql'`,
+		`CREATE TYPE route_type AS ENUM ('http', 'tcp')`,
+		`CREATE TABLE routes (
+	route_id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+	parent_ref text NOT NULL,
+	type route_type NOT NULL,
+	config hstore NOT NULL,
```
bgentry: I don't think hstore is actually appropriate for this column. That type essentially stores a `map[string]string`, but our config structs contain some `bool` and `int` fields. Maybe we should just use the JSON type?

router/tcp.go
```diff
@@ -34,6 +34,21 @@ type TCPListener struct {
 	closed   bool
 }
 
+func (s *TCPListener) List() ([]*router.Route, error) {
```
bgentry: I can't figure out why this function was added. Doesn't look like it's being used anywhere?

bgentry:
I wrote 010169ff807133bde720a4640746ffa51a174d16 (in bgentry/flynn@router-pg-datastore-blake ) to switch from hstore to json type for the config column.

@titanous one question that I probably should have cleared up sooner.. Are we focused right now on just changing the data store for service metadata (routes & domains)? We're not changing anything to do with discoverd & its service discovery, right?

router/schema.go
benburkert: The bools & number values keep their type on the way out thanks to `hstore_to_json_loose`. The values are stored as unescaped strings, so that function produces the same JSON going at as going in.

bgentry: Yeah, you're right that we don't lose anything from the type conversions. But what do we gain from the additional complexity of using hstore?

titanous:
No changes to service discovery.

benburkert:
I don't really have an opinion on using `hstore` vs `json` on that column, I picked `hstore` b/c that's what `controller/schema.go` used for a similar field.

@titanous is there a design/plan for how the config data will evolve? b/c that will most likely dictate the column type.

router/tcp.go
benburkert: `pgDataStore.List` doesn't filter on http vs tcp routes, it returns all the routes. This was a hack to filter out routes by type, there is an equivalent method in `HTTPListener`. Without these, the API returns duplicate routes (which causes `TestAPIListRoutes` to fail).

titanous:
I think JSON makes sense. We could end up with more nested structures in the future.

benburkert:
@bgentry I pushed a few more fixes, but the tests are still very broken. I don

benburkert:
@titanous this is ready for :eyes: but you might want to hold off on merging this to master. @bgentry is working on a follow up PR to break up the single `routes` table into `http_routes` & `tcp_routes` as part of the work to address domain conflicts documented in a TODO. The schema changes will either require a second migration, or changes to the migration introduced here. Altering a migration after it's on master is pretty risky, so moving this PR to a long lived branch would let us avoid that.

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
+	return &pgDataStore{
+		pgx:   pgx,
+		rt:    routeType,
+		donec: make(chan struct{}),
+	}
 }
 
-var ErrExists = errors.New("router: route already exists")
-var ErrNotFound = errors.New("router: route not found")
+var sqlAddRoute = `INSERT INTO routes (parent_ref, type, config) VALUES ($1, $2, $3::json) RETURNING route_id, created_at, updated_at`
```
titanous: These query strings can all be constants.

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
-		return err
+func (d *pgDataStore) Add(r *router.Route) error {
+	if r.CreatedAt == nil {
+		r.CreatedAt = new(time.Time)
 	}
-	_, err = s.etcd.Create(s.path(r.ID), string(data), 0)
-	if e, ok := err.(*etcd.EtcdError); ok && e.ErrorCode == 105 {
-		err = ErrExists
+	if r.UpdatedAt == nil {
+		r.UpdatedAt = new(time.Time)
```
titanous: We don't use `new` for structs.

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
 	}
-	return err
+	return d.pgx.QueryRow(sqlAddRoute, r.ParentRef, r.Type, r.Config).Scan(&r.ID, r.CreatedAt, r.UpdatedAt)
 }
 
-func (s *etcdDataStore) Set(r *router.Route) error {
-	data, err := json.Marshal(r)
-	if err != nil {
-		return err
+var sqlSetRoute = `UPDATE routes SET parent_ref = $1, type = $2, config = $3::json WHERE route_id = $4 RETURNING updated_at`
```
titanous: `updated_at` isn't getting bumped?

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
+
+func (d *pgDataStore) Set(r *router.Route) error {
+	if r.UpdatedAt == nil {
+		r.UpdatedAt = new(time.Time)
+	}
+	err := d.pgx.QueryRow(sqlSetRoute, r.ParentRef, r.Type, r.Config, r.ID).Scan(r.UpdatedAt)
```
titanous: Does pgx not handle `**time.Time` properly?

router/schema.go
```diff
@@ -0,0 +1,42 @@
...
+		`CREATE FUNCTION set_updated_at_column() RETURNS TRIGGER AS $$
+	BEGIN
+		NEW.updated_at = CURRENT_TIMESTAMP AT TIME ZONE 'UTC';
+		RETURN NEW;
+	END;
+$$ language 'plpgsql'`,
+		`CREATE TYPE route_type AS ENUM ('http', 'tcp')`,
+		`CREATE TABLE routes (
+	route_id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+	parent_ref text NOT NULL,
```
titanous: This is not actually required, so shouldn't require a `NOT NULL` constraint?

router/server.go
```diff
@@ -85,27 +84,31 @@ func main() {
...
+		shutdown.Fatal(err)
 	}
-	etcdc := etcd.NewClient(etcdAddrs)
 
-	prefix := os.Getenv("ETCD_PREFIX")
-	if prefix == "" {
-		prefix = "/router"
+	pgxpool, err := pgx.NewConnPool(pgx.ConnPoolConfig{
+		ConnConfig: pgx.ConnConfig{
+			Database: os.Getenv("PGDATABASE"),
```
titanous: This is insufficient to connect to postgres when running inside of flynn. I believe the deployer has an example of setting up a pgx connection pool.

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
-	r := &router.Route{}
-	err = json.Unmarshal([]byte(res.Node.Value), r)
-	return r, err
+
+	return r, nil
 }
 
-func (s *etcdDataStore) List() ([]*router.Route, error) {
-	res, err := s.etcd.Get(s.prefix, false, true)
+var sqlListRoutes = `SELECT route_id, parent_ref, type, config, created_at, updated_at FROM routes WHERE type = $1 AND deleted_at IS NULL`
```
titanous: This query probably needs an index, but we can skip it if the table split is coming next.

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
 		return nil, err
 	}
 
-	routes := make([]*router.Route, len(res.Node.Nodes))
-	for i, node := range res.Node.Nodes {
-		r := &router.Route{}
-		if err := json.Unmarshal([]byte(node.Value), r); err != nil {
+	routes := []*router.Route{}
+	for rows.Next() {
+		r := &router.Route{Config: new(router.Config), CreatedAt: new(time.Time), UpdatedAt: new(time.Time)}
```
titanous: This route initialization seems to happen a lot (I assume to work around pgx's scanner). Perhaps make a helper?

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
-			if res.Action == "delete" {
-				delete(keys, id)
-				err = h.Remove(id)
-			} else {
-				keys[id] = res.Node.ModifiedIndex
-				route := &router.Route{}
-				if err = json.Unmarshal([]byte(res.Node.Value), route); err != nil {
-					goto fail
+	if err := h.Set(route); err != nil {
+		log.Printf("router: sync handler set error: %s", err)
```
titanous: Should probably log the route ID for all of these.

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
+	conn, err := d.pgx.Acquire()
+	if err != nil {
+		return err
+	}
+	if err = conn.Listen("routes"); err != nil {
+		d.pgx.Release(conn)
+		return err
+	}
+
+	mu := sync.Mutex{}
```
titanous: Could do `var mu sync.Mutex`

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
+		for {
+			notification, err := conn.WaitForNotification(time.Second)
+			if err == pgx.ErrNotificationTimeout {
+				mu.Lock()
+				if done {
+					mu.Unlock()
+					return
 				}
-				err = h.Set(route)
+				mu.Unlock()
```
titanous: Should probably check `done` whether or not we get a timeout?

router/data_store.go
```diff
@@ -39,196 +33,206 @@ type SyncHandler interface {
...
-				log.Printf("Error while processing update from etcd: %s, %#v", err, res.Node)
+				log.Printf("router: notifier error: %s", err)
+				d.StopSync()
+				return
+			}
+
+			data := strings.Split(notification.Payload, ":")
+			if len(data) != 2 {
+				log.Printf("router: invalid route notification: %s", notification.Payload)
+				d.StopSync()
```
titanous: Could defer this since it's wrapped in a sync.Once

router/data_store.go
bgentry: yeah, I noticed the same and made the change in my branch. Feel free to change here too if you want.

router/data_store.go
benburkert: it's updated by the `set_updated_at_routes` trigger on `UPDATE`s.

router/data_store.go
benburkert: unfortunately not.

router/schema.go
benburkert: Right now it will default to storing an empty string, so dropping that constraint adds another possible state that ends up being an empty string in the `Route` object anyways.

titanous: Okay, that's fine.

router/data_store.go
titanous: Ah, missed that.

router/data_store.go
benburkert: yup, and it looks like that was what caused the tests to slow down. :smile:

benburkert:
ok, I addressed the feedback. Unfortunately CI is still failing and I'm not sure why.

bgentry:
@benburkert: so aside from a couple of TODOs and figuring out the sqlstruct situation, this is ready to go. Should I bring these two commits into router-pg-datastore? https://github.com/benburkert/flynn/compare/benburkert:router-pg-datastore...router-pg-datastore-more-schema

Avoiding sqlstruct should be pretty straightforward. Just need two lists of columns, and a single function to scan a list of columns into a `Route` struct.

benburkert:
> Should I bring these two commits into router-pg-datastore?

:+1:

bgentry:
almost done ditching sqlstruct, will bring them in after that

bgentry:
ok, there are a couple of TODOs outstanding, but otherwise I think this is in pretty good shape.

Oh, it also needs JSON schema updates. Maybe somebody can review whether the API changes are ok before I alter the relevant JSON schema.

router/http_test.go
```diff
@@ -1005,6 +1004,7 @@ func (s *S) TestDefaultServerKeypair(c *C) {
 	assertGet(c, "https://"+l.TLSAddr, "foo.example.com", "2")
 }
 
+// TODO(bgentry): validate this in Postgres??
 func (s *S) TestCaseInsensitiveDomain(c *C) {
```
bgentry: thoughts on this one? It's great that the router can work with non-lowercase domains, but maybe we should enforce that before shoving the data into Postgres.

router/types/types.go
```diff
@@ -1,88 +1,138 @@
...
 
+// Route is a struct that combines the fields of HTTPRoute and TCPRoute
+// for easy JSON marshaling.
 type Route struct {
-	ID        string `json:"id,omitempty"`
+	// Type is the type of Route, either "http" or "tcp".
+	Type string `json:"type"`
+	// ID is the unique ID of this route.
+	ID string `json:"id,omitempty"`
+	// ParentRef is ... TODO
```
bgentry: Struggling to come up with the right description for this field. Ideas?

router/types/types.go
```diff
@@ -1,88 +1,138 @@
...
+	default:
+		panic(fmt.Sprintf("unknown table name: %q", r.Type))
+	}
+}
+
+// HTTPRoute is an HTTP Route.
 type HTTPRoute struct {
-	*Route  `json:"-"`
+	// TODO(bgentry): remove json tags here, or make this serialize properly with
+	// a Type field.
```
bgentry: Basically, in order for either `HTTPRoute` or `TCPRoute` to serialize properly, they need to also have a `type` field added to the fields listed below. There are 2 ways I can think of to do that:

* Add a `MarshalJSON()` method that uses a private (anonymous?) struct specifically for serialization, which also includes a `type` field
* Add a `MarshalJSON()` method that calls `ToRoute()` and then marshals the result of that

Otherwise we can just eliminate these tags altogether, which means you always need to serialize directly with a `Route`.

bgentry:
@benburkert @titanous all of the issues I'm aware of with this are pending feedback.

It also hasn't been reviewed since I split it into 2 postgres tables, so please look over the resulting diff as well.

router/http_test.go
titanous: Yeah, it's worth enforcing domains are lowercase domain-looking strings.

router/types/types.go
titanous: `ParentRef` is an external opaque identifier used by the route creator for filtering and correlation. That being said, it's kind of a hack and we should come up with a better way of doing this (the controller puts the app ID in there)

router/types/types.go
benburkert: > Add a MarshalJSON() method that calls ToRoute() and then marshals the result of that

I like that idea.

router/http_test.go
benburkert: I think this test is still useful b/c the domain in the request still has to be downcase'd.

benburkert:
> That being said, it's kind of a hack and we should come up with a better way of doing this (the controller puts the app ID in there)

Then perhaps we should add a `meta hstore` column that is stored in `Route` as `map[string]string`, like with the `App` type in controller. That way clients can add any associative tracking data there.

router/client/client.go
```diff
@@ -50,13 +50,13 @@ func NewWithAddr(addr string) Client {
...
-	DeleteRoute(id string) error
-	// GetRoute returns a route with the specified id.
-	GetRoute(id string) (*router.Route, error)
+	// UpdateRoute updates an existing route by overwriting all fields on the route
+	// except ID and Domain.
+	UpdateRoute(*router.Route) error
+	// DeleteRoute deletes the route with the specified routeType and id.
+	DeleteRoute(routeType, id string) error
+	// GetRoute returns a route with the specified routeType and id.
+	GetRoute(routeType, id string) (*router.Route, error)
```
titanous: What do you think about making `routeType` a custom type with constants for the valid options?

titanous:
> Then perhaps we should add a `meta hstore` column that is stored in `Route` as `map[string]string`, like with the `App` type in controller. That way clients can add any associative tracking data there.

Yeah, I like that, please open an issue.

router/client/client.go
```diff
@@ -66,17 +66,17 @@ func (c *client) CreateRoute(r *router.Route) error {
...
-	return c.Delete("/routes/" + id)
+func (c *client) DeleteRoute(routeType, id string) error {
+	return c.Delete("/routes/" + routeType + "/" + id)
 }
 
-func (c *client) GetRoute(id string) (*router.Route, error) {
+func (c *client) GetRoute(routeType, id string) (*router.Route, error) {
 	res := &router.Route{}
-	err := c.Get("/routes/"+id, res)
+	err := c.Get("/routes/"+routeType+"/"+id, res)
```
titanous: Might be easier to `fmt.Sprintf` these.

router/data_store.go
```diff
@@ -39,196 +33,315 @@ type SyncHandler interface {
...
+const sqlAddRouteTCP = `
+INSERT INTO %s (parent_ref, service, port)
+	VALUES ($1, $2, $3)
+	RETURNING id, created_at, updated_at`
+
+func (d *pgDataStore) Add(r *router.Route) (err error) {
+	switch d.tableName {
+	case "http_routes":
+		err = d.pgx.QueryRow(
+			fmt.Sprintf(sqlAddRouteHTTP, d.tableName),
```
titanous: Any reason to interpolate the table name? The queries are different for each table.

router/data_store.go
```diff
@@ -39,196 +33,315 @@ type SyncHandler interface {
...
+			&route.ParentRef,
+			&route.Service,
+			&route.Domain,
+			&route.Sticky,
+			&route.TLSCert,
+			&route.TLSKey,
+			&route.CreatedAt,
+			&route.UpdatedAt,
+		)
+	case "tcp_routes":
```
titanous: Seems like we should use constants for these table names.

benburkert:
@bgentry, @titanous I added an enum style type for `RouterType`, but it's a bit awkward and i'm not convinced it simplifies anything. Think it's worth including? https://github.com/benburkert/flynn/commit/cc739543c03920118fd30f8b4fa50a966e21d444

benburkert:
the controller tests are failing because they try to set a value for `service` on http routes, which is accounted for in the current schema. Should http routes have a service field? or should the controller test remove the service field?

titanous:
@benburkert Yeah, agree about the awkwardness. Perhaps just a constant for the table names?

titanous:
I think the controller tests are failing because the schemas haven't been updated for the new layout.

router/data_store.go
bgentry: yeah, good call, didn't even realize how much these were being repeated. I'm also changing the `routeType` values to constants (`http` and `tcp`).

router/data_store.go
bgentry: Well, the only drawback of not doing that is that we end up with `"http_routes"` strewn all over this file. The benefit would be that I can drop the `d.tableName` field altogether and just switch on `d.routeType`. Think that's a net improvement?

titanous: It seems like some queries are specific to the table, and some aren't? All I'm suggesting is that the queries that are specific to a table have the table name in the query constant.

bgentry: Something like this would at least avoid redefining that string in multiple places:

```go
const sqlAddRouteHTTP = `
INSERT INTO ` + tableNameHTTP + ` (parent_ref, service, domain, tls_cert, tls_key, sticky)
	VALUES ($1, $2, $3, $4, $5, $6)
	RETURNING id, created_at, updated_at`

const sqlAddRouteTCP = `
INSERT INTO ` + tableNameTCP + ` (parent_ref, service, port)
	VALUES ($1, $2, $3)
	RETURNING id, created_at, updated_at`
```

that ok?

titanous: Yep, fine with me.

router/http_test.go
bgentry: So I'm not actually sure how I should go about doing this validation. There's many different places where it could be added, from the schema and Controller API all the way down to Postgres itself. Is there a convention for this sort of thing across the project?

bgentry:
Ok, I think I've fixed everything except for the schema and the domain validation.

router/http_test.go
titanous: We're still establishing conventions. One layer where we should apply it is via the JSON schemas. I think they exist for the router, so we should probably just open a ticket for adding them to the API. The other place that might make sense is as a postgres constraint, but that doesn't seem to important as all writes will go through the API.

bgentry:
I made an attempt to fix the schema, though I'm pretty sure there's some stuff missing from it. One thing in particular that's causing some of the remaining failures is that Flynn's `random.UUID()` [returns a UUID](https://github.com/flynn/flynn/blob/9d35ea6015a2f9c767bb8f9349521f81daa15814/pkg/random/random.go#L37-L44) formatted without separators. That's used by some of these tests to generate a fake route ID ([see here](https://github.com/flynn/flynn/blob/9d35ea6015a2f9c767bb8f9349521f81daa15814/controller/routes_test.go#L28)).

However, since we're using Postgres's UUID type as the primary key in these tables, that free-form UUID won't work as input:

```
STATEMENT:  SELECT release_id, artifact_id, data, created_at FROM releases WHERE release_id = $1 AND deleted_at IS NULL
ERROR:  invalid input syntax for uuid: "3703ff6d2cf04afab7108f765473a56cfail"
```

This results in some controller test logs like this:

```
52 method=GET path=/apps/4ca4ff7561a44cc98c56d7db4af7367a/routes/tcp/ee4177aed1eb4be3b0f4964300a69328 client_ip=127.0.0.1
INFO[02-11|15:20:10] request completed                        component=controller req_id=72b71ff0793045b8ba4b446481723052 status=404 duration=210.086µs
```

What's the best way to fix that? Make another function in `random` that returns a "proper" UUID?

titanous:
> What's the best way to fix that?

We have a helper for that:

bgentry:
I'm guessing this is a sign of the router not starting properly, maybe because the postgres database hasn't been created during the bootstrap?

```
bootstrap ===> controller-cert start
bootstrap ===> controller-cert done
bootstrap ===> router-wait start
bootstrap ===> router-wait error
bootstrap: timed out waiting for http://router-api.discoverd, last response "Get http://router-api.discoverd: dial tcp: lookup router-api.discoverd: no such host"
```

titanous:
@bgentry Correct. Bootstrap should be capable of creating the database, I can walk you through it on IRC>

titanous:
Ping @jvatic, please review the schema changes.

titanous:
I made a change to the host service that should fix the DNS errors on CI.

router/example-generator/examples.go
```diff
@@ -30,7 +30,7 @@ func main() {
 
 	examples := []g.Example{
 		{"route_create", e.createRoute},
-		{"route_set", e.setRoute},
+		{"route_set", e.updateRoute},
```
jvatic: `s/route_set/route_update/` to match the JSON schema

bgentry: :+1:

bgentry:
Made the schema change and gave this a fresh rebase on master.

Now struggling to figure out where [this CI failure](https://ci.flynn.io/builds/20150212044346-827ca63d) is coming from:

```
[router] 2015/02/12 04:45:20.953209 Started GET /
[router] 2015/02/12 04:45:20.953542 Completed 404 Not Found in 1.311955ms
2015/02/12 04:45:20 router: sync handler set error: 558e4fb7-4b1a-4241-a146-6f60c9a4e5ca, http: read on closed response body
2015/02/12 04:45:21 router: sync handler set error: 9022515e-5712-4c2a-adac-6ce69309c322, http: read on closed response body
```

I think I've traced it down in to the discoverd client and `httpclient.Stream()`. Not sure what could have broken there.

router/data_store.go
```diff
@@ -39,196 +33,322 @@ type SyncHandler interface {
...
+			r.TLSKey,
+			r.Sticky,
+		).Scan(&r.ID, &r.CreatedAt, &r.UpdatedAt)
+	case tableNameTCP:
+		err = d.pgx.QueryRow(
+			sqlAddRouteTCP,
+			r.ParentRef,
+			r.Service,
+			r.Port,
+		).Scan(&r.ID, &r.CreatedAt, &r.UpdatedAt)
```
titanous: In the controller we strip the dashes from UUIDs before they hit the presentation layer. I can't remember the rationale for it, and I'm open to reversing that decision, but that should happen on another ticket.

bgentry: so you want us to change this branch to strip dashes when returning from the datastore?

titanous: Well I don't think anything will break, so we can leave it. I just wanted to note the inconsistency, as we should pick one or the other at some point in the future.

bgentry: opened #995 for that.

bgentry:
So I'm pretty much stuck on this. Can't think of how to reproduce this issue locally, nor has my digging turned up anything obvious. The only thing that seems to be making an external request from within `httpSyncHandler.Set()` is discoverd-related, and I don't see how anything we've changed would affect that. Thoughts? :confused:

benburkert:
I think the difference is that the existing code [does not stop the sync loop on an error](https://github.com/flynn/flynn/blob/master/router/data_store.go#L144). Perhaps we should log & ignore errors returned from `Set`?

bgentry:
hmm, it seems like a strange behavior to simply continue after the error. That, and doing so doesn't actually fix this problem. Seems like something is preventing discoverd from getting going.

titanous:
It looks liked discoverd is being given an empty service name:

```text
watch events ""
2015/02/12 22:51:04 router: sync handler set error: d8da57f7-af53-4010-b1c0-95361589047d, http: read on closed response body
```

I've filed #998 to track not getting a 404 back.

bgentry:
ok, progress. I think the remaining failures are being caused by routes not being deleted properly. The problem is that we split up the route's ID and type into different fields, and the router + controller API actions for deleting a route are under `DELETE /routes/:route_type/:id`. But the CLI (and the controller.Client) don't yet know about those separate fields, so they try to `DELETE /routes/:id`.

I'm not sure if I should attempt to make sure that we only expose a combined ID to the outside (`:route_type/:id`) or if we should just change the CLI so that you have to specify both of those fields.

titanous:
> I'm not sure if I should attempt to make sure that we only expose a combined ID to the outside (:route_type/:id) or if we should just change the CLI so that you have to specify both of those fields.

Currently the type information is shown as part of the ID in the CLI, not sure where that happens, but we can keep it:

```text
$ flynn route
ROUTE                                            SERVICE                      ID
http:wolves-fascinate-bruton.dev.localflynn.com  wolves-fascinate-bruton-web  http/62a362c62ceac49208bae272601d1bb3

$ flynn route remove http/62a362c62ceac49208bae272601d1bb3
Route http/62a362c62ceac49208bae272601d1bb3 removed.
```

bgentry:
down to one last failure, `ControllerSuite.TestExampleOutput`

bgentry:
oh hey, the tests passed! :beers:

benburkert:
:boom: :metal:

router/http_test.go
titanous: Lets drop this TODO since we have an issue about JSON schemas.

router/schema.go
```diff
@@ -0,0 +1,85 @@
...
+	FOR EACH ROW EXECUTE PROCEDURE notify_tcp_route_update()`,
+
+		// http routes
+
+		`
+CREATE TABLE http_routes (
+	id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+	parent_ref varchar(255) NOT NULL,
+	service varchar(255) NOT NULL CHECK (service <> ''),
+	domain varchar(255) NOT NULL,
```
titanous: Seems like a few other fields should get the not empty check, (`domain` here, `service`/`port` for TCP)

router/types/types.go
```diff
@@ -2,87 +2,155 @@ package router
...
-	json.Unmarshal(*r.Config, route)
-	return route
 }
 
-func (r *Route) TCPRoute() *TCPRoute {
-	rCopy := *r
-	if rCopy.Config == nil {
-		empty := json.RawMessage(`{}`)
-		rCopy.Config = &empty
+func (r Route) routeForDB() interface{} {
```
titanous: This doesn't appear to be used?

bgentry:
Fixed everything you commented on so far. :v:

router/schema.go
```diff
@@ -0,0 +1,85 @@
...
+	FOR EACH ROW EXECUTE PROCEDURE notify_tcp_route_update()`,
+
+		// http routes
+
+		`
+CREATE TABLE http_routes (
+	id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+	parent_ref varchar(255) NOT NULL,
+	service varchar(255) NOT NULL CHECK (service <> ''),
+	domain varchar(255) NOT NULL  CHECK (service <> ''),
```
titanous: I'm not sure this check is correct, should it say `service`?

bgentry: Ugh, sloppy. Sorry about that. Can fix after dinner. Probably a similar
mistake in tcp_routes

bgentry:
Fixed those check constraints. Tests passing again.

titanous:
LGTM. Give it a rebase/squash and we can merge!

bgentry:
@titanous rebased and squashed out a couple commits, lmk if you want me to do more on that front.

titanous:
I'd like to see this squashed more if possible. In general any fixups to code added in the pull request should be squashed.

bgentry:
Not sure what broke in the tests, all I did was rebase on master and squash a couple small things. Was working as of this build: https://ci.flynn.io/builds/20150213042709-f852e40f (commit 4e0fe6462c2c50d0203a283e560ad3fc211515c1)

titanous:
@bgentry Looks like it's an intermittent failure, I restarted and the unit tests passed.

bgentry:
squashed it down to 6 commits and rebased on master again.

titanous:
LGTM

titanous:
Awesome! :balloon:

PR #961, opened 2015-02-05
controller: Fix deployments with zero processes.

### LINKED ISSUE #942: controller: Successive deployments fail after timeout

archSeer:
We were checking that map length was 0, when we could have a key
that is set to zero (and then len(map) != 0). Count all of the
processes instead.

Fixes #942.


titanous:
The test for this is in #946 but we will expedite merging the fix.

titanous:
LGTM

PR #964, opened 2015-02-05
router: detect early termination of downstream TCP connections

benburkert:
Wrap downstream TCP connections in a net.Conn that implements CloseNotify so that early termination can be detected before dialing a backend.

I'm not sure how to write a test for this, any ideas?

### DIFF
--- a/router/proxy/reverseproxy_test.go
+++ b/router/proxy/reverseproxy_test.go
@@ -1,0 +1,43 @@
+package proxy
+
+import (
+	"errors"
+	"net"
+	"net/http"
+	"testing"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/net/context"
+)
+
+func TestServeConnClientGone(t *testing.T) {
+	control, conn := net.Pipe()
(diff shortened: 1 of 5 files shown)


titanous:
The primary testing approach I can thing of is a "hanging dialer" that allows us to hang, then fail the first dial, and while that's happening we'd close the client connection and assert that dial was not called again.

benburkert:
test added, this is ready for another :eyes:

router/proxy/reverseproxy_test.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+	}
+
+	fn := func() []string { return []string{"127.0.0.1:0", "127.0.0.1:0"} }
+	prox := NewReverseProxy(fn, nil, false)
+
+	prox.ServeConn(context.Background(), cnConn)
+}
+
+type testDialer struct {
```
titanous: Could do a func type instead.

titanous:
Looks good. Please squash down to one commit.

benburkert:
done. :+1:

titanous:
LGTM

PR #967, opened 2015-02-08
Update installation script and docs

lmars:
@domserio please review the docs change.

### DIFF
--- a/website/source/docs/installation.html.md
+++ b/website/source/docs/installation.html.md
@@ -63,61 +63,31 @@ cluster, you should boot at least 3 nodes to keep etcd efficient
 (see [here](https://github.com/coreos/etcd/blob/v0.4.6/Documentation/optimal-cluster-size.md) for
 an explanation).
 
-### Dependencies
-
-Flynn uses Docker images to store job filesystems, and by default uses the AUFS filesystem driver. To
-check if your system supports AUFS, run this command:
-
-```
-$ sudo modprobe aufs
-```
-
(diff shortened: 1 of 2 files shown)


website/source/docs/installation.html.md
```diff
@@ -63,61 +63,31 @@ cluster, you should boot at least 3 nodes to keep etcd efficient
...
 ### Installation
 
-Flynn is available as a Debian package from the Flynn apt repository.
-
-First, add the Flynn repository key to your list of trusted apt keys:
+Download and run the Flynn installer script:
 
 ```
-$ sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv BC79739C507A9B53BB1B0E7D820A5489998D827B
+$ sudo bash < <(curl -fsSL https://dl.flynn.io/install-flynn)
```
titanous: Is the file redirection necessary? This works for me, as process substitution provides a file path:

```text
bash <(echo ls)
```

Also, why process substitution over "curl pipe bash"?

lmars: @titanous ah yes I'm just used to always passing it to stdin rather than as an argument, updated.

WRT `curl | sudo bash`, when I am required to use `sudo` I like it being clear at the start of the command. Just a personal preference, happy to change if you think the alternative is clearer? Both side by side for reference:

```
sudo bash <(curl -fsSL https://dl.flynn.io/install-flynn)

curl -fsSL https://dl.flynn.io/install-flynn | sudo bash
```

titanous:
LGTM

PR #968, opened 2015-02-08
Modify ZFS installation instructions to keep them consistent

jaysh:
Hello,

I've made 2 changes:
- Use `sudo` in the commands and make their presentation consistent with other commands (which assume the running user is unprivileged)
- Remove the 'assume yes'  from `apt-get install` flag for consistency

I'm happy to spit/remove one of the changes if necessary.

Thanks,
Jay


titanous:
Thanks! Both commit messages need a `website: ` prefix. These instructions will be going away shortly, as we're moving to an installation script.

jaysh:
Thanks for the quick reply! I've added the prefix now. The script sounds great - is there a branch for that already? I'd love to take a look.

titanous:
Check out https://github.com/flynn/flynn/blob/3380a443d6abe00eccb51cd190b2a67792c404bf/script/install-flynn.tmpl and #967.

titanous:
LGTM

PR #969, opened 2015-02-08
dashboard: Don't create release artifact when deploying

jvatic:
Releases no longer require an artifact as of 6be3d4376bb4d2405b622f9b7f2444cd4c0743de and the artifact being created here is not used.


titanous:
LGTM

PR #971, opened 2015-02-09
test: Fix testCluster.Size()

lmars:
`TestOmniJobs` uses `testCluster.Size()` to determine how many jobs to wait for, so it should not include hosts which have been removed.


titanous:
LGTM

archSeer:
LGTM

PR #973, opened 2015-02-09
Flynn host cli volume-destroy command

heavenlyhash:
Adds a command to `flynn-host` which destroys volumes and their configuration in an attempt to return the host to a clean state for a future run.

### DIFF
--- a/host/cli/destroy-volumes.go
+++ b/host/cli/destroy-volumes.go
@@ -1,0 +1,124 @@
+package cli
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
+	"github.com/flynn/flynn/host/volume"
+	"github.com/flynn/flynn/host/volume/manager"
+	"github.com/flynn/flynn/pkg/shutdown"
+)
+
+func init() {
+	Register("destroy-volumes", runVolumeDestroy, `
+usage: flynn-host destroy-volumes [options]
+
+options:
+  --volpath=PATH     directory to create volumes in [default: /var/lib/flynn/volumes]
+  --include-data     actually destroy data in backends *this is dangerous* [default: false]
+
+Destroys all data volumes on this host.  This is a dangerous operation: data
+may be permanently discarded.
+
+If the '--include-data' flag is given, the volume database will be loaded, and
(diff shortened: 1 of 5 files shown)


host/cli/destroy-volumes.go
```diff
@@ -0,0 +1,107 @@
...
+options:
+  --volpath=PATH         directory to create volumes in [default: /var/lib/flynn/volumes]
+
+Destroys all data volumes on this host.  This is a dangerous operation: data may be permanently discarded.
+
+If the volume database can be loaded, data in the referenced systems will be removed (i.e., with a zfs backend, the datasets will be destroyed).
+
+If the volume database cannot be loaded, it will simply be removed.  In this event, data may remain behind in the backend storage engines and require manual cleanup, but the flynn-host daemon's next launch will be like a fresh launch.
+
+Major features of backends will not be adjusted (i.e., zfs datasets will be destroyed, but zpools will not be touched).`)
```
titanous: These docs should be wrapped to 80 columns.

host/cli/destroy-volumes.go
```diff
@@ -0,0 +1,107 @@
...
+		},
+	)
+	if err != nil {
+		if strings.HasSuffix(err.Error(), "timeout") { //bolt.ErrTimeout
+			fmt.Println("volume state db is locked by another process; aborting.")
+			shutdown.ExitWithCode(4)
+		}
+		return fmt.Errorf("warning: the previous volume database could not be loaded; any data in backends may need manual removal\n  (error was: %s)", err)
+	}
+	fmt.Println("volume state db restored.")
```
titanous: s/restored/opened/

titanous:
This looks good. What do you think about defaulting to just removing the state database, and adding a flag that does the zfs nuke as well? That way if you accidentally run it, there is no real data loss.

heavenlyhash:
Sounds good, adding flag.

heavenlyhash:
Added as '--include-data'.

I made the command do a restore and acquire a lock on the boltdb even if not including backend's data in the destroy; it doesn't technically need any of the data, but I'm thinking it's worth avoiding the surprise of dropping data from a running host.  Sound right?

titanous:
Yeah, grabbing a lock makes sense.

titanous:
LGTM pending squash of https://github.com/flynn/flynn/commit/32f3185e4dd599c0739274bb5cfdc6f9d561221a and CI pass.

PR #978, opened 2015-02-11
Volume snapshots

heavenlyhash:
Added permanent snapshots.  Snapshot mounts are readonly, but otherwise act mostly like other volumes.  Snapshots can be forked again into live filesystems.

Snapshots can now be serialized and transmitted to other matching providers (once connected to the http API, this will make snapshot transfer possible between hosts).  The provider interface supports negotiation for existing data objects so that transmission can use incremental deltas when possible.

### DIFF
--- a/host/volume/backend.go
+++ b/host/volume/backend.go
@@ -17,6 +17,8 @@ type Provider interface {
 
 	NewVolume() (Volume, error)
 	DestroyVolume(Volume) error
+	CreateSnapshot(Volume) (Volume, error)
+	ForkVolume(Volume) (Volume, error)
 
 	MarshalGlobalState() (json.RawMessage, error)
 	MarshalVolumeState(volumeID string) (json.RawMessage, error)
--- a/host/volume/zfs/zfs.go
+++ b/host/volume/zfs/zfs.go
@@ -1,14 +1,15 @@
 package zfs
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
+	"strings"
 	"syscall"
-	"time"
 
 	zfs "github.com/flynn/flynn/Godeps/_workspace/src/github.com/mistifyio/go-zfs"
 	"github.com/flynn/flynn/host/volume"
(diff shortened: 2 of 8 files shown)


heavenlyhash:
This should now be ready for a review.  Incremental transfers are now supported.

host/volume/backend.go
```diff
@@ -17,6 +18,12 @@ type Provider interface {
 
 	NewVolume() (Volume, error)
 	DestroyVolume(Volume) error
+	CreateSnapshot(Volume) (Volume, error)
+	ForkVolume(Volume) (Volume, error)
+
+	ReportHaves(Volume) ([]json.RawMessage, error) // Report known data addresses; this can be given to `SendSnapshot` to attempt deduplicated/incrememntal transport.
```
titanous: Does `ListSnapshots` make sense here? Also, why `[]json.RawMessage` as a return type? I'd expect `[]string` or something similar.

host/volume/zfs/zfs.go
```diff
@@ -161,6 +177,196 @@ func (b *Provider) DestroyVolume(vol volume.Volume) error {
 	return nil
 }
 
+func (b *Provider) CreateSnapshot(vol volume.Volume) (volume.Volume, error) {
+	zvol, err := b.owns(vol)
+	if err != nil {
+		return nil, err
+	}
+	id := random.UUID()
+	v2 := &zfsVolume{
```
titanous: Maybe `s/v2/snap/`?

host/volume/backend.go
heavenlyhash: I borrowed the term `haves` from [git's transport packing spec](https://github.com/git/git/blob/18d0fec24027ac226dc2c4df2b955eef2a16462a/Documentation/technical/pack-protocol.txt#L283).  I'd be open to different names, but specifically opposed to calling it `ListSnapshots` because that's an implementation detail of ZFS (and a bad one at that; insert entire rant here about how ZFS fails miserably to make a distinction between the object layer and the file tree layer, which goes a long way to explaining why its dedup is such here-be-dragons territory to begin with).

I can't currently think of any providers or protocols where much beyond a string objectID is needed here, so I guess I tossed a json object in mostly to leave room for future upgrades.  I could see arguments for either cutting to `[]string` or doing one big `json.RawMessage`.  Thoughts?  At the end of the day, the only constraint is that it's going in an API post body shortly.

titanous: Ah, I see. `ListHaves` returning a `json.RawMessage` is fine with me.

heavenlyhash: Renamed function.

Maybe a type aliasing `json.RawMessage` with a more local name would also improve clarity, thoughts?  I don't think it would actually give us much more for static type checking, so I'm on the fence.

PR #979, opened 2015-02-11
volume persistence deletion tests

heavenlyhash:
Noticed some issues with consistency around deletion; fixed and added missing tests.

n.b. This should have gone in with #973 -- I'd be happy to squash it into that branch after a quick review if desired.

### DIFF
--- a/host/volume/manager/manager.go
+++ b/host/volume/manager/manager.go
@@ -142,6 +142,23 @@ func (m *Manager) DestroyVolume(id string) error {
 		return err
 	}
 	delete(m.volumes, id)
+	// clean up named volumes
+	namesToRemove := make([]string, 0, 1)
+	for name, volID := range m.namedVolumes {
+		if volID == id {
+			namesToRemove = append(namesToRemove, name)
+		}
+		delete(m.namedVolumes, name)
+	}
+	// commit both changes
+	m.persist(func(tx *bolt.Tx) error {
+		for _, id := range namesToRemove {
+			if err := m.persistNamedVolume(tx, id); err != nil {
+				return err
+			}
+		}
+		return m.persistVolume(tx, vol)
+	})
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


host/volume/manager/manager.go
```diff
@@ -279,12 +296,12 @@ func (m *Manager) getProviderBucket(tx *bolt.Tx, providerID string) (*bolt.Bucke
...
-func (m *Manager) persistVolume(tx *bolt.Tx, id string) error {
+// Called to sync changes to disk when a volume is updated
+func (m *Manager) persistVolume(tx *bolt.Tx, vol volume.Volume) error {
 	// Save the general volume info
 	volumesBucket := tx.Bucket([]byte("volumes"))
-	k := []byte(id)
-	vol, ok := m.volumes[id]
+	k := []byte(vol.Info().ID)
+	_, ok := m.volumes[vol.Info().ID]
 	if !ok {
```
titanous: Could merge the ok check into this if statement

host/volume/manager/manager.go
```diff
@@ -279,12 +296,12 @@ func (m *Manager) getProviderBucket(tx *bolt.Tx, providerID string) (*bolt.Bucke
...
 
-// Called to sync changes to disk when a VolumeInfo is updated
-func (m *Manager) persistVolume(tx *bolt.Tx, id string) error {
+// Called to sync changes to disk when a volume is updated
+func (m *Manager) persistVolume(tx *bolt.Tx, vol volume.Volume) error {
 	// Save the general volume info
 	volumesBucket := tx.Bucket([]byte("volumes"))
-	k := []byte(id)
-	vol, ok := m.volumes[id]
+	k := []byte(vol.Info().ID)
```
titanous: How expensive is calling `Info`?

heavenlyhash: somewhere in the area of a vtable jump, but if you're suggesting this reads like a stutter I'd have to concede; will replace.

titanous: Yeah, not a big deal, but I'd use a temp as it's called three times.

host/volume/manager/manager.go
heavenlyhash: It's used again further down in the branch for whether or not the provider storage is doing a deletion.  I'll rename so we don't have an 'ok' floating around that long.

heavenlyhash:
Comments addressed

titanous:
LGTM

PR #980, opened 2015-02-11
Installation script fixes

### LINKED ISSUE #976: installation script nits
@lmars 

- Running the script again succeeds, but prints this error message:
```text
ln: failed to create symbolic link ‘/usr/bin/flynn’: File exists
```
- We should probably be installing into `/usr/local/bin` not `/usr/bin`
- The URL used in manifests is `https://dl.flynn.io/images` but that doesn't reflect reality, the TUF repo is at `https://dl.flynn.io/tuf`. Any reason for this?
- Installing the ZFS kernel module can take a while, we should indicate that the installer hasn't hung somehow (could show the apt-get output)

lmars:
Closes #976


titanous:
LGTM.

PR #983, opened 2015-02-11
Deployer updates

lmars:
The main change here is to make the deployer log more verbosely, as currently it only really logs on error. I also made the log messages more consistent by always using lower case, and starting error messages with the word "error" to make the logs easier to scan down.

### DIFF
--- a/controller/deployer/strategies/one-by-one.go
+++ b/controller/deployer/strategies/one-by-one.go
@@ -7,36 +7,40 @@ import (
 )
 
 func oneByOne(l log15.Logger, client *controller.Client, d *ct.Deployment, events chan<- ct.DeploymentEvent) error {
-	log := l.New("fn", "oneByOne")
-	log.Info("Starting")
+	log := l.New("fn", "oneByOne", "deployment_id", d.ID, "app_id", d.AppID)
+	log.Info("starting one-by-one deployment")
 
+	log.Info("getting job event stream")
 	jobStream := make(chan *ct.JobEvent)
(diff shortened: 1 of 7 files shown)


controller/deployer/strategies/one-by-one.go
```diff
@@ -7,36 +7,40 @@ import (
...
 	if err != nil {
-		log.Error("Failed fetching the old formation", "at", "get_formation", "err", err)
+		olog.Error("error getting old formation", "err", err)
 		return err
 	}
 
-	oldFormation := f.Processes
-	newFormation := map[string]int{}
+	oldProcesses := f.Processes
+	newProcesses := map[string]int{}
```
titanous: This should use `make`

test/test_deployer.go
```diff
@@ -41,14 +41,14 @@ func (s *DeployerSuite) createDeployment(t *c.C, process, strategy string) *ct.D
 }
 
 func waitForDeploymentEvents(t *c.C, stream chan *ct.DeploymentEvent, expected []*ct.DeploymentEvent) {
-	// wait for an event with no release to mark the end of the deployment,
-	// collecting events along the way
-	events := []*ct.DeploymentEvent{}
+	debugf(t, "waiting for %d deployment events", len(expected))
+	actual := []*ct.DeploymentEvent{}
```
titanous: This should use make and have a capacity of `len(expected)`

lmars:
@titanous comments addressed

titanous:
LGTM

PR #985, opened 2015-02-12
host: Use discoverd resolv.conf if available

titanous:
This allows services like the router to use discoverd DNS. Previously all jobs in the host network namespace used the host resolv.conf, but only jobs started before discoverd is up should use it.

This should fix issues connecting to postgres in #959.


lmars:
LGTM

PR #986, opened 2015-02-12
postgres appliance state machine

titanous:
First, I'd like to apologize for the size of this. You're not reading the diff wrong, there are indeed over 4k lines of new code here. Most of that is the tests and the simulator supporting the tests, the state machine itself weighs in at 1k lines. I couldn't come up with an incremental way to deliver this, as it is a direct translation that doesn't make much sense in a partial state.

---

This adds a Go translation of the [manatee-state-machine project](https://github.com/joyent/manatee-state-machine/tree/d441fe941faddb51d6e6237d792dd4d7fae64cc6) from Joyent. It is translated verbatim with the following minor differences:

- A few extra tests were added for completeness
- Changes were made to translate from the Node.js async model to a more
  straightforward single-threaded model that uses channels for coordination.
- The term 'one-node-write-mode' (abbreviated as ONWM) was replaced
  with 'singleton' as they are used interchangeably upstream.
- Assertions were translated into panics wrapped in if statements.
- Discoverd is used instead of Zookeeper.

The state machine provides a fault-tolerant system for running Postgres clusters that automatically fail over with minimal downtime and no possibility for data loss or split brain during failovers. The README linked above has a good overview and some diagrams. I expect we'll change this a bit to fit containers better, but this is a really good starting point.

A simulator is used for testing and is also available in an interactive REPL for experimenting with various scenarios.

Test coverage is at 87.9% with only assertions, a few error paths, and very rare cases not covered.

### DIFF
--- a/appliance/postgresql2/simulator/simulator.go
+++ b/appliance/postgresql2/simulator/simulator.go
@@ -1,0 +1,804 @@
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+//
+// This file is derived from
+// https://github.com/joyent/manatee-state-machine/blob/d441fe941faddb51d6e6237d792dd4d7fae64cc6/lib/sim.js
+// https://github.com/joyent/manatee-state-machine/blob/d441fe941faddb51d6e6237d792dd4d7fae64cc6/lib/sim-zk.js
+// https://github.com/joyent/manatee-state-machine/blob/d441fe941faddb51d6e6237d792dd4d7fae64cc6/lib/sim-pg.js
+//
+// Copyright (c) 2014, Joyent, Inc.
+// Copyright (c) 2015, Prime Directive, Inc.
+//
+
+package simulator
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"sync"
+	"text/tabwriter"
+	"time"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
+	"github.com/flynn/flynn/appliance/postgresql2/state"
+	"github.com/flynn/flynn/appliance/postgresql2/xlog"
+	"github.com/flynn/flynn/discoverd/client"
+)
+
+type commandFunc func([]string)
+type command struct {
+	Name     string
+	Help     string
+	Args     string
+	Func     commandFunc
+	WaitRest bool
+}
+
+const opLag = 1 * time.Millisecond
+
+type Simulator struct {
+	singleton bool
+	nextPeer  int
+
+	allIdents map[string]*discoverd.Instance
+
+	out       io.Writer
+	log       log15.Logger
+	postgres  *postgresSimulator
+	discoverd *discoverdSimulator
+	peer      *simPeer
+	restCh    chan struct{}
+	retryCh   chan struct{}
+	started   bool
+
+	commands       []*command
+	commandsByName map[string]*command
+}
+
+func New(singleton bool, out io.Writer, logOut io.Writer) *Simulator {
+	if logOut == nil {
+		logOut = out
+	}
+	s := &Simulator{
+		singleton: singleton,
+		allIdents: make(map[string]*discoverd.Instance),
+		out:       out,
+		log:       log15.New(),
+		restCh:    make(chan struct{}),
+		retryCh:   make(chan struct{}, 1),
+	}
+	s.discoverd = newDiscoverdSimulator(s.log.New("component", "discoverd"))
+	s.postgres = newPostgresSimulator(s.discoverd, s.log.New("component", "pg"))
+	s.peer = s.createSimPeer()
+	s.initCommands()
+	s.log.SetHandler(log15.StreamHandler(logOut, log15.LogfmtFormat()))
+	return s
+}
+
+func (s *Simulator) initCommands() {
+	s.commands = []*command{
+		{"addpeer", "simulate a new peer joining the discoverd cluster", "[NAME]", s.AddPeer, true},
+		{"bootstrap", "simulate initial setup", "PRIMARY [SYNC]", s.Bootstrap, true},
+		{"catchUp", "simulate peer's postgres catching up to primary", "", s.CatchUp, false},
+		{"depose", "simulate a takeover from the current config", "", s.Depose, true},
+		{"rebuild", "simulate rebuilding a deposed peer", "NAME", s.Rebuild, true},
+		{"echo", "emit the string to stdout", "STR", s.Echo, false},
+		{"freeze", "freeze cluster", "", s.Freeze, true},
+		{"help", "show help output", "", s.Help, false},
+		{"ident", "print the identity of the peer being tested", "", s.Ident, false},
+		{"lspeers", "list simulated peers", "", s.LsPeers, false},
+		{"peer", "dump peer's current state", "", s.Peer, false},
+		{"rmpeer", "simulate a peer being removed from the discoverd cluster", "ID", s.RmPeer, true},
+		{"setClusterState", "simulate a write to the cluster state stored in discoverd", "STATE", s.SetClusterState, true},
+		{"startPeer", "start the peer state machine", "", s.StartPeer, true},
+		{"unfreeze", "unfreeze the cluster", "", s.Unfreeze, true},
+		{"discoverd", "print simulated discoverd state", "", s.Discoverd, false},
+		{"exit", "exit the simulator", "", s.Exit, false},
+	}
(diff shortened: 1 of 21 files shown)


titanous:
ping @lmars for review

appliance/postgresql2/simulator/simulator.go
```diff
@@ -0,0 +1,804 @@
...
+type discoverdSimulatorClient struct {
+	sync.RWMutex
+	d      *discoverdSimulator
+	inst   *discoverd.Instance
+	events chan *state.DiscoverdEvent
+	init   bool
+}
+
+func (d *discoverdSimulatorClient) SetState(s *state.DiscoverdState) error {
+	time.Sleep(opLag)
```
lmars: Why the `sleep`?

lmars: Oh, I'm guessing to simulate doing some I/O?

appliance/postgresql2/state/state.go
```diff
@@ -0,0 +1,1019 @@
...
+
+type Postgres interface {
+	XLogPosition() (xlog.Position, error)
+	Reconfigure(*PgConfig) error
+	Start() error
+	Stop() error
+
+	// Events returns a channel that returns a single event when the interface
+	// is ready.
+	Events() <-chan PostgresEvent
```
lmars: Would this be better named `Ready()`?

appliance/postgresql2/state/state.go
```diff
@@ -0,0 +1,1019 @@
...
+			p.handlePgInit(e)
+		case <-p.evalStateCh:
+			p.evalClusterState()
+		case <-p.applyConfCh:
+			p.pgApplyConfig()
+		case <-p.workDoneCh:
+			// There is no work to do, we are now at rest
+			p.rest()
+		case <-p.stopCh:
+			return
```
lmars: Should the other selects check `stopCh` too?

appliance/postgresql2/state/state.go
```diff
@@ -0,0 +1,1019 @@
...
+func (p *Peer) handleDiscoverdState(e *DiscoverdEvent) {
+	log := p.log.New("fn", "handleDiscoverdState", "index", e.State.Index)
+	log.Info("got discoverd state", "state", e.State.State != nil)
+	if p.clusterPeers == nil {
+		panic("received discoverd state before init")
+	}
+	if p.decodeState(e) {
+		p.evalClusterState()
+	} else {
+		log.Info("already have this state", "at", "skip_eval")
```
lmars: I'm not sure if the `at=xxx` context provides a benefit, given the log message usually says the same thing, and there is a `fn=xxx` context, wdyt?

appliance/postgresql2/state/state.go
```diff
@@ -0,0 +1,1019 @@
...
+
+func (p *Peer) moving() {
+	if !p.atRest() {
+		return
+	}
+	p.log.Debug("moving")
+
+	// Set up a channel that the select in the run loop can use to detect
+	// when there is no more work to do.
+	p.workDoneCh = make(chan struct{}, 1)
```
lmars: Does this need to be wrapped in a mutex?

appliance/postgresql2/state/state.go
```diff
@@ -0,0 +1,1019 @@
...
+
+func (p *Peer) atRest() bool {
+	return p.workDoneCh == nil
+}
+
+// Examine the current cluster state and determine if new actions need to be
+// taken. For example, if we're the primary, and there's no sync present, then
+// we need to declare a new generation.
+func (p *Peer) evalClusterState() {
+	p.moving()
```
lmars: I'm not sure I follow the moving -> rest workflow here, I was expecting this to be deferred, what is the purpose?

appliance/postgresql2/state/state.go
lmars: Comment added to discuss: https://github.com/flynn/flynn/issues/34#issuecomment-75448708

appliance/postgresql2/state/state.go
```diff
@@ -0,0 +1,1019 @@
...
+		} else if p.role != RoleUnassigned {
+			p.assumeUnassigned()
+		}
+
+		return
+	}
+
+	// Bail out if we're configured for singleton mode but the cluster is not.
+	if p.singleton && !p.clusterState.Singleton {
+		shutdown.Fatal("configured for singleton mode but cluster found in normal mode")
```
lmars: Any reason for `shutdown.Fatal` rather than `panic` like the rest of the method?

appliance/postgresql2/state/state.go
```diff
@@ -0,0 +1,1019 @@
...
+		p.updatingState.Freeze = NewFreezeDetails("cluster started in singleton mode")
+	} else {
+		p.updatingState.Sync = p.clusterPeers[1]
+		p.updatingState.Async = p.clusterPeers[2:]
+	}
+	log := p.log.New("fn", "startInitialSetup")
+	log.Info("creating initial cluster state", "generation", 1, "at", "create_state")
+
+	if err := p.putClusterState(); err != nil {
+		log.Error("failed to create cluster state", "at", "create_state", "err", err)
```
lmars: Should this return?

appliance/postgresql2/state/state.go
```diff
@@ -0,0 +1,1019 @@
...
+		Generation: p.clusterState.Generation,
+		Primary:    p.clusterState.Primary,
+		Sync:       p.clusterState.Sync,
+		Async:      newAsync,
+		Deposed:    p.clusterState.Deposed,
+		InitWAL:    p.clusterState.InitWAL,
+	}
+	log.Info("updating list of asyncs", "at", "update_state")
+	err := p.putClusterState()
+	if err != nil {
```
lmars: `if err := p.putClusterState(); err != nil {`

appliance/postgresql2/state/state_test.go
```diff
@@ -0,0 +1,2137 @@
...
+}
+
+var fakeTime = time.Now().UTC()
+
+func init() {
+	state.TimeNow = func() time.Time { return fakeTime }
+}
+
+func runSteps(t *testing.T, singleton bool, steps []step) {
+	logOut := &bytes.Buffer{}
```
lmars: `var logOut bytes.Buffer`

lmars: Oh, I realise it's a pointer.

appliance/postgresql2/simulator/simulator.go
titanous: yeah, with the thought that it could help the race detector find interesting things.

appliance/postgresql2/state/state.go
titanous: Yep, will change.

appliance/postgresql2/state/state.go
titanous: No, the eval loop is single-threaded.

appliance/postgresql2/state/state.go
titanous: This is used by the simulator/test suite to figure out when all of the triggered work has been completed. At the start of an evaluation, moving will get called, and one or more evaluation iterations may happen. When there is no more work to do, the select will end up receiving from `workDoneCh` and trigger the resting state.

appliance/postgresql2/state/state.go
titanous: Nope, will change.

appliance/postgresql2/state/state.go
titanous: Yeah, looks like this error path needs some cleanup, I'll figure out what it should do (probably schedule another evaluation after a second).

appliance/postgresql2/state/state.go
titanous: Yep, will drop `at`.

appliance/postgresql2/state/state.go
titanous: Can't hurt.

titanous:
Comments addressed.

appliance/postgresql2/state/state.go
lmars: Ok I think I was confused thinking there were multiple goroutines at play, thanks for clarifying.

appliance/postgresql2/state/state.go
```diff
@@ -0,0 +1,1027 @@
...
+		Generation: p.clusterState.Generation,
+		Primary:    p.clusterState.Primary,
+		Sync:       p.clusterState.Sync,
+		Async:      newAsync,
+		Deposed:    p.clusterState.Deposed,
+		InitWAL:    p.clusterState.InitWAL,
+	}
+	log.Info("updating list of asyncs")
+	if err := p.putClusterState(); err != nil {
+		log.Error("failed to update cluster state", "err", err)
```
lmars: Should this also be followed by `evalLater`?

lmars:
LGTM pending last comment

appliance/postgresql2/state/state.go
titanous: No, there is a `triggerEval` call below on line 895.

PR #987, opened 2015-02-12
Add PGHOST to postgres app config

titanous:
This allows apps that are not aware of service discovery to connect to postgres.


archSeer:
LGTM

PR #990, opened 2015-02-12
host: Allocate IP addresses on job restore.

### LINKED ISSUE #60: Allocate IPs in libvirt-lxc Restore
Currently the libvirt-lxc backend doesn't properly allocate IPs from the tracker when [restoring job state](https://github.com/flynn/flynn/blob/72a76d33e546059c197e4cb17411895153d4ccbd/host/libvirt_lxc_backend.go#L702).

archSeer:
Allocate IP addresses on job restore. Setup networking on restarts.

IPs from running containers should be marked as used. Log any errors in doing so. Additionally, ConfigureNetworking needs to be ran regardless of whether flannel was already running or not.

Note that for this functionality to work correctly, --force should not be set on flynn-host, as that will run Cleanup() before starting it, destroying all of the old containers.

Closes #60.

### DIFF
--- a/host/manifest.go
+++ b/host/manifest.go
@@ -132,9 +132,10 @@ func (m *manifestRunner) runManifest(r io.Reader) (map[string]*ManifestData, err
 	m.state.mtx.Unlock()
 
 	var netInfo NetworkInfo
-	for _, service := range services {
+
+	runService := func(service *manifestService) error {
 		if _, exists := serviceData[service.ID]; exists {
-			continue
+			return nil
 		}
 
 		data := &ManifestData{
@@ -149,7 +150,7 @@ func (m *manifestRunner) runManifest(r io.Reader) (map[string]*ManifestData, err
 		for _, port := range service.TCPPorts {
 			port, err := strconv.Atoi(port)
 			if err != nil {
-				return nil, err
+				return err
 			}
 			data.TCPPorts = append(data.TCPPorts, port)
 		}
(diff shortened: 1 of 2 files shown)


archSeer:
@titanous I've confirmed this is functioning correctly by doing `kill -9` on `flynn-host`, then running 
```
sudo host/bin/flynn-host daemon --manifest host/bin/manifest.json --external 172.31.20.134 --backend libvirt-lxc --state /tmp/flynn-host-state.bolt --flynn-init host/bin/flynn-init
```

(I inserted a bunch of debug statements to see the IP allocation happening).

host/manifest.go
```diff
@@ -258,8 +259,24 @@ func (m *manifestRunner) runManifest(r io.Reader) (map[string]*ManifestData, err
...
+			return nil, err
+		}
 		if service.ID == "flannel" {
+			var job *host.Job
+			for _, j := range m.state.jobs {
+				if j.ManifestID != service.ID {
+					continue
+				}
+				job = j.Job
+			}
```
titanous: Add a `break`

host/manifest.go
```diff
@@ -258,8 +259,24 @@ func (m *manifestRunner) runManifest(r io.Reader) (map[string]*ManifestData, err
...
+			if job.ID == "" {
```
titanous: This will cause a nil panic, `if job == nil` is correct.

archSeer:
@titanous Addressed.

titanous:
LGTM pending CI

PR #991, opened 2015-02-12
host: Remove unused port allocator

titanous:
It became obsolete with the move to per-container IPs, and the last use is an unused code path in the layer 0 bootstap runner.


archSeer:
LGTM, pending CI

PR #994, opened 2015-02-12
deployer: Support resuming deployments

### LINKED ISSUE #799: deployer: Handle partially running formations due to crash
When the deployer starts a deployment, it should check the current status of a formations jobs and then determine what changes it needs to make in case a previous deployer crashed mid deployment.

lmars:
The main changes here are determining what job events to wait for based on currently running jobs, and ensuring the deployer reconnects to the controller when waiting for job events.

Apologies for the big diff, but I did not commit along the way and splitting things out would take time.

Closes #799.

### DIFF
--- a/controller/deployer/strategies/main.go
+++ b/controller/deployer/strategies/main.go
@@ -7,20 +7,132 @@ import (
 	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/pkg/attempt"
+	"github.com/flynn/flynn/pkg/cluster"
+	"github.com/flynn/flynn/pkg/stream"
 )
 
-type PerformFunc func(log15.Logger, *controller.Client, *ct.Deployment, chan<- ct.DeploymentEvent) error
+type UnknownStrategyError struct {
+	Name string
+}
+
+func (e UnknownStrategyError) Error() string {
+	return fmt.Sprintf(`deployer: unknown strategy "%s"`, e.Name)
+}
+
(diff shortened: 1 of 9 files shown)


controller/deployer/strategies/main.go
```diff
@@ -7,20 +7,132 @@ import (
...
+	if err := deploy.connectEventStream(); err != nil {
+		log.Error("error connecting job event stream", "err", err)
+		return err
+	}
+	defer deploy.closeEventStream()
+
+	log.Info("determining deployment state")
+	c, err := cluster.NewClient()
+	if err != nil {
+		log.Error("error determining deployment state", "err", err)
```
titanous: This should be "error connecting to cluster" or similar

controller/deployer/strategies/main.go
```diff
@@ -7,20 +7,132 @@ import (
...
+		return err
 	}
-	return nil, fmt.Errorf("Unknown strategy '%s'!", strategy)
+	hosts, err := c.ListHosts()
+	if err != nil {
+		log.Error("error determining deployment state", "err", err)
+		return err
+	}
+	for _, host := range hosts {
+		for _, job := range host.Jobs {
```
titanous: If possible, we should use the controller APIs to get this information. Are we working around a flaw/shortcoming of those APIs?

lmars: I avoided using the controller APIs because the test I wrote for this was all about deploying the controller, so if the deployer starts during a scheduler restart, there is potential for the deployer to get an incorrect view of actual jobs running (because the scheduler is still itself syncing with the cluster, so the fact that it itself is up may not yet be visible in the controller).

One way to get around that is to ensure that the scheduler leader is in sync with the cluster, then we can be more confident the controller has the correct view of what jobs are running.

WDYT?

lmars:
Closing this in favour of smaller incremental patches.

PR #997, opened 2015-02-12
discoverd: Handle bare IPv6 upstream resolver addresses

titanous:
Before this patch, passing a bare IPv6 address with no port resulted in the error "too many colons in address". If we get this error, we should treat it as an IPv6 address with no port.


lmars:
LGTM

Issue #1002, opened 2015-02-13
router: recover missed route deletions after sync failure


benburkert:
`DataStore` uses a `LISTEN` connection to watch for route changes. This connection will occasionally fail, and to recover the `DataStore` must restart the `LISTEN` connection then do a full-sync of all routes to pickup routes added while the `LISTEN` connection dropped.

This recovery strategy currently has a flaw which causes routes to never be removed if the remove is issued during a window when the `LISTEN` connection has failed and the `DataStore` has not detected the failure.

In order to solve this, the Sync API, which is currently asynchronous (`StartSync`/`StopSync`), needs to be synchronous (`Sync`), so that the caller (`HTTPListener`/`TCPListener`) can be notified of errors. This allows the caller to rebuild it's routing state from scratch when `Sync` fails. Furthermore, the caller should be capable of rebuilding the routing state without having to lock routing.

benburkert:
Shelving my work this for now, for reference the commits are here: https://github.com/benburkert/flynn/tree/router-route-table

bgentry:
FYI this issue applies to any updates that happen while the sync is disconnected, not just deletions. I'll leave my version of the test here for reference and hopefully we can come back to fix this for real https://github.com/bgentry/flynn/commit/634de1b522b6d0f8efc16966c21d8a3869ddd115

PR #1003, opened 2015-02-13
router: datastore sync

benburkert:
Switch to a single, synchronous `Sync` method for `DataStore`. This is part of the work to address #1002, but doesn't include the fix.

### DIFF
--- a/router/data_store.go
+++ b/router/data_store.go
@@ -3,11 +3,10 @@ package main
 import (
 	"errors"
 	"fmt"
-	"log"
-	"sync"
 	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/jackc/pgx"
+	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/net/context"
 	"github.com/flynn/flynn/router/types"
 )
 
@@ -19,8 +18,7 @@ type DataStore interface {
 	Get(id string) (*router.Route, error)
 	List() ([]*router.Route, error)
 	Remove(id string) error
-	Sync(h SyncHandler, started chan<- error)
-	StopSync()
+	Sync(ctx context.Context, h SyncHandler, startc chan<- struct{}) error
 }
 
 type DataStoreReader interface {
@@ -38,9 +36,6 @@ type pgDataStore struct {
 
 	routeType string
 	tableName string
-
-	doneo sync.Once
-	donec chan struct{}
 }
 
 const (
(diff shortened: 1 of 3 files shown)


router/data_store.go
```diff
@@ -199,72 +193,69 @@ func (d *pgDataStore) List() ([]*router.Route, error) {
...
+	if err != nil {
+		return err
 	}
-	defer d.StopSync()
 
 	initialRoutes, err := d.List()
 	if err != nil {
-		started <- err
-		return
+		cancel()
```
bgentry: doesn't look like there's a need for this (or the `context.WithCancel()` above). The only thing this will do is cancel child contexts like the new `ctx` you've created above. The block below that switches on `ctx.Done()` will only execute when a parent has canceled.

Maybe just drop the `context.WithCancel()` and all uses of `cancel()` inside this method?

router/http.go
```diff
@@ -84,23 +89,57 @@ func (s *HTTPListener) Start() error {
 		s.cookieKey = &[32]byte{}
 	}
 
-	started := make(chan error)
-
 	// TODO(benburkert): the sync API cannot handle routes deleted while the
```
bgentry: outdated TODO?

bgentry:
Overall this looks great. My main concern is the fact that the non-trivial sync loop is ~completely duplicated between `HTTPListener` and `TCPListener`. The only difference I can see between them is the use of `httpSyncHandler` vs `tcpSyncHandler` as an argument to `Sync()`.

Maybe we can share all the rest of this logic and just have a way to provide a different sync handler to whatever is running the sync loops?

router/data_store.go
benburkert: This call to `cancel` stops the `LISTEN` goroutine started in `startListener`.

router/http.go
benburkert: it's still a problem until routing state (`domains` & `routes` maps) are entirely rebuilt on each call to `Sync`.

router/data_store.go
bgentry: so it does, my mistake :v:

benburkert:
> My main concern is the fact that the non-trivial sync loop is ~completely duplicated between HTTPListener and TCPListener. The only difference I can see between them is the use of httpSyncHandler vs tcpSyncHandler as an argument to Sync().

Yes, it's definitely a problem, but I think it's better to address it in a follow up PR for #1002 b/c the `runSync` methods will need some big changes, and it will be more straight forward to DRY it up after that work.

bgentry:
> This is part of the work to address #1002, but does include the fix.

*doesn't* include the fix, right?

benburkert:
> doesn't include the fix, right?

oops, right. it **doesn't** include the fix.

@titanous this is read for :eyeglasses:

titanous:
LGTM

titanous:
Please rebase and I'll merge.

PR #1004, opened 2015-02-14
util/packer: Disable cron before starting work

titanous:
The default daily crontab has a bunch of apt/dpkg commands and runs at ~06:00 which can coincide with releases and one-off builds. This will cause lock errors, as only one command at a time can manipulate the dpkg/apt databases.

This should resolve the intermittent errors during releases like this:

    Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)


archSeer:
LGTM

PR #1005, opened 2015-02-14
deployer: Only rollback if jobs fail to start

lmars:
A crashed job means the job exited non-zero, but was running at some point, so the only case when a deployment will receive a crashed event is when scaling down an old formation, in which case it doesn't matter that the job crashed as we want it to be down anyway.


titanous:
LGTM

PR #1006, opened 2015-02-16
script: Print usage when too many arguments given

lmars:
Missing a flag like `script/run-integration-tests DeployerSuite` is ignored and runs all the tests (it should be `script/run-integration-tests -f DeployerSuite`).


titanous:
LGTM

PR #1007, opened 2015-02-16
deployer: Support waiting for service discovery events

### LINKED ISSUE #849: deployer: Wait for discoverd events instead of job events
Waiting for jobs allows us to catch run failures, but we are unable to detect crashes. By waiting for discoverd services to come up, we'd take care of both scenarios.

lmars:
Closes #849.

### DIFF
--- a/controller/deployer/strategies/main.go
+++ b/controller/deployer/strategies/main.go
@@ -7,21 +7,123 @@ import (
 	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/discoverd/client"
 )
 
-type PerformFunc func(log15.Logger, *controller.Client, *ct.Deployment, chan<- ct.DeploymentEvent) error
+type UnknownStrategyError struct {
+	Strategy string
+}
+
+func (e UnknownStrategyError) Error() string {
+	return fmt.Sprintf("deployer: unknown strategy %q", e.Strategy)
+}
+
+type Deploy struct {
+	*ct.Deployment
+	client        *controller.Client
+	deployEvents  chan<- ct.DeploymentEvent
+	jobEvents     chan *ct.JobEvent
+	serviceEvents chan *discoverd.Event
+	useJobEvents  map[string]struct{}
+	logger        log15.Logger
+}
+
+type PerformFunc func(d *Deploy) error
 
 var performFuncs = map[string]PerformFunc{
 	"all-at-once": allAtOnce,
 	"one-by-one":  oneByOne,
 }
 
-func Get(strategy string) (PerformFunc, error) {
-	if f, ok := performFuncs[strategy]; ok {
-		return f, nil
+func Perform(d *ct.Deployment, client *controller.Client, deployEvents chan<- ct.DeploymentEvent, logger log15.Logger) error {
+	log := logger.New("fn", "Perform", "deployment_id", d.ID, "app_id", d.AppID)
+
+	log.Info("validating deployment strategy")
+	performFunc, ok := performFuncs[d.Strategy]
+	if !ok {
+		err := UnknownStrategyError{d.Strategy}
+		log.Error("error validating deployment strategy", "err", err)
+		return err
 	}
-	return nil, fmt.Errorf("Unknown strategy '%s'!", strategy)
+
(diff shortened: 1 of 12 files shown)


controller/deployer/strategies/main.go
```diff
@@ -39,60 +141,87 @@ func jobEventsEqual(expected, actual jobEvents) bool {
...
+			}
+			if _, ok := d.useJobEvents[typ]; ok {
+				continue
+			}
+			log.Info("got service event", "type", typ, "state", event.Kind)
+			switch event.Kind {
+			case discoverd.EventKindUp:
+				handleEvent(typ, "up")
+			case discoverd.EventKindDown:
+				handleEvent(typ, "down")
```
titanous: I think we need to always use job events to track the "down" event as an instance could go down in discoverd and then come back up.

host/containerinit/init.go
```diff
@@ -412,20 +413,32 @@ func monitor(port host.Port, container *ContainerInit, env map[string]string) (d
...
 		}
 	}
-	addr := fmt.Sprintf("%s:%v", env["EXTERNAL_IP"], port.Port)
+	inst := &discoverd.Instance{
+		Addr:  fmt.Sprintf("%s:%v", env["EXTERNAL_IP"], port.Port),
+		Proto: port.Proto,
+	}
+	// add any Flynn environment variables to the meta
+	for k, v := range env {
+		if strings.HasPrefix(k, "FLYNN_") {
```
titanous: I think we probably need a whitelist, this will pick up user-define variables as well as some other stuff that is not relevant like `FLYNN_POSTGRES`.

controller/deployer/strategies/main.go
```diff
@@ -7,22 +7,124 @@ import (
...
+					break outer
+				}
+			case <-time.After(5 * time.Second):
+				log.Error("error creating service discovery watcher, timeout reached", "service", proc.Service)
+				return fmt.Errorf("deployer: could not create watcher for service: %s", proc.Service)
+			}
+		}
+		go func() {
+			for {
+				event, ok := <-events
```
titanous: Use `range` here

controller/deployer/strategies/main.go
```diff
@@ -39,60 +141,87 @@ func jobEventsEqual(expected, actual jobEvents) bool {
...
+func (d *Deploy) waitForJobEvents(releaseID string, expected jobEvents, log log15.Logger) error {
 	actual := make(jobEvents)
-outer:
+
+	handleEvent := func(typ, state string) {
+		if _, ok := actual[typ]; !ok {
+			actual[typ] = make(map[string]int)
+		}
+		actual[typ][state] += 1
+		d.deployEvents <- ct.DeploymentEvent{
```
titanous: Should we dedup using the job ID? It seems like there are a variety of edge cases where we could get duplicate events.

controller/deployer/strategies/main.go
lmars: `events` is a channel, so I can't both use `range` and check for a closed channel. If you are proposing dropping the closed check, I am going to actually handle that case in a subsequent PR.

lmars:
@titanous comments addressed

controller/deployer/strategies/main.go
titanous: The range loop ends when the channel closes, which seems like it would work fine?

lmars: Ok, so I am going to be changing this to something like:

```go
events := make(chan *discoverd.Event)
reconnect := func() {
	stream, err := discoverd.NewService(proc.Service).Watch(events)
}
go func() {
	for {
		event, ok := <-events
		if !ok {
			reconnect()
			continue
		}
	}
}()
```

I don't mind changing for now though.

titanous: Ah, that's fine. You can leave it.

controller/deployer/strategies/main.go
```diff
@@ -7,22 +7,124 @@ import (
...
+	"github.com/flynn/flynn/discoverd/client"
 )
 
-type PerformFunc func(log15.Logger, *controller.Client, *ct.Deployment, chan<- ct.DeploymentEvent) error
+type UnknownStrategyError struct {
+	Strategy string
+}
+
+func (e UnknownStrategyError) Error() string {
+	return fmt.Sprintf(`deployer: unknown strategy "%s"`, e.Strategy)
```
titanous: Could use `%q`

titanous:
LGTM pending minor comment.

PR #1011, opened 2015-02-16
Fix router schema and examples

jvatic:
/cc @bgentry @benburkert


bgentry:
That looks like it makes much more sense :) Sorry I missed the examples.json previously!

LGTM

titanous:
LGTM

PR #1017, opened 2015-02-17
[WIP] router: log HTTP requests

### LINKED ISSUE #168: router needs better logging
@ianneub ran into an issue recently and we have no logs to track it down.

We should log every route change as well as every single request. This would be a good time to switch to log15 (see #34).

bgentry:
This is an attempt to add structured logging to the router for HTTP requests. This fixes #168 for HTTP requests. No idea if/how this should work for the TCP listener, but that could also be done as a separate PR if it's important.

It turned out to be a bit more complicated than expected because of the fact that different bits of knowledge about a request / response pair are generated at many different places throughout the code base. To overcome that, I stuck a `Log` struct in the context at the beginning of the request pipeline. Anywhere in the code where we need to add something to that struct, we can fetch it out of the context and change a value on it.

Some items, in particular the `StatusCode` and `BodyBytes` (count of bytes in the response body) were easiest to log at the top level using a wrapped `ResponseWriter`. Others, like the time the response is received, are measured deeper in the pipeline as soon as we receive headers from the backend.

Here's a sample of the resulting logs:

```
INFO[02-16|18:50:00] request                                  app=router method=GET path=/body host=example.com fwd=127.0.0.1 backend= service=1 status=200 bytes.body=1 request_id=3a9fc75fbc8b4f76859c88c77a6682b9
INFO[02-16|18:50:00] request                                  app=router method=GET path=/header host=example.com fwd=127.0.0.1 backend= service=0 status=200 bytes.body=0 request_id=946f5e5a83434982beae7d512ce219a2
```

The format was designed to mimc the [Heroku router log format](https://devcenter.heroku.com/articles/http-routing#heroku-router-log-format). It's not a complete match, mainly because of the log15 prefixes for log level, timestamp, and message. Another key difference is that instead of recording `bytes` to represent the entire size of the response, we record `bytes.body` to record only the size of the body. It turns out it's not really feasible to record the total number of bytes for the entire response (save for hijacked conns), so hopefully this is good enough for now.

Please do let me know how you want the format to end up looking.

There are some pieces of data that we're not yet recording. One is the backend that the request was served by. Should I be logging the `ip:port` of the backend for that?

Another piece of data is the size of the request. Heroku's logs don't include this, but many will consider it valuable. We could easily wrap the request body in a `countingReadCloser`, similar to the `serveUpgrade()` path, if we wanted that data.

One result of this change is that the test logs become very noisy. Would you prefer that logs be disabled in test mode? Maybe there's already a convention for this elsewhere in Flynn?

Finally, there are no logs yet for things like failed connections to backends. Also, all 503s will look the same at this point. We could add a `code` field similar to Heroku's logs if you wanted to give more detail about why 503s are happening.

There are a few TODOs that I've intentionally left in this PR because I wasn't sure how to address them. Please give feedback on those items so I can finish it up.

### DIFF
--- a/router/reqlog/reqlog.go
+++ b/router/reqlog/reqlog.go
@@ -1,0 +1,102 @@
+package reqlog
+
+import (
+	"net/http"
+	"sync/atomic"
+	"time"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/net/context"
+	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
+)
+
+type Log struct {
+	ID         string
+	Method     string
+	Path       string
+	Host       string
+	RemoteAddr string
+	Backend    string
+	Start      time.Time
+	Finish     time.Time
+	StatusCode int
+	BodyBytes  uint64
+
+	// heroku-isms:
+
+	Level string
+	Code  string
+}
+
+func (rl *Log) WriteTo(logger log15.Logger) {
+	logger.Info("request",
+		"method", rl.Method,
+		"path", rl.Path,
+		"host", rl.Host,
+		"fwd", rl.RemoteAddr,
+		"backend", rl.Backend,
+		"service", durationMS(rl.Start, rl.Finish),
+		"status", rl.StatusCode,
+		"bytes.body", rl.BodyBytes,
+		"request_id", rl.ID,
+	)
+}
+
+type ctxKey int
+
+const (
+	ctxKeyLog ctxKey = iota
+)
+
+// NewContext creates a new context that carries the provided request log.
+func NewContext(ctx context.Context, log Log) context.Context {
+	return context.WithValue(ctx, ctxKeyLog, &log)
+}
+
+// StartTimeFromContext extracts a start time from a context.
+func FromContext(ctx context.Context) (log *Log, ok bool) {
+	log, ok = ctx.Value(ctxKeyLog).(*Log)
+	return
+}
+
+// TODO: maybe add a FromRequest function
+
+type ResponseWriterPlus interface {
+	http.CloseNotifier
+	http.Flusher
+	http.Hijacker
+	http.ResponseWriter
+}
+
+// ResponseTracker is wrapper of http.ResponseWriter that keeps track of its HTTP
+// status code and body size.
+type ResponseTracker struct {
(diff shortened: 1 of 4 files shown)


archSeer:
@jvatic Would be good if we do a similar format for the controller, it's currently really hard to read as we split it into two lines (one that contains the path and an ID, and one that contains the ID and the params)

benburkert:
woah, this looks awesome. I really like the idea of using contexts directly with logging. One thing I can see cropping up is that there will be a need for lots of `reqlogger.Log` style structs for setting up & holding to-be-logged state. If possible I'd like to see a general solution that allows state to be held before a call to the logger. What do you think about making the context a logger, and allow it to associate state for logging:

```go
ctx := ctx.WithLoggedValue(Info, ctxKeyMethod, req.Method)
// ...
ctx.Info("request processed", "bytes.body", res.Body.Size)
```

jvatic:
It would be great if we could merge this with the existing stuff in [`pkg/httphelper`](https://github.com/flynn/flynn/tree/master/pkg/httphelper) (which is currently used in the controller and host).

router/reqlog/reqlog.go
```diff
@@ -0,0 +1,102 @@
...
+		"method", rl.Method,
+		"path", rl.Path,
+		"host", rl.Host,
+		"fwd", rl.RemoteAddr,
+		"backend", rl.Backend,
+		"service", durationMS(rl.Start, rl.Finish),
+		"status", rl.StatusCode,
+		"bytes.body", rl.BodyBytes,
+		"request_id", rl.ID,
+	)
```
bgentry: oh, I meant to mention on here last night… This is currently logging synchronously, which is a problem for a high performance HTTP router. Basically we want to router to give up connections and return them to the pool as fast as possible, but we might be introducing a substantial delay by blocking on a write to stdout. Ideally we would throw these messages into a queue of some kind so they can just be logged asynchronously by another goroutine.

@titanous was mentioning log15's [Handlers](https://godoc.org/gopkg.in/inconshreveable/log15.v2#hdr-Handlers) as a possible solution for this. I haven't spent enough time looking at that to know if it's appropriate, but if there are other examples of this elsewhere in Flynn, please lmk!

titanous:
Closing this for now, we should do this in a new PR that integrates with the log aggregation tier.

PR #1021, opened 2015-02-17
scheduler: Make all schedulers watch events

lmars:
This is so that the down event of the scheduler leader is forwarded to the controller by the non-leaders (previously they would miss the event as only the leader watched for events).

This also ensures that deploying a scheduler will wait for the new scheduler to be watching host events before killing old schedulers.


archSeer:
LGTM

titanous:
LGTM

Issue #977, opened 2015-02-10
Add instance-store AMIs


titanous:
We should build and publish S3-backed AMIs, allowing EC2 instances to be booted without EBS, which removes a failure mode for running instances. Packer supports this with the `amazon-instance` builder.

titanous:
I have this half-implemented, but ran into issues booting the resulting AMI. Let me know if you'd like to pick it up.

lmars:
Push a branch and I may take a look.

titanous:
https://github.com/flynn/flynn/compare/ec2-ephemeral

This is relevant: https://github.com/Lumida/packer/wiki/Building-Ubuntu-12.04-and-14.04-HVM-Instance-Store-AMIs

PR #1022, opened 2015-02-17
Vagrantfile: Adding AWS launch capabilities (via vagrant-aws)

temujin9:


Vagrantfile
```diff
@@ -12,20 +12,53 @@ end
...
+
+    # RFC 5737 TEST-NET-1 used to avoid DNS rebind protection
+    override.vm.network "private_network", ip: "192.0.2.100"
+  end
+
+  config.vm.provider "aws" do |aws, override|
+    aws.access_key_id = ENV["AWS_ACCESS_KEY_ID"]
+    aws.secret_access_key = ENV["AWS_SECRET_ACCESS_KEY"]
+    aws.session_token = ENV["AWS_SESSION_TOKEN"]
+    aws.keypair_name = ENV["FLYNN_VAGRANT_KEY"].match("/([^/]*).pem")[1] rescue nil
```
lmars: It may be worth adding a comment with a list of the environment variables along with their purpose.

lmars: This could be:

```ruby
aws.keypair_name = File.basename(ENV["FLYNN_VAGRANT_KEY"] || "")
```

temujin9: Unfortunately, we still need to strip the `*.pem` off the back of the name, as well.

temujin9: And yes, commentary about the variables would be good.

temujin9:
Okay, requested updates made. Let me know if there are any further tweaks to cover.

Vagrantfile
```diff
@@ -12,20 +12,63 @@ end
...
+  # AWS_DEFAULT_REGION      - AWS EC2 region to launch instance in
+  # FLYNN_VAGRANT_KEY       - SSH key (.pem file) used to connect to the instance
+  # FLYNN_VAGRANT_SUBNET    - AWS VPC subnet to launch instance in
+  # FLYNN_VAGRANT_IAM       - AWS IAM role to give instance permissions
+  # FLYNN_VAGRANT_TAGS      - AWS EC2 tags to apply to instance (comma separated key-value pairs)
+  config.vm.provider "aws" do |aws, override|
+    aws.access_key_id = ENV["AWS_ACCESS_KEY_ID"]
+    aws.secret_access_key = ENV["AWS_SECRET_ACCESS_KEY"]
+    aws.session_token = ENV["AWS_SESSION_TOKEN"]
+    aws.keypair_name = File.basename(ENV["FLYNN_VAGRANT_KEY"]).split('.')[0] rescue nil
```
archSeer: `File#basename` takes a second argument that strips the suffix.
```
File.basename(ENV["FLYNN_VAGRANT_KEY"], File.extname(file))
```
Using `".*"` as the second argument would also work.

Vagrantfile
```diff
@@ -12,20 +12,63 @@ end
...
+
+    get_amis().each_pair { |ami,region| aws.region_config region, :ami => ami }
+
+    aws.region = ( ENV["AWS_DEFAULT_REGION"] || "us-east-1" )
```
archSeer: No need for parentheses.

Vagrantfile
```diff
@@ -12,20 +12,63 @@ end
...
+
+    aws.instance_type = "m3.large"
+    aws.subnet_id = ENV["FLYNN_VAGRANT_SUBNET"]
+    aws.iam_instance_profile_arn = ENV["FLYNN_VAGRANT_IAM"]
+    added_tags = Hash[ ENV["FLYNN_VAGRANT_TAGS"].split(",").map {|kv| kv.split("=")} ] rescue {}
```
archSeer: `Hash#[]` always returns `{}`. It will of course fail if there's an odd number of arguments, but I'd argue returning an error there is more intuitive than hiding the failure.

Vagrantfile
titanous: I think this should be the name of a keypair already registered in AWS, not a file?

titanous: Ah, I see that the same file is used to connect below. I don't think we can assume that the name in AWS is the same as the filename locally, so I think we need two variables.

Vagrantfile
titanous: Maybe `AWS_REGION`, is this environment variable used elsewhere?

Vagrantfile
titanous: Why the `FLYNN_VAGRANT_` prefix? It seems like we also have `AWS_*`.

Vagrantfile
```diff
@@ -12,20 +12,63 @@ end
...
+    aws.tags = { "Name" => 'flynn-build' }.merge(added_tags)
```
titanous: Maybe `flynn-vagrant`?

Vagrantfile
```diff
@@ -12,20 +12,63 @@ end
...
 
+def get_amis
+  require "open-uri"
+  require "json"
+
+  images_uri = "https://dl.flynn.io/ec2/images.json"
+  json = JSON.parse(open(images_uri).read())
+  latest = json["versions"][0]["images"]
+
+  Hash[ latest.map {|ami| [ ami["region"], ami["id"] ] } ]
```
titanous: This assumes that there will only be one AMI per region. When we add instance-store AMIs (#977, this will not be the case). I think we need to divide it up by disk type as well.

Vagrantfile
temujin9: It's a standard AWS environment variable, which I tried to use (where available).

Vagrantfile
temujin9: To distinguish between things set specifically for this program from things set in a standard AWS way.

temujin9: @archSeer Fair enough, it's a cheap hack. I'll default to an empty string (instead of nil, which causes the error I'm swallowing).

Vagrantfile
temujin9: Sure, can do.

Vagrantfile
temujin9: Sure: I can pull that logic out into our wrapper, instead.

Vagrantfile
temujin9: Done.

Vagrantfile
```diff
@@ -12,20 +12,70 @@ end
...
+  # AWS_ACCESS_KEY_ID       - AWS IAM public ID
+  # AWS_SECRET_ACCESS_KEY   - AWS IAM secret key
+  # AWS_SESSION_TOKEN       - AWS IAM role session token
+  # AWS_DEFAULT_REGION      - AWS EC2 region to launch instance in
+  # FLYNN_VAGRANT_KEY       - SSH key (.pem file) used to connect to the instance
+  # FLYNN_VAGRANT_KEYNAME   - AWS EC2 key pair name
+  # FLYNN_VAGRANT_SUBNET    - AWS VPC subnet to launch instance in
+  # FLYNN_VAGRANT_IAM       - AWS IAM role (by name) to give instance permissions
+  # FLYNN_VAGRANT_IAM_ARN   - AWS IAM role (by ARN) to give instance permissions
+  # FLYNN_VAGRANT_TAGS      - AWS EC2 tags to apply to instance (comma separated key-value pairs)
```
titanous: What do you think about `VAGRANT_AWS_*` or `FLYNN_AWS_*` for the custom AWS-specific variables?

temujin9: FLYNN_AWS_* seems good.

lmars:
@temujin9 please drop 3f718ee by rebasing on master and I'll merge.

lmars:
Thanks!

temujin9:
Thank you.

PR #1028, opened 2015-02-18
test: Don't pull the scratch Docker image

lmars:
The scratch image can no longer be pulled (see https://github.com/docker/docker/issues/10737).

This was originally added as building many images at once in CI was failing with a claim that the scratch image did not exist. Given it is no longer an image which gets downloaded, there is no need to pull it.


heavenlyhash:
LGTM

PR #1029, opened 2015-02-18
slugbuilder: Fix buildpack ordering (to explicit ordering set in buildpacks.txt)

temujin9:
Signed-off-by: Nathaniel Eliot <nathaniel.eliot@bazaarvoice.com>

### DIFF
--- a/slugbuilder/Dockerfile
+++ b/slugbuilder/Dockerfile
@@ -1,6 +1,7 @@
 FROM flynn/cedarish
 
 ADD ./builder/ /tmp/builder
-RUN xargs -L 1 /tmp/builder/install-buildpack /tmp/buildpacks < /tmp/builder/buildpacks.txt
+# Explicitly number the buildpacks directory based on the order of buildpacks.txt
+RUN nl -nrz /tmp/builder/buildpacks.txt | awk '{print $2 "\t" $1}' | xargs -L 1 /tmp/builder/install-buildpack /tmp/buildpacks
 ENTRYPOINT ["/tmp/builder/build.sh"]
 CMD []
--- a/slugbuilder/builder/build.sh
+++ b/slugbuilder/builder/build.sh
@@ -92,13 +92,14 @@ export CURL_TIMEOUT=600
 
 ## Buildpack detection
 
+# Ordering here is in line number order from buildpacks.txt
 buildpacks=(${buildpack_root}/*)
 selected_buildpack=
 
 if [[ -n "${BUILDPACK_URL}" ]]; then
   echo_title "Fetching custom buildpack"
 
-  buildpack="${buildpack_root}/custom"
+  buildpack="${buildpack_root}/custom*"
   rm -rf "${buildpack}"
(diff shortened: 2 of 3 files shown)


temujin9:
Basically, we lay down the buildpacks in an explicit order, but then pick them up in (the default) alphabetical order. We should at least pick them up in the order laid down; a better (but more code intense) solution would be to lay them down in explicit order.

lmars:
@temujin9 we could prefix the buildpacks with a number?

slugbuilder/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM flynn/cedarish
 
 ADD ./builder/ /tmp/builder
-RUN xargs -L 1 /tmp/builder/install-buildpack /tmp/buildpacks < /tmp/builder/buildpacks.txt
+# buildpacks.txt has number-prefixed names to provide explicit buildpack order
+RUN xargs -L 2 /tmp/builder/install-buildpack /tmp/buildpacks < /tmp/builder/buildpacks.txt
```
lmars: I think this should remain `-L 1`, from xargs man page:

```
       -L max-lines
              Use  at most max-lines nonblank input lines per command line.  Trailing blanks cause an input line to be logically continued on the next
              input line.  Implies -x.
```

titanous:
This approach is fine, but will require a parsing change in https://github.com/flynn/flynn/blob/master/util/bump-buildpacks/bump.go

temujin9:
@titanous Good point. Changed the plan of attack: instead of altering buildpacks, altered the Dockerfile to create explicit numeric order from the implicit file order. That means we don't need to alter the format of builtpacks.txt.

titanous:
@lmars please handle reviewing this.

slugbuilder/builder/build.sh
```diff
@@ -92,21 +92,23 @@ export CURL_TIMEOUT=600
...
 
+# Ordering here is in line number order from buildpacks.txt
 buildpacks=(${buildpack_root}/*)
 selected_buildpack=
 
 if [[ -n "${BUILDPACK_URL}" ]]; then
   echo_title "Fetching custom buildpack"
 
-  buildpack="${buildpack_root}/custom"
+  buildpack="${buildpack_root}/000000*"
```
lmars: Was there a need to change this? If so, why the `*`?

temujin9: install-buildpack now takes an order instead of a name as parameter 3, and prefixes the name with that ordering component. To get back the custom buildpack, we need to search for the one with that prefix (rather than assuming its the whole of the name).

I suppose I could use "custom*" instead of "000000*", as it looks like the custom one is the only one invoked if its set. Either way, the `*` would need to remain.

temujin9: Changed back to `custom`, to make it more legible.

lmars:
Thanks!

temujin9:
My pleasure.

Issue #1030, opened 2015-02-18
dashboard: Deploy logs do not work in Chrome/Safari


titanous:
The log field is empty, and these errors are in the console:

![screen shot 2015-02-18 at 2 25 58 pm](https://cloud.githubusercontent.com/assets/13026/6258063/6898dbf6-b77a-11e4-9240-51c260c10e99.png)

PR #1032, opened 2015-02-18
dashboard: Use new route JSON structure

jvatic:
Properties previously found in `config` are now in the root object

refs 4ebafaa3078786fb6561b5ca61a541ca57e0442f

refs #1030


titanous:
LGTM

PR #1035, opened 2015-02-19
gitreceived: Cache repos in blobstore.

### LINKED ISSUE #212: Better gitreceived repo caching
Currently if I delete an app that I've pushed to, create a new one with the same name, and push the same commits, I'll get `Everything up-to-date` instead of a deploy. The app ID should be part of the cache key (this probably depends on #211).

Also, the cache is currently per-daemon, not for the entire cluster. This may cause behavior that the user isn't expected when running multiple daemons (re-deploying the same commit, pushing extra objects). This also prevents us from allowing pulls from the gitreceived endpoint.

archSeer:
Allows us to consistently cache repos across different gitreceived instances. Fixes #212

### DIFF
--- a/gitreceived/gitreceived.go
+++ b/gitreceived/gitreceived.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"archive/tar"
 	"bytes"
 	"crypto/x509"
 	"encoding/pem"
@@ -11,6 +12,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -20,6 +22,7 @@ import (
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-shlex"
 	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/crypto/ssh"
+	"github.com/flynn/flynn/pkg/archiver"
 )
 
 const PrereceiveHookTmpl = `#!/bin/bash
@@ -31,6 +34,7 @@ done
 var prereceiveHook []byte
 
 var port = flag.String("p", "22", "port to listen on")
+var useBlobstore = flag.Bool("b", true, "use the blobstore for repo cache")
 var repoPath = flag.String("r", "/tmp/repos", "path to repo cache")
 var noAuth = flag.Bool("n", false, "disable client authentication")
 var keys = flag.String("k", "", "pem file containing private keys (read from SSH_PRIVATE_KEYS by default)")
@@ -197,13 +201,27 @@ func handleChannel(conn *ssh.ServerConn, newChan ssh.NewChannel) {
 				ch.Stderr().Write([]byte("Invalid repo.\n"))
 				return
 			}
-
-			if err := ensureCacheRepo(cmdargs[1]); err != nil {
+			tempDir := *repoPath
+			if *useBlobstore {
+				path, err := ioutil.TempDir("", "")
+				if err != nil {
+					fail("generateTempDir", err)
+					return
+				}
+				tempDir = path
+			}
+			if err := ensureCacheRepo(tempDir, cmdargs[1]); err != nil {
 				fail("ensureCacheRepo", err)
 				return
 			}
+			if *useBlobstore {
+				if err := restoreBlobstoreCache(tempDir, cmdargs[1]); err != nil {
+					fail("restoreBlobstoreCache", err)
+					return
+				}
+			}
 			cmd := exec.Command("git-shell", "-c", cmdargs[0]+" '"+cmdargs[1]+"'")
-			cmd.Dir = *repoPath
+			cmd.Dir = tempDir
 			cmd.Env = append(os.Environ(),
 				"RECEIVE_USER="+conn.User(),
 				"RECEIVE_REPO="+cmdargs[1],
@@ -223,8 +241,14 @@ func handleChannel(conn *ssh.ServerConn, newChan ssh.NewChannel) {
 				fail("exitStatus", err)
 				return
 			}
+			if *useBlobstore {
+				if err := uploadCache(tempDir, cmdargs[1]); err != nil {
+					fail("uploadCache", err)
+				}
+			}
 			if _, err := ch.SendRequest("exit-status", false, ssh.Marshal(&status)); err != nil {
 				fail("sendExit", err)
+				return
 			}
 			return
 		case "env":
@@ -302,11 +326,29 @@ func exitStatus(err error) (exitStatusMsg, error) {
 
 var cacheMtx sync.Mutex
 
-func ensureCacheRepo(path string) error {
+func restoreBlobstoreCache(tempDir, path string) error {
+	cachePath := tempDir + "/" + path
+
+	res, err := http.Get("http://blobstore.discoverd/cache/" + path + ".tar")
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == 200 {
+		// cache hit
+		r := tar.NewReader(res.Body)
+		archiver.Untar(cachePath, r)
+	}
+	return nil
+}
+
+func ensureCacheRepo(tempDir, path string) error {
 	cacheMtx.Lock()
 	defer cacheMtx.Unlock()
 
-	cachePath := *repoPath + "/" + path
+	cachePath := tempDir + "/" + path
+
 	if _, err := os.Stat(cachePath); os.IsNotExist(err) {
 		os.MkdirAll(cachePath, 0755)
 		cmd := exec.Command("git", "init", "--bare")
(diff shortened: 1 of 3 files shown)


archSeer:
Would testing a git repository with two pushes and detecting the second one reports
```
Counting objects: 1, done.
Writing objects: 100% (1/1), \n+ bytes | 0 bytes/s, done.
Total 1 (delta 0), reused 0 (delta 0)
```
be sufficient?

titanous:
Yeah, that sounds like a reasonable testing strategy.

gitreceived/gitreceived.go
```diff
@@ -307,14 +313,134 @@ func ensureCacheRepo(path string) error {
...
+	}
+	os.MkdirAll(cachePath, 0755)
+
+	res, err := http.Get("http://blobstore.discoverd/cache/" + path + ".tar")
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == 200 {
```
titanous: There should be a flag to toggle this and the old behavior so that we don't have a dependency on blobstore for uses outside of flynn.

gitreceived/gitreceived.go
```diff
@@ -307,14 +313,134 @@ func ensureCacheRepo(path string) error {
...
+						return err
+					}
+				}
+				if err := os.Chmod(filename, os.FileMode(header.Mode)); err != nil {
+					return err
+				}
+			case tar.TypeReg, tar.TypeRegA:
+				// if the files are out of order, the dir might not exist yet
+				if _, err := os.Stat(filepath.Dir(filename)); os.IsNotExist(err) {
+					if err := os.MkdirAll(filepath.Dir(filename), os.FileMode(header.Mode)); err != nil {
```
titanous: No need for the stat, mkdirall should be a no-op if the directory exists.

gitreceived/gitreceived.go
```diff
@@ -307,14 +313,134 @@ func ensureCacheRepo(path string) error {
...
+				}
+				writer, err := os.Create(filename)
+				if err != nil {
+					return err
+				}
+				defer writer.Close()
+
+				io.Copy(writer, r)
+
+				if err := os.Chmod(filename, os.FileMode(header.Mode)); err != nil {
```
titanous: Use OpenFile to set the mode and create at the same time: http://golang.org/pkg/os/#OpenFile

gitreceived/gitreceived.go
```diff
@@ -307,14 +313,134 @@ func ensureCacheRepo(path string) error {
...
+			filename := cachePath + "/" + header.Name
+
+			switch header.Typeflag {
+			case tar.TypeDir:
+				if _, err := os.Stat(filename); os.IsNotExist(err) {
+					if err := os.MkdirAll(filename, os.FileMode(header.Mode)); err != nil {
+						return err
+					}
+				}
+				if err := os.Chmod(filename, os.FileMode(header.Mode)); err != nil {
```
titanous: No need for the chmod or the stat

gitreceived/gitreceived.go
```diff
@@ -307,14 +313,134 @@ func ensureCacheRepo(path string) error {
 	defer cacheMtx.Unlock()
 
 	cachePath := *repoPath + "/" + path
-	if _, err := os.Stat(cachePath); os.IsNotExist(err) {
-		os.MkdirAll(cachePath, 0755)
+
+	// remove old copy
+	if _, err := os.Stat(cachePath); err == nil {
+		os.RemoveAll(cachePath)
```
titanous: When in blobstore cache mode, we should use a random tempdir for each push.

gitreceived/gitreceived.go
```diff
@@ -307,14 +313,134 @@ func ensureCacheRepo(path string) error {
...
+	}); err != nil {
+		return err
+	}
+	if err := tw.Close(); err != nil {
+		return err
 	}
-	return ioutil.WriteFile(cachePath+"/hooks/pre-receive", prereceiveHook, 0755)
+
+	// upload the tarball to the blobstore
+	req, err := http.NewRequest("PUT", "http://blobstore.discoverd/cache/"+path+".tar", buf)
```
titanous: The tar operation should write into a pipe from a goroutine with the read side in the http request.

gitreceived/gitreceived.go
```diff
@@ -307,14 +313,134 @@ func ensureCacheRepo(path string) error {
...
+
+	client := &http.Client{}
```
titanous: Use `http.DefaultClient`

gitreceived/gitreceived.go
```diff
@@ -307,14 +313,134 @@ func ensureCacheRepo(path string) error {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
```
titanous: No need to defer.

gitreceived/gitreceived.go
archSeer: Chmod should be there, in case we created the directory under ``case tar.TypeReg`, so that we now set it to the correct permissions.

titanous: Ah, yep.

archSeer:
@titanous All comments addressed.

archSeer:
@titanous Tests fixed.

gitreceived/gitreceived.go
```diff
@@ -302,19 +320,139 @@ func exitStatus(err error) (exitStatusMsg, error) {
...
+	defer cacheMtx.Unlock()
+
+	cachePath := tempDir + "/" + path
+
+	r, w := io.Pipe()
+	tw := tar.NewWriter(w)
+
+	go func() {
+		defer w.Close()
+		if err := filepath.Walk(cachePath, func(path string, file os.FileInfo, err error) error {
```
titanous: This and the untar should be split out into a utility package since they are generic.

gitreceived/gitreceived.go
```diff
@@ -302,19 +320,139 @@ func exitStatus(err error) (exitStatusMsg, error) {
...
+
+			if _, err = io.Copy(tw, f); err != nil {
+				return err
+			}
+			return nil
+		}); err != nil {
+			//return err
+		}
+		if err := tw.Close(); err != nil {
+			// return err
```
titanous: These comments need resolving

archSeer:
@titanous Addressed yet again. Created a `pkg/archiver` with `Tar(path, writer)/Untar(path, reader)`

gitreceived/gitreceived.go
```diff
@@ -302,19 +321,70 @@ func exitStatus(err error) (exitStatusMsg, error) {
...
+
+	cachePath := tempDir + "/" + path
+
+	r, w := io.Pipe()
+	tw := tar.NewWriter(w)
+
+	go func() {
+		defer tw.Close()
+		defer w.Close()
+		archiver.Tar(cachePath, tw)
```
titanous: This should use a channel to pass any error back so that it can be logged.

pkg/archiver/archiver.go
```diff
@@ -0,0 +1,85 @@
...
+	for {
+		header, err := r.Next()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return err
+		}
+
+		filename := dir + "/" + header.Name
```
titanous: Use `filepath.Join`

pkg/archiver/archiver.go
```diff
@@ -0,0 +1,85 @@
...
+			// if the files are out of order, the dir might not exist yet
+			if err := os.MkdirAll(filepath.Dir(filename), os.FileMode(header.Mode)); err != nil {
+				return err
+			}
+			writer, err := os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, os.FileMode(header.Mode))
+			if err != nil {
+				return err
+			}
+			defer writer.Close()
+			io.Copy(writer, r)
```
titanous: This needs an error check

gitreceived/gitreceived.go
```diff
@@ -302,19 +321,70 @@ func exitStatus(err error) (exitStatusMsg, error) {
...
+func ensureCacheRepo(tempDir, path string) error {
 	cacheMtx.Lock()
 	defer cacheMtx.Unlock()
 
-	cachePath := *repoPath + "/" + path
-	if _, err := os.Stat(cachePath); os.IsNotExist(err) {
-		os.MkdirAll(cachePath, 0755)
+	cachePath := tempDir + "/" + path
+
+	if !*useBlobstore {
```
titanous: Should probably split this into two functions since the blobstore path doesn't need a mutex.

gitreceived/gitreceived.go
```diff
@@ -197,13 +201,22 @@ func handleChannel(conn *ssh.ServerConn, newChan ssh.NewChannel) {
 				ch.Stderr().Write([]byte("Invalid repo.\n"))
 				return
 			}
-
-			if err := ensureCacheRepo(cmdargs[1]); err != nil {
+			tempDir := *repoPath
+			if *useBlobstore {
+				path, err := ioutil.TempDir("", "")
```
titanous: This could be moved into the blobstore function

gitreceived/gitreceived.go
```diff
@@ -197,13 +201,22 @@ func handleChannel(conn *ssh.ServerConn, newChan ssh.NewChannel) {
...
+				if err != nil {
+					fail("generateTempDir", err)
+					return
+				}
+				fmt.Println("tmpdir is", path)
```
titanous: extraneous log

gitreceived/gitreceived.go
archSeer: I need the `tempDir` variable available later when I start uploading the new cache.

archSeer:
@titanous Addressed yet again.

gitreceived/gitreceived.go
```diff
@@ -315,6 +357,31 @@ func ensureCacheRepo(path string) error {
...
+	return nil
+}
+
+func uploadCache(tempDir, path string) error {
+	cachePath := tempDir + "/" + path
+
+	r, w := io.Pipe()
+	tw := tar.NewWriter(w)
+
+	fail := make(chan error)
```
titanous: Call this `errCh` or similar.

titanous:
LGTM pending CI

PR #1038, opened 2015-02-20
test: Don't start daemons when building the rootfs

### LINKED ISSUE #1026: Can't unmount CI rootfs after build
Building the CI rootfs involves loop mounting the image, running a setup script via chroot, then unmounting the image, but currently the unmounting fails, e.g. during the latest restart of CI:

```
+ cleanup
+ sudo umount /tmp/tmp.AMG3m2HRQI
umount: /tmp/tmp.AMG3m2HRQI: device is busy.
        (In some cases useful info about processes that use
         the device is found by lsof(8) or fuser(1))
```

Looking in the process tree, there are 3 processes which were started inside the chroot which I suspect are stopping the image from unmounting:

```
root     15002  0.0  0.0  19484   176 ?        S    08:03   0:00 upstart-udev-bridge --daemon
root     15005  0.0  0.0  51612  3420 ?        Ss   08:03   0:00 /lib/systemd/systemd-udevd --daemon
syslog   15165  0.0  0.0 255852  2932 ?        Ssl  08:03   0:00 rsyslogd
```

We should modify the setup script so no daemons start inside the chroot (e.g. by diverting `initctl` to `/bin/true` - see http://blog.opperschaap.net/2009/11/16/chroot-with-upstart/).

lmars:
Some daemons were starting in the chroot (e.g. upstart-udev-bridge, docker, rsyslogd) and preventing unmounting the rootfs.img.

The `initctl` + `policy-rc.d` hacks were good enough for most daemons, but for some reason `docker` was managing to circumvent that, so `jchroot` is needed so that `docker` gets killed by the kernel when the PID namespace goes away.

Fixes #1026.

### DIFF
--- a/test/rootfs/build.sh
+++ b/test/rootfs/build.sh
@@ -19,7 +19,9 @@ trap cleanup ERR
 image="http://cdimage.ubuntu.com/ubuntu-core/releases/14.04.1/release/ubuntu-core-14.04.1-core-amd64.tar.gz"
 curl -L ${image} | sudo tar -xzC ${dir}
 
-sudo chroot ${dir} bash < "${src_dir}/setup.sh"
+# use jchroot (https://github.com/vincentbernat/jchroot) which uses a PID
+# namespace so daemons do not outlive the setup and prevent unmounting rootfs.img
+sudo jchroot ${dir} bash < "${src_dir}/setup.sh"
 
 sudo cp ${dir}/boot/vmlinuz-* ${build_dir}/vmlinuz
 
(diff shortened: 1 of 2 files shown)


titanous:
Did you try diverting the `service` binary? https://github.com/docker/docker/blob/e22487db370e431016b53b4044e452856ac72d10/project/make/ubuntu#L103

lmars:
I need `service` to start `postgresql` though

titanous:
LGTM

PR #1040, opened 2015-02-21
logging: Add a log aggregator component

bgentry:
This is my first batch of progress on a log aggregator component. This component takes syslog-over-TCP input (RFC 6587 / RFC 5424), parses the messages, and buckets the logs into individual buffers based on the `AppName` field in the syslog header.

Buffers currently have a fixed size of 10,000 log messages. I think I'd like to change this to be configurable and maybe make it work based on the number of bytes in the buffered messages, rather than be based on the count of the messages. It would probably be good if all buffers weren't pre-allocated with a slice of capacity 10,000.

Logs will be sent to this aggregator by the “collectors” running on each instance, a component that @benburkert is working on.

The log aggregator will provide an API that the controller will use to fetch logs out of the buffer. It will also replicate the raw log stream to another aggregator for failover purposes. Eventually it should be able to drain logs for a given channel to syslog or HTTP destinations.

Any comments on the direction of this at this early stage would be appreciated!

### DIFF
--- a/logaggregator/main.go
+++ b/logaggregator/main.go
@@ -1,0 +1,186 @@
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"flag"
+	"fmt"
+	"net"
+	"os"
+	"strconv"
+	"sync"
+
+	"github.com/flynn/flynn/pkg/shutdown"
+)
+
+func main() {
+	defer shutdown.Exit()
+
+	listenPort := os.Getenv("PORT")
+	if listenPort == "" {
+		listenPort = "5000"
+	}
+
+	listenAddr := flag.String("listenaddr", ":"+listenPort, "syslog input listen address")
+
+	a := NewAggregator(*listenAddr)
+	if err := a.Start(); err != nil {
+		shutdown.Fatal(err)
+	}
+	shutdown.BeforeExit(a.Shutdown)
+	defer shutdown.Exit()
+}
+
+// Aggregator is a log aggregation server that collects syslog messages.
+type Aggregator struct {
+	// Addr is the address (host:port) to listen on for incoming syslog messages.
+	Addr string
+
+	listener     net.Listener
+	logc         chan []byte
+	numConsumers int
+	consumerwg   sync.WaitGroup
+	producerwg   sync.WaitGroup
+
+	once     sync.Once // protects the following:
+	shutdown chan struct{}
+}
+
+// NewAggregator creates a new unstarted Aggregator that will listen on addr.
+func NewAggregator(addr string) *Aggregator {
+	return &Aggregator{
+		Addr:         "127.0.0.1:0",
+		logc:         make(chan []byte),
+		numConsumers: 10,
+		shutdown:     make(chan struct{}),
+	}
+}
+
+// Start starts the Aggregator on Addr.
+func (a *Aggregator) Start() error {
+	var err error
+	a.listener, err = net.Listen("tcp", a.Addr)
+	if err != nil {
+		return err
+	}
+	a.Addr = a.listener.Addr().String()
+
+	for i := 0; i < a.numConsumers; i++ {
+		a.consumerwg.Add(1)
+		go func() {
+			defer a.consumerwg.Done()
+			a.consumeLogs()
+		}()
+	}
+
+	a.producerwg.Add(1)
+	go func() {
+		defer a.producerwg.Done()
+		a.accept()
+	}()
+	return nil
+}
+
+// Shutdown shuts down the Aggregator gracefully by closing its listener,
+// and waiting for already-received logs to be processed.
+func (a *Aggregator) Shutdown() {
+	a.once.Do(func() {
+		close(a.shutdown)
+		a.listener.Close()
+		a.producerwg.Wait()
+		close(a.logc)
+		a.consumerwg.Wait()
+	})
+}
+
+func (a *Aggregator) accept() {
+	defer a.listener.Close()
+
+	for {
+		select {
+		case <-a.shutdown:
+			return
+		default:
+		}
+		conn, err := a.listener.Accept()
+		if err != nil {
+			continue
+		}
+
+		a.producerwg.Add(1)
+		go func() {
+			defer a.producerwg.Done()
+			a.readLogsFromConn(conn)
+		}()
+	}
+}
+
+// testing hook:
+var afterMessage func()
+
+func (a *Aggregator) consumeLogs() {
+	for line := range a.logc {
+		// TODO: forward message to follower aggregator
+		// TODO: parse the message, send it to the right bucket
+		fmt.Printf("message received: %q\n", string(line))
+		if afterMessage != nil {
+			afterMessage()
+		}
+	}
+}
+
+func (a *Aggregator) readLogsFromConn(conn net.Conn) {
+	defer conn.Close()
+
+	connDone := make(chan struct{})
+	defer close(connDone)
+
+	go func() {
+		select {
+		case <-connDone:
+		case <-a.shutdown:
+			conn.Close()
+		}
+	}()
+
+	s := bufio.NewScanner(conn)
+	s.Split(rfc6587Split)
+	for s.Scan() {
+		msg := s.Bytes()
+		// slice in msg could get modified on next Scan(), need to copy it
+		msgCopy := make([]byte, len(msg))
+		copy(msgCopy, msg)
+		a.logc <- msgCopy
+	}
+}
+
+func rfc6587Split(data []byte, atEOF bool) (advance int, token []byte, err error) {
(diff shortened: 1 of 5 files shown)


titanous:
This looks good! I'll make some style comments so that we can get this reviewed in bite-sized pieces.

logaggregator/main.go
```diff
@@ -0,0 +1,216 @@
...
+	listenPort := os.Getenv("PORT")
+	if listenPort == "" {
+		listenPort = "5000"
+	}
+
+	listenAddr := flag.String("listenaddr", ":"+listenPort, "syslog input listen address")
+
+	a := NewAggregator(*listenAddr)
+	if err := a.Start(); err != nil {
+		log.Fatal(err)
```
titanous: `shutdown.Fatal`

logaggregator/main.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	defer a.Shutdown()
```
titanous:
```go
shutdown.BeforeExit(a.Shutdown)
defer shutdown.Exit()
```

logaggregator/main.go
```diff
@@ -0,0 +1,216 @@
...
+	bmu          sync.Mutex // protects buffers
+	buffers      map[string]*ring.Buffer
+	listener     net.Listener
+	logc         chan []byte
+	numConsumers int
+	consumerwg   sync.WaitGroup
+	filerwg      sync.WaitGroup
+	producerwg   sync.WaitGroup
+
+	smu      sync.Mutex // protects the following:
```
titanous: Seems like a `sync.Once` could also be used here?

logaggregator/main.go
```diff
@@ -0,0 +1,216 @@
...
+		}()
+	}
+}
+
+func (a *Aggregator) consumeLogs() {
+	for line := range a.logc {
+		// TODO: forward message to follower aggregator
+		msg, err := rfc5424.Parse(line)
+		if err != nil {
+			fmt.Println("ERROR PARSING:", err)
```
titanous: `log` or even better use  `log15`.

logaggregator/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+var _ = Suite(&S{})
+
+func (s *S) SetUpSuite(c *C) {
+}
+
+func (s *S) TearDownSuite(c *C) {
+}
+
+func (s *S) TearDownTest(c *C) {
+}
```
titanous: Drop these methods until they are used.

logaggregator/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+func newAggregator(c *C) *Aggregator {
+	a := NewAggregator("127.0.0.1:0")
+	if err := a.Start(); err != nil {
+		c.Fatal(err)
+	}
+	return a
+}
+
+func (s *S) TestAggregatorListensOnAddr(c *C) {
```
titanous: If `s` is unused do: `func (S)`

logaggregator/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+	a := newAggregator(c)
+	defer a.Shutdown()
+
+	parts := strings.Split(a.Addr, ":")
+	ip, port := parts[0], parts[1]
```
titanous: Does `net.SplitHostPort` make sense?

logaggregator/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (s *S) TearDownTest(c *C) {
+}
+
+func newAggregator(c *C) *Aggregator {
+	a := NewAggregator("127.0.0.1:0")
+	if err := a.Start(); err != nil {
+		c.Fatal(err)
+	}
```
titanous: `c.Assert(err, IsNil)` (and ditto for the rest of the err checks)

logaggregator/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+	select {
+	case <-a.logc:
+	default:
+		c.Errorf("logc was not closed")
+	}
+}
+
+func (s *S) TestAggregatorBuffersMessages(c *C) {
+	a := newAggregator(c)
+	defer a.Shutdown()
```
titanous: Looks like this setup is done for every test could do this in setup/teardown methods and stash it in the suite struct.

logaggregator/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+	defer conn.Close()
+
+	if _, err := conn.Write([]byte(sampleLogLine1)); err != nil {
+		c.Fatal(err)
+	}
+	if _, err := conn.Write([]byte(sampleLogLine2)); err != nil {
+		c.Fatal(err)
+	}
+	conn.Close()
+	time.Sleep(10 * time.Millisecond) // time for messages to be received
```
titanous: I don't like sleeping in tests unless it's absolutely necessary, please fix it or if it's complicated add a TODO for a better test harness.

logaggregator/rfc5424/rfc5424.go
```diff
@@ -0,0 +1,172 @@
...
+
+func Parse(buf []byte) (*Message, error) {
+	cursor := 0
+	msg := &Message{}
+	if err := parseHeader(buf, &cursor, msg); err != nil {
+		return nil, err
+	}
+
+	if err := parseStructuredData(buf, &cursor, msg); err != nil {
+		return nil, err
```
titanous: Looks like these errors need a prefix. I'd suggest wrapping it in a error struct that includes the cursor so useful debug error messages can be constructed.

logaggregator/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"testing"
+	"time"
+
+	. "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
+)
+
+// Hook gocheck up to the "go test" runner
+func Test(t *testing.T) { TestingT(t) }
+
+type S struct {
```
heavenlyhash: Giving this struct a fuller name makes it possible to use with gocheck's '-check.f' flag a la `go test -check.f MyTestSuite`.  (For unit tests it's fairly moot since package specifiers are also handy, but I use this a lot in running the integration tests.)

logaggregator/main.go
```diff
@@ -0,0 +1,216 @@
...
+	// Addr is the address (host:port) to listen on for incoming syslog messages.
+	Addr string
+
+	bmu          sync.Mutex // protects buffers
+	buffers      map[string]*ring.Buffer
+	listener     net.Listener
+	logc         chan []byte
+	numConsumers int
+	consumerwg   sync.WaitGroup
+	filerwg      sync.WaitGroup
```
titanous: This looks unused.

benburkert:
Looking good so far :+1: :smile:

I liked the way you used a scanner to implement RFC6587 & the buffer based parser for `rfc5424.Message`, so I spent some time merging our syslog related stuff into a `syslog` package: https://github.com/benburkert/flynn/commit/0b77d817b5aef9353a5c9e67520ab04e7e42369a & https://github.com/benburkert/flynn/commit/af119288be765572707aa3a63266bf1379c68e7d

My plan is to open a PR for the `pkg/syslog` code once I get a bit further on the collector stuff and verify that the interfaces are correct. But I don't expect that work to cause a significant changes to the interfaces on the "server" side, so it should remain easy to pull into this branch (as the second commit shows).

logaggregator/main_test.go
bgentry: yeah, meant for there to be a TODO on this. Will see if I can get rid of it altogether.

bgentry:
I think I've fixed everything but the structured & prefixed parsing error messages.

bgentry:
@titanous is e01051b what you were thinking re: structured error messages from the syslog parser?

titanous:
@bgentry Yep, the Error method should prefix the package name, but exactly what I was thinking.

bgentry:
> the Error method should prefix the package name

@titanous sorry, can you elaborate? I'm not sure what you mean.

logaggregator/rfc5424/rfc5424.go
```diff
@@ -0,0 +1,180 @@
...
+	return msg, nil
+}
+
+type ParseError struct {
+	Cursor  int
+	Message string
+}
+
+func (p *ParseError) Error() string {
+	return p.Message
```
titanous: Convention is to prefix the error string with the name of the package it came from:

```go
return "rfc5424: "+p.Message
```

bgentry: oh right, thanks! :v:

titanous:
I don't have time to review this today, but if @benburkert signs off on it feel free to merge.

benburkert:
:+1: :shipit:

Issue #907, opened 2015-01-30
host: Running jobs stuck in "starting" state


titanous:
```text
++ 23:11:26.811 /home/ubuntu/go/src/github.com/flynn/flynn/test/../cli/flynn-cli -a starlings-humble-arnprior scale --no-wait env=1
scaling env: 0=>1


++ 23:11:26.904 waiting for job events: map[env:map[up:1]]
++ 23:11:26.928 got job event: env d9adc29f-e8c86151cc70472c9b195a77fd4ab885 starting
test_cli.go:533:
    _, jobID := app.waitFor(jobEvents{"env": {"up": 1}})
test_scheduler.go:101:
    t.Fatal("timed out waiting for job events: ", expected)
... Error: timed out waiting for job events: map[env:map[up:1]]
```

https://s3.amazonaws.com/flynn-ci-logs/20150130231002-b003fe21-build-376e150ab4de7365232b447e3333080047a3d17c-2015-01-30-23-15-07.txt

titanous:
Looks like e8c86151cc70472c9b195a77fd4ab885 got stuck in the starting state.

titanous:
Looks similar: https://s3.amazonaws.com/flynn-ci-logs/20150204072119-5e0e9d49-build-e74cfcb4b41b5525bd667cfba376e9387b98f9c2-2015-02-04-07-26-33.txt

titanous:
Might be similar: https://s3.amazonaws.com/flynn-ci-logs/20150204183814-94124f24-build-5638a35ac18798e82fc95aeed0e1d98d2f4ab26b-2015-02-04-18-44-32.txt

lmars:
Some similar failures:

* [20150220220653-3bc055e1](https://s3.amazonaws.com/flynn-ci-logs/20150220220653-3bc055e1-build-0f51e339666b3cc1d2999f8fd383f960b96448fc-2015-02-20-22-16-45.txt) - Job `c3f192a4-809a0256adb7428e8b79958ec20c9a84`stuck in starting state
* [20150220133653-f340ee49](https://s3.amazonaws.com/flynn-ci-logs/20150220133653-f340ee49-build-0f51e339666b3cc1d2999f8fd383f960b96448fc-2015-02-20-13-47-29.txt) - Job `86e58712-f0f998eb92b94ba89e19b2c21482897c` stuck in starting state

In both cases, the jobs have output so are running.

lmars:
[...]

PR #1043, opened 2015-02-21
Expose containerinit logs

lmars:
In order to debug #907, it would be useful to know what is happening internally in containerinit, and exposing logs is a start at that.

My first aim was to redirect the console of the containers to a file, but that is not supported ([the console must be a PTY](http://libvirt.org/git/?p=libvirt.git;a=blob;f=src/lxc/lxc_process.c;h=f1b36f1;hb=7212990#l1060)), and the consoles can't be connected after a container has exited to get the log, so instead I have exposed the logs via RPC and used logbuf to pipe them to a file.

Ideally I would like to call something like `flynn-host log --init ID` to see these logs but would like to get feedback on how that can be achieved? A new flag for attach to return the init logs?

### DIFF
--- a/host/attach.go
+++ b/host/attach.go
@@ -89,6 +89,9 @@ func (h *attachHandler) attach(req *host.AttachReq, conn io.ReadWriteCloser) {
 	if req.Flags&host.AttachFlagStderr != 0 {
 		opts.Stderr = newFrameWriter(2, w, writeMtx)
 	}
+	if req.Flags&host.AttachFlagInitLog != 0 {
+		opts.InitLog = newFrameWriter(3, w, writeMtx)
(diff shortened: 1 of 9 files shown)


titanous:
Since this is just for debugging, we could redirect this log into the stdout of the container. I think we just need to do a bit of juggling so that we have the write side of the stdout pipe available to write into.

lmars:
@titanous I tried just copying to stdout but a few tests which check stdout failed, and the output was intermingled, so I changed the approach to add the init logs to stream 3 in the job log, and support outputting it to stderr by passing `--init` to `flynn-host log`.

titanous:
LGTM

PR #1044, opened 2015-02-21
Volume destroy

heavenlyhash:
Add volume destroy to API, add retries to zfs destroy commands to improve stability, handle zfs child filesystems correctly in the host cli destroy-volumes command.


titanous:
LGTM

PR #1047, opened 2015-02-21
controller: Add job state constraint

lmars:
Since multiple schedulers watch for events, job state updates can get intermingled and erroneously leave the job in an incorrect state, so this change ensures that a job never returns to a previous state.


titanous:
LGTM

PR #1048, opened 2015-02-22
pkg/sse: Fix writer.Error

lmars:
`writer.Write` also tries to lock the mutex, so `writer.Error` was previously causing a deadlock.


titanous:
LGTM

PR #1053, opened 2015-02-22
Replace App.Protected with app metadata

lmars:
This changes the concept of a "protected" app to a "system" app, and lifts the restriction that such apps cannot be scaled to zero (so that they can be deployed).

@jvatic this requires a change to the dashboard, could you push a change to this PR?

### DIFF
--- a/bootstrap/manifest_template.json
+++ b/bootstrap/manifest_template.json
@@ -105,6 +105,7 @@
     "from_step": "controller",
     "app": {
       "name": "controller",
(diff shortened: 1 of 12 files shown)


jvatic:
@lmars pushed the dashboard change

lmars:
@jvatic ok thanks, I have squashed that change into my commit. This is ready for review.

titanous:
LGTM

PR #1059, opened 2015-02-23
discoverd: Implement manual leader election

titanous:
This patch adds support for creating services with a configuration that selects a leader election type. It also adds a manual leader election type that can be used instead of the default oldest instance election.

Manual leader election is useful if an external state machine is responsible for choosing the leader, but the discoverd API (DNS and HTTP) should be kept in sync. The first use case for this is the postgres appliance state machine (#986) which takes into account more than the instance age when doing leader election.


lmars:
LGTM

PR #1061, opened 2015-02-23
test: Use Basic Auth with cluster API

lmars:
Putting the auth key in the URL gets leaked in the test logs because the run script is executed with trace on.


titanous:
LGTM

PR #1062, opened 2015-02-23
host: Don't respawn in upstart.conf

lmars:
It is currently unlikely that respawning flynn-host will fix anything, it usually results in the daemon continuously flapping and filling up the logs.


titanous:
LGTM

PR #1067, opened 2015-02-23
Use new postgres appliance

titanous:
These patches replace the old postgres appliance with the new one based on the new state machine and postgres daemon manager.


titanous:
Tests are failing because the Go example app needs new vendored dependencies which we can't easily do until we merge to master.

lmars:
LGTM (very nice!)

PR #1068, opened 2015-02-24
logging: send container stdout/stderr to logaggregator service

benburkert:
Send container output on STDOUT, STDERR, and containerinit to the logaggregator service over TCP as rfc5424 encoded syslog messages.

### TODO
- [x] replace `panic`s in `LogMux` methods.
- [x] figure out a better way to test this stuff.
- [x] bootstrap process for the `logaggregator` service.
- [x] make `Follow` compatible with `Close`.

### DIFF
--- a/host/libvirt_lxc_backend.go
+++ b/host/libvirt_lxc_backend.go
@@ -27,6 +27,7 @@ import (
 	"github.com/flynn/flynn/host/containerinit"
 	lt "github.com/flynn/flynn/host/libvirt"
 	"github.com/flynn/flynn/host/logbuf"
+	"github.com/flynn/flynn/host/logmux"
 	"github.com/flynn/flynn/host/types"
 	"github.com/flynn/flynn/host/volume/manager"
 	"github.com/flynn/flynn/pinkerton"
@@ -43,7 +44,7 @@ const (
 	imageRoot      = "/var/lib/docker"
 )
 
-func NewLibvirtLXCBackend(state *State, vman *volumemanager.Manager, volPath, logPath, initPath string) (Backend, error) {
+func NewLibvirtLXCBackend(state *State, vman *volumemanager.Manager, volPath, logPath, initPath string, mux *logmux.LogMux) (Backend, error) {
 	libvirtc, err := libvirt.NewVirConnection("lxc:///")
 	if err != nil {
 		return nil, err
@@ -65,6 +66,7 @@ func NewLibvirtLXCBackend(state *State, vman *volumemanager.Manager, volPath, lo
 		logs:       make(map[string]*logbuf.Log),
 		containers: make(map[string]*libvirtContainer),
 		resolvConf: "/etc/resolv.conf",
+		mux:        mux,
 	}, nil
 }
 
@@ -84,6 +86,7 @@ type LibvirtLXCBackend struct {
 
 	logsMtx sync.Mutex
 	logs    map[string]*logbuf.Log
+	mux     *logmux.LogMux
 
 	containersMtx sync.RWMutex
 	containers    map[string]*libvirtContainer
@@ -669,11 +672,28 @@ func (c *libvirtContainer) watch(ready chan<- error) error {
 			g.Log(grohl.Data{"at": "get_streams", "status": "error", "err": err.Error()})
 			return err
 		}
+
 		log := c.l.openLog(c.job.ID)
 		defer log.Close()
-		// TODO: log errors from these
-		go log.Follow(1, stdout)
-		go log.Follow(2, stderr)
+
+		muxConfig := logmux.Config{
+			AppID:   c.job.Metadata["flynn-controller.app"],
+			HostID:  c.l.state.id,
+			JobType: c.job.Metadata["flynn-controller.type"],
+			JobID:   c.job.ID,
+		}
+
+		// TODO(benburkert): remove file logging once attach proto uses logaggregator
+		streams := []io.Reader{stdout, stderr}
+		for i, stream := range streams {
+			fd := i + 1
+			pr, pw := io.Pipe()
+			tr := io.TeeReader(stream, pw)
+
+			go log.Follow(fd, tr)
+			go c.l.mux.Follow(pr, fd, muxConfig)
+		}
+
 		go log.Follow(3, initLog)
 	}
 
--- a/host/logmux/logmux.go
+++ b/host/logmux/logmux.go
@@ -1,0 +1,144 @@
+package logmux
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"sync"
+	"time"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/technoweenie/grohl"
+
+	"github.com/flynn/flynn/discoverd/client"
+	"github.com/flynn/flynn/pkg/syslog/rfc5424"
+	"github.com/flynn/flynn/pkg/syslog/rfc6587"
+)
+
+// LogMux collects log lines from multiple readers and forwards them to a log
+// aggregator service registered in discoverd. Log lines are buffered in memory
+// and are dropped in LIFO order.
+type LogMux struct {
+	logc chan *rfc5424.Message
+
+	sc *serviceConn
+
+	producerwg *sync.WaitGroup
+
+	shutdowno sync.Once
+	shutdownc chan struct{}
+
+	doneo sync.Once
+	donec chan struct{}
+}
+
+// New returns an instance of LogMux ready to follow io.Reader producers. The
+// log messages are buffered internally until Connect is called.
+func New(bufferSize int) *LogMux {
+	return &LogMux{
+		logc:       make(chan *rfc5424.Message, bufferSize),
+		producerwg: &sync.WaitGroup{},
+		shutdownc:  make(chan struct{}),
+		donec:      make(chan struct{}),
+	}
+}
+
+// Connect opens a connection to the named log aggregation service in discoverd
+// and creates a goroutine that drains the log message buffer to the connection.
+func (m *LogMux) Connect(discd *discoverd.Client, name string) error {
+	conn, err := connect(discd, name, m.donec)
+	if err != nil {
+		return err
+	}
+	m.sc = conn
+
+	go m.drainTo(conn)
+	return nil
+}
+
+func (m *LogMux) drainTo(w io.Writer) {
+	defer close(m.donec)
+
+	g := grohl.NewContext(grohl.Data{"at": "logmux_drain"})
+
+	for {
+		msg, ok := <-m.logc
+		if !ok {
+			return // shutdown
+		}
+
+		_, err := w.Write(rfc6587.Bytes(msg))
+		if err != nil {
+			g.Log(grohl.Data{"status": "error", "err": err.Error()})
+
+			// write logs to local logger when the writer fails
+			g.Log(grohl.Data{"msg": msg.String()})
+			for msg := range m.logc {
+				g.Log(grohl.Data{"msg": msg.String()})
+			}
+
+			return // shutdown
+		}
+	}
+}
+
+// Close blocks until all producers have finished, then terminates the drainer,
+// and blocks until the backlog in logc has been processed.
+func (m *LogMux) Close() {
+	m.producerwg.Wait()
+
+	m.doneo.Do(func() { close(m.logc) })
+
+	select {
+	case <-m.donec:
+	case <-time.NewTimer(3 * time.Second).C:
+		// logs did not drain to logaggregator in 3 seconds, drain them to the local logger
+		close(m.sc.closec)
+		<-m.donec
+	}
+}
+
+type Config struct {
+	AppID, HostID, JobID, JobType string
+}
+
+// Follow forwards log lines from the reader into the syslog client. Follow
+// runs until the reader is closed or an error occurs. If an error occurs, the
+// reader may still be open.
+func (m *LogMux) Follow(r io.Reader, fd int, config Config) {
+	m.producerwg.Add(1)
+
+	hdr := &rfc5424.Header{
+		Hostname: []byte(config.HostID),
+		AppName:  []byte(config.AppID),
+		ProcID:   []byte(config.JobType + "." + config.JobID),
+		MsgID:    []byte(fmt.Sprintf("ID%d", fd)),
+	}
+
+	go m.follow(r, hdr)
+}
+
+func (m *LogMux) follow(r io.Reader, hdr *rfc5424.Header) {
+	defer m.producerwg.Done()
+
+	g := grohl.NewContext(grohl.Data{"at": "logmux_follow"})
+	bufr := bufio.NewReader(r)
+
+	for {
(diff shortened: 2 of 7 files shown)


pkg/syslog/rfc5424/message_test.go
```diff
@@ -0,0 +1,62 @@
...
+		msg := Format(&test.hdr, nil, test.msg)
+
+		got := write(msg)
+		c.Assert(got, Equals, test.want)
+	}
+}
+
+func write(msg *Message) string {
+	buf := &bytes.Buffer{}
+	w := NewWriter(buf)
```
bgentry: Rather than this standalone `write()` function, why not just have `Message.Bytes()` and `Message.String()` functions? Fewer allocations, no extra buffers, and more descriptive naming.

pkg/syslog/client.go
```diff
@@ -0,0 +1,44 @@
...
+func Dial(address string, hdr *rfc5424.Header) (*Client, error) {
+	conn, err := net.Dial("tcp", address)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{
+		c: conn,
+		h: hdr,
+		w: rfc5424.NewWriter(rfc6587.NewWriter(conn)),
```
bgentry: This seems a little over-abstracted, especially with this 3-level wrapped `Writer`.

Also, it doesn't feel idiomatic for a `func Dial()` to take a struct arg like `hdr` here. Maybe you want a constructor that returns the `*Client`, then a separate method you can call for it to `Dial()` or `Connect()` ?

You can also easily just serialize the `hdr` into `[]byte` once during construction, rather than on every write, since it never changes.

pkg/syslog/client.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (c *Client) Close() error {
+	return c.c.Close()
+}
+
+func (c *Client) Write(b []byte) (int, error) {
+	return c.WriteFormat(c.h, nil, b)
+}
+
+func (c *Client) WriteFormat(hdr *rfc5424.Header, sd rfc5424.StructuredData, msg []byte) (int, error) {
```
bgentry: Doesn't look like this is used anywhere other `Write()` above. Is there a reason it needs to be a separate function, especially exported?

pkg/syslog/client_test.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		for _, want := range test.want {
+			got := <-msgc
+			c.Assert(got.String(), Equals, want)
+		}
+	}
+}
+
+func testChanHandler(msgc chan *rfc5424.Message) messageHandler {
```
bgentry: can you inline this as an anonymous func in the one test that uses it?

pkg/syslog/rfc6587/writer.go
```diff
@@ -0,0 +1,18 @@
...
+type Writer struct {
+	w io.Writer
+}
+
+func NewWriter(w io.Writer) *Writer {
+	return &Writer{w}
+}
+
+func (w *Writer) Write(p []byte) (int, error) {
+	return fmt.Fprintf(w.w, "%d %s", len(p), p)
```
bgentry: these `Writer` structs feel like a heavy way to get what you need. 18 lines of code and a new file here just for a single `fmt` call?

pkg/syslog/rfc5424/reader.go
```diff
@@ -0,0 +1,80 @@
...
+		Header: Header{
+			Priority:  priority,
+			Version:   version,
+			Timestamp: timestamp,
+			Hostname:  parseMaybeNil(parts[3]),
+			AppName:   parseMaybeNil(parts[4]),
+			ProcID:    parseMaybeNil(parts[5]),
+			MsgID:     parseMaybeNil(parts[6]),
+		},
+		StructuredData: parseMaybeNil(parts[7]),
```
bgentry: I don't think this works, because structured data contains spaces. May be better to explicitly disallow it until we can support it properly (if we ever even need to).

bgentry:
Does any of this syslog stuff need to be general enough to write both RFC5424 and RFC6587 to the wire? Are we ever going to write anything directly in RFC5424?

There's a number of abstractions and exported structs introduced that it doesn't look like we'll need. You could probably trim a couple hundred lines off of this PR by dropping them.

pkg/syslog/client.go
bgentry: ah, you can't pre-serialize the header fully because the timestamp changes. In that case it's a little odd that you're passing in an `*rfc5424.Header` here at all since not all of its fields are actually used in the resulting message.

host/logmux/logmux.go
```diff
@@ -0,0 +1,138 @@
...
+	for {
+		line, _, err := bufr.ReadLine()
+		if err == io.EOF {
+			return
+		}
+		if err != nil {
+			panic(err)
+		}
+
+		m.bufc <- rfc5424.Format(hdr, nil, line)
```
bgentry: This seems like the ideal place to timestamp this message: as soon as it's read off the file descriptor.

host/logmux/logmux.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		select {
+		case m.logc <- msg:
+		default:
+			select {
+			case m.logc <- msg:
+			case <-m.logc:
+				m.logc <- msg
+			}
```
bgentry: Something is off here. Why is the inner select also selecting on the same thing as the outer select?



AFAICT this code is here to make sure that you're reading off the FD as fast as possible, throwing away log lines if you can't process them fast enough.

It seems like a better place to do that is in `follow()`, where you can do the following to just throw away log lines if `bufc` doesn't have space for them:

```go
select {
case m.bufc <- msg:
default:
	// throw away msg
}
```

Then maybe you don't even need 2 separate channels (`bufc` and `logc`).

pkg/syslog/rfc6587/writer.go
benburkert: yeah, I started off with a `type WriterFunc func([]byte) (int, error)` a la `http.HandlerFunc`, but I later switched to `fmt.Fprintf` and that broke it. It's a smell but it shouldn't stick around long.

host/logmux/logmux.go
benburkert: > AFAICT this code is here to make sure that you're reading off the FD as fast as possible, throwing away log lines if you can't process them fast enough.

Yes, but it's throwing out the oldest message in the buffer, not the new one that just arrived. I'll add some docs on why the inner select is necessary.

bgentry: Is there a reason you need to throw away the oldest message instead of the
newest message? Does it make any difference which logs you drop if you've
got to drop some?

bgentry:
@benburkert I think the rebase got borked on this, it's showing a bunch of extra commits

benburkert:
This is ready for another look, although there are still some undefined/untested behaviors. The `serviceConn` watcher cannot recover when the `discoverd` client/watch dies. The current behavior on a shutdown is to keep draining the streams over the connection, and if that connection fails, dump the remaining log lines to the grohl logger. If the logaggregator service is down when a shutdown is initiated, it hangs. Fixing this will add complexity, so i'd like first like to figure out what should happen on a proper shutdown vs. one where the logaggregator service is down or flapping.

bgentry:
Did you manage to try this with the changes from #1107? The CLI tests in there should just magically start working with these changes

titanous:
> The current behavior on a shutdown is to keep draining the streams over the connection, and if that connection fails, dump the remaining log lines to the grohl logger. If the logaggregator service is down when a shutdown is initiated, it hangs. Fixing this will add complexity, so i'd like first like to figure out what should happen on a proper shutdown vs. one where the logaggregator service is down or flapping.

Draining to the aggregator before shutting down definitely makes sense. If the aggregator is down when shutting down, I think dumping unflushed log buffers to a file somewhere make sense, but I don't have strong opinions other than that we shouldn't drop them on the floor.

Anything else need clarifying?

benburkert:
@bgentry I haven't had much success, I started a branch here: https://github.com/flynn/flynn/tree/log-aggregator-integration-tests

@titanous cool, that answers my questions.

benburkert:
this is ready for :eyes:

titanous:
Does this handle reconnecting if the TCP connection drops but the endpoint is still the same (no discoverd changes?).

titanous:
I also don't see any retry logic for connecting.

bgentry:
@benburkert I asked @titanous about initLogs and he said they're just for debugging purposes. We probably don't even need to send them to the aggregator, as long as they're visible from `flynn-host log`. @lmars can confirm if that seems sane. At least that would let us avoid shipping those around just to end up filtering them out in 99% of cases.

host/libvirt_lxc_backend.go
```diff
@@ -663,12 +666,27 @@ func (c *libvirtContainer) watch(ready chan<- error) error {
...
+
 		log := c.l.openLog(c.job.ID)
 		defer log.Close()
-		// TODO: log errors from these
-		go log.Follow(1, stdout)
-		go log.Follow(2, stderr)
-		go log.Follow(3, initLog)
+
+		muxConfig := logmux.Config{
+			AppName: c.job.Metadata["flynn-controller.app_name"],
```
bgentry: just noticed this.. we don't want to log into a channel by app name, we want to use something that doesn't change (app ID).

benburkert: want me to swap `JobID` and `AppName` so that `ProcID=<config.AppName>.<config.JobType>` and `AppName=<config.JobID>`?

bgentry: No, I don't think appname should be used anywhere, as it can change at any time if somebody renames their app. The other values (`JobID`, `JobType`, and `AppID`) are fixed for a given job, and those values never become stale.

So I think just changing to `AppName: c.job.Metadata["flynn-controller.app_id"]` is sufficient.

benburkert:
> Does this handle reconnecting if the TCP connection drops but the endpoint is still the same (no discoverd changes?).

Handled & tested in the latest commit.

> I also don't see any retry logic for connecting.

This is tricky b/c the leader can change at any time, so the retry logic for connecting can't block incoming discoverd events. I think it's handled implicitly in the `EventKindLeader` case, which performs a reconnect. Does this provide enough of a fallback? Or is it possible for the service to transition from unreachable to reachable without triggering the discoverd `EventKindLeader` event.

titanous:
> Does this provide enough of a fallback? Or is it possible for the service to transition from unreachable to reachable without triggering the discoverd EventKindLeader event.

No, this is not enough. The discoverd leader is heartbeat based, so a service could bounce from unreachable (network disruption) to reachable without the TTL on the instance expiring.

benburkert:
This should be ready for another round of :eyes:, the latest commit attempts to reconnect every 100ms.

host/libvirt_lxc_backend.go
```diff
@@ -669,11 +672,28 @@ func (c *libvirtContainer) watch(ready chan<- error) error {
...
 		}
+
 		log := c.l.openLog(c.job.ID)
 		defer log.Close()
-		// TODO: log errors from these
-		go log.Follow(1, stdout)
-		go log.Follow(2, stderr)
+
+		muxConfig := logmux.Config{
+			AppName: c.job.Metadata["flynn-controller.app"],
```
titanous: This should be the app ID.

host/logmux/service_conn.go
```diff
@@ -0,0 +1,187 @@
...
+
+				if err := c.reset(); err != nil {
+					g.Log(grohl.Data{"status": "error", "err": err.Error()})
+				}
+
+				if err := c.connect(srv); err != nil {
+					g.Log(grohl.Data{"status": "error", "err": err.Error()})
+					reconc = time.After(100 * time.Millisecond)
+				}
+			case discoverd.EventKindDown:
```
titanous: This check should be dropped or changed, as it will trigger for any instance of the service that goes down, not just the leader. Since there is reconnect logic in place, I don't think it's necessary.

host/libvirt_lxc_backend.go
bgentry: I think it is? here's the metadata I see from flynn-host for a local process:

```
flynn-controller.app_name  logaggregator
flynn-controller.release   9af2f285a3694981a629684c89856eb6
flynn-controller.type      app
flynn-controller.app       1f9325db7de94a4d82a091b456d70ceb
```

titanous: Ah, right. I was just confused because the field is called `AppName`.

bgentry: actually, I've noticed an issue with metadata when writing integration tests. Here's another bit of metadata for a local job:

```
ENV[FLYNN_JOB_ID]          96a3d2f495414019a73f7254fbc225fc
```

We're sending logs to the log aggregator with the job ID as `96a3d2f495414019a73f7254fbc225fc`, but afaict the external representation of that is actually `flynn-96a3d2f495414019a73f7254fbc225fc`. So as of now, in order to actually view the logs for a job, I have to run:

```
flynn log -j 96a3d2f495414019a73f7254fbc225fc
```

Rather than using the identifier shown in `flynn ps`, like this:

```
flynn log -j flynn-96a3d2f495414019a73f7254fbc225fc
```

Where should that transformation take place? Should we be adding a `flynn-` prefix before the logs get sent to the aggregator, or should we be stripping it when requesting logs from the aggregator via the controller?

titanous: The `flynn` prefix is the host id, which should already be in the message somewhere? If not, it should be added.

lmars: I have [a commit to fix this](https://github.com/flynn/flynn/commit/529ff1c2fd60382d89b3ab4b72bdffa4d52605a5) on a branch, I will PR it individually soon.

bgentry: @lmars nice, thanks :+1:

benburkert: ok, to clarify, here is the current format for syslog messages:

`<1>0 {timestamp} {hostID} {appID} {jobType}.{jobID} ID{fd} - {message}`

here is a sample of a message from a CI run:

`<0>1 2015-03-12T19:46:42.003397093Z flynn 68f43e1007584aba8d29b26286178aca printer.9ec352554d864355ae3835d645d26a56 ID1 - I like to print`

@titanous et al, does this look correct?

titanous: That looks good! The only question I have is about the `ID1` part. Does it make sense to call stdout and stderr `FD1` and `FD2`?

benburkert: `ID` was initially used because the initlog was being sent too, and it's not an FD. I kept it b/c i figured that in the future we may collect from some other source than STDOUT/STDERR (e.g. `/dev/log`). The underlying FD number is not constant like with STDOUT/STDERR, so `ID` (which the example prefix used in the RFC) avoids the possibly faulty assumption that this number value is the FD number the process sees.

bgentry:
Anything left to do on this PR? It's working great with #1158 including the new integration tests that I'm still finishing up.

titanous:
LGTM

PR #1069, opened 2015-02-24
Postgres appliance v2

titanous:
All of the code in this PR has already been reviewed. This lands the new postgres appliance based on the [manatee state machine](https://github.com/joyent/manatee-state-machine) that is "correct" in that it has automatic failovers and will sacrifice write availability for consistency (CP) by using synchronous replication.

I will update the Go example as soon as this is merged so that CI goes green.


lmars:
LGTM pending CI

PR #1086, opened 2015-02-25
logaggregator: move syslog stuff to pkg/syslog/*

bgentry:
This is inspired by the syslog packaging stuff from #1068, specifically ae839fd51ecf25b91fcc7eb379482e4c8e455ca2. This version maintains history by relocating existing code out of the `logaggregator` and `logaggregator/rfc5424` packages.

It also drops some of the items from that commit that I didn't think were pulling their weight. In particular, the [`Client` struct](https://github.com/flynn/flynn/commit/ae839fd51ecf25b91fcc7eb379482e4c8e455ca2#diff-ef981dc4d02042ae0ab0e1376b821960) has been dropped, as have the `Reader` and `Writer` types in that commit. Otherwise it's mostly the same as that commit.

As in ae839fd51, the `Message` struct now uses `[]byte` instead of `string` for many of the fields, which will save us a lot of garbage. It doesn't really complicate its usage either as of now.


bgentry:
oh, /cc @benburkert

logaggregator/main.go
```diff
@@ -178,7 +175,7 @@ func (a *Aggregator) readLogsFromConn(conn net.Conn) {
 	}()
 
 	s := bufio.NewScanner(conn)
-	s.Split(rfc6587Split)
+	s.Split(rfc6587.Split)
```
benburkert: did you mean to use `rfc6587.Scanner`? or should that be :bomb: too?

bgentry: Well, I left that in because I thought you wanted to use it (maybe for tests or something). But can easily just nuke it.

benburkert: nah, nuke it for now, i'll add it back if i need it.

bgentry: cool, it's gone. :v:

bgentry:
Anything left to do on this or should I merge it? I could use this stuff in my API work.

benburkert:
LGTM. :+1:

PR #1089, opened 2015-02-25
Run HA clusters

titanous:
This changes bootstrap to detect whether it should run in singleton mode by first checking the `SINGLETON` environment variable and if not set, checking if the number of expected nodes is 1.

If singleton mode is not enabled, enough instances of each process are run to make the control plane tolerate the loss of a node without having to start new instances.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -2,6 +2,7 @@ package bootstrap
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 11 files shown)


bootstrap/bootstrap.go
```diff
@@ -108,6 +115,10 @@ func Run(manifest []byte, ch chan<- *StepInfo, minHosts int) (err error) {
 	state := &State{
 		StepData:  make(map[string]interface{}),
 		Providers: make(map[string]*ct.Provider),
+		Singleton: minHosts == 1,
+	}
+	if s := os.Getenv("SINGLETON"); s != "" {
+		state.Singleton = s == "true"
```
lmars: `state.Singleton = os.Getenv("SINGLETON") == "true"`

lmars:
LGTM pending minor comment

bootstrap/bootstrap.go
titanous: I'm not sure that would work, as it would override the `minHosts == 1` check above even if unset.

lmars: Ah totally missed that assignment.

PR #1090, opened 2015-02-26
Installer

jvatic:


cli/install.go
```diff
@@ -0,0 +1,86 @@
...
+usage: flynn install aws [-n <instances>] [--aws-access-key-id <key-id>] [--aws-secret-access-key <secret>] [--aws-region <region>]
+
+"aws" is currently the only supported value for <target>
+
+Commands:
+	aws  creates a flynn cluster on EC2
+
+Options:
+  -n, --instances <instances> Number of instances to launch [default: 1]
+	--aws-access-key-id <key-id>  AWS access key ID
```
titanous: indentation

cli/install.go
```diff
@@ -0,0 +1,86 @@
...
+  --aws-secret-access-key <secret> AWS access key secret
+  --aws-region <region> AWS region [default: us-east-1]
+
+Examples:
+
+	$ flynn install aws --aws-access-key-id=asdf --aws-secret-access-key=fdsa
+`)
+}
+
+var InvalidInstallTarget = errors.New("Invalid install target")
```
titanous: no need for this to be a var

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+		EventChan: make(chan *Event),
+		ErrChan:   make(chan error),
+		Done:      make(chan struct{}),
+
+		InstanceIPs: make([]string, 0, instances),
+		instances:   instances,
+
+		Creds: creds,
+
+		ec2Client: ec2.New(creds, region, nil),
```
titanous: call this `ec2`?

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+		var imageID string
+		for _, i := range latestVersion.Images {
+			if i.Region == s.Region {
+				imageID = i.ID
+				break
+			}
+		}
+		if imageID == "" {
+			err := errors.New(fmt.Sprintf("No image found for region %s", s.Region))
+			return err
```
titanous: return `err` directly

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+		}
+		s.ImageID = imageID
+		return nil
+	}(); err != nil {
+		go s.HandleEvent("image_id_fetch_error", err.Error())
+		if s.ImageID != "" {
+			go s.HandleEvent("image_id_load", "Falling back to saved Image ID")
```
titanous: It looks like these should be channel sends, otherwise the ordering is not guaranteed.

titanous: You can use a function to do the send

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+	go s.HandleEvent("key_create", "Creating key pair")
+	keypair, err := sshkeygen.Generate()
+	if err != nil {
+		return err
+	}
+
+	var publicKey bytes.Buffer
+	encoder := base64.NewEncoder(base64.StdEncoding, &publicKey)
+	encoder.Write(keypair.PublicKey)
+	encoder.Close()
```
titanous: Use `Encode` with a `[]byte` sized with `EncodedLen`

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+	s.SSHKeyName = *res.KeyName
+
+	err = saveSSHKey(keypairName, keypair)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+var stackTemplate = template.Must(template.ParseFiles("stack-template.json"))
```
titanous: We need to compile this in by using an inline variable or `go-bindata`

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+
+type stackTemplateData struct {
+	Instances []struct{}
+}
+
+func (s *Stack) createStack() error {
+	go s.HandleEvent("start_script_gen", "Generating start script")
+	startScript, discoveryToken, err := genStartScript(1)
```
titanous: This needs to use the actual number of instances, up to a maximum of five

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+	"time"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/awslabs/aws-sdk-go/aws"
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/awslabs/aws-sdk-go/gen/cloudformation"
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/awslabs/aws-sdk-go/gen/ec2"
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/awslabs/aws-sdk-go/gen/route53"
+	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/crypto/ssh"
+	"github.com/flynn/flynn/pkg/etcdcluster"
+	"github.com/flynn/flynn/pkg/sshkeygen"
+	r "github.com/flynn/flynn/util/release"
```
titanous: `r` seems short, why not `release`?

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+	client := &http.Client{}
+	resp, err := client.Get("https://dl.flynn.io/ec2/images.json")
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.New(fmt.Sprintf("Failed to fetch list of flynn images: %s", resp.Status))
+	}
+	dec := json.NewDecoder(resp.Body)
+	var manifest *r.EC2Manifest
```
titanous:
```go
manifest := release.EC2Manifest{}
```

titanous: Actually, can just drop the `*` and what you have will work fine.

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+				}
+				var desc string
+				if se.ResourceStatusReason != nil {
+					desc = " (" + *se.ResourceStatusReason + ")"
+				}
+				name := *se.ResourceType
+				if se.LogicalResourceID != nil {
+					name = name + "(" + *se.LogicalResourceID + ")"
+				}
+				go s.HandleEvent(name+"_"+*se.ResourceStatus, name+"\t"+*se.ResourceStatus+desc)
```
titanous: Use `fmt.Sprintf` for all these

installer/installer.go
```diff
@@ -0,0 +1,743 @@
...
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
+
+	return nil
+}
+
+func (s *Stack) fetchStack() error {
+	cf := s.cf
```
titanous: Why?

installer/stack-template.json
```diff
@@ -0,0 +1,227 @@
...
+            "IpProtocol": "icmp",
+            "FromPort": "3",
+            "ToPort": "-1",
+            "CidrIp": "0.0.0.0/0"
+          }
+        ]
+      }
+    },
+
+    {{range $i, $_ := .Instances}}
```
titanous: Is the `$_` necessary?

jvatic: Yes, otherwise `$i` will be the value.

cli/install.go
jvatic: it doesn't like being a const

jvatic:
@titanous comments addressed, ready for another review pass

installer/installer.go
```diff
@@ -0,0 +1,821 @@
...
+
+func (s *Stack) HandleEvent(description string) {
+	s.eventMutex.Lock()
+	defer s.eventMutex.Unlock()
+	s.EventChan <- &Event{description}
+}
+
+func (s *Stack) HandleError(err error) {
+	s.errMutex.Lock()
+	defer s.errMutex.Unlock()
```
titanous: Why the mutexes for this and the above method? Channels are thread-safe

installer/installer.go
```diff
@@ -0,0 +1,821 @@
...
+		return nil
+	}(); err != nil {
+		s.HandleEvent(err.Error())
+		if s.ImageID != "" {
+			s.HandleEvent("Falling back to saved Image ID")
+			return nil
+		} else {
+			return err
+		}
+	}
```
titanous: Instead of an anonymous function for processing, you can do this:

```go
func (s *Stack) fetchImageID() (err error) {
  defer func() {
    if err == nil {
      return
    }
    // error handling here
  }()

File: installer/installer.go [hunk 16]
```diff
@@ -0,0 +1,821 @@
...
+	enc := base64.StdEncoding
+	publicKeyBytes := make([]byte, enc.EncodedLen(len(keypair.PublicKey)))
+	enc.Encode(publicKeyBytes, keypair.PublicKey)
+
+	res, err := s.ec2.ImportKeyPair(&ec2.ImportKeyPairRequest{
+		KeyName:           aws.String(keypairName),
+		PublicKeyMaterial: publicKeyBytes,
+	})
+	if err != nil {
+		if strings.Contains(err.Error(), "already exists") {
```
titanous: Is the error a structured type? If so we should try to match against the actual error code instead of the message.

cli/install.go
```diff
@@ -0,0 +1,109 @@
...
+			return err
+		}
+	}
+
+	stack := &installer.Stack{
+		NumInstances: instances,
+		InstanceType: instanceType,
+		Region:       region,
+		Creds:        creds,
+		YesNoPromt:   promptYesNo,
```
titanous: `s/Promt/Prompt/`

installer/installer.go
titanous: Also, this check should be moved up into the above if statement to reduce indentation.

installer/installer.go
```diff
@@ -0,0 +1,821 @@
...
+	}
+	s.DiscoveryToken = discoveryToken
+	s.persist()
+
+	validInstanceTypesBytes, err := json.Marshal(ValidEC2InstanceTypes)
+	if err != nil {
+		return err
+	}
+	var stackTemplateBuffer bytes.Buffer
+	err = stackTemplate.Execute(&stackTemplateBuffer, &stackTemplateData{make([]struct{}, s.NumInstances), string(validInstanceTypesBytes), ValidEC2InstanceTypes[0]})
```
titanous: Use field names and one line per field

installer/installer.go
```diff
@@ -0,0 +1,821 @@
...
+			}
+			if err := s.persist(); err != nil {
+				s.HandleError(err)
+			}
+		}
+	}()
+	return nil
+}
+
+func (s *Stack) HandleEvent(description string) {
```
titanous: Call this `SendEvent`

installer/installer.go
```diff
@@ -0,0 +1,821 @@
...
+	for {
+		status, err := s.Domain.Status()
+		if err != nil {
+			return err
+		}
+		if status == "applied" {
+			break
+		}
+		time.Sleep(time.Second)
+	}
```
titanous: This step should be done after bootstrap, it isn't required to bootstrap, so we just waste time if we do it here.

installer/installer.go
```diff
@@ -0,0 +1,821 @@
...
+func (s *Stack) configureDNS() error {
+	s.HandleEvent("Configuring DNS")
+	r53 := route53.New(s.Creds, s.Region, nil)
+	res, err := r53.GetHostedZone(&route53.GetHostedZoneRequest{ID: aws.String(s.DNSZoneID)})
+	if err != nil {
+		return err
+	}
+	if err := s.Domain.Configure(res.DelegationSet.NameServers); err != nil {
+		return err
+	}
```
titanous: We can technically do this as soon as we get the stack event showing the zone is created by cloudformation. Add a TODO about it.

installer/installer.go
```diff
@@ -0,0 +1,821 @@
...
+	defer sshConn.Close()
+
+	done := make(chan struct{})
+	go func() {
+		select {
+		case <-time.After(10 * time.Minute):
+			sshConn.Close()
+		case <-done:
+		}
+	}()
```
titanous: No need for this.

cli/install.go
```diff
@@ -0,0 +1,109 @@
...
+usage: flynn install <target> [-n <instances>] [-t <instance-type>] [--aws-access-key-id <key-id>] [--aws-secret-access-key <secret>] [--aws-region <region>]
+
+Targets:
+	aws  creates a flynn cluster on EC2
+
+Options:
+  -n, --instances <instances> Number of instances to launch [default: 1]
+  -t, --type <instance-type> Type of instances to launch (%s) [default: %s]
+      --aws-access-key-id <key-id>  AWS access key ID
+      --aws-secret-access-key <secret> AWS access key secret
```
titanous: We should note the env var defaults here.

installer/installer.go
jvatic: Oh, neat!

installer/installer.go
jvatic: Good point!

installer/installer.go
jvatic: Comment added

installer/installer.go
jvatic: Done.

jvatic:
@titanous comments addressed

installer/installer.go
```diff
@@ -0,0 +1,807 @@
...
+			}
+		case "dashboard-login-token":
+			if err := json.Unmarshal(*step.Data, &loginTokenData); err != nil {
+				return err
+			}
+		case "log-complete":
+			if err := json.Unmarshal(*step.Data, &logCompletionData); err != nil {
+				return err
+			}
+			s.SendEvent(logCompletionData.Message)
```
jvatic: Hmm, should probably defer printing this out until after DNS is live.

cli/install.go
```diff
@@ -0,0 +1,108 @@
...
+usage: flynn install <target> [-n <instances>] [-t <instance-type>] [--aws-access-key-id <key-id>] [--aws-secret-access-key <secret>] [--aws-region <region>]
+
+Targets:
+	aws  creates a flynn cluster on EC2
+
+Options:
+  -n, --instances <instances> Number of instances to launch [default: 1]
+  -t, --type <instance-type> Type of instances to launch (%s) [default: %s]
+      --aws-access-key-id <key-id>  AWS access key ID [default: AWS_ACCESS_KEY_ID]
+      --aws-secret-access-key <secret> AWS access key secret [default: AWS_SECRET_ACCESS_KEY]
```
titanous: I think these default statements will override the env vars, which is not what we want. Better to do it in prose: `Defaults to $AWS_ACCESS_KEY_ID`

cli/install.go
```diff
@@ -0,0 +1,108 @@
...
+      --aws-region <region> AWS region [default: us-east-1]
+
+Examples:
+
+	$ flynn install aws --aws-access-key-id=asdf --aws-secret-access-key=fdsa
+`, validInstanceTypes, defaultInstanceType))
+}
+
+var InvalidInstallTarget = errors.New("Invalid install target")
```
titanous: Why does this need to be a global var?

installer/installer.go
```diff
@@ -0,0 +1,807 @@
...
+func (s *Stack) validateInputs() error {
+	if s.NumInstances <= 0 {
+		return fmt.Errorf("You must specify at least one instance")
+	}
+
+	if s.Region == "" {
+		return fmt.Errorf("No region specified")
+	}
+
+	if s.NumInstances > 5 {
```
titanous: We should disallow two instances.

installer/installer.go
```diff
@@ -0,0 +1,807 @@
...
+						isFailed = true
+					}
+				}
+				var desc string
+				if se.ResourceStatusReason != nil {
+					desc = " (" + *se.ResourceStatusReason + ")"
+				}
+				name := *se.ResourceType
+				if se.LogicalResourceID != nil {
+					name = name + "(" + *se.LogicalResourceID + ")"
```
titanous: Use `fmt.Sprintf` for this and the above

installer/installer.go
```diff
@@ -0,0 +1,807 @@
...
+
+	attempts := 0
+	maxAttempts := 3
+	var sshConn *ssh.Client
+	for {
+		sshConn, err = ssh.Dial("tcp", ipAddress+":22", sshConfig)
+		if err != nil {
+			if attempts < maxAttempts {
+				attempts += 1
+				continue
```
titanous: sleep for a second between attempts

installer/installer.go
```diff
@@ -0,0 +1,807 @@
...
+		Pin string `json:"pin"`
+	}
+	var loginTokenData struct {
+		Token string `json:"data"`
+	}
+	var logCompletionData struct {
+		Message string `json:"message"`
+	}
+	var buf bytes.Buffer
+	output := json.NewDecoder(io.TeeReader(stdout, &buf))
```
titanous: `buf` looks unused.

installer/persistance.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if err := os.MkdirAll(filepath.Dir(dataPath), 0755); err != nil {
+		return err
+	}
+	file, err := os.Create(dataPath)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	io.Copy(file, &buf)
```
titanous: `json.Encode` directly into the file

installer/persistance.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if err := pem.Encode(f, &pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(key.PrivateKey)}); err != nil {
+		return err
+	}
+
+	f, err = os.Create(filepath.Join(keysDir, name+".pub"))
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(key.PublicKey)
```
titanous: Use `ioutil.WriteFile`

installer/installer.go
titanous: Yeah.

installer/installer.go
jvatic: Added check below

jvatic:
@titanous comments addressed, testing it out to make sure nothing broke

titanous:
LGTM

jvatic:
CI failures are unrelated, merging

PR #1091, opened 2015-02-26
appliance/postgresql: Fix xlog position scan

titanous:
We need to scan into a string, otherwise we’ll get an error.

Also added tests for both paths through the method.


lmars:
LGTM

PR #1093, opened 2015-02-26
discoverd: Add json tags to ServiceMeta

titanous:
Before this the fields were capitalized in the JSON, which doesn’t match our conventions.


lmars:
LGTM

PR #1101, opened 2015-02-27
logging: fix syslog parser

benburkert:
The syslog message parser was including an extra space at the beginning of the message byte slice. It was `panic`ing when the syslog message did not contain a `MSG` field, in which case there is `SP` after `STRUCTURED-DATA`. For fields containing a nil value ('-'), the header member is set to nil instead of `[]byte{'-'}`.

/cc @bgentry

### DIFF
--- a/pkg/syslog/rfc5424/parser.go
+++ b/pkg/syslog/rfc5424/parser.go
@@ -137,6 +137,10 @@ func parseNextField(buf []byte, cursor *int) ([]byte, error) {
 	}
 	res := buf[*cursor:nextSpace]
 	*cursor = nextSpace + 1
+
+	if bytes.Equal(res, nilValue) {
+		return nil, nil
+	}
 	return res, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/syslog/rfc5424/parser.go
```diff
@@ -19,7 +19,7 @@ func Parse(buf []byte) (*Message, error) {
 	}
 
 	if cursor < len(buf) {
-		msg.Msg = buf[cursor:]
+		msg.Msg = buf[cursor+1:]
```
bgentry: I think for consistency w/ the other parse funcs, this should actually be fixed inside `parseStructuredData()` by changing [this line](https://github.com/benburkert/flynn/blob/logging-fix-parser/pkg/syslog/rfc5424/parser.go#L155) to a `*cursor += 2`. The other parse methods skip over the trailing space (like [this one](https://github.com/benburkert/flynn/blob/d51e5120de0838e2f8d0f91cb71eda6907bb70a3/pkg/syslog/rfc5424/parser.go#L106))

bgentry:
Aside from the small nit about where the cursor is incremented, :+1: LGTM

pkg/syslog/rfc5424/parser.go
benburkert: This can cause `cursor > len(buf)` when the message is not present. It doesn't cause a panic, but it feels wrong to put the cursor in a bad position. So it would be better to check for that:

```go
if len(buf) > *cursor+2 {
	*cursor++
}
*cursor++
```

(these types of edge cases are why long ago compilers started splitting up tokenizing & parsing into two steps :stuck_out_tongue_winking_eye:)

PR #1106, opened 2015-02-27
discoverd/client: Expose automatic metadata to register caller

titanous:
This allows us to read the generated metadata from an Instance after registering it.


lmars:
LGTM

PR #1107, opened 2015-02-27
log aggregation: Log aggregator API, controller actions, and associated clients

bgentry:
This adds an API to the logaggregator, as well as a Client for that API. It also adds a controller action to be able to access that API and return logs to external clients, and attempts to add JSON schema for that action.

#### What should the output to the user look like?

Currently, I'm somewhat mimicking the output of Heroku's log tails:

```
2009-11-10T23:00:00.123450Z app[web.12345]: testing message
```
    
That's what is actually returned by the logaggregator API as of now. It's nice that it's concise, but it loses a lot of metadata that might be important to be able to get from places like the controller. For example, the hostname and stream ID (as encoded in the syslog message's `SEVERITY` field) are both lost before the controller gets ahold of the messages. That means the controller can't currently do things like return SSE events that distinguish between `STDOUT`/`STDERR` as the job log API currently offers.

#### Are the JSON schema changes correct and complete?

I'm still a bit lost on this and not sure if I made the right additions to the schema in the right places, including the examples. Don't know how I'm supposed to test that either.

#### For testing, do I need to create a fake logaggregator API?

Some components like the host service and router API have fake implementations in the controller for testing against. What's the current best practice for that sort of thing? The good news is that the logaggregator API is very simple, currently just a single action (and may stay that way).

#### Follow, wait, and filtering features

None of these have been added yet. Not sure if I should try to add them to this PR or just tackle them separately once we get the bare essentials merged. Leaning towards punting on it until we've got a very basic end-to-end setup going.

### Remaining:

* [x] Refactor to preserve metadata outside of logaggregator
* [x] Add logaggregator service to bootstrap process
* [x] Stub service for logaggregator API to use in controller tests
* [x] Tests for controller action
* [ ] Fetch logs from CLI (moved to #1158)
* [ ] Integration / CI tests (moved to #1158)

### DIFF
--- a/controller/client/client.go
+++ b/controller/client/client.go
@@ -267,6 +267,43 @@ func (c *Client) GetApp(appID string) (*ct.App, error) {
 	return app, c.Get(fmt.Sprintf("/apps/%s", appID), app)
 }
 
+// GetAppLog returns a ReadCloser log stream of the app with ID appID. If lines is
+// above zero, the number of lines returned will be capped at that value.
+// Otherwise, all available logs are returned. If follow is true, new log lines
+// are streamed after the buffered log.
+func (c *Client) GetAppLog(appID string, lines int, follow bool) (io.ReadCloser, error) {
+	return c.getAppLog(appID, lines, follow, false)
+}
+
+// GetAppLogWithWait is just like GetAppLog, except that it will wait for a
+// non-existent log stream to be created rather than simply returning empty
+// results.
+func (c *Client) GetAppLogWithWait(appID string, lines int, follow bool) (io.ReadCloser, error) {
+	return c.getAppLog(appID, lines, follow, true)
+}
+
+func (c *Client) getAppLog(appID string, lines int, follow, wait bool) (io.ReadCloser, error) {
+	path := fmt.Sprintf("/apps/%s/log", appID)
+	query := url.Values{}
+	if lines > 0 {
+		query.Add("lines", strconv.Itoa(lines))
+	}
+	if follow {
+		query.Add("follow", "true")
+	}
+	if wait {
+		query.Add("wait", "true")
+	}
+	if encodedQuery := query.Encode(); encodedQuery != "" {
+		path = fmt.Sprintf("%s?%s", path, encodedQuery)
+	}
+	res, err := c.RawReq("GET", path, nil, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	return res.Body, nil
+}
+
 // GetDeployment returns a deployment queued on the deployer.
 func (c *Client) GetDeployment(deploymentID string) (*ct.Deployment, error) {
 	res := &ct.Deployment{}
--- a/controller/examples/examples.go
+++ b/controller/examples/examples.go
@@ -67,6 +67,7 @@ func main() {
 		{"app_initial_release_get", e.getInitialAppRelease},
 		{"app_get", e.getApp},
 		{"app_list", e.listApps},
+		{"app_log", e.getAppLog},
 		{"app_update", e.updateApp},
 		{"app_resource_list", e.listAppResources},
 		{"route_create", e.createRoute},
@@ -207,6 +208,14 @@ func (e *generator) updateApp() {
 		},
 	}
 	e.client.UpdateApp(app)
+}
+
+func (e *generator) getAppLog() {
+	res, err := e.client.GetAppLogWithWait(e.resourceIds["app"], 0, false)
+	if err == nil {
+		defer res.Close()
+		io.Copy(ioutil.Discard, res)
+	}
 }
 
 func (e *generator) listAppResources() {
--- a/docs/api-examples/controller.json
+++ b/docs/api-examples/controller.json
@@ -64,6 +64,22 @@
 			"body": "{\"id\":\"7406a4d71a0c43d3ac4b39f006cb0342\",\"name\":\"my-app-1422557606845347930\",\"protected\":false,\"strategy\":\"all-at-once\",\"created_at\":\"2015-01-29T18:53:26.845896Z\",\"updated_at\":\"2015-01-29T18:53:26.845896Z\"}"
 		}
 	},
+	"app_log": {
+		"request": {
+			"method": "GET",
+			"url": "/apps/7406a4d71a0c43d3ac4b39f006cb0342/log?wait=true",
+			"headers": {
+				"Authorization": "Basic OnMzY3IzdA==",
+				"Content-Type": "application/json"
+			}
+		},
+		"response": {
+			"headers": {
+				"Content-Type": "text/plain"
+			},
+			"body": "{\"hostname\":\"my.flynn.local\",\"job_id\":\"flynn-cef74685c83b47889c69fa95451e75b3\",\"msg\":\"a log message\",\"process_type\":\"web\",\"source\":\"app\",\"stream\":\"stderr\",\"timestamp\":\"2009-11-10T23:00:00.123450Z\"}"
+		}
+	},
 	"app_initial_release_get": {
 		"request": {
 			"method": "GET",
--- a/logaggregator/api.go
+++ b/logaggregator/api.go
@@ -48,7 +48,6 @@ func (a *aggregatorAPI) GetLog(ctx context.Context, w http.ResponseWriter, req *
 		}
 	}
 
-	// TODO(bgentry): sort here, or sort w/ heap in buffer...
 	w.WriteHeader(200)
 	messages := a.agg.ReadLastN(channelID, lines)
 	enc := json.NewEncoder(w)
--- a/logaggregator/api.go
+++ b/logaggregator/api.go
@@ -1,0 +1,107 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"strconv"
+
+	"github.com/flynn/flynn/logaggregator/client"
+	"github.com/flynn/flynn/pkg/ctxhelper"
+	"github.com/flynn/flynn/pkg/httphelper"
+	"github.com/flynn/flynn/pkg/syslog/rfc5424"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/julienschmidt/httprouter"
+	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/net/context"
+	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
+)
+
+func apiHandler(agg *Aggregator) http.Handler {
+	api := aggregatorAPI{agg: agg}
+	r := httprouter.New()
+
+	r.GET("/log/:channel_id", httphelper.WrapHandler(api.GetLog))
+	return httphelper.ContextInjector(
+		"logaggregator-api",
+		httphelper.NewRequestLogger(r),
+	)
+}
+
+type aggregatorAPI struct {
+	agg *Aggregator
+}
+
+func (a *aggregatorAPI) GetLog(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	params, _ := ctxhelper.ParamsFromContext(ctx)
+	channelID := params.ByName("channel_id")
+	vals := req.URL.Query()
+
+	// TODO(bgentry): tail -f support
+
+	lines := 0
+	if strLines := vals.Get("lines"); strLines != "" {
+		var err error
+		lines, err = strconv.Atoi(strLines)
+		if err != nil || lines < 0 || lines > 10000 {
+			respondWithError(w, "lines", "lines must be an integer between 0 and 10000")
+			return
+		}
+	}
+
+	// TODO(bgentry): sort here, or sort w/ heap in buffer...
+	w.WriteHeader(200)
+	messages := a.agg.ReadLastN(channelID, lines)
+	enc := json.NewEncoder(w)
+	for _, syslogMsg := range messages {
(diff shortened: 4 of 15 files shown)


titanous:
Ping @jvatic for schema review.

titanous:
> Some components like the host service and router API have fake implementations in the controller for testing against. What's the current best practice for that sort of thing?

Yeah, current practice is to implement a stub of the client in the controller tests to avoid dependencies on other daemons. The goal is to have unit tests that run quickly so that we can refactor/modify the controller without having to run the entire integration suite after every local change.

> Follow, wait, and filtering features

Leave these for another PR.

bgentry:
@jvatic regarding the schema changes.. I couldn't find any examples in existing schema where query params were used for validation. Is that something that's currently supported?

Currently I've got the following params:

* `lines`, an int ≥ 0
* `follow`, a bool
* `wait`, a bool

There should more in the future as we add filtering for things like process type, process ID, and maybe hostname or stream type.

titanous:
> What should the output to the user look like?

This isn't actually the question we need to answer here. Presentation should be handled by the controller clients (cli/dashboard). I think we should provide as much information as possible to raw API clients. One option definitely needs to be JSON (parsing anything else in JS is crazy), probably SSE since it gets us streaming out of the box and we have all of the client/server code to handle that.

 I'm not sure how tightly coupled the syslog message format is to our internals? Would it make sense to expose it via the API as well? I guess if a drain type available in the future is syslog then we'll need to have a standard format to emit, so exposing the same format via this API would make sense, but we don't need to do that now.

bgentry:
> This isn't actually the question we need to answer here. Presentation should be handled by the controller clients (cli/dashboard). I think we should provide as much information as possible to raw API clients. One option definitely needs to be JSON (parsing anything else in JS is crazy), probably SSE since it gets us streaming out of the box and we have all of the client/server code to handle that.

Yeah, that makes sense and is what I started realizing as I was writing the controller stuff. So the logaggregator API and the controller API both need to preserve as much info as possible.

> I'm not sure how tightly coupled the syslog message format is to our internals? Would it make sense to expose it via the API as well? I guess if a drain type available in the future is syslog then we'll need to have a standard format to emit, so exposing the same format via this API would make sense, but we don't need to do that now.

That's a good question. We're using syslog for transport between the collectors (running on each instance) and the logaggregator service. We don't preserve all available metadata through that transformation right now, though we could probably do so if we adopted syslog's structured data. There are also some things we can't easily do yet, for example distinguishing based on the source of a log (such as a user's app vs. a router log for that app or a resource for that app). That's because the only identifier we have right now is the app's ID; we can't also identify which component is injecting logs into that app's stream. Injecting logs in general is not something that we've accounted for yet.

If we want, we can avoid propagating syslog outside of the logaggregator (save for eventual syslog drains) and instead expose a structured JSON format to consumers of that API, specifically the controller.  I'm leaning towards this approach because it will let us offer multiple external formats, and it would be much easier to extend with arbitrary metadata for individual log messages.

That would let us keep the details of the syslog format confined to the logaggregator and the collectors. It would also make it easier to switch to something like protobufs in the future.

jvatic:
@bgentry RE: schema changes. This is fine for now. We're not currently validating query params directly, though we should probably start having an example for each meaningful combination of them (assuming the responses can be validated via the schema, which is not currently the case for SSE).

titanous:
> That would let us keep the details of the syslog format confined to the logaggregator and the collectors. It would also make it easier to switch to something like protobufs in the future.

Sounds good. Any unanswered questions left?

bgentry:
Nope, not at the moment. Will see if I have any trouble figuring out how to write controller or integration tests for this.

bgentry:
Another question to ask. Should I be merging the "job log" API into a new "app log" API that takes a job parameter, or should that API be left as-is for attaching to a running job?

Depending on that answer, how should the CLI implementation look? Two separate commands, or a single `flynn log` command that requires an app param but also accepts a job ID param?

titanous:
> Should I be merging the "job log" API into a new "app log" API that takes a job parameter, or should that API be left as-is for attaching to a running job?

I think we can drop the "job log" API, as the one-off job API already handles attaching.

bgentry:
@titanous cool. Regarding the SSE stuff that API currently does, I'm assuming that's used by browser clients like the dashboard? We probably do want to preserve that functionality in the new API, right?

Also, where in the bootstrap process should the log aggregator go? I'm thinking near the beginning since the controller has a (weak) dependency on it.

titanous:
> Regarding the SSE stuff that API currently does, I'm assuming that's used by browser clients like the dashboard? We probably do want to preserve that functionality in the new API, right?

Yes, we need SSE for the dashboard. When the API solidifies, we'll have @jvatic rework the UI for the new API.

> Also, where in the bootstrap process should the log aggregator go? I'm thinking near the beginning since the controller has a (weak) dependency on it.

Yeah, I'd put it right after the controller is fully booted/configured (step name `scheduler`).

bgentry:
This PR should be ready for a final review.

I've moved the integration tests into a separate PR (#1158) because they won't work until #1068 is also merged. That being said, I've tested #1158 manually on a local cluster by injecting logs and using the upcoming CLI changes to fetch them. That all works :v:

controller/client/client.go
```diff
@@ -267,6 +267,52 @@ func (c *Client) GetApp(appID string) (*ct.App, error) {
...
+	path := fmt.Sprintf("/apps/%s/log", appID)
+	query := url.Values{}
+	query.Add("wait", "true")
+	if lines > 0 {
+		query.Add("lines", strconv.Itoa(lines))
+	}
+	if follow {
+		query.Add("follow", "true")
+	}
+	if encodedQuery := query.Encode(); encodedQuery != "" {
```
titanous: No need for this conditional, it will always evaluate to true. That being said, perhaps it makes sense to share the core of these two methods since the only difference is a single URL param?

controller/examples/examples.go
```diff
@@ -209,6 +210,13 @@ func (e *generator) updateApp() {
 	e.client.UpdateApp(app)
 }
 
+func (e *generator) getAppLog() {
+	res, err := e.client.GetAppLogWithWait(e.resourceIds["app"], 0, false)
+	if err == nil {
+		io.Copy(ioutil.Discard, res)
+	}
+}
```
titanous: Looks like res needs a Close?

docs/api-examples/controller.json
```diff
@@ -64,6 +64,22 @@
...
+			"headers": {
+				"Authorization": "Basic OnMzY3IzdA==",
+				"Content-Type": "application/json"
+			}
+		},
+		"response": {
+			"headers": {
+				"Content-Type": "text/plain"
+			},
+			"body": "\u0000\u0003\u0001\u0000\u0000\u0000\u0000\u0003\u0002\u0000\u0000\u0000\u0000"
```
titanous: Is this body correct? Shouldn't it be JSON?

logaggregator/api.go
```diff
@@ -0,0 +1,107 @@
...
+func (a *aggregatorAPI) GetLog(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	params, _ := ctxhelper.ParamsFromContext(ctx)
+	channelID := params.ByName("channel_id")
+	vals := req.URL.Query()
+
+	// TODO(bgentry): tail -f support
+
+	lines := 0
+	strLines := vals.Get("lines")
+	if strLines != "" {
```
titanous: Could combine the above assignment into this if.

logaggregator/api_test.go
```diff
@@ -0,0 +1,131 @@
...
+	c.Assert(string(b), Equals, expected)
+}
+
+func assertAllLogsEquals(c *C, r io.Reader, expected string) {
+	donec := make(chan struct{})
+	go func() {
+		logb, err := ioutil.ReadAll(r)
+		c.Assert(err, IsNil)
+		result := string(logb)
+		c.Assert(result, Equals, expected)
```
titanous: Calling gocheck assertions is not safe in goroutines.

logaggregator/client/client.go
```diff
@@ -0,0 +1,95 @@
...
+			ErrNotFound: ErrNotFound,
+			URL:         url,
+			HTTP:        http,
+		},
+	}
+	return c
+}
+
+// NewClient creates a new Client pointing at uri and using key for
+// authentication.
```
titanous: The key is a lie. :cake:

logaggregator/client/client.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil, err
+	}
+	return res.Body, nil
+}
+
+// Message represents a single log message.
+type Message struct {
+	// Hostname is the host that the job was running on when this log message was
+	// emitted.
+	Hostname string `json:"hostname,omitempty"`
```
titanous: We call this `HostID`, as it is not always the hostname.

logaggregator/client/client.go
```diff
@@ -0,0 +1,95 @@
...
+	Msg string `json:"msg,omitempty"`
+	// ProcessType is the type of process that emitted this log message.
+	ProcessType string `json:"process_type,omitempty"`
+	// Source is the source of this log message, such as "app" or "router".
+	Source string `json:"source,omitempty"`
+	// Stream is the I/O stream that emitted this message, such as "stdout" or
+	// "stderr".
+	Stream string `json:"stream,omitempty"`
+	// Timestamp is the time that this log line was emitted.
+	Timestamp time.Time `json:"timestamp,omitempty"`
```
titanous: What do you think about using milliseconds since the unix epoch here? It would avoid leap second nonsense and be more compact.

logaggregator/api.go
```diff
@@ -0,0 +1,107 @@
...
+			respondWithError(w, "lines", "lines must be an integer between 0 and 10000")
+			return
+		}
+	}
+
+	w.WriteHeader(200)
+	messages := a.agg.ReadLastN(channelID, lines)
+	enc := json.NewEncoder(w)
+	for _, syslogMsg := range messages {
+		if err := enc.Encode(NewMessageFromSyslog(syslogMsg)); err != nil {
```
titanous: We use SSE everywhere else internally for streams (and have client and server libraries that handle the JSON automatically). Other than the ~8 byte overhead per message, is there any reason not to use it here?

controller/client/client.go
bgentry: yeah, I'll pull pretty much the whole method out into a separate one and call it from both `GetAppLog()` and `GetAppLogWithWait()`. Thanks :v:

controller/examples/examples.go
bgentry: Thanks. I also wasn't 100% sure what to do with these "examples" or how they're used. The rest of it looks ok though?

docs/api-examples/controller.json
bgentry: yep, though that may change depending on whether we want this API to be SSE only.

controller/examples/examples.go
titanous: The examples are used in the API docs: https://flynn.io/docs/api/controller#create-app

@jvatic can explain in detail.

logaggregator/api_test.go
bgentry: Yeah, why exactly is that? Isn't `testing` thread-safe? Why can't gocheck be?

bgentry: Wait, is that true here even though there are no concurrent checks going on? How should I set up a timeout on this instead? Do I need to send the result of `ioutil.ReadAll()` over a chan and assert on it from elsewhere?

titanous: I think this is true because assert can be called after the test goroutine exits (timeout case). It will only trigger the race if the test fails, so maybe not a big deal. The solution is to send the result to the test goroutine, yes.

titanous: We should probably just fix the thread safety issue, we already have a fork of gocheck...

logaggregator/client/client.go
bgentry: I much prefer the RFC3339 format for timestamps in JSON as it's more human friendly, though I guess the leap second scenario could be a concern. Do you care much about the precision, and whether or not it's ms vs μs?

logaggregator/api.go
bgentry: I definitely agree with offering SSE streams for browsers and supporting clients, though I question whether that's the right default in an API that you might want to `curl`.

logaggregator/client/client.go
titanous: I don't think microseconds is exposed in `time`? Nanoseconds should be fine. I agree about the human readability, but I think this is one of the few cases where having precise unambiguous time is useful.

logaggregator/api.go
titanous: Hmm, yeah, that's reasonable. The same case could be argued for all of the other internal streams that are not consumed by JS. I'll open an issue.

logaggregator/client/client.go
bgentry: @titanous so in JSON, that'd have to be `UnixNano()` encoded as a string rather than a number in order to avoid floating point issues.

The border of a leap second doesn't strike me as a particularly big concern in terms of the design of this API, certainly not enough to warrant the worsened UX of the alternative. But if that's what you'd prefer, I'll make the change.

titanous: Ugh, JS. Leave it.

bgentry:
@titanous I think I've fixed everything except the gocheck race. Are you ok with fixing that in the package itself rather than complicating tests over it?

titanous:
Yep, LGTM.

PR #1111, opened 2015-02-28
Revert "test: Support booting multiple clusters"

lmars:
Reverts flynn/flynn#1110

This change is being reverted due to `AddHosts` not working (new hosts do not get any flynn-host peers). The patch needs re-thinking.


titanous:
LGTM

PR #1112, opened 2015-02-28
script: Destroy volumes before running integration tests

lmars:
I was getting strange errors in postgres tests and I realised it was because service meta was persisting between test runs. I don't think there is any benefit in doing that.


script/bootstrap-flynn
```diff
@@ -14,14 +14,16 @@ OPTIONS:
   -b BACKEND    The job backend to use [default: `libvirt-lxc`]
   -d DOMAIN     The default domain to use [default: `dev.localflynn.com`]
   -i IP         The external IP address to bind to [default: the IP assigned to `eth0`]
+  -z            Destroy volumes
```
titanous: Let's make destroying volumes the default and a flag to disable it.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #1114, opened 2015-02-28
cli: Add installed cluster as default

titanous:
After we install a cluster, it makes sense to configure the CLI to talk to it. Before this we just printed the config command, but this is unnecessary work for the user.

/cc @jvatic


installer/installer.go
```diff
@@ -768,7 +779,7 @@ func (s *Stack) bootstrap() error {
 	if err := s.waitForDNS(); err != nil {
 		return err
 	}
-	s.SendEvent(logCompletionData.Message)
```
jvatic: We can get rid of the `json.Unmarshal` for this above if we're not using it.

titanous: Fixed.

jvatic:
LGTM

PR #1118, opened 2015-03-01
test: Support booting multiple clusters

lmars:
This is similar to #1110 but updated so that `AddHosts` adds hosts to the default cluster.

### DIFF
--- a/test/cluster/cluster.go
+++ b/test/cluster/cluster.go
@@ -21,6 +21,13 @@ import (
 	"github.com/flynn/flynn/pkg/random"
 )
 
+type ClusterType uint8
+
+const (
+	ClusterTypeDefault ClusterType = iota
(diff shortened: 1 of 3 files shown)


test/cluster/cluster.go
```diff
@@ -166,17 +184,35 @@ func (c *Cluster) Boot(rootFS string, count int, dumpLogs io.Writer, killOnFailu
 	}()
 
 	c.log("Booting", count, "VMs")
-	_, err = c.startVMs(rootFS, count, true, true)
+	var instances []*Instance
+	instances, err = c.startVMs(c.rootFS, count, true)
```
titanous: Can just use `:=` here?

titanous:
LGTM pending comment

PR #1122, opened 2015-03-02
bootstrap: Use cluster domain wildcard cert as router fallback

titanous:
This gets us TLS for default routes out of the box if the CA certificate is installed via the dashboard. It also changes the behavior when there is no valid TLS certificate for a route to serving the wildcard, which will result in an untrusted certificate warning, instead of just closing the connection which causes cryptic errors in clients.


lmars:
LGTM

jvatic:
This may break logic in the dashboard for detecting if the certificate is installed.

jvatic:
Never mind, dashboard logic still works.

PR #1129, opened 2015-03-02
Added the FreeTDS library to the default cedarish image.

thomasfedb:
Added the FreeTDS library, which is used for connecting to SQL Server.

PR #1132, opened 2015-03-02
Revert "Added the FreeTDS library to the default cedarish image."

lmars:
Reverts flynn/flynn#1129

We need to keep the `cedarish` image inline with Heroku's `cedar-14` image.


titanous:
LGTM

PR #1137, opened 2015-03-02
CLI support for listing out current app resources

### LINKED ISSUE #1136: I should be able to see a list of resources on my app from the CLI.

patcoll:
With corresponding provider names. Fixes #1136

### DIFF
--- a/cli/resource.go
+++ b/cli/resource.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"log"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
@@ -11,12 +10,15 @@ import (
 
 func init() {
 	register("resource", runResource, `
-usage: flynn resource add <provider>
+usage: flynn resource
+       flynn resource add <provider>
 
 Manage resources for the app.
 
 Commands:
-	add  provisions a new resource for the app using <provider>.
+	With no arguments, shows a list of resources.
+
+	add   provisions a new resource for the app using <provider>.
 `)
 }
 
@@ -24,7 +26,27 @@ func runResource(args *docopt.Args, client *controller.Client) error {
 	if args.Bool["add"] {
 		return runResourceAdd(args, client)
 	}
-	return fmt.Errorf("Top-level command not implemented.")
+
+	resources, err := client.AppResourceList(mustApp())
+	if err != nil {
+		return err
+	}
+
+	w := tabWriter()
+	defer w.Flush()
+
+	var provider *ct.Provider
+
+	listRec(w, "ID", "Provider ID", "Provider Name")
+	for _, j := range resources {
+		provider, err = client.GetProvider(j.ProviderID)
+		if err != nil {
+			return err
+		}
+		listRec(w, j.ID, j.ProviderID, provider.Name)
+	}
+
+	return err
 }
 
 func runResourceAdd(args *docopt.Args, client *controller.Client) error {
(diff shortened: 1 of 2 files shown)


titanous:
Thanks! The commit message needs a `cli: ` prefix.

cli/resource.go
```diff
@@ -24,7 +26,24 @@ func runResource(args *docopt.Args, client *controller.Client) error {
...
+	if err != nil {
+		return err
+	}
+
+	w := tabWriter()
+	defer w.Flush()
+
+	var provider *ct.Provider
+
+	listRec(w, "ID", "ProviderID", "ProviderName")
```
titanous: I think we should add spaces between Provider and ID/Name, so `Provider ID`, etc.

patcoll:
Should I amend it to be this then? `cli:Add support for listing out current app resources, with corresponding provider names.`

titanous:
The first line should be `cli: Add provider list command`, you can describe in more detail starting on the third line. The commit should be amended and force-pushed to the same branch.

cli/resource.go
```diff
@@ -24,7 +26,24 @@ func runResource(args *docopt.Args, client *controller.Client) error {
...
+	}
+
+	w := tabWriter()
+	defer w.Flush()
+
+	var provider *ct.Provider
+
+	listRec(w, "ID", "Provider ID", "Provider Name")
+	for _, j := range resources {
+		provider, err = client.GetProvider(j.ProviderID)
```
titanous: This error should be handled otherwise the next line could trigger a nil panic.

patcoll: Right indeed. Added a clause to continue if error found.

titanous: The new code has the potential to ignore errors (error, then success would hide the issue). I'd suggest just returning the error right away.

patcoll: Should CLI commands halt on first error? Asking earnestly. I wonder if it's
better to output as much as you can in a streaming format. Perhaps tabular
output to stdout, errors to stderr.

On Monday, March 2, 2015, Jonathan Rudenberg <notifications@github.com>
wrote:

> In cli/resource.go
> <https://github.com/flynn/flynn/pull/1137#discussion_r25658959>:
>
> > @@ -24,7 +26,24 @@ func runResource(args *docopt.Args, client *controller.Client) error {
> >  	if args.Bool["add"] {
> >  		return runResourceAdd(args, client)
> >  	}
> > -	return fmt.Errorf("Top-level command not implemented.")
> > +
> > +	resources, err := client.AppResourceList(mustApp())
> > +	if err != nil {
> > +		return err
> > +	}
> > +
> > +	w := tabWriter()
> > +	defer w.Flush()
> > +
> > +	var provider *ct.Provider
> > +
> > +	listRec(w, "ID", "Provider ID", "Provider Name")
> > +	for _, j := range resources {
> > +		provider, err = client.GetProvider(j.ProviderID)
>
> The new code has the potential to ignore errors (error, then success would
> hide the issue). I'd suggest just returning the error right away.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/flynn/flynn/pull/1137/files#r25658959>.
>


-- 
Patrick Collins

titanous: > Should CLI commands halt on first error?

Until we add retry support, I think that makes sense.

> Perhaps tabular output to stdout, errors to stderr.

I'm pretty sure we already do this.

patcoll: Fixed.

titanous:
Looks good! Please squash the last commit and rebase on the latest master and I'll merge.

patcoll:
Done and ready for merge.

titanous:
Thanks!

PR #1138, opened 2015-03-02
logaggregator: bootstrap

benburkert:
Add logaggregator as a layer1 service.

### DIFF
--- a/bootstrap/manifest_template.json
+++ b/bootstrap/manifest_template.json
@@ -269,6 +269,28 @@
     "status": 401
   },
   {
+    "id": "logaggregator",
+    "action": "deploy-app",
+    "app": {
+      "name": "logaggregator",
+      "meta": {"flynn-system-app": "true"}
+    },
+    "artifact": {
+      "type": "docker",
+      "uri": "$image_repository?name=flynn/logaggregator&id=$image_id[logaggregator]"
+    },
+    "release": {
+      "processes": {
+        "app": {
+          "cmd": ["logaggregator", "-listenaddr", ":1514"]
+        }
+      }
+    },
+    "processes": {
+      "app": 1
+    }
+  },
+  {
     "id": "taffy",
     "action": "deploy-app",
     "app": {
--- a/logaggregator/Dockerfile
+++ b/logaggregator/Dockerfile
@@ -1,0 +1,5 @@
+FROM flynn/busybox
+
+ADD ./bin/logaggregator /bin/logaggregator
+
+ENTRYPOINT ["/bin/logaggregator"]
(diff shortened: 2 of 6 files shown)


titanous:
Will also need adding to `util/release/version_template.json`

lmars:
This will need adding to [the updater's system apps](https://github.com/flynn/flynn/blob/master/updater/updater.go#L32-L38) (which will shortly move, see #1139) so that it gets updated.

bgentry:
Thanks for getting started on this, it's exactly what I needed to do next anyway :)

bootstrap/manifest_template.json
```diff
@@ -67,6 +67,34 @@
     "url": "http://postgres-api.discoverd/ping"
   },
   {
+    "id": "logaggregator",
```
bgentry: @titanous I think you asked me to put this component after all of the controller stuff (i.e. after the last controller step, `scheduler`), but @benburkert wouldn't have known that. Is the current ordering fine or do you want it after `scheduler`?

titanous: It should come after the scheduler. `run-app` is only used to bootstrap hard dependencies of the controller, `deploy-app` should be used after the controller+scheduler are running as it uses the controller API instead of emulating the behavior of the controller/scheduler in bootstrap.

benburkert:
this should be ready to go. /cc @titanous, @bgentry

bootstrap/manifest_template.json
```diff
@@ -269,6 +269,33 @@
...
+      "name": "logaggregator",
+      "meta": {"flynn-system-app": "true"}
+    },
+    "artifact": {
+      "type": "docker",
+      "uri": "$image_repository?name=flynn/logaggregator&id=$image_id[logaggregator]"
+    },
+    "release": {
+      "env": {
+        "PORT": "1514"
```
titanous: Any reason for this?

bootstrap/manifest_template.json
```diff
@@ -269,6 +269,33 @@
...
+      },
+      "processes": {
+        "app": {
+          "host_network": true,
```
titanous: What is the rationale for using the host network namespace?

bootstrap/manifest_template.json
```diff
@@ -269,6 +269,33 @@
...
+          "cmd": ["logaggregator", "-listenaddr", ":1514"],
+          "omni": true
```
titanous: I don't think this needs to be an omni job? Omni jobs run on every host in the cluster.

bootstrap/manifest_template.json
benburkert: can `flynn-host daemon` connect to it otherwise?

titanous: Yes, absolutely, the routes for the overlay network are in the host namespace.

titanous:
LGTM

PR #1139, opened 2015-03-02
updater: Deploy taffy as part of the update process

lmars:
Although taffy does not have any processes, it needs to be deployed so that it's release uses the latest image.


titanous:
LGTM

lmars:
@titanous updated to also update the builder / runner image URIs.

titanous:
LGTM++

PR #1141, opened 2015-03-02
cli: Don't scale unspecified types

### LINKED ISSUE #1140: scaling command should only target the specified processes
When I scale a service, Flynn sets all other services to zero. For example, 
`flynn scale mq=2`
sets all other services to 0, forcing me to type
`flynn scale mq=2 db=1 strat=1 etc.`
to keep everything else running each time I scale.

titanous:
If a process type is not specified when running `flynn scale foo=2` it should not touch it, otherwise we need to specify every process type for every scale action.

Closes #1140

### DIFF
--- a/cli/scale.go
+++ b/cli/scale.go
@@ -77,7 +77,11 @@ func runScale(args *docopt.Args, client *controller.Client) error {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


lmars:
LGTM

Issue #1143, opened 2015-03-03
Installation complete but flynn command not found.


jeffkit:
The installation logs show as bellow:

```text
===> 10:07:08.674 installing runtime dependencies
===> 10:07:08.676 running "apt-get install --yes aufs-tools iptables libvirt-bin ubuntu-zfs"
Reading package lists... Done
Building dependency tree       
Reading state information... Done
iptables is already the newest version.
aufs-tools is already the newest version.
libvirt-bin is already the newest version.
ubuntu-zfs is already the newest version.
The following packages were automatically installed and are no longer required:
  libjansson4 sqlite3 uwsgi-core
Use 'apt-get autoremove' to remove them.
0 upgraded, 0 newly installed, 0 to remove and 105 not upgraded.
===> 10:07:09.404 downloading flynn-host binary to tmp dir
===> 10:09:36.954 verifying flynn-host binary checksum
===> 10:09:36.999 moving verified flynn-host binary to /usr/local/bin
===> 10:09:37.003 running "gunzip /tmp/tmp.RrqFFY1NZJ/flynn-host.gz"
===> 10:09:37.325 running "mv /tmp/tmp.RrqFFY1NZJ/flynn-host /usr/local/bin/flynn-host"
===> 10:09:37.341 running "chmod +x /usr/local/bin/flynn-host"
===> 10:09:37.346 downloading Flynn components
10:09:37.520270 host.go:122: tuf: no root keys found in local meta store
===> 10:09:37.524 installing Upstart job
cp: cannot stat ‘/etc/flynn/upstart.conf’: No such file or directory
===> 10:09:37.589 installation complete!
root@i-aoggstf5:~# flynn
flynn: command not found
root@i-aoggstf5:~# flynn-host 
usage: flynn-host [-h|--help] <command> [<args>...]
```

lmars:
@jeffkit was this the first time you ran the script? If not, what was the previous output?

jeffkit:
The last 4 lines of log is the same, no matter the first time or the second time, third time.  After first time, the installation is more quickly.

My OS is ubuntu 14.04. install script is: 
bash < <(curl -fsSL https://dl.flynn.io/install-flynn)

lmars:
@jeffkit do you by any chance have the complete output from the first time you ran the installer? It looks like the first install failed in some way and left your system in an odd state where further attempts will also fail. Seeing the complete output of the first install will help me understand what went wrong.

jeffkit:
[...]

PR #1146, opened 2015-03-03
host: Fix initializing the TUF client

lmars:
Calling `tuf.FileLocalStore` creates the local TUF db, so if a method after that fails then just checking for its's existence isn't enough to determine whether it needs initializing.

Instead we should check for `ErrNoRootKeys` from the TUF client to determine whether it needs initializing.

This was causing installations to fail in #1143 after an initial failure due to a network timeout fetching the remote root.json.


titanous:
LGTM

PR #1153, opened 2015-03-03
Updated flynn install cmd line usage

arekkas:
Instead of spaces, double-hyphen args need to be set with equal signs:

```
  -n <instances>,     --instances=<instances>            Number of instances to launch [default: 1]
  -t <instance-type>, --type=<instance-type>             Type of instances to launch [default: %s]
                      --aws-access-key-id=<key-id>       AWS access key ID. Defaults to $AWS_ACCESS_KEY_ID
                      --aws-secret-access-key=<secret>   AWS access key secret. Defaults to $AWS_SECRET_ACCESS_KEY
                      --aws-region=<region>              AWS region [default: us-east-1]
```

### DIFF
--- a/cli/install.go
+++ b/cli/install.go
@@ -13,17 +13,17 @@ import (
 
 func init() {
 	register("install", runInstaller, fmt.Sprintf(`
-usage: flynn install <target> [-n <instances>] [-t <instance-type>] [--aws-access-key-id <key-id>] [--aws-secret-access-key <secret>] [--aws-region <region>]
+usage: flynn install <target> [-n <instances>] [-t <instance-type>] [--aws-access-key-id=<key-id>] [--aws-secret-access-key=<secret>] [--aws-region=<region>]
 
 Targets:
 	aws  creates a flynn cluster on EC2
 
 Options:
-  -n, --instances <instances>            Number of instances to launch [default: 1]
-  -t, --type <instance-type>             Type of instances to launch [default: %s]
-      --aws-access-key-id <key-id>       AWS access key ID. Defaults to $AWS_ACCESS_KEY_ID
-      --aws-secret-access-key <secret>   AWS access key secret. Defaults to $AWS_SECRET_ACCESS_KEY
-      --aws-region <region>              AWS region [default: us-east-1]
+  -n <instances>,     --instances=<instances>            Number of instances to launch [default: 1]
+  -t <instance-type>, --type=<instance-type>             Type of instances to launch [default: %s]
+                      --aws-access-key-id=<key-id>       AWS access key ID. Defaults to $AWS_ACCESS_KEY_ID
+                      --aws-secret-access-key=<secret>   AWS access key secret. Defaults to $AWS_SECRET_ACCESS_KEY
+                      --aws-region=<region>              AWS region [default: us-east-1]
 
 Examples:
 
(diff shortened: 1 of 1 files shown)


titanous:
Thanks! Please amend the commit message to have a `cli: ` prefix (you can force push to the same branch after amending the commit).

arekkas:
Damn, I used `git commit --amend` and `git push --force` but it seems to heave created a new commit. Should I recreate the PR?

titanous:
The PR looks good (reload the page, GitHub push doesn't remove the old commit). The message needs a bit of work still (it is missing the first character and is too long, take a look at my first comment for suggestions).

arekkas:
Better? :)

titanous:
Yep! LGTM.

arekkas:
Tests failed - my fault?

titanous:
Nope, known issue with CI. @jvatic will test this out locally and merge.

arekkas:
Great!

cli/install.go
```diff
@@ -13,17 +13,17 @@ import (
...
 Targets:
 	aws  creates a flynn cluster on EC2
 
 Options:
-  -n, --instances <instances>            Number of instances to launch [default: 1]
-  -t, --type <instance-type>             Type of instances to launch [default: %s]
-      --aws-access-key-id <key-id>       AWS access key ID. Defaults to $AWS_ACCESS_KEY_ID
-      --aws-secret-access-key <secret>   AWS access key secret. Defaults to $AWS_SECRET_ACCESS_KEY
-      --aws-region <region>              AWS region [default: us-east-1]
+  -n <instances>,     --instances=<instances>            Number of instances to launch [default: 1]
```
jvatic: The parser doesn't like the spaces between the `,` and `--instances`: `--instances=3` causes usage to be printed out.

archSeer: Yeah, docopt requires a single space there, a double space means that the next column is the usage.

arekkas: So instead it should look like:

```
usage: flynn install <target> [-n <instances>] [-t <instance-type>] [--aws-access-key-id=<key-id>] [--aws-secret-access-key=<secret>] [--aws-region=<region>]
Targets:
	aws  creates a flynn cluster on EC2
Options:
  -n <instances>, --instances=<instances>                Number of instances to launch [default: 1]
  -t <instance-type>, --type=<instance-type>             Type of instances to launch [default: %s]
                      --aws-access-key-id=<key-id>       AWS access key ID. Defaults to $AWS_ACCESS_KEY_ID
                      --aws-secret-access-key=<secret>   AWS access key secret. Defaults to $AWS_SECRET_ACCESS_KEY
                      --aws-region=<region>              AWS region [default: us-east-1]
```

?

arekkas:
cli usage for other parts of cli: are formatted with spaces as well. If you think it's useful to update these I'd create PRs on those as well. Here are some examples:

* https://github.com/flynn/flynn/blob/master/cli/cluster.go#L24
* https://github.com/flynn/flynn/blob/master/cli/app.go#L25

Reason I'm pushing this is because it took me about 10 minutes to figure out why my AWS credentials were not recognized and since everything else is set up in 5 minutes (**cudos to that**) it would be sad to waste double the time on this :)

cli/install.go
jvatic: Correct.

titanous:
It looks like this commit is missing the DCO sign-off.

arekkas:
Yep sorry, it should be fine now

titanous:
Thanks!

PR #1154, opened 2015-03-03
docs: Document postgres appliance

titanous:
Describes how to provision and connect to databases as well as the high-level design of the appliance.

### DIFF
--- a/docs/content/postgres.html.md
+++ b/docs/content/postgres.html.md
@@ -1,0 +1,163 @@
+---
+title: PostgreSQL
+layout: docs
+---
+
+# PostgreSQL
+
+The Flynn Postgres appliance provides PostgreSQL 9.4 in a highly-available
+configuration with automatic provisioning. It automatically fails over to
+a synchronous replica with no loss of data if the primary server goes down.
+
+## Usage
+
+### Adding a database to an app
+
+Postgres comes ready to go as soon as you've installed Flynn. After you create
+an app, you can provision a database for your app by running:
+
+```text
+flynn resource add postgres
+```
+
+This will provision a database on the Postgres cluster and configure your
+application to connect to it.
+
+### Connecting to the database
+
+Provisioning the database will add a few environment variables to your app
+release. `PGDATABASE`, `PGUSER`, `PGPASSWORD`, and `PGHOST` provide connection
+details for the database and are used automatically by many Postgres clients.
+
+#### DATABASE_URL
+
+If your application expects a `DATABASE_URL` environment variable, you can
+create one by combining the `PG*` variables:
+
+```text
+flynn env set DATABASE_URL=$(. <(flynn env); echo "postgres://$PGUSER:$PGPASSWORD@$PGHOST:5432/$PGDATABASE")
+```
+
+### Connecting to a console
+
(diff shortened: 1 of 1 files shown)


bgentry:
This looks great.

The main questions I have after reading it are about provisioning replicas. Does `flynn resource add postgres` provision all 3 nodes, or just a single primary? How do I add additional replicas in either sync or async mode, especially to replace nodes which were lost during a previous failover?

titanous:
@bgentry Cool. I'll add more detail to answer those questions.

docs/content/postgres.html.md
```diff
@@ -0,0 +1,154 @@
...
+- The primary has synchronous replication to a single member called the _sync_.
+  Write transactions are not acknowledged to client until they have been added
+  to the sync's transaction log.
+- Replicating from the sync is a daisy chain of one or more _async_ peers, which
+  replicate changes asynchronously from their upstream link in the chain.
+- If possible, the system automatically reconfigures itself after failures to
+  maximize uptime and never lose data.
+
+In the face of an arbitrary failure or maintenance action, the cluster can
+temporarily lose the ability to accept writes and read consistently. Eventually
```
lmars: s/read/reads

titanous:
@bgentry @lmars Comments addressed.

docs/content/postgres.html.md
```diff
@@ -0,0 +1,160 @@
...
+current primary instance.
+
+This design is heavily based on the prior work done by Joyent on the [Manatee
+state machine](https://github.com/joyent/manatee-state-machine).
+
+Flynn comes with a cluster configured with three instances by default. If an
+instance fails, the scheduler will create a new instance and the cluster will be
+reconfigured by the primary without operator intervention. When a user runs
+`flynn resource add postgres`, a new user and database is created on the default
+cluster.
```
bgentry: I think the only issue with this is that it's buried at the end of this doc. Maybe it would benefit from a mention in the "Adding a database to an app" section, where you could say that the command `flynn resource add postgres` adds a highly-available postgres cluster. To me, it wasn't obvious that that was all done with a single simple command (coming from the Heroku world where this stuff takes several steps).

titanous: Okay, I'll note it. Just for clarification: this creates a database on the default cluster, we don't currently have a resource provider that creates a new cluster (that will come later). I do say that in this graf, do you think I should make it more explicit by stating the negative as well somewhere?

bgentry: oooh, I think that's what I was being confused by then. So basically the cluster is already set up and managed as part of Flynn itself, and you can add databases to the cluster using `flynn resource add postgres`. There's no way to add new clusters at all, which does mean there's a limit on how much pressure you can put through your Postgres databases.

Is all that correct? Probably worth noting if so.

titanous: Yeah, there's no command to add clusters. It can be done pretty easily, but I don't want to document the curl/jq hack that I'm using currently. The plan is to resolve this with Flynnfiles that can provision backing services (including dedicated postgres clusters)

titanous: Added a note to the 'adding' section.

lmars:
LGTM

PR #1155, opened 2015-03-03
cli: Allow arbitrary arguments to psql

titanous:
This makes it easier to get machine-readable results out of psql, so that it can be used in shell pipelines, etc.


lmars:
LGTM

PR #1156, opened 2015-03-04
Use less for psql paging

titanous:
`psql` uses `more` by default, most distros/people use `less`. This patch adds environment variables that configure `psql` to use `less` for paging along with some basic, sane `less` configuration defaults.


lmars:
LGTM

PR #1158, opened 2015-03-04
log aggregation: app-level logs, CLI changes and integration tests

bgentry:
**Not ready yet.**

This will be a supplement to #1107 and #1068. This PR makes changes to the `flynn log` command to retrieve the log for an entire app rather than just a single job.

This PR will be the first place we can actually run end-to-end integration tests against the entire new log aggregation system. We won't be able to make these tests pass until the 2 other PRs have been merged into the `staging-log-aggregation` branch. Most of the commits currently shown here are also included in #1107, so the actual diff should be fairly small by the time this is ready for review.

/cc @benburkert 

#### TODO

* [x] Merge #1107 
* [ ] Merge #1068 
* [x] Use SSE in new app log API
* [x] Delete anything related to old job log API
* [x] Add follow support to logaggregator API
* [x] Remove any special filtering / handling of `initLog` messages, just don't ship them around at all
* [x] Add process type filtering to logaggregator API
* [x] Add job ID filtering to logaggregator API
* [x] Don't use `0` to indicate "give me all lines", it should mean 0.
* [x] Use app ID, not app Name, when asking aggreggator for logs


bgentry:
@titanous I'm working on adding follow support right now, and I started wondering about the relationship between the `follow` and `wait` flags. If somebody specifies `follow=true` and `wait=false` together, what's the behavior for a stream where no log messages have been buffered yet? How is that different from `follow=true` with `wait=true`?

Actually, maybe I don't even fully understand how `wait` is supposed to work and why it's necessary.

titanous:
Hmm, I *hink* we can drop wait as long as follow works on a buffer that doesn't exist yet.

bgentry:
Ok. I think the behavior within the logaggregator would have to be the same in either case.. That is, it would have to initialize a buffer for the nonexistent stream in order to subscribe to it.

That could potentially lead to a leak if you're initializing buffers for lots of nonexistent streams, but the controller would prevent you from doing this to nonexistent apps. So it's only a problem if you're checking logs for large numbers of apps, and that's already a scalability issue for a single logaggregator, so nothing new there :)

titanous:
Okay, sounds good.

bgentry:
Progress, I've got some working app-level streaming logs coming in (when run w/ @benburkert's commit from #1068). Tests are still failing because of the display format, though. Here's what that currently looks like:

```
2015-03-05T22:15:33.225168Z app[8cc65d61052e4f838ced2d90fbfbc1bf]: t=2015-03-05T22:15:33+0000 lvl=info msg="registering RPC server" app=containerinit fn=containerInitApp
2015-03-05T22:15:33.225187Z app[8cc65d61052e4f838ced2d90fbfbc1bf]: t=2015-03-05T22:15:33+0000 lvl=info msg="getting stdout pipe" app=containerinit fn=containerInitApp
2015-03-05T22:15:33.225189Z app[8cc65d61052e4f838ced2d90fbfbc1bf]: t=2015-03-05T22:15:33+0000 lvl=info msg="getting stderr pipe" app=containerinit fn=containerInitApp
2015-03-05T22:15:33.225191Z app[8cc65d61052e4f838ced2d90fbfbc1bf]: t=2015-03-05T22:15:33+0000 lvl=info msg="creating stdin pipe" app=containerinit fn=containerInitApp
2015-03-05T22:15:33.225193Z app[8cc65d61052e4f838ced2d90fbfbc1bf]: t=2015-03-05T22:15:33+0000 lvl=info msg="waiting to be resumed" app=containerinit fn=containerInitApp
```

I was attempting to do heroku-like display formatting on log output. I forgot to include the process type, so it was supposed to look like:

```
2015-03-05T22:15:33.225193Z app[web.8cc65d61052e4f838ced2d90fbfbc1bf]: t=2015-03-05T22:15:33+0000 lvl=info msg="waiting to be resumed" app=containerinit fn=containerInitApp
```

Should I give up on that and just print the raw content of the message? Maybe we want to do the formatted output by default, and offer a flag to show raw output?

Or we could offer some other sort of output that includes more of the structured info coming back from the [Message struct](https://github.com/flynn/flynn/blob/log-aggregator-cli-and-integration-tests/logaggregator/client/client.go#L77-L95)

titanous:
> Maybe we want to do the formatted output by default, and offer a flag to show raw output?

That seems reasonable, however the job IDs are pretty huge. Perhaps we should just show the first N characters of the ID by default?

bgentry:
Something like:

```
2015-03-05T22:15:33.225193Z app[web.8cc65d6…]: t=2015-03-05T22:15:33+0000 lvl=info msg="waiting to be resumed" app=containerinit fn=containerInitApp
```

Or without the ellipsis:

```
2015-03-05T22:15:33.225193Z app[web.8cc65d6]: t=2015-03-05T22:15:33+0000 lvl=info msg="waiting to be resumed" app=containerinit fn=containerInitApp
```

That's using 7 chars. I'll proceed with the 7 char no-ellipsis version unless you want something else. I'll also add a `-r / --raw` flag to display raw log message output.

bgentry:
@titanous what do you want done with initLogs? Should they be collected and sent to the log aggregator? Should they be returned in a `flynn log` call by default, or does there need to be a flag & API option to include them? Which FD should they be written to?

bgentry:
I did a CI run with #1068 on top of this and all the CLISuite integration tests are passing https://ci.flynn.io/builds/20150306042139-142e1656

bgentry:
Another CI run that includes #1068: https://ci.flynn.io/builds/20150306195356-5d931e87

The only failure appears to be unrelated. :v:

bgentry:
I've finished up the filtering stuff, though I'm having trouble getting the integration tests to pass again even with #1068. AFAICT the logs aren't arriving at the aggregator anymore even w/ that commit. If I manually send some logs over telnet, I can view and filter those logs using the CLI.

bgentry:
So I believe this should be ready to go, but I can't fully test it until #1068 is ready.

It can be reviewed now, though.

logaggregator/ring/ring.go
```diff
@@ -10,10 +10,13 @@ import (
...
 // clear them; messages merely get moved out of the buffer when they are
 // replaced by new messages.
+//
+// A Buffer also offers the ability to subscribe to new incoming messages.
 type Buffer struct {
+	mu       sync.RWMutex // protects all of the following:
 	messages []*rfc5424.Message
 	start    int
-	mu       sync.RWMutex
+	subs     map[chan *rfc5424.Message]bool
```
titanous: Is the `bool` value used? If not, use `struct{}` as it is zero-width.

titanous:
Overall this looks great! Once the log sender is working, I think we could do with more integration tests covering some of the new features.

logaggregator/api.go
```diff
@@ -73,24 +124,27 @@ func NewMessageFromSyslog(m *rfc5424.Message) client.Message {
...
-	case 3:
-		return "stderr"
-	case 6:
+	switch string(m.MsgID) {
+	case "ID1":
 		return "stdout"
+	case "ID2":
+		return "stderr"
+	case "ID3":
+		return "initLog"
```
titanous: Are we sending the init log to the aggregator? I thought we decided not to.

logaggregator/ring/ring.go
bgentry: nope, I'll change that. Never used a `struct{}` in a map before, only chans :)

logaggregator/api.go
bgentry: We are not supposed to be doing that, no. Looks like I missed deleting this line when I removed the special handling of initLog stuff yesterday. Will nuke it :v:

bgentry:
Awesome. Yeah, I was thinking about some better integration testing on the filtering stuff. It's a bit tough figuring out how to write those without lots of duplication, but it will be a lot easier to try once I can again run integration tests successfully locally :)

titanous:
61b971da95c08194fd0dcbd0e07fcf271ff8904a needs a `controller: `prefix in the commit message.

logaggregator/main_test.go
```diff
@@ -81,11 +86,191 @@ func (s *LogAggregatorTestSuite) TestAggregatorBuffersMessages(c *C) {
...
+		case <-timeout:
+			c.Fatalf("timeout waiting for send on msgc")
+		}
+	}
+	select {
+	case msg := <-msgc:
+		c.Fatalf("unexpected message received: %+v", msg)
+	default:
+	}
+	cancel()
```
bgentry: I added the above ~18 lines to fix a panic I discovered when `s.agg.ReadLastNAndSubscribe()` was called with `n == -1`. This test is getting a bit out of hand now so it needs refactoring before a merge.

bgentry:
there is one CI failure that doesn't appear to be related to anything in this branch: https://ci.flynn.io/builds/20150311013347-14e09007

I'll finish up the remaining integration tests tomorrow now that those are working again, then we should be ready to merge this.

bgentry:
I cleaned up a number of test cases on the aggregator. Still probably need to add a little more integration coverage in the form of additional CLI tests.

bgentry:
OK, I've run the CI on this with #1068 and everything looks good. As soon as @benburkert merges that, we can rebase this branch and merge it as well as long as everything looks good.

bgentry:
Actually, sorry, I just went ahead and rebased this on top of #1068 so I can show some better proof of passing CI.

bgentry:
@titanous just wanted to make sure you knew this was waiting for your sign-off. We'll be working on the logaggregator replication next, but were thinking it would probably make the most sense to get this PR and the [staging-log-aggregation](https://github.com/flynn/flynn/tree/staging-log-aggregation) branch merged into master first.

titanous:
LGTM! Let's :ship: this!

PR #1164, opened 2015-03-04
Support for non-default regions and CIDRs in flynn aws install

### LINKED ISSUE #1162: Allowing specifying the subnet to launch in
`flynn install aws` automatically places instances in 10.0.0.0/21, which conflicts with the address space being used in our office, and so causes some hassle when routing traffic to the cluster over a VPN connection via the office.

If it were possible to specify the subnet CIDR that would be great - I'm quite happy to do so, but having never written any Go in anger, I'm not really sure where to start.

### LINKED ISSUE #1163: Installing to AWS in eu-west-1 fails while configuring DNS
When using flynn install aws at the point of configuring DNS I get the error:

```
Oops, something went wrong: Credential should be scoped to a valid region, not 'eu-west-1'.
```

Googling shows this tends to be due to setting a region other than us-east-1 for services which are global, such as Route53.

jellybob:
I've added `--vpc-cidr` and `--subnet-cidr` options to `flynn aws install` to allow specifying the IP range to use. This resolves #1162

I've also hardcoded the region to `us-east-1` when configuring DNS, since it was previously failing at that point when using other regions. Resolves #1163

### DIFF
--- a/cli/install.go
+++ b/cli/install.go
@@ -13,7 +13,7 @@ import (
 
 func init() {
 	register("install", runInstaller, fmt.Sprintf(`
-usage: flynn install <target> [-n <instances>] [-t <instance-type>] [--aws-access-key-id <key-id>] [--aws-secret-access-key <secret>] [--aws-region <region>]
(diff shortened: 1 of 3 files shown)


jvatic:
LGTM

titanous:
@jellybob Thanks for the contribution! Please add `cli: `prefixes to each of the commit messages and force push to the same branch, then we'll merge.

titanous:
Merged, thanks!

jvatic:
Merged via #1173. Thanks!

PR #1166, opened 2015-03-05
test: Timeout the unit tests after 5 minutes

lmars:
CI is intermittently hanging indefinitely running `go test`. We need to investigate why, but we should also not allow the test to hang indefinitely.


titanous:
LGTM

PR #1178, opened 2015-03-06
controller: Retry requests after dial errors

lmars:
There is a chance that service discovery could return an address of a controller and that instance subsequently goes down, meaning when a HTTP request is actually made, there will be an i/o timeout.

Previously the client would use a connect timeout of 30s (set by `http.DefaultTransport`) which is too long to wait in this case.

This change lowers the connect timeout to 1s, and retries dialling the controller multiple times if an error occurs, giving up after 30s.

### DIFF
--- a/controller/client/client.go
+++ b/controller/client/client.go
@@ -48,10 +48,32 @@ func newClient(key string, url string, http *http.Client) *Client {
 	return c
 }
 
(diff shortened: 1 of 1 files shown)


titanous:
LGTM. If this works out we should move it to the main http client library.

PR #1180, opened 2015-03-06
host/containerinit: Log state changes

lmars:
This is to help debug jobs which remain in the starting state in CI, even though they are running and have output in their logs.

Relates to #907.


titanous:
LGTM

PR #1184, opened 2015-03-06
test: Don't kill the cluster if adding release hosts fails

lmars:


titanous:
LGTM, but please add an explanation to the commit message.

PR #1186, opened 2015-03-06
installer,cli: Add installer web app

jvatic:
screen shot 2015-03-06 at 5 24 47 pm
screen shot 2015-03-06 at 5 24 58 pm
screen shot 2015-03-06 at 5 25 18 pm
screen shot 2015-03-06 at 5 23 26 pm


jvatic:
Also, the dashboard doesn't support being passed the login token (I'll open another pull request for that).

cli/install.go
```diff
@@ -13,7 +13,9 @@ import (
 
 func init() {
 	register("install", runInstaller, fmt.Sprintf(`
-usage: flynn install <target> [-n <instances>] [-t <instance-type>] [--aws-access-key-id=<key-id>] [--aws-secret-access-key=<secret>] [--aws-region=<region>] [--vpc-cidr=<cidr>] [--subnet-cidr=<cidr>]
+usage: flynn install [<target>] [-n <instances>] [-t <instance-type>] [--aws-access-key-id=<key-id>] [--aws-secret-access-key=<secret>] [--aws-region=<region>] [--vpc-cidr=<cidr>] [--subnet-cidr=<cidr>]
+
+Running flynn install without a target will start a web server for the installer app
```
titanous: Delete the CLI interface for installing. The web UI obsoletes it.

titanous:
Any way to avoid duplicating the cert images and code?

jvatic:
@titanous comments addressed

jvatic:
I don't like my current solution for sharing the cert install view as it requires updating the `Gemfile.lock` to propagate changes (i.e. always two commits and delayed feedback as to any dashboard breakage). I think copying the files via the `Tupfile` is be best.

titanous:
Symlinks might also work.

jvatic:
Symlinks don't work on their own as the files they point to become unavailable.

cli/install.go
```diff
@@ -1,135 +1,28 @@
...
-func promptInput(msg string) (result string) {
-	fmt.Print(msg)
-	fmt.Print(": ")
-	for {
-		var answer string
-		fmt.Scanln(&answer)
-		return answer
-	}
+	port := args.String["--port"]
+	return installer.ServeHTTP(port)
```
titanous: This should also run `open` or the OS equivalent on the URL. Look around for a list of the equivalent on Linux.

titanous:
When I open the same URL in another tab while deploying it appears to be pretty broken.

titanous:
Loading the install URL after a successful install prompts me to install the CA certificate again. It should send me to the dashboard.

titanous:
The HTTP server should bind to `127.0.0.1:0` so that the port is unpredictable and available.

jvatic:
The dashboard will need to setup CORS for localhost in order to properly test the certificate before redirect (the other option is setting flag in localStorage indicating the cert has been installed, but I think having a verification step is better). This will also allow the installer login to the dashboard instead of redirecting with the token in the URL. One possibility is for the installer to set an ENV variable with the localhost address on the dashboard app.

jvatic:
@titanous ready for another review pass

installer/http.go
```diff
@@ -0,0 +1,541 @@
...
+		}
+	}
+	s.logger.Info(s.Stack.DashboardLoginMsg())
+}
+
+var httpInstallerPrompts = make(map[string]*httpPrompt)
+var httpInstallerPromtsMux sync.Mutex
+var httpInstallerStacks = make(map[string]*httpInstaller)
+var httpInstallerStackMux sync.Mutex
+var awsEnvCreds aws.CredentialsProvider
```
titanous: I prefer using a struct with fields and handler methods over globals.

installer/http.go
```diff
@@ -0,0 +1,541 @@
...
+
+func openAddr(addr string) error {
+	cmds := []string{
+		"open",
+		"xdg-open",
+		"firefox",
+		"google-chrome",
+	}
+	for _, cmdStr := range cmds {
+		if err := exec.Command("which", cmdStr).Run(); err == nil {
```
titanous: No need call which, just try calling each of them until one succeeds.

installer/http.go
```diff
@@ -0,0 +1,541 @@
...
+			exec.Command(cmdStr, addr).Start()
+			return nil
+		}
+	}
+	return errors.New(fmt.Sprintf("unable to open %s", addr))
+}
+
+func corsHandler(main http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		httphelper.CORSAllowAllHandler(w, r)
```
titanous: This handler should limit requests to the same origin its being served from.

installer/http.go
```diff
@@ -0,0 +1,541 @@
...
+			}
+		}
+	}()
+
+	stream.Wait()
+}
+
+func promptHandler(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
+	id := params.ByName("id")
+	prompt := httpInstallerPrompts[id]
```
titanous: Looks like this needs a mutex lock?

installer/http.go
```diff
@@ -0,0 +1,541 @@
...
+		s := httpInstallerStacks[params.ByName("id")]
+		if s == nil && len(httpInstallerStacks) > 0 {
+			for id := range httpInstallerStacks {
+				s = httpInstallerStacks[id]
+				break
+			}
+		}
+		if s == nil {
+			w.WriteHeader(404)
+		} else {
```
titanous: Return after the 404 instead of using `else`.

titanous:
What's left on building the assets and putting them into the binary?

installer/http.go
jvatic: The last two will not return if the browser isn't open already.

titanous: Okay, then use [`exec.LookPath`](https://golang.org/pkg/os/exec/#LookPath)

jvatic:
@titanous comments addressed

Assets are bundled into `bindata.go` with `go generate` (requires `bundle exec rake compile` to be run in the app directory first).

installer/http.go
```diff
@@ -0,0 +1,566 @@
...
+
+func (s *httpInstaller) YesNoPrompt(msg string) bool {
+	prompt := &httpPrompt{
+		ID:      random.Hex(16),
+		Type:    "yes_no",
+		Message: msg,
+		resChan: make(chan *httpPrompt),
+		api:     s.api,
+	}
+	prompt.api.InstallerPromptsMux.Lock()
```
titanous: s/Mux/Mtx/

jvatic:
@titanous asset compilation via go-bindata is moved into the build process.

jvatic:
@titanous ready for final review

titanous:
LGTM

PR #1188, opened 2015-03-07
appliance/postgresql: Don't error if sync xlog is greater than primary

titanous:
There are some scenarios where this can be the case, and it isn't an error. See https://smartos.org/bugview/MANATEE-263 for details.

This most likely happens when postgres is setting up replication and trying to figure out exactly what it has most recently sent to the remote peer.

Upstream change: https://github.com/joyent/manatee/commit/fb8d0f788000cdc24be5eb7b43a286c19f38fb69


lmars:
LGTM

Issue #235, opened 2014-09-22
Cluster persistence


codykrieger:
I'm manually installing on Ubuntu 14.04 based on the instructions in the readme.

I've got everything set up to the point where I can create and push apps, but I'm a little lost with respect to how to boot up all of the flynn services again once I've rebooted my VM.

I can manually start flynn-host, but when I do that, I only end up with etcd and discoverd jobs, and gitreceived, the router, and all of the other services aren't there:

```
# start flynn-host
flynn-host start/running, process 1917
# flynn-host ps
JOB ID                                            STATE    STARTED         CONTROLLER APP  CONTROLLER TYPE
flynn-discoverd-a775fb62609245abb224dd4fa72263fe  running  24 seconds ago                  
flynn-etcd-0f46161aba704d24a07bfd065702e436       running  24 seconds ago
```

Is there a way to get these services to start again after I've rebooted my VM?

geovanisouza92:
:+1: 

@titanous On Deis, they uses Ceph to produce a distributed blob storage across the cluster. Maybe using Ceph or the blobstore component this can be addressed?

titanous:
@geovanisouza92 the blobstore component currently proxies to Postgres, which is fine. The combination of #534 and the new postgres appliance will resolve this completely.

geovanisouza92:
I hope so! :smile:

codykrieger:
@titanous Now that #534 is done, is this effectively fixed? (I couldn't find a bug representing the new Postgres appliance you were referring to, but the Flynn Gamma announcement said “completely rewriting the Postgres appliance,” so...)

titanous:
Almost, we just need a bit more glue, I'm currently working on resolving this.

codykrieger:
\o/

geovanisouza92:
Will be shipped in next release?

titanous:
@geovanisouza92 It already shipped in `v20150307.0`!

OwlyCode:
So I have my server that went down. When I look at /var/lib/flynn I see the state of the cluster and the .bolt file contains my jobs names.

When I run `start flynn-host`. The controller is still unreachable. When I do a `flynn-host ps` I only see this : 

```
ID                                        STATE    STARTED        CONTROLLER APP  CONTROLLER TYPE
sd74131-b21d0950c5434a75993a3548cc554bcc  running  9 minutes ago                  
sd74131-3fe52a1b7b6f407a89063a61e784f856  running  9 minutes ago                  
[...]

PR #1190, opened 2015-03-07
host: Don't save state database in /tmp

titanous:
The upstart job saves the state database in `/tmp`, which will not persist between reboots. Since the default is in `/var/lib/flynn` we can just remove the flag to fix this.

Refs #235.


lmars:
LGTM

PR #1197, opened 2015-03-08
test: Dump the Docker log if it doesn't come up

lmars:
The timeout was hit in CI (see [here](https://s3.amazonaws.com/flynn-ci-logs/20150308031603-ad79e92d-build-5305600de6dc5095280db6a4fda22d304f20adf3-2015-03-08-03-19-38.txt)), which indicates Docker is failing to start, rather than just being slow to start.


titanous:
LGTM

PR #1209, opened 2015-03-11
host,pkg/cluster: Add SignalJob

### LINKED ISSUE #1196: Add API to signal job
The following code doesn't result in a signal being sent to the job:

```go
clusterClient, err := cluster.NewClient()
if err != nil {
	return err
}
hostClient, err := clusterClient.DialHost(hostID)
if err != nil {
	return err
}
attachClient, err := hostClient.Attach(&host.AttachReq{JobID: jobID}, false)
if err != nil {
	return err
}
defer attachClient.Close()
if err := attachClient.Signal(int(syscall.SIGUSR1)); err != nil {
	return err
}
```

Calling [attachClient.Wait](https://github.com/flynn/flynn/blob/master/pkg/cluster/attach.go#L129-L134) at the start of [attachClient.Signal](https://github.com/flynn/flynn/blob/master/pkg/cluster/attach.go#L220-L228) seems to fix this, we should add a test and determine why this is the case.

lmars:
Signalling a job can be achieved via the attach protocol but only when also requesting logs.

Sending a signal without requesting logs currently fails as the remote end hangs up the connection, but in this case establishing an attach connection is overkill, so I've added a simpler host API method.

I also refactored `pkg/httphelper` along the way.

Fixes #1196.

### DIFF
--- a/host/http.go
+++ b/host/http.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/julienschmidt/httprouter"
@@ -39,6 +40,14 @@ func (h *Host) StopJob(id string) error {
 	default:
 		return errors.New("host: job is already stopped")
 	}
+}
+
+func (h *Host) SignalJob(id string, sig int) error {
+	job := h.state.GetJob(id)
+	if job == nil {
+		return errors.New("host: unknown job")
+	}
+	return h.backend.Signal(id, sig)
 }
 
 func (h *Host) streamEvents(id string, w http.ResponseWriter) error {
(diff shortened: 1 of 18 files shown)


host/http.go
```diff
@@ -128,6 +156,7 @@ func (h *jobAPI) RegisterRoutes(r *httprouter.Router) error {
 	r.GET("/host/jobs", h.ListJobs)
 	r.GET("/host/jobs/:id", h.GetJob)
 	r.DELETE("/host/jobs/:id", h.StopJob)
+	r.POST("/host/jobs/:id/signal", h.SignalJob)
```
titanous: How about `PUT /host/jobs/:id/signals/:signal`?

lmars:
@titanous comments addressed

titanous:
LGTM

Issue #1087, opened 2015-02-25
Distribute CLI using TUF


titanous:
The current [cli distribution tools](https://github.com/flynn/flynn-cli-dist) are from the hk project and kinda hacked together. We should use the TUF tooling we've built out to distribute the updates instead.

PR #1226, opened 2015-03-11
CLI cross-compilation

titanous:
Incremental patch for #1087.

This adds a cross-platform toolchain build to the Makefile and uses it to cross-compile the CLI for linux and darwin (amd64 and 386). Other platforms can be easily added as long as the CLI is compatible with them.

 An added bonus is that the Makefile implicitly causes the correct version of Go to be used to compile everything, as the same toolchain is used everywhere, the system Go is only used to compile gonative.


lmars:
LGTM

PR #1229, opened 2015-03-12
host/containerinit: Bump max open files to 10k

### LINKED ISSUE #1205: Too Many Open Files in Router and Apps
I'm currently receiving `too many open files` errors from the router and both my application after using it for a while. This could occur between a couple hours to a few days depending on server load.  Current temporary fix is to restart router via `virsh` and scale apps 1->0->1.

## Router Log
```
2015/03/10 15:34:53 http: Accept error: accept tcp 0.0.0.0:80: too many open files; retrying in 1s
2015/03/10 15:34:54 http: Accept error: accept tcp 0.0.0.0:80: too many open files; retrying in 1s
2015/03/10 15:34:55 http: Accept error: accept tcp 0.0.0.0:80: too many open files; retrying in 1s
```

## Router Limits
```
...
Limit                     Soft Limit           Hard Limit           Units
Max open files            1024                 4096                 files
...
```

## Sysctl.conf
`fs.file-max = 2097152`

## Ulimits
```
~  ulimit -n
500000
```

titanous:
This is a temporary hard-coded measure until we have a good system for tuning options like this.

Also dropped the max number of processes to 256 to match Heroku, as a sanity measure.

Closes #1205

### DIFF
--- a/host/containerinit/init.go
+++ b/host/containerinit/init.go
@@ -390,7 +390,7 @@ func getCredential(c *Config) (*syscall.Credential, error) {
 	return &syscall.Credential{Uid: uint32(users[0].Uid), Gid: uint32(users[0].Gid)}, nil
 }
 
-func setupCommon(c *Config) error {
+func setupCommon(c *Config, log log15.Logger) error {
 	if err := setupHostname(c); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


host/containerinit/init.go
```diff
@@ -399,6 +399,32 @@ func setupCommon(c *Config) error {
...
+		{syscall.RLIMIT_NOFILE, 10000},
+		// drop max processes to 256
+		{RLIMIT_NPROC, 256},
+	} {
+		if err := syscall.Setrlimit(r.resource, &syscall.Rlimit{Max: r.value, Cur: r.value}); err != nil {
+			log.Error("error setting rlimit", "err", err)
+		}
+	}
+
 	return nil
```
lmars: Could drop the `error` return value, given this method never returns one.

lmars:
LGTM pending minor comment

PR #1233, opened 2015-03-12
controller: Include host ID in job env FLYNN_JOB_ID

lmars:
Outside of flynn-host, job IDs must contain the host ID so that operations like attaching to the job know which host to dial.


titanous:
LGTM

PR #1240, opened 2015-03-12
Log aggregation

### LINKED ISSUE #98: App-wide logs
Application-wide logs, returning a combined log stream of all app containers. Would simplify this usecase:
```
$ flynn log router
controller: not found
$ flynn -a router ps
e5eee5d32f1345ebb6c4de6220fc433f
$ flynn log e5eee5d32f1345ebb6c4de6220fc433f
...
```
This would simply become `flynn log router`

bgentry:
This PR adds a log aggregation component to Flynn.

The flynn-host process is responsible for collecting log lines from running jobs, labeling them with the appropriate metadata, and shipping them via syslog over TCP (RFC6587/RFC5424) to a new logaggregator component. Along with that, it has to buffer a certain amount of messages in case the aggregator is unreachable, or in case a new primary takes over. Lines are split at 10,000 bytes if they happen to exceed that length.

The logaggregator component receives TCP syslog input and organizes messages into buffers by app ID. For each app ID that it sees, it allocates a ring buffer with a 10,000 line capacity. The logaggregator exposes an API that the controller uses to fetch logs for API requests.

The CLI command `flynn log` has been altered to return app-level log messages instead of for only a single job. By default, `flynn log` returns all buffered logs for a given app. The user can limit the number of lines returned with the `-n <lines>` option. They can also filter based on job ID (`-j <job>`) or process type (`-t <type>`). Any of the filters can also be used with the `-f / --follow` option to receive real-time log messages. All such filtering happens at the aggregator itself, and it would be trivial to add more server-side filters in the future. The CLI outputs messages in a Heroku-like format by default, but also has a `--raw-output` option to display only the raw messages.

This PR does not include any replication or failover of logaggregators yet. We think we can still get that done this week, though, because it will involve much of the same work as was done in flynn-host to handle disappearing/changing primary logaggregators.

This also does not include any sort of forwarding or archival features. That sort of thing shouldn't be too tough to add, but will require some additional API surface area. But one of the nice things about this design is that it would be trivial to forward only logs that match a certain set of filters.

Finally, router logs are not shown in an application's log stream at this point, only in the log stream for the `router` app. There might have to be some integration between the router and the aggregator in order to achieve that.

/cc @benburkert 

Closes #98


titanous:
Looks like this needs a rebase on the latest master.

bgentry:
@titanous should have mentioned, @benburkert is working on a rebase and squashing down some of the commits

benburkert:
I've regrouped the commits into four sections: the syslog package, logaggregator, host/logmux package, and the various bootstrap/controller/updater/etc related changes. Grouped together in this order, it should make it easier to squash out the minor commits.

(note: there is a bug in GitHub's commit display, where the series of commits can be displayed out of order (seems to happen when the PR branch is force pushed), so don't trust the ordering on this page)

bgentry:
I'll take another pass at this this morning as I think I can squash away a couple more commits. Anything specific you'd like us to look fix up @titanous ?

titanous:
Nothing specific, just get rid of any commits that are obviously fixups.

bgentry:
OK, I've gotten rid of all of the obvious ones I could fine and dropped it from 30 -> 25 commits.

bgentry:
oh, crap, I didn't even think about @jvatic needing to update the dashboard to support the new logs. Do we need to block this on it? Apparently there aren't any tests that cover that functionality or we would have noticed it.

titanous:
He'll do it today, so no need to block on it. We can hold the binary release until the dashboard changes land in master.

bgentry:
Cool. Waiting for sign-off or feedback then :v:

titanous:
LGTM. Ship it! Awesome work!

bgentry:
:+1: :beers: :fireworks: :boom:

PR #1249, opened 2015-03-13
test: Fix tracking VMs

lmars:
`Cluster.Boot` calls `DumpLogs` if bootstrapping fails, so the booted instances need to be added to `Cluster.defaultInstances` as soon as possible so that they are available for grabbing logs (and so that `DumpLogs` does not panic accessing `Cluster.defaultInstances[0]`)


titanous:
LGTM

PR #1261, opened 2015-03-14
test: Remove unnecessary scale after git push

lmars:
The receiver already adds a `web=1` formation, so the scale command after the push always exits with `requested scale equals current scale, nothing to do!`.


titanous:
LGTM

titanous:
Adding the resource after deploying will result in the initial process crashing, is this intentional?

lmars:
@titanous hmm yes, it's not really intentional but the app gets deployed when adding the resource so the test will pass. If we don't want to deploy a crashed process though I'll probably just leave the code as it is.

titanous:
It's probably fine, but worth adding a comment explaining.

Issue #1264, opened 2015-03-14
Postgres connection leak


titanous:
```text
FATAL:  remaining connection slots are reserved for non-replication superuser connections
```

https://s3.amazonaws.com/flynn-ci-logs/20150314172535-17cdc78e-build-5fd0d5fac00f1baec4124e602b88a3e030e8fe49-2015-03-14-17-41-36.txt

lmars:
This was fixed in #1537 by not opening a DB connection for client stream request.

PR #1265, opened 2015-03-14
appliance/postgresql: Log client connections

titanous:
This is useful when debugging connection leaks.

Will help debug #1264.


lmars:
LGTM

PR #1266, opened 2015-03-14
all: Add Bats for script unit testing

lmars:
I am making this pull request before adding tests so that `bats` is available in CI.


titanous:
LGTM

PR #1267, opened 2015-03-14
logaggregator: register one discoverd service

benburkert:
The logaggregator server listens on two ports, but only registers a single discoverd service for the syslog port. The api port should be inferred by the client.


titanous:
LGTM

PR #1250, opened 2015-03-13
CLI updates with TUF

titanous:
This removes the old distribution toolchain inherited from hk and replaces it with an updater that uses TUF. The same TUF repo that is used to distribute the rest of the container images and binaries is used.

`script/export-components` is updated to add the cross-compiled flynn CLI binaries to the repo tagged with their version. The CLI will be released and versioned as part of the nightly build process instead of using a separate version tag and build system.

Closes #1087 

**TODO**

- [x] Set `User-Agent` for flynn-host and flynn-cli TUF clients
- [ ] Implement redirect server and replace CLI installation snippets in docs

PR #1273, opened 2015-03-14
all: Use new redirect endpoint for CLI download

titanous:
The `cli.flynn.io` server is deprecated and will serve one last update, the first release of the CLI with the TUF updater. A new redirect endpoint is provisioned at `https://dl.flynn.io/cli` that serves the same function and reads from the TUF repo.

Refs #1250

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -12,7 +12,7 @@ available.
 To install a pre-built binary release, run the following one-liner:
 
 ```bash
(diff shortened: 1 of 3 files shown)


lmars:
LGTM

titanous:
The deprecated server has been updated. I confirmed that the update worked from `v20150305 -> v20150314.1`. I have also confirmed that the new snippet works on linux and darwin amd64.

PR #1276, opened 2015-03-15
Revert "test: Remove unnecessary scale after git push"

lmars:
`TestGoBuildpack` started failing intermittently, and I think it is related to the fact that the inital release crashes. This change is not critical so I am reverting and will open an issue to investigate any issues with deploying a crashed release.

Reverts flynn/flynn#1261.


titanous:
LGTM

PR #1277, opened 2015-03-15
slugrunner: Fix command quoting

### LINKED ISSUE #656: slugrunner run arguments are parsed incorrectly
```
$ flynn run bash -c 'echo foo'
```

Only returns a single newline, the quoting is incorrect by the time it execs. I have verified that the argument quoting is correct all the way through to slugrunner. The best bet is probably to rewrite slugrunner in Go after we land #145.

titanous:
This fixes running commands with simple quoting and spaces like:

    flynn run bash -c "echo 'foo bar'"

I have not tested more advanced cases, as I suspect that this is good enough for the majority of use cases (even this does much quoting does not work on Heroku).

The leading space in the `printf` format string is necessary for some unknown reason that I didn't track down.

The test is in the git deploy suite because it requires an SSH key to be set up and we don't use git anywhere else in the test suite.

Closes #656

### DIFF
--- a/slugrunner/runner/init
+++ b/slugrunner/runner/init
@@ -51,7 +51,7 @@ case "$1" in
     fi
     ;;
   *)
-    command="$@"
+    printf -v command " %q" "$@"
     ;;
 esac
 
--- a/test/test_git_deploy.go
+++ b/test/test_git_deploy.go
@@ -160,3 +160,13 @@ func (s *GitDeploySuite) runBuildpackTest(t *c.C, name string, resources []strin
 
 	t.Assert(r.flynn("scale", "web=0"), Succeeds)
 }
+
+func (s *GitDeploySuite) TestRunQuoting(t *c.C) {
+	r := s.newGitRepo(t, "empty")
+	t.Assert(r.flynn("create"), Succeeds)
+	t.Assert(r.git("push", "flynn", "master"), Succeeds)
+
+	run := r.flynn("run", "bash", "-c", "echo 'foo bar'")
+	t.Assert(run, Succeeds)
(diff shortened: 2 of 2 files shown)


lmars:
@titanous 

> The leading space in the printf format string is necessary for some unknown reason that I didn't track down.

`"$@"` expands to `"$1" "$2" ...`, so with two args `echo 'foo bar'`, `printf " %q" "$@"` is expanded to `printf " %q" echo 'foo bar'` which prefixes a space to each argument:

```
$ bash -xc 'bash -c "$(printf " %q" "$@")"' -- echo 'foo bar'
++ printf ' %q' echo 'foo bar'
+ bash -c ' echo foo\ bar'
foo bar
```

Removing the leading space will just join the args together:

```
$ bash -xc 'bash -c "$(printf "%q" "$@")"' -- echo 'foo bar'
++ printf %q echo 'foo bar'
+ bash -c 'echofoo\ bar'
bash: echofoo bar: command not found
```

slugrunner/runner/init
```diff
@@ -49,13 +50,9 @@ case "$1" in
         puts (YAML.load_file('.release')['default_process_types'] || {})['$2']"
       )"
     fi
+    exec setuidgid nobody bash -c "${command}"
     ;;
   *)
-    command="$@"
+    exec setuidgid nobody bash -c "$(printf " %q" "$@")"
```
lmars: `printf` supports assigning variables with the `-v` flag, so this could be:

```
printf -v command " %q" "$@"
```

titanous:
@lmars Thanks! Comments addressed.

lmars:
LGTM

PR #1280, opened 2015-03-15
host: Don’t include env vars in debug info by default

### LINKED ISSUE #954: ENV secrets filter for upload-debug-info
If we have to share debug info it would be handy to have a reference / option to change how 'secret' `ENV` values are handled.

One thought would be to exclude all non-flynn related `ENV` values in the debug dump.

Or alternatively allow for a convention of a prefix/postfix we could use with our `ENV` keys (e.g. `SECRET_XYZ`

titanous:
There are many pieces of sensitive information that we shouldn’t collect that are in the environment. Also added a flag to `flynn-host inspect` to omit the environment.

Closes #954

### DIFF
--- a/host/cli/collect-debug-info.go
+++ b/host/cli/collect-debug-info.go
@@ -32,10 +32,11 @@ var debugCmds = [][]string{
 
 func init() {
 	Register("collect-debug-info", runCollectDebugInfo, `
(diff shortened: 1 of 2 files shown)


lmars:
LGTM

PR #1281, opened 2015-03-15
host/cli: Collect iptables rules as part of debug info

### LINKED ISSUE #719: upload-debug-info should include iptables info
We should dump the full iptables rules list, some issues are caused by conflicting manually created rules.

titanous:
Closes #719


lmars:
LGTM

PR #1282, opened 2015-03-15
util/release: Move manifest types into separate package

titanous:
The installer (and other programs in the future) want to read these manifests. Currently the installer imports the main release package, which is brittle and adds 1-2MB to the binary size.

This patch moves the manifest types into a separate release package to make them easy to work with. No code is changed, just moved.


lmars:
LGTM

PR #1283, opened 2015-03-15
host/cli: Error when creating discovery token for single node

### LINKED ISSUE #1262: Error on --init-discovery=1
Using a discovery token for a single node doesn't make sense and can result in confusing etcd breakage.

titanous:
This is unnecessary, and can result in etcd failure when restarting the host.

Closes #1262


lmars:
LGTM

PR #1284, opened 2015-03-16
router: Show useful error when listen fails

### LINKED ISSUE #1149: Router should report port collisions better
Currently when the router is unable to bind to a port, it prints a generic error that does not reference the port. It should instead print the port it was trying to bind to, and suggest finding and stopping any services listening on that port.

titanous:
Errors from listen don't include the bound address, so wrap it in an error that prints the address along with an error.

Closes #1149


lmars:
LGTM

PR #1287, opened 2015-03-16
host: Store logs in /var/log/flynn rather than /tmp

lmars:
`/tmp` is commonly a tmpfs so rebooting a node can lose logs which are helpful for debugging purposes.


titanous:
This requires a change to CI: https://github.com/flynn/flynn/blob/c0a5adc07ae7cdc1cab2f9712405d1b0de941ce8/test/cluster/cluster.go#L630

lmars:
@titanous comments addressed

titanous:
LGTM

PR #1289, opened 2015-03-16
slugbuilder: Fix parsing release files with no default types

lmars:
For example the apt buildpack makes a release file containing `---` which is parsed to `nil` by Ruby:

https://github.com/ddollar/heroku-buildpack-apt/blob/master/bin/release


titanous:
LGTM

lmars:
@titanous updated the slugrunner too (the initial changes were to the slugbuilder)

titanous:
LGTM++

Issue #1170, opened 2015-03-05
Refactor layer 0 bootstrap


titanous:
This issue describes a series of steps that will refactor the layer 0 bootstrap system, with the following results:

- Layer 0 bootstrap will be only used when starting a cluster for the first time.
- Node additions and reboots will not run bootstrap, the layer 1 scheduler will be responsible for starting services on them.
- Layer 0 services will be registered in the controller by layer 1 bootstrap, allowing management/updates via the controller API.
- Networking and the default environment variables (containing the discoverd address) will be configured via API endpoints.
- Core services will be resurrected with their existing data volumes if the whole cluster is simultaneously  rebooted.

**Steps:**

- [ ] Change layer 0 bootstrap format to match the layer 1 bootstrap JSON.
- [ ] Generate a consistent identifier for each artifact/release based on a hash of the JSON and annotate the job with controller info.
- [ ] Read layer 0 jobs from hosts in layer 1 bootstrap and add them to the controller.
- [ ] Add API to flynn-host for configuring networking and change the flannel container to use it.
- [ ] Add API to flynn-host for configuring default environment variables and change the discoverd container to use it.
- [ ] Change flynn-host to persist cluster details after first bootstrap starts a working scheduler. (TODO: how to signal that a scheduler is initialized?)
- [x] Add concept of 'resurrection' to jobs on flynn-host, resurrect jobs instead of bootstrapping if there is persisted cluster state. (#1187)
- [ ] Change resurrection to try to contact discoverd on any other hosts. If can contact, register host and wait for scheduler to send jobs over.
- [ ] Update flynn-host to reconnect to local discoverd when it comes up after the scheduler starts it.
- [ ] Change flynn-host to not run jobs requiring container networking until it is configured. (this will later be handled by tag-based scheduling)

/cc @lmars am I missing anything major here?

lmars:
@titanous SGTM

titanous:
Superceded by #1373.

PR #1291, opened 2015-03-16
Use start script to pass arguments to etcd

titanous:
This moves the argument building logic out of the bootstrap manifest and into the container image, reducing the knowledge of etcd in the manifest.

Incremental patch for #1170.


lmars:
LGTM

PR #1292, opened 2015-03-17
discoverd,host: Move args to start script

titanous:
Move the configuration flags for discoverd from the layer 0 bootstrap manifest to the start script. The start script now waits for the bridge to appear and parses the IP address from it.

This removes knowledge of the flags used by discoverd from the layer 0 manifest and allows the same environment to be used for the discoverd instance on each node.

Incremental patch for #1170.


lmars:
LGTM

PR #1293, opened 2015-03-17
logaggregator: buffer snapshots

benburkert:
Implement snapshotting to disk as a stepping stone to full replication from leader to follower(s). The snapshot format & internal mechanism for dumping/loading can be reused for part of replication.


titanous:
As discussed yesterday, it looks like flatbuffers are a good fit, as the deserialization cost is near-zero.

benburkert:
> As discussed yesterday, it looks like flatbuffers are a good fit, as the deserialization cost is near-zero.

I actually went the other direction and swapped out the protobufs code for the `encoding/gob` package. It should be performant enough for the time being, and the serialization format is not used by anything outside logaggregator, so it should be easy to replace `gob` when/if that's needed.

I need to squash this down, do a bit of cleanup, and manually test this some more. But it's ready for :eyes:

titanous:
LGTM!

PR #1299, opened 2015-03-17
installer: Fix input data

### LINKED ISSUE #1297: Installer only boots one instance
I requested a three instance cluster with the installer, but a one instance cluster booted.

jvatic:
Fields for instance type and num instances were camelCased instead of snake_cased.

fixes #1297


titanous:
LGTM

PR #1301, opened 2015-03-18
deployer: Support deploying postgres

lmars:
To minimise the impact of deploying postgres (i.e. to minimise the time in read-only mode), I have added a new deploy strategy which stops postgres jobs in a particular order, waiting for specific service discovery events before moving on.

To summarise the strategy:

* While the head async is from the old release, replace it
* If sync is present, replace it
* Replace the primary

Replacing a job involves:

* calling `Stop` via the API to stop postgres and unregister the service
* waiting for a service down event
* scaling up the new release (the new job will either become the sync or be appended to the asyncs)
* waiting for the new job to set `PG_ONLINE=true` in service discovery (which happens once it has run `pg_basebackup`)

Ideally this would just be a one-by-one deployment with the scheduler being flexible enough to be able to stop jobs in a particular order, and the deployer's job waiting flexible enough to wait for specific instance metadata before declaring a job is up, but this will need to wait until the scheduler is refactored.

### DIFF
--- a/appliance/postgresql/client/client.go
+++ b/appliance/postgresql/client/client.go
@@ -57,3 +57,7 @@ func (c *Client) Status() (*Status, error) {
 	res := &Status{}
 	return res, c.c.Get("/status", res)
 }
+
+func (c *Client) Stop() error {
+	return c.c.Post("/stop", nil, nil)
+}
--- a/appliance/postgresql/client/client.go
+++ b/appliance/postgresql/client/client.go
@@ -17,6 +17,7 @@ type PostgresInfo struct {
 	Running          bool                `json:"running"`
 	SyncedDownstream *discoverd.Instance `json:"synced_downstream"`
 	XLog             string              `json:"xlog,omitempty"`
+	UserExists       bool                `json:"user_exists,omitempty"`
 	Replicas         []*Replica          `json:"replicas,omitempty"`
 }
 
--- a/appliance/postgresql/client/client.go
+++ b/appliance/postgresql/client/client.go
@@ -19,6 +19,7 @@ type PostgresInfo struct {
 	XLog             string              `json:"xlog,omitempty"`
 	UserExists       bool                `json:"user_exists,omitempty"`
 	Replicas         []*Replica          `json:"replicas,omitempty"`
+	ReadWrite        bool                `json:"read_write,omitempty"`
 }
 
 type Replica struct {
@@ -67,17 +68,31 @@ func (c *Client) Stop() error {
 }
 
 func (c *Client) WaitForReplSync(downstream *discoverd.Instance, timeout time.Duration) error {
+	return c.waitFor(func(status *Status) bool {
+		return status.Postgres.SyncedDownstream != nil && status.Postgres.SyncedDownstream.ID == downstream.ID
+	}, timeout)
+}
+
+func (c *Client) WaitForReadWrite(timeout time.Duration) error {
+	return c.waitFor(func(status *Status) bool {
+		return status.Postgres.ReadWrite
+	}, timeout)
+}
+
+var ErrTimeout = errors.New("timeout waiting for expected status")
+
+func (c *Client) waitFor(expected func(*Status) bool, timeout time.Duration) error {
 	start := time.Now()
 	for {
 		status, err := c.Status()
 		if err != nil {
 			return err
-		} else if status.Postgres.SyncedDownstream != nil && status.Postgres.SyncedDownstream.ID == downstream.ID {
+		} else if expected(status) {
 			return nil
 		}
 		time.Sleep(100 * time.Millisecond)
 		if time.Now().Sub(start) > timeout {
-			return errors.New("timed out waiting for replication sync")
+			return ErrTimeout
 		}
 	}
 }
--- a/appliance/postgresql/client/client.go
+++ b/appliance/postgresql/client/client.go
@@ -1,20 +1,23 @@
 package pgmanager
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
 	"time"
 
 	"github.com/flynn/flynn/appliance/postgresql/state"
+	"github.com/flynn/flynn/discoverd/client"
 	"github.com/flynn/flynn/pkg/httpclient"
 )
 
 type PostgresInfo struct {
-	Config   *state.PgConfig `json:"config"`
-	Running  bool            `json:"running"`
-	XLog     string          `json:"xlog,omitempty"`
-	Replicas []*Replica      `json:"replicas,omitempty"`
+	Config           *state.PgConfig     `json:"config"`
+	Running          bool                `json:"running"`
+	SyncedDownstream *discoverd.Instance `json:"synced_downstream"`
+	XLog             string              `json:"xlog,omitempty"`
+	Replicas         []*Replica          `json:"replicas,omitempty"`
 }
 
 type Replica struct {
@@ -60,4 +63,20 @@ func (c *Client) Status() (*Status, error) {
 
 func (c *Client) Stop() error {
 	return c.c.Post("/stop", nil, nil)
+}
+
+func (c *Client) WaitForReplSync(downstream *discoverd.Instance, timeout time.Duration) error {
+	start := time.Now()
+	for {
+		status, err := c.Status()
+		if err != nil {
+			return err
+		} else if status.Postgres.SyncedDownstream != nil && status.Postgres.SyncedDownstream.ID == downstream.ID {
+			return nil
+		}
+		time.Sleep(100 * time.Millisecond)
+		if time.Now().Sub(start) > timeout {
+			return errors.New("timed out waiting for replication sync")
+		}
+	}
 }
--- a/appliance/postgresql/postgres.go
+++ b/appliance/postgresql/postgres.go
@@ -541,11 +541,6 @@ func (p *Postgres) stop() error {
 		}
 	}
 
-	// Wait for all clients to terminate before shutting down cleanly
-	if tryExit(syscall.SIGTERM) {
-		return nil
-	}
-
 	// Forcefully disconnect all clients and shut down cleanly
 	if tryExit(syscall.SIGINT) {
 		return nil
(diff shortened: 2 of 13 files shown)


discoverd/client/heartbeat.go
```diff
@@ -111,6 +112,16 @@ func (h *heartbeater) SetMeta(meta map[string]string) error {
...
 }
 
+func (h *heartbeater) UpdateMeta(key, val string) error {
+	h.Lock()
+	defer h.Unlock()
+	if h.inst.Meta == nil {
+		h.inst.Meta = make(map[string]string, 1)
+	}
+	h.inst.Meta[key] = val
+	return h.c.c.Put(fmt.Sprintf("/services/%s/instances/%s", h.service, h.inst.ID), h.inst, nil)
```
titanous: Looks like this can use `SetMeta`?

appliance/postgresql/state/state.go
```diff
@@ -211,15 +213,18 @@ type Peer struct {
 	workDoneCh  chan struct{}
 	retryCh     chan struct{}
 	stopCh      chan struct{}
+
+	closed bool
```
titanous: Use a `sync.Once` instead of a `bool`.

controller/deployer/strategies/postgres.go
```diff
@@ -0,0 +1,171 @@
...
+	if state.Sync != nil {
+		numProcs++
+	}
+	if numProcs != d.Processes["postgres"] {
+		return loggedErr("pg cluster state does not match expected processes")
+	}
+	if d.newReleaseState["postgres"] > 0 {
+		return loggedErr("pg cluster in unexpected state")
+	}
+
```
titanous: This should probably bail if it is in singleton mode.

controller/deployer/strategies/postgres.go
```diff
@@ -0,0 +1,171 @@
...
+			case event := <-d.serviceEvents:
+				if event.Kind == discoverd.EventKindDown && event.Instance.ID == inst.ID {
+					d.deployEvents <- ct.DeploymentEvent{
+						ReleaseID: d.OldReleaseID,
+						JobState:  "down",
+						JobType:   "postgres",
+					}
+					break loop
+				}
+			case <-time.After(30 * time.Second):
```
titanous: Bump this up to a minute, it can take a while to shut down in some cases.

controller/deployer/strategies/postgres.go
```diff
@@ -0,0 +1,171 @@
...
+					event.Instance.Meta["FLYNN_PROCESS_TYPE"] == "postgres" &&
+					event.Instance.Meta["PG_ONLINE"] == "true" {
+					d.deployEvents <- ct.DeploymentEvent{
+						ReleaseID: d.NewReleaseID,
+						JobState:  "up",
+						JobType:   "postgres",
+					}
+					return nil
+				}
+			case <-time.After(30 * time.Second):
```
titanous: This should probably be bumped up to a few minutes, starting can take a while.

titanous:
> waiting for the new job to set `PG_ONLINE=true` in service discovery (which happens once it has run `pg_basebackup`)

I don't think this is correct. The wait should be for the replication to catch up (sent xlog == flushed xlog), which happens later.

lmars:
@titanous 

>> waiting for the new job to set PG_ONLINE=true in service discovery (which happens once it has run pg_basebackup)

> I don't think this is correct. The wait should be for the replication to catch up (sent xlog == flushed xlog), which happens later.

Yes I forgot to clarify this. Currently only the Primary node waits for the downstream to catch up, so in the case of replacing asyncs there is no indication of when they catch up to their upstreams.

It seems there are two options:

1) Continue to set `PG_ONLINE=true` after `pg_basebackup` has finished on asyncs, but only once replication has caught up on syncs (which requires the primary notifying the sync that it has caught up)
2) Always set `PG_ONLINE=true` after replication has caught up (which requires all nodes to call `waitForSync` and notify their downstreams)

Which do you think is the best approach?

titanous:
There is a third option, which seems like it might be better: use the new status endpoint to query the replication status from the upstream (see `waitForUpstream` for an example of this). That way the upstream doesn't need to do any notification.

I wonder if we should just do the check from the deployer since nothing else in the appliance cares about the flag? (with the plan to move the deploy strategy into the appliance in the future)

lmars:
@titanous I have added `pgmanager.WaitForReplSync` which waits for the postgres manager to set `PostgresInfo.SyncedInst` (see 4ef1001), WDYT?

titanous:
Yeah, that looks good. I'll do a more thorough review in a bit.

lmars:
@titanous the test pass for me locally but intermittently failing in CI.

It seems the asyncs are restarting postgres whenever they see a new downstream, which isn't necessary, and I thought I had that handled [here](https://github.com/flynn/flynn/blob/deploy-postgres/appliance/postgresql/postgres.go#L276-L282).

Could you take a look at commit 2d0dc74 to see if there are any holes in the logic?

appliance/postgresql/postgres.go
```diff
@@ -253,14 +264,23 @@ func (p *Postgres) reconfigure(config *state.PgConfig) (err error) {
...
+	p.setSyncedInst(nil)
 
 	// If we're already running and this is only a sync change, we just need to update the config.
 	if p.running() && p.config() != nil && config != nil && p.config().Role == state.RolePrimary && config.Role == state.RolePrimary {
 		return p.updateSync(config.Downstream)
 	}
 
+	// If we're already running and this is only a downstream change, just wait for the new downstream to catch up
+	if p.running() && p.config() != nil && config != nil &&
+		p.config().Role == config.Role && p.config().Upstream == config.Upstream &&
```
titanous: Both sides of the second check on this line should be `Upstream.ID`.

lmars:
@titanous the integration tests now pass (in custom CI, requires 895a3d4), but the unit tests are broken due to f1ba89e. Can you give this a pass before I start fixing up the unit tests?

appliance/postgresql/client/client.go
```diff
@@ -1,20 +1,25 @@
...
 )
 
 type PostgresInfo struct {
-	Config   *state.PgConfig `json:"config"`
-	Running  bool            `json:"running"`
-	XLog     string          `json:"xlog,omitempty"`
-	Replicas []*Replica      `json:"replicas,omitempty"`
+	Config     *state.PgConfig     `json:"config"`
+	Running    bool                `json:"running"`
+	SyncedInst *discoverd.Instance `json:"synced_inst"`
```
titanous: I'm not a huge fan of the `SyncedInst` field name. Is `SyncedDownstream` accurate? I don't have any great ideas.

appliance/postgresql/postgres.go
```diff
@@ -660,15 +728,18 @@ func (p *Postgres) waitForSync(inst *discoverd.Instance) {
...
+			if err := p.writeConfig(configData{Sync: inst.ID}); err != nil {
+				log.Error("error writing postgres.conf", "err", err)
+				return
+			}
 
-		if err := p.sighup(); err != nil {
-			log.Error("error calling sighup")
-			return
+			if err := p.sighup(); err != nil {
+				log.Error("error calling sighup")
```
titanous: Error needs logging?

titanous:
Overall this looks good! I'm curious about what the cause of the unit test failures is.

test/test_postgres.go
```diff
@@ -38,3 +46,275 @@ func (s *PostgresSuite) TestDumpRestore(t *c.C) {
...
+		case <-time.After(30 * time.Second):
+			t.Fatal("timed out waiting for postgres formation")
+		}
+	}
+
+	// connect to the db so we can test writes
+	db := postgres.Wait(d.name, fmt.Sprintf("dbname=postgres user=flynn password=%s", release.Env["PGPASSWORD"]))
+	dbname := "deploy-test"
+	t.Assert(db.Exec(fmt.Sprintf(`CREATE DATABASE "%s" WITH OWNER = "flynn"`, dbname)), c.IsNil)
+	db, err = postgres.Open(d.name, fmt.Sprintf("dbname=%s user=flynn password=%s", dbname, release.Env["PGPASSWORD"]))
```
titanous: `db` needs a Close.

lmars:
@titanous the issue with the unit tests is that the peer transitions to being at rest before it has applied the configuration because it will now evaluate the cluster state after receiving service events (whereas previously it would always apply config first). I tried not being at rest at the end of state evaluation, but things broke in other ways. Thoughts?

titanous:
This fixes most of the issues: 

```diff
diff --git a/appliance/postgresql/state/state.go b/appliance/postgresql/state/state.go
index a314703..07e81b7 100644
--- a/appliance/postgresql/state/state.go
+++ b/appliance/postgresql/state/state.go
@@ -270,6 +270,8 @@ func (p *Peer) Run() {
 		case <-p.evalStateCh:
 			p.evalClusterState()
 			continue
+		case <-p.applyConfCh:
+			p.pgApplyConfig()
 		case <-p.stopCh:
 			return
 		default:
```

(previously there was a race with the rest channel)

There are a still a bunch of issues, I'm poking at it now but might not have it figured out before you get up.

titanous:
Ah, missed a `continue`. This fixes the unit tests:

```diff
diff --git a/appliance/postgresql/state/state.go b/appliance/postgresql/state/state.go
index a314703..1c879e7 100644
--- a/appliance/postgresql/state/state.go
+++ b/appliance/postgresql/state/state.go
@@ -270,6 +270,9 @@ func (p *Peer) Run() {
 		case <-p.evalStateCh:
 			p.evalClusterState()
 			continue
+		case <-p.applyConfCh:
+			p.pgApplyConfig()
+			continue
 		case <-p.stopCh:
 			return
 		default:
```

lmars:
@titanous comments addressed, I have run this a few times through CI and the postgres tests passed each time.

appliance/postgresql/client/client.go
```diff
@@ -1,20 +1,25 @@
...
 )
 
 type PostgresInfo struct {
-	Config   *state.PgConfig `json:"config"`
-	Running  bool            `json:"running"`
-	XLog     string          `json:"xlog,omitempty"`
-	Replicas []*Replica      `json:"replicas,omitempty"`
+	Config           *state.PgConfig     `json:"config"`
+	Running          bool                `json:"running"`
+	SyncedDownstream *discoverd.Instance `json:"synced_inst"`
```
titanous: The JSON tag for this field hasn't been updated.

lmars:
@titanous comments addressed

titanous:
LGTM

Issue #975, opened 2015-02-10
demo: Update Vagrantfile for new installation workflow


lmars:
The installation and update procedure now uses a TUF repository (see #974).

PR #1304, opened 2015-03-18
demo: Add Makefile to simplify steps and self document

simple10:
Signed-off-by: Joe Johnston <joe@simple10.com>

This will hopefully make it a lot easier for people to get quickly up and running with flynn.

It also fixes https://github.com/flynn/flynn/issues/975

### DIFF
--- a/demo/Makefile
+++ b/demo/Makefile
@@ -1,0 +1,86 @@
+
+# Providers: virtualbox, vmware_fusion, aws
+PROVIDER := virtualbox
+
+# Domain to use when starting flynn
+CLUSTER_DOMAIN := demo.localflynn.com
+CLUSTER := demo-${PROVIDER}
+
+# Location to install flynn CLI bin
+CLI_INSTALL_BIN := /usr/local/bin/flynn
+
+# ENV to set when running vagrant
+# See https://github.com/mitchellh/vagrant/issues/4367 for use of SSH_AUTH_SOCK=''
+VAGRANT_ENV := CLUSTER_DOMAIN="${CLUSTER_DOMAIN}" SSH_AUTH_SOCK="" FLAG_BOOTSTRAP=false
+
+
+.PHONY: init install-cli bootstrap up down halt update ssh reset token destroy dashboard ssh-cmd
+
+
+# Initialize a new cluster
+init: update
+	@$(MAKE) up
+	@$(MAKE) bootstrap
+	flynn -c ${CLUSTER} key add
+	@printf "\nFlynn is now running in the VM.\nOpen dashboard: make dashboard\n\n"
+
+# Install flynn cli on host
+install-cli:
+	curl -sL -A "`uname -sp`" https://dl.flynn.io/cli | zcat >${CLI_INSTALL_BIN}
+	chmod +x ${CLI_INSTALL_BIN}
+
+# Bootstrap the cluster
+# This should only be called once per cluster
+# Runs bootstrap in VM and cluster add on host
+bootstrap:
+	@$(MAKE) CMD='CLUSTER_DOMAIN="${CLUSTER_DOMAIN}" flynn-host bootstrap /etc/flynn/bootstrap-manifest.json 2>&1' ssh-cmd | tee bootstrap.log
+	@(grep '^flynn cluster add .*$$' bootstrap.log || echo exit 1) | sed "s/ default / ${CLUSTER} /" | sed "s/$$(printf '\r')\$$/ --force --default/" | bash
+	@rm bootstrap.log
+
+# Start the VM and flynn
+up:
+	${VAGRANT_ENV} vagrant up --provider ${PROVIDER}
+	@$(MAKE) CMD='sudo start flynn-host || echo flynn-host already running or unable to start' ssh-cmd
+
+# Stop the VM
+down:
+	${VAGRANT_ENV} vagrant halt
+
+halt: down
+
+# Download the latest flynn VM image
+# This does not update already initialized VMs
+update:
+	${VAGRANT_ENV} vagrant box update
+
+# SSH into VM
+ssh:
+	${VAGRANT_ENV} vagrant ssh
+
+# Destroy and rebuilt the VM
+reset:
+	@$(MAKE) destroy
+	@$(MAKE) update
+	@$(MAKE) init
+
+# Destroy VM
+# Run this to delete the VM and free up disk space
+destroy:
+	${VAGRANT_ENV} vagrant destroy -f
+	flynn cluster remove ${CLUSTER}
+
+# Get token for API and dashboard
+token:
+	@flynn -a dashboard -c ${CLUSTER} env get LOGIN_TOKEN
+
+# Open dashboard in browser
+dashboard:
+	@echo Login token:
+	@$(MAKE) token
+	@sleep 1 # Allow user to see token output before browser opens
+	@open http://dashboard.${CLUSTER_DOMAIN}
+
+# Run a command in the VM via SSH
+ssh-cmd:
+	${VAGRANT_ENV} vagrant ssh -c '${CMD}'
+
--- a/demo/Vagrantfile
+++ b/demo/Vagrantfile
@@ -1,6 +1,11 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+$cluster_domain = ENV["CLUSTER_DOMAIN"] || "demo.localflynn.com"
+$flags_gui = ENV["FLAG_GUI"] || "false"
+$flag_update = ENV["FLAG_UPDATE"] || "false"
+$flag_bootstrap = ENV["FLAG_BOOTSTRAP"] || "true"
+
 # Fail if Vagrant version is too old
 begin
   Vagrant.require_version ">= 1.6.0"
@@ -22,15 +27,29 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   config.vm.synced_folder ".", "/vagrant", disabled: true
 
-  config.vm.provision "shell", privileged: false, inline: <<SCRIPT
-    ## Uncomment these lines to update flynn before starting
-    # sudo apt-get update -qq
-    # sudo apt-get install -y flynn-host
-    # sudo flynn-host download /etc/flynn/version.json
+  # Show VM GUI as a visual reminder that the VM is running
+  if $flag_gui == "true"
+    config.vm.provider "virtualbox" do |v|
+      v.gui = true
+    end
+    config.vm.provider "vmware_fusion" do |v|
+      v.gui = true
+    end
+  end
 
-    sudo start flynn-host
-
-    CLUSTER_DOMAIN=demo.localflynn.com \
-    flynn-host bootstrap /etc/flynn/bootstrap-manifest.json 2>&1
-SCRIPT
+  # Provisioners
+  if $flag_update == "true"
+    config.vm.provision "shell", inline: <<-SCRIPT
+      apt-get update -qq
+      apt-get install -y flynn-host
+      flynn-host download /etc/flynn/version.json
+    SCRIPT
+  end
+  if $flag_bootstrap == "true"
+    config.vm.provision "shell", privileged: false, inline: <<-SCRIPT
+      sudo start flynn-host
+      CLUSTER_DOMAIN="#{$cluster_domain}" \
+        flynn-host bootstrap /etc/flynn/bootstrap-manifest.json 2>&1
+    SCRIPT
+  end
 end
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ bin
 /version.json
 /log
 script/custom-vagrant
+demo/*.log
--- a/demo/README.md
+++ b/demo/README.md
@@ -1,6 +1,6 @@
 # Flynn Demo Environment
 
-This repo contains a Vagrantfile that boots up Flynn layer 0 and then bootstraps
+This repo contains a Vagrantfile and Makefile that boots up Flynn layer 0 and then bootstraps
 Flynn layer 1.
 
 ## Usage
--- a/docs/content/installation.html.md
+++ b/docs/content/installation.html.md
@@ -46,22 +46,61 @@ Clone the Flynn git repository:
 $ git clone https://github.com/flynn/flynn
 ```
 
+A Makefile is provided for convenience as a simple wrapper around common flynn
+commands. For systems without `make` installed, manually run the `vagrant` commands.
+
 Change to the `demo` directory and bring up the Vagrant box:
 
 ```
 $ cd flynn/demo
+
+# Provision the VM and bootstrap a flynn cluster
+# Init should only be called once
+$ make init
+# Print login token and open dashboard in browser
+$ make dashboard
+
+# OR
+
 $ vagrant up
+# Follow the instructions output by vagrant up, then...
+$ flynn key add
+$ flynn -a dashboard env get LOGIN_TOKEN
+# Copy the login token
+# Open the dashboard in a browser
+$ open http://dashboard.demo.localflynn.com
+```
+
+Additional make commands:
+
+```
+# Halt the VM
+$ make down
+
+# Bring up the VM and flynn
+$ make up
+
+# SSH into the VM
+$ make ssh
+
+# Destroy the VM
+$ make destroy
+
+# See Makefile for other useful commands
 ```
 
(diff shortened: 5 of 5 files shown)


demo/Makefile
```diff
@@ -0,0 +1,78 @@
...
+
+# Providers: virtualbox, vmware_fusion, aws
+PROVIDER := virtualbox
+
+
+.PHONY: install init bootstrap up down halt update ssh reload reset token destroy dashboard ssh-cmd
+
+
+# Install flynn cli on host
+install:
```
titanous: I'd call this target `cli`

demo/Makefile
```diff
@@ -0,0 +1,78 @@
...
+	@$(MAKE) CMD='sudo stop flynn-host' ssh-cmd
+
+# Halt the VM
+halt:
+	vagrant halt
+
+# Update flynn in the VM
+# VM must already be up and running
+update:
+	@$(MAKE) CMD='sudo bash < <(curl -fsSL https://dl.flynn.io/install-flynn) && sudo flynn-host update' ssh-cmd
```
titanous: The second command is unnecessary, currently the updater is incomplete so installing the latest version requires restarting from scratch.

demo/Vagrantfile
```diff
@@ -22,15 +24,17 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
 
-  config.vm.provision "shell", privileged: false, inline: <<SCRIPT
-    ## Uncomment these lines to update flynn before starting
-    # sudo apt-get update -qq
-    # sudo apt-get install -y flynn-host
-    # sudo flynn-host download /etc/flynn/version.json
-
-    sudo start flynn-host
+  config.vm.provider "virtualbox" do |v|
+    v.gui = true
```
titanous: Why this flag?

demo/Makefile
simple10: flynn-host does not currently auto start after halting the VM and restarting it. Thus the addition of `start flynn-host` in `make up`.

Should I just add a comment that this should be removed once `flynn-host` is added to the init system?

demo/Vagrantfile
simple10: Glad you asked. ;-)

I've worked with a lot of teams on-boarding them onto Vagrant. One of the really annoying things about Vagrant to someone who hasn't used it much (like a developer moving from Heroku to Flynn) is how it somewhat obfuscates the fact that `vagrant up` actually started a VM that is hogging a lot of resources. Or more precise, the fact that the dev has to run a `vagrant halt` to stop the VM from hogging resources. This is obvious to anyone who has Vagrant experience, but surprisingly easy to forget for non-Vagrant users.

Adding the gui flag causes VirtualBox and VMWare to show the normal KVM'ish interface and serves as a visual reminder that the VM is running.

Basically, I'm in favor of as little magic as possible when people are demoing things.

demo/Makefile
simple10: I might have misunderstood your comment @titanous. Did you mean `make down` is unnecessary or the `start flynn-host` in `make up` is unnecessary?

titanous: This is about `make update`: I think it's probably less confusing to replace the command with `vagrant box update` for now until the in-place updater is finished.

simple10: Got it. Pushed.

demo/Makefile
```diff
@@ -0,0 +1,78 @@
...
+	@$(MAKE) up
+	@$(MAKE) bootstrap
+	flynn key add
+
+# Bootstrap the cluster
+# This should only be called once per cluster
+# Runs bootstrap in VM and cluster add on host
+bootstrap:
+	@$(MAKE) CMD='CLUSTER_DOMAIN="${CLUSTER_DOMAIN}" flynn-host bootstrap /etc/flynn/bootstrap-manifest.json 2>&1' ssh-cmd | tee ${BOOTSTRAP_LOG_FILE}
+	cat ${BOOTSTRAP_LOG_FILE} | awk '/flynn cluster add/' | awk 'BEGIN {RS=""}{gsub(/\r/,"",$$0); print $$0}' | sh -s
```
lmars: We prefer using `bash` rather than `sh`, and `grep` for matching, so I suggest:

```
grep '^flynn cluster add' ${BOOTSTRAP_LOG_FILE} | bash
```

It may also be worth exiting non-zero if bootstrapping fails, e.g.:

```
(grep '^flynn cluster add' bootstrap.log || echo exit 1) | bash
```

demo/Vagrantfile
titanous: Ah, I see where you are coming from. I agree that for new users having no indicator that it is running can lead to forgetting about it. The current downside is that the window shows a login terminal that is basically useless, so we need to do something about that. A text or graphical message instead of the login prompt that explains how to access the dashboard would be one solution to this.

simple10: Yeah, help text output in the box terminal would be really cool.

How about I comment out the gui flag for now and open a separate issue for reinstating it once the flynn box can be updated with help text output?

I'm not sure right off hand how to customize the actual login prompt on Debian/Ubuntu.

titanous: Sounds good.

demo/Makefile
simple10: I'd recommend sticking with `sh` when possible since it's universal where `bash` is sometimes not installed (mostly in cases where it's been replaced with zsh and removed from the system or locked down). But `bash` is fine for the demo. I'll update.

demo/Vagrantfile
lmars: FYI the contents of `/etc/issue` are printed before the login prompt, see [`login(5)`](http://linux.die.net/man/5/issue)

demo/Makefile
lmars: > I'd recommend sticking with sh when possible since it's universal where bash is sometimes not installed

Yes I agree when portability is an issue, but here we have control of what is installed in the VM so I much prefer using `bash` whenever possible (I would even consider going as far as setting the default `sh` to `bash` rather than `dash`, not right now though :smile:)

demo/Vagrantfile
simple10: Oh sweet @lmars. How do I rebuild and test the box image? Looks like you're using packer.

demo/Makefile
simple10: In the case with the Makefile, it's actually running on the host and not in the VM. So portability is a concern. But I've updated it to bash anyway.

demo/Makefile
```diff
@@ -0,0 +1,77 @@
...
+	vagrant box update
+
+# SSH into VM
+ssh:
+	vagrant ssh
+
+# Stop, start, and reprovision VM
+# Run to fix configuration or provisioning problems
+reload:
+	vagrant reload --provision
```
titanous: This doesn't redo any provisioning, could call it `reboot` or change it to `destroy` + `init`.

demo/Makefile
lmars: Ah yes of course, my mistake.

demo/Makefile
simple10: `vagrant reload --provision` reruns the provisioners. It's equivalent to ...

```bash
vagrant halt
vagrant up --provision
```

I only included it because `vagrant reload` is mentioned in the docs. But I've always thought "reload" is a confusing term.

How about calling it "reset" and just keeping it in there for troubleshooting purposes. i.e. Whenever someone shows up on Github with an issue related to the demo, the first response can be "Well, did you try running `make reset` ?"  ;-)

simple10: Oh, and we should probably add the box update as part of the reset.

titanous: Sounds good. There are no provisioners that aren't commented out btw.

simple10: Yeah, strictly there for future proofing because everybody forgets to update the Makefile after changing Vagrantfiles.  ;-)

demo/Makefile
```diff
@@ -0,0 +1,79 @@
...
+	vagrant box update
+
+# SSH into VM
+ssh:
+	vagrant ssh
+
+# Destroy and rebuilt the VM
+reset:
+	@$(MAKE) destroy
+	@$(MAKE) init
```
titanous: I think we discussed adding an `make update` in here?

demo/Makefile
```diff
@@ -0,0 +1,79 @@
...
+
+# Stop flynn
+down:
+	@$(MAKE) CMD='sudo stop flynn-host' ssh-cmd
+
+# Halt the VM
+halt:
+	vagrant halt
+
+# Update the VM image to the lastest flynn vagrant box
```
titanous: s/lastest/latest

demo/Makefile
simple10: `make update` is currently doing a `vagrant box update`. Can you clarify your comment?

titanous: https://github.com/flynn/flynn/pull/1304#discussion_r26723070

> Oh, and we should probably add the box update as part of the reset.

titanous:
Is there anything left on this? I think it needs a squash and rebase on master before we can merge.

demo/Makefile
```diff
@@ -0,0 +1,80 @@
...
+# Get token for API and dashboard
+token:
+	@flynn -a dashboard env get LOGIN_TOKEN
+
+# Open dashboard in browser
+dashboard:
+	@echo Login token:
+	@$(MAKE) token
+	@sleep 1 # Allow user to see token output before browser opens
+	@open https://dashboard.${CLUSTER_DOMAIN}
```
titanous: This should be `http`, no `s` so that the CA certificate flow works.

demo/Makefile
```diff
@@ -0,0 +1,80 @@
...
+# Providers: virtualbox, vmware_fusion, aws
+PROVIDER := virtualbox
+
+
+.PHONY: install-cli init bootstrap up down halt update ssh reset token destroy dashboard ssh-cmd
+
+
+# Install flynn cli on host
+install-cli:
+	L=/usr/local/bin/flynn && curl -sL -A "`uname -sp`" https://dl.flynn.io/cli | zcat >$L && chmod +x $L
```
titanous: This doesn't work for me:

```text
L=/usr/local/bin/flynn && curl -sL -A "`uname -sp`" https://dl.flynn.io/cli | zcat > && chmod +x
/bin/sh: -c: line 0: syntax error near unexpected token `&&'
/bin/sh: -c: line 0: `L=/usr/local/bin/flynn && curl -sL -A "`uname -sp`" https://dl.flynn.io/cli | zcat > && chmod +x '
make: *** [install-cli] Error 2
```

demo/Makefile
```diff
@@ -0,0 +1,80 @@
...
+
+# Initialize a new cluster
+init:
```
titanous: `init` should be the first target in the file so that a bare `make` triggers it.

demo/Makefile
```diff
@@ -0,0 +1,80 @@
...
+	@$(MAKE) bootstrap
+	flynn key add
+	@printf "\nFlynn is now running in the VM.\nOpen dashboard: make dashboard\n\n"
+
+# Bootstrap the cluster
+# This should only be called once per cluster
+# Runs bootstrap in VM and cluster add on host
+bootstrap:
+	@$(MAKE) CMD='CLUSTER_DOMAIN="${CLUSTER_DOMAIN}" flynn-host bootstrap /etc/flynn/bootstrap-manifest.json 2>&1' ssh-cmd | tee ${BOOTSTRAP_LOG_FILE}
+	@(grep '^flynn cluster add .*$$' ${BOOTSTRAP_LOG_FILE} || echo exit 1) | sed "s/$$(printf '\r')\$$//" | bash
```
titanous: I think the `cluster add` command should be a separate target. It currently fails when I already have a cluster named `default` configured, so I think we should call the cluster `virtualbox` and set the `--force --default` flags when adding.

demo/Makefile
```diff
@@ -0,0 +1,80 @@
...
+	rm ${BOOTSTRAP_LOG_FILE}
+
+# Start the VM and flynn
+up:
+	CLUSTER_DOMAIN="${CLUSTER_DOMAIN}" vagrant up --provider ${PROVIDER}
```
titanous: This `vagrant up` doesn't work for me, as I need to set `SSH_AUTH_SOCK=''` to work around mitchellh/vagrant#4367. I think we should set that env var for all vagrant commands, there's no downside as we're using the insecure key, so the SSH agent is not needed.

demo/Makefile
```diff
@@ -0,0 +1,80 @@
...
+	@$(MAKE) CMD='sudo start flynn-host || echo flynn-host already running or unable to start' ssh-cmd
+
+# Stop flynn
+down:
+	@$(MAKE) CMD='sudo stop flynn-host' ssh-cmd
```
titanous: This `down` doesn't have symmetry with `up`. I'd drop it and rename `halt` to `down`.

titanous:
I added a few more comments after testing this out locally.

demo/Makefile
```diff
@@ -0,0 +1,80 @@
+
+# Domain to use when starting flynn
+CLUSTER_DOMAIN := demo.localflynn.com
+CLUSTER := default
+BOOTSTRAP_LOG_FILE := bootstrap.${CLUSTER}.log
```
titanous: I don't think we need the cluster name in the log file. Also, it should be gitignored.

demo/Makefile
simple10: How about calling the cluster "demo-virtualbox"?

titanous: Fine with me.

simple10: How do I set the name of the cluster? It seems like the `flynn-host bootstrap` command should include the cluster name since it outputs the command that's fed into bash. But I don't see a way in the help output to change the cluster name.

We could grep for `default` in the output but this seems kludgy.

titanous: Best bet is to rewrite the `cluster add` command since we need to add the force/default flags anyway. That message is only read by people using the manual install script, all of the orchestration tools use the JSON output from bootstrap and build the command using data from the relevant steps.

demo/Makefile
```diff
@@ -0,0 +1,82 @@
...
+
+# Destroy VM
+# Run this to delete the VM and free up disk space
+destroy:
+	vagrant destroy -f
+	flynn cluster remove ${CLUSTER}
+
+# Get token for API and dashboard
+token:
+	@flynn -a dashboard env get LOGIN_TOKEN
```
titanous: This should specify `-c ${CLUSTER}`.

demo/Makefile
```diff
@@ -0,0 +1,82 @@
...
+
+
+.PHONY: init install-cli bootstrap up down halt update ssh reset token destroy dashboard ssh-cmd
+
+
+# Initialize a new cluster
+init:
+	@$(MAKE) up
+	@$(MAKE) bootstrap
+	flynn key add
```
titanous: Ditto on the cluster specification.

simple10:
I think this is ready for merge.

demo/Makefile
```diff
@@ -0,0 +1,82 @@
...
+# Open dashboard in browser
+dashboard:
+	@echo Login token:
+	@$(MAKE) token
+	@sleep 1 # Allow user to see token output before browser opens
+	@open http://dashboard.${CLUSTER_DOMAIN}
+
+# Run a command in the VM via SSH
+ssh-cmd:
+	vagrant ssh -c '${CMD}'
```
titanous: `SSH_AUTH_SOCK` needs to be specified here and in the `ssh` target. Might make sense to make a variable.

simple10:
This should be ready to merge.

titanous:
@simple10 We just shipped Windows support for the CLI, and the current Vagrant setup (without this Makefile) should work fine on Windows. The Makefile will probably work fine if you have installed `make` somehow on Windows, but this is not ideal. Any thoughts on what we can do to keep the Windows support we already have while retaining the improved experience?

simple10:
@titanous  In the past with other Vagrant projects, I've passed in ENV flags to the Vagrantfile to enable/disable features. So by default, the provisioners would run everything unless the disable flags were present. The Makefile would then pass in the disable flags to take finer grain control of provisioning steps.

I suggest we use this approach for now since it requires minimal changes to code and docs. Then, if someone wants to write a bat script for Windows, they can submit another pull request.

Sound good?

titanous:
Yeah, that's fine.

simple10:
Pushed. Works fine with or without the Makefile now.

titanous:
LGTM! Thanks for your patience!

PR #1305, opened 2015-03-18
installer: Configure CLI

### LINKED ISSUE #1303: Installer doesn't configure CLI cluster
After the installer succeeds, it should configure the CLI for access to the cluster. This is a regression, as it used to do this before the web GUI landed.

jvatic:
refs 67e39ebebccc5dbb550ae694b4027dc249b779d7
fixes #1303

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -6,6 +6,8 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"os/user"
+	"path/filepath"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/BurntSushi/toml"
 )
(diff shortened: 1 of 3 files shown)


cli/config/config.go
```diff
@@ -23,6 +25,23 @@ type Config struct {
 	Clusters []*Cluster `toml:"cluster"`
 }
 
+// copied from cli/main.go
+func homedir() string {
```
jvatic: Is there any reason not to use [`os/user.Current()`](http://golang.org/pkg/os/user/#Current) here?

titanous: I don't think so, the new cross-compilation setup should allow it to work. Just make sure you test with a cross-compiled binary and it works.

titanous:
LGTM

Issue #1145, opened 2015-03-03
router: sync error: cannot acquire from closed pool logspam


titanous:
There is some scenario where the router can fail into a loop infinitely printing this error:

```text
router: sync error: cannot acquire from closed pool
```

I *think* this happens when postgres goes away.

/cc @benburkert @bgentry

bgentry:
hm, looked into this briefly just now, but I can't see anywhere we're closing the `pgx.ConnPool` except for on shutdown:

https://github.com/flynn/flynn/blob/9574f9aea07dd0e28caeea56284f596d64036af2/router/server.go#L116

Do you have any more details on when this happened? Was the server up and running for awhile?

titanous:
I'm pretty sure I saw this after booting a cluster with `SINGLETON=false` and killing the postgres primary, but I didn't have time to look at it in detail or try again. It was just reported in IRC by a user that had a bunch of issues getting the router going, possibly port collisions.

titanous:
@monaka is running into this: https://gist.github.com/f29f850ffcd3f871cb9d

titanous:
I haven't seen this in quite a while, closing.

PR #1306, opened 2015-03-18
router: add sleep backoff on datastore sync error

benburkert:
Limit logspam on a datastore sync errors, see #1145. I wasn't able to reproduce the `cannot acquire from closed pool`, but I was able to trigger logspam for `the database system is in recovery mode (SQLSTATE 57P03)`, and this change helped recovery.


titanous:
LGTM

PR #1307, opened 2015-03-18
Fix complaints of various linters

titanous:
Fixes various complaints of `varcheck` and `vet`. Also ran `grind`, which moved a few variables around, see the commit message for details.


controller/log_test.go
```diff
@@ -223,7 +223,7 @@ func (s *S) TestGetAppLogFollow(c *C) {
 
 	select {
 	case msg := <-msgc:
-		c.Fatalf("unexpected message received:", msg)
+		c.Fatal("unexpected message received:", msg)
```
lmars: This should have a space after the `:`

lmars:
LGTM

PR #1309, opened 2015-03-19
slugrunner: Don't use a proxy for .discoverd hosts

lmars:
discoverd domains are internal to the cluster, so will not work if the user has set http_proxy, as the proxy is most likely not within the cluster.

I tested this by starting a proxy and setting `http_proxy`, then seeing that the build uses the proxy (by viewing the proxy logs) but the slugrunner does not.


titanous:
LGTM

Issue #737, opened 2015-01-15
controller: Add SSE firehose


jvatic:
Add a single endpoint that supports server sent events and streams changes to all objects. This will things like the dashboard always showing the most recent information.

jvatic:
This was implemented in #1636

PR #1311, opened 2015-03-19
dashboard: integration tests WIP

jvatic:


jvatic:
Closing this for now as the current approach is essentially build on race conditions. I think it makes most sense to come back to this once #737 has landed and use that endpoint to give descriptive failures and not rely on timeouts.

The github api calls will need to be wrapped in some way to get rid of the timeouts for those steps as well (this approach could also work for most of the other steps).

PR #1312, opened 2015-03-19
router: shutdown router before datastore

benburkert:
Shutdown the `Router` before `DataStore` to avoid logspamming pgx pool errors.

fix for #1145.


titanous:
LGTM

PR #1315, opened 2015-03-20
Bump Ubuntu

titanous:
14.04.2 includes the backported Utopic 3.16 kernel.

### DIFF
--- a/script/release-vm-images
+++ b/script/release-vm-images
@@ -33,7 +33,6 @@ source "${ROOT}/script/lib/ui.sh"
 source "${ROOT}/script/lib/aws.sh"
 source "${ROOT}/script/lib/util.sh"
 
-UBUNTU_BOX_NAME="ubuntu-trusty-20150204.1" # used in filenames for caching, bump at the same time as UBUNTU_BOX_URL
 UBUNTU_BOX_URL="http://cloud-images.ubuntu.com/vagrant/trusty/20150316/trusty-server-cloudimg-amd64-vagrant-disk1.box"
 UBUNTU_BOX_SHA="e71422e9fa0ae1a1e7a84a8afee5c6177ee7a84e077addfd329696d4c5b9a3c9"
 UBUNTU_USEAST_AMI="ami-6089d208" # us-east-1 | trusty | 14.04 LTS | amd64 | hvm:ebs-ssd | 20150305 | hvm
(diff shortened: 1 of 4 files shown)


util/packer/ubuntu-14.04/scripts/upgrade.sh
```diff
@@ -15,11 +15,6 @@ if [[ "${PACKER_BUILDER_TYPE}" == "virtualbox-ovf" ]]; then
   apt-get remove --purge -y virtualbox-guest-dkms virtualbox-guest-utils virtualbox-guest-x11
 fi
 
-apt-get install --install-recommends linux-generic-lts-utopic \
-  -y \
-  -o Dpkg::Options::="--force-confdef" \
-  -o Dpkg::Options::="--force-confold"
-
```
lmars: Could we now drop this whole file and just use the provided kernel?

script/release-vm-images
```diff
@@ -34,9 +34,9 @@ source "${ROOT}/script/lib/aws.sh"
 source "${ROOT}/script/lib/util.sh"
 
 UBUNTU_BOX_NAME="ubuntu-trusty-20150204.1" # used in filenames for caching, bump at the same time as UBUNTU_BOX_URL
```
lmars: This variable needs bumping to invalidate the cache (we could just use `UBUNTU_BOX_SHA` for this).

util/packer/ubuntu-14.04/scripts/upgrade.sh
titanous: I'm not sure I follow? The `dist-upgrade` below can still result in a new kernel.

script/release-vm-images
titanous: Changed to use the SHA.

util/packer/ubuntu-14.04/scripts/upgrade.sh
lmars: Well, I meant just don't upgrade the kernel and use the one from the release image.

lmars:
LGTM

util/packer/ubuntu-14.04/scripts/upgrade.sh
titanous: Ah, I think we should always update the kernel.

PR #1318, opened 2015-03-20
Upgrade kernel (again)

titanous:
For two completely different reasons we actually need to keep these lines that I removed in #1315.


lmars:
LGTM

PR #1319, opened 2015-03-20
script: Don't force release tag

titanous:
If the tag already exists, we end up overwriting the previous release and the tag doesn't end up being written to GitHub.


lmars:
LGTM

PR #1325, opened 2015-03-21
Add support for controller key rotation

### LINKED ISSUE #1324: Allow changing the controller key
Imagine I want to change the controller key. I'd expect to be able to do something like this:

```text
$ flynn -a controller env set AUTH_KEY=$(openssl rand -hex 16)
timed out waiting for deployment completion
```

It looks like the deployer starts a deployer with the new release, which doesn't have the old key and so can't fetch the deploy record from the old controller web instance with the new controller key. I think the solution is to support multiple keys so that the transition looks like:

- Create new release with old+new keys accepted, wait for successful deploy.
- Create new release with only new key.

titanous:
Changing the controller key looks like this:

```text
NEW_KEY=$(openssl rand -hex 16)
flynn -a controller env set -t web AUTH_KEY=$NEW_KEY,$(flynn -a controller env get AUTH_KEY)
flynn -a gitreceive env set CONTROLLER_KEY=$NEW_KEY
flynn -a taffy env set CONTROLLER_KEY=$NEW_KEY
flynn -a dashboard env set CONTROLLER_KEY=$NEW_KEY
# change key in ~/.flynnrc
flynn -a controller env set AUTH_KEY=$NEW_KEY
flynn -a controller env unset -t web AUTH_KEY
```

Closes #1324

### DIFF
--- a/bootstrap/manifest_template.json
+++ b/bootstrap/manifest_template.json
@@ -222,6 +222,12 @@
       "uri": "$image_repository?name=flynn/receiver&id=$image_id[receiver]"
     },
     "release": {
+      "env": {
+        "SSH_PRIVATE_KEYS": "{{ (index .StepData \"gitreceive-key\").PrivateKeys }}",
+        "CONTROLLER_KEY": "{{ (index .StepData \"controller-key\").Data }}",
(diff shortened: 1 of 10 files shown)


lmars:
LGTM

Issue #1298, opened 2015-03-17
Installer should check SSH key fingerprint


titanous:
I had a local SSH key from a previous install, so the installer said:

```text
INFO[03-17|15:08:04] Using saved key pair (flynn)
```

However, the public key on EC2 was not the same as the local one, so it failed to SSH to the server:

```text
INFO[03-17|15:13:39] ssh: handshake failed: ssh: unable to authenticate, attempted methods [publickey none], no supported methods remain
INFO[03-17|15:13:39] ssh: handshake failed: ssh: unable to authenticate, attempted methods [publickey none], no supported methods remain
```

The installer should look for the key on EC2 based on the fingerprint, not the filename.

PR #1326, opened 2015-03-22
Installer key fingerprint

jvatic:
#1298

### DIFF
--- a/installer/installer.go
+++ b/installer/installer.go
@@ -23,6 +23,7 @@ import (
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/awslabs/aws-sdk-go/gen/route53"
 	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/crypto/ssh"
 	cfg "github.com/flynn/flynn/cli/config"
+	"github.com/flynn/flynn/pkg/awsutil"
 	"github.com/flynn/flynn/pkg/etcdcluster"
 	"github.com/flynn/flynn/pkg/sshkeygen"
 	"github.com/flynn/flynn/util/release/types"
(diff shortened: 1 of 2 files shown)


pkg/awsutil/awsutil.go
```diff
@@ -0,0 +1,36 @@
...
+func FingerprintPublicKey(key []byte) (string, error) {
+	sshPublicKey, _, _, _, err := ssh.ParseAuthorizedKey(key)
+	if err != nil {
+		return "", err
+	}
+	v := reflect.Indirect(reflect.ValueOf(sshPublicKey))
+	publicKey := &rsa.PublicKey{
+		N: v.FieldByName("N").Interface().(*big.Int),
+		E: v.FieldByName("E").Interface().(int),
+	}
```
titanous: Reaching into unexported structs is a no-no. Instead, you'll have to parse it yourself using (simplified) logic based on the internals of `ssh.ParseAuthorizedKey`.

jvatic: Actually, seeing as we always have the private key for this use case, we can just use that to get the public key in the form of an `rsa.PublicKey`.

installer/installer.go
```diff
@@ -269,19 +270,24 @@ func (s *Stack) createKeyPair() error {
...
+				Filters: []ec2.Filter{{
+					Name:   aws.String("key-name"),
+					Values: []string{keypairName},
+				},
+					{
+						Name:   aws.String("fingerprint"),
+						Values: []string{fingerprint},
+					}},
+			})
+			if err == nil && len(res.KeyPairs) > 0 && *res.KeyPairs[0].KeyName == keypairName {
```
titanous: I don't really like these nested `err == nil` checks. Is there a way to invert the logic or split it into separate functions?

installer/installer.go
```diff
@@ -269,19 +270,24 @@ func (s *Stack) createKeyPair() error {
...
+			s.SSHKeyName = keypairName
+			res, err := s.ec2.DescribeKeyPairs(&ec2.DescribeKeyPairsRequest{
+				Filters: []ec2.Filter{{
+					Name:   aws.String("key-name"),
+					Values: []string{keypairName},
+				},
+					{
+						Name:   aws.String("fingerprint"),
+						Values: []string{fingerprint},
+					}},
```
titanous: Split these braces onto two lines (and the same for the opening)

installer/installer.go
```diff
@@ -269,19 +270,24 @@ func (s *Stack) createKeyPair() error {
...
+			})
+			if err == nil && len(res.KeyPairs) > 0 && *res.KeyPairs[0].KeyName == keypairName {
+				// key exists, we're good to go
```
titanous: This comment is superfluous

jvatic:
@titanous comments addressed

installer/installer.go
```diff
@@ -263,26 +264,46 @@ func (s *Stack) allocateDomain() error {
...
+	}
+	fingerprint, err := awsutil.FingerprintImportedKey(keypair.PrivateKey)
+	if err != nil {
+		return err
+	}
+	res, err := s.ec2.DescribeKeyPairs(&ec2.DescribeKeyPairsRequest{
+		Filters: []ec2.Filter{
+			{
+				Name:   aws.String("key-name"),
+				Values: []string{name},
```
titanous: Is the name filter necessary?

installer/installer.go
```diff
@@ -263,26 +264,46 @@ func (s *Stack) allocateDomain() error {
...
+			{
+				Name:   aws.String("fingerprint"),
+				Values: []string{fingerprint},
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+	if len(res.KeyPairs) <= 0 {
```
titanous: `==`

jvatic:
@titanous comments addressed

installer/installer.go
```diff
@@ -263,26 +264,42 @@ func (s *Stack) allocateDomain() error {
...
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+	if len(res.KeyPairs) == 0 {
+		return errors.New("No matching key found")
+	}
+	s.SSHKey = keypair
```
jvatic: This should use the name from the response as there could be a matching key of another name.

jvatic: Done.

titanous:
LGTM

PR #1332, opened 2015-03-23
controller/client: Use DialTLS with pinned

titanous:
This avoids the need to parse the URL and rewrite it to confuse net/http.


benburkert:
LGTM

PR #1339, opened 2015-03-23
logaggregator: replication

benburkert:
Replicate logs from leader instance to follower(s). The follower dials the replication port of the leader and receives a stream of `gob` encoded `rfc5424.Message` objects. By default, a flynn cluster now runs two logaggregator app processes: one will become the leader, the other will follow.


benburkert:
This is ready for :eyes:

/cc @titanous

logaggregator/aggregator.go
```diff
@@ -0,0 +1,130 @@
...
+	bmu     sync.Mutex // protects buffers
+	buffers map[string]*ring.Buffer
+
+	msgc chan *rfc5424.Message
+
+	pmu    sync.Mutex
+	pausec chan struct{}
+}
+
+// NewAggregator creates a new unstarted Aggregator that will listen on addr.
```
titanous: This comment references `addr`, which doesn't appear to be an argument.

logaggregator/server.go
```diff
@@ -0,0 +1,349 @@
...
+
+	api      http.Handler
+	shutdown chan struct{}
+}
+
+type ServerConfig struct {
+	syslogAddr, replicationAddr, apiAddr string
+
+	serviceName string
+	discoverd   *discoverd.Client
```
titanous: Any fields that are expected to be set before calling `NewServer` should be capitalized/exported.

benburkert:
This is ready for another :eyes:

test/test_logaggregator.go
```diff
@@ -0,0 +1,103 @@
...
+
+	ldrc := make(chan *discoverd.Instance)
+	stream, err := srv.Leaders(ldrc)
+	t.Assert(err, c.IsNil)
+	defer stream.Close()
+
+	hostID, jobID, err := cluster.ParseJobID(leader.Meta["FLYNN_JOB_ID"])
+	t.Assert(err, c.IsNil)
+
+	for len(ldrc) > 0 {
```
titanous: What does this do?

test/test_logaggregator.go
```diff
@@ -0,0 +1,103 @@
...
+func (s *LogAggregatorSuite) TestLogReplication(t *c.C) {
+	app := s.newCliTestApp(t)
+
+	scale := app.flynn("scale", "sequential-printer=1")
+	app.waitFor(jobEvents{"sequential-printer": {"up": 1}})
+	t.Assert(scale, Succeeds)
+
+	for {
+		if app.flynn("log", "--raw-output").Output != "" {
+			break
```
lmars: Is it worth waiting for a few lines of output before checking they are sequential?

lmars:
@benburkert I will be modifying the updater to handle this, what I will need is some way to know that the new jobs have all the data before killing the old ones, what do you think is the best approach?

From my quick look through, given the replication looks async, one approach to avoid data loss is to pause / read-only the leader for the duration of the update, WDYT?

titanous:
Closing until this is picked up again.

PR #1350, opened 2015-03-24
Build and release Windows CLI binaries

lmars:
The docs also need updating with instructions on how to install the CLI on Windows, but I do not have the relevant knowledge on that.

Closes #1349.


titanous:
LGTM! I'll handle testing and docs updates.

PR #1352, opened 2015-03-25
logaggregator: refactor filter interface

benburkert:
Add the `filterFunc` & `filterSlice` types and the `Filter` func to the `Filter` interface.


titanous:
LGTM

PR #1356, opened 2015-03-25
Windows fixes for CLI

titanous:
- Allow installer to open the web UI on Windows
- Shuffle paths a bit so that everything ends up in `%APPDATA%\flynn`
- Move `~/.flynn-installer` to `~/.flynn/installer`
- Bump go-tuf to fix some path issues on Windows
- Add install snippet for Windows to docs

I've tested the install script on Windows 7, Windows 8.1, and Windows Server 2012 R2. I've tested the auto-updater, installer, and many commands on Windows 8.1. The only known broken command as of this PR is `flynn run`, which I will fix in a subsequent PR.


jvatic:
LGTM

PR #1363, opened 2015-03-26
script,test: Add flynnbr0 to /etc/resolv.conf before running tests

lmars:
This is so that the tests can use discoverd DNS (for example to connect to a postgres DB at `leader.postgres.discoverd`).

Extracted from #1301.


titanous:
LGTM

PR #1364, opened 2015-03-26
Static binaries

titanous:
All but a few binaries can be built as fully static binaries with no cgo.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -6,11 +6,11 @@ import (
 	"net"
 	"net/url"
 	"os"
-	"os/user"
 	"path/filepath"
 	"runtime"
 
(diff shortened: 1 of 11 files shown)


cli/config/config.go
```diff
@@ -33,8 +32,7 @@ func HomeDir() string {
 	}
 	home := os.Getenv(key)
 	if home == "" {
-		user, _ := user.Current()
-		home = user.HomeDir
+		panic(fmt.Sprintf("unknown home directory, ensure $%s is set", key))
```
alex: https://github.com/mitchellh/go-homedir/ is also an option

titanous: Neat, thanks!

lmars:
LGTM

PR #1370, opened 2015-03-28
Select next available cluster when default is removed

### LINKED ISSUE #1358: cli: Removing default cluster doesn't update default
If I remove the default cluster and have another cluster available, it should be marked as the default.

geovanisouza92:
Closes #1358

### DIFF
--- a/cli/cluster.go
+++ b/cli/cluster.go
@@ -100,11 +100,19 @@ func runClusterRemove(args *docopt.Args) error {
 	name := args.String["<cluster-name>"]
 
 	if config.Remove(name) {
+		msg := fmt.Sprintf("Cluster %q removed.", name)
+
+		// Select next available cluster as default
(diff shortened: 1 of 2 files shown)


cli/cluster.go
```diff
@@ -100,11 +100,19 @@ func runClusterRemove(args *docopt.Args) error {
 	name := args.String["<cluster-name>"]
 
 	if config.Remove(name) {
+		msg := fmt.Sprintf("Cluster %q removed.", name)
+
+		// Select next available cluster as default
+		if len(config.Clusters) > 0 {
+			config.SetDefault(config.Clusters[0].Name)
```
titanous: This logic should only be triggered if the cluster being removed is the default.

geovanisouza92: Yeah, you right, I forget it.

titanous:
Thanks!

PR #1372, opened 2015-03-28
host/cli: Unlink destination path during download

### LINKED ISSUE #1371: Install script should remove files before writing
The install script should remove destination files before writing new ones, otherwise if flynn is running it will fail with errors like:

```text
03:20:39.065205 host.go:131: open /usr/local/bin/flynn-init: text file busy
===> 03:20:39.069 ERROR: failed to run "/usr/local/bin/flynn-host download --repository https://dl.flynn.io/tuf - -tuf-db /etc/flynn/tuf.db --config-dir /etc/flynn --bin-dir /usr/local/bin", exit status 1
```

This is just a hypothesis, we should test it before committing.

lmars:
This avoids trying to replace a file which is in use.

Fixes #1371.

### DIFF
--- a/host/cli/download.go
+++ b/host/cli/download.go
@@ -100,6 +100,10 @@ func downloadGzippedFile(client *tuf.Client, path, dir string) (string, error) {
 	}
 	defer file.Close()
 	dst := filepath.Join(dir, path)
+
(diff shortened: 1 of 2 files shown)


titanous:
LGTM

Issue #276, opened 2014-10-02
Flynn updater


titanous:

lmars:
All layer 1 components can now be updated by running `flynn-host update`, namely:

* blobstore
* taffy
* dashboard
* router
* gitreceive
* controller
* logaggregator (possible data loss until replication lands #1339 and deployer updated)
* postgres

What remains is updating layer 0 components, namely:

* flynn-host
* discoverd
* etcd
* flannel

Updating those needs #1373 to land first.

We will add full documentation on the process once this work is complete.

titanous:
This is done!

lmars:
:heart_eyes:

PR #1376, opened 2015-03-30
updater: Update postgres

lmars:
WIP. Opening this as an early pull request for CI.

Things which need fixing:

- [x] Don't rollback when a postgres deploy fails, just bail as rolling back will not help the situation
- [x] Handle starting new postgres jobs when postgres is read-only (cannot scale using formations), or avoid trying to start jobs in that state
- [x] Retry creating deployment event records when postgres is down / read-only
- [x] Investigate the updater succeeding when a deployment fails (I suspect [`e.Err()`](https://github.com/flynn/flynn/blob/master/controller/client/client.go#L376) is returning nil)


lmars:
I have dealt with the initial issues by avoiding getting stuck in a read-only state, and retrying requests when Postgres is read-only, but another issue has revealed itself:

When killing the existing Primary, the deployers all get disconnected, and upon connection to the new Primary one of them starts to work the already in progress Postgres deploy (because there are no longer any locks on it), detects it is in a bad state and fails the deploy. Meanwhile though the initial deploy still runs, and eventually succeeds, but by this time the deployment has already been marked as failed.

Possible solutions:

* Revisit adding a `locked_until` column on deployments (see #603), ensuring the deployer connects to the new Primary to continue bumping the lock
* Handle updating Postgres directly in the updater rather than using the deployer (i.e. move the logic in postgres deploy strategy into the updater)

@titanous WDYT?

lmars:
@titanous ok I'll give the `locked_until` approach a try, unless you have other ideas?

titanous:
@lmars One thing to evaluate is doing this in que-go as we may have jobs in the future that we want the same protection for.

lmars:
This is part of #276.

lmars:
@titanous ready for a review.

Most changes are around reconnecting and handling transient postgres errors, and whilst there is still potential for errors to occur during the update, I have verified the cluster returns to a normal state once the update is complete.

controller/deployer/main.go
```diff
@@ -179,5 +180,17 @@ func (c *context) createDeploymentEvent(e ct.DeploymentEvent) error {
...
+
+var execAttempts = attempt.Strategy{
+	Total: 10 * time.Second,
+	Delay: 100 * time.Millisecond,
+}
+
+// Retry db queries in case postgres has been deployed
+func (c *context) execWithRetries(query string, args ...interface{}) error {
+	return execAttempts.Run(func() error {
+		return c.db.Exec(query, args...)
```
titanous: Do we need to filter errors so that this doesn't trigger on data validation errors, etc?

controller/client/client.go
```diff
@@ -421,3 +423,29 @@ func (c *Client) ProviderList() ([]*ct.Provider, error) {
...
+}
+
+func (c *Client) Delete(path string) error {
+	return c.send("DELETE", path, nil, nil)
+}
+
+func (c *Client) send(method, path string, in, out interface{}) (err error) {
+	for startTime := time.Now(); time.Since(startTime) < 10*time.Second; time.Sleep(100 * time.Millisecond) {
+		err = c.Send(method, path, in, out)
+		if !httphelper.IsPostgresError(err) {
```
titanous: I'd prefer if the error was marked as a retry-able error instead of exposing internal details about the use of postgres.

controller/deployer/main.go
lmars: On my journey I have seen quite a few different errors coming back from the DB during the update, and I found it hard to collate an appropriate list of retry-able errors, so I just went for a blanket retry here as no errors are usually expected.

We could make a whitelist of errors which should not be retried here, but given we don't usually expect an error here, what would be in the list?

titanous: Okay, I think we can leave it as-is.

lmars:
@titanous comments addressed

pkg/httphelper/httphelper.go
```diff
@@ -64,6 +69,18 @@ func IsValidationError(err error) bool {
...
+// IsRetryableError indicates whether a HTTP request can be safely retried.
+//
+// We consider all postgres errors as retry-able as they usually occur when
+// postgres is read-only (for example during a system update). Data related
+// postgres errors should in general not be retried (because a retry will
+// likely result in the same error), but it is expected that such errors are
+// caught and a validation error returned to the client rather than the
+// postgres error.
+func IsRetryableError(err error) bool {
+	return isJSONErrorWithCode(err, PostgresErrorCode)
```
titanous: The error response should be marked as retry-able in the JSON message, and we should not send a detailed postgres error (a 500 with the retry flag is sufficient).

titanous:
LGTM

PR #1385, opened 2015-04-01
cli: Add response time for each command

### LINKED ISSUE #202: Response times in CLI
Show response time for each action in the CLI (a la Google search)

geovanisouza92:
Closes #202

Would be useful have a toggle configuration for this in ```.flynnrc```?

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strings"
 	"text/tabwriter"
+	"time"
 	"unicode"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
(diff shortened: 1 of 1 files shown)


lmars:
@geovanisouza92 an option would be good, personally I don't need to see the response time for each command.

This also breaks a lot of tests which assert the CLI output (e.g. most of the [CLISuite tests](https://github.com/flynn/flynn/blob/master/test/test_cli.go)).

geovanisouza92:
@lmars Ok, I'm looking at the tests right now.

titanous:
I think we should probably just close the original issue, we already have a time elapsed for scaling, and I don't feel a pressing need for it anywhere else.

PR #1388, opened 2015-04-01
cli: Fix log message

lmars:
The string in `msg` is generated with `Sprintf` so it is not a format, which leads to the following being logged:

```
Cluster "default" removed.%!(EXTRA string=default)
```


titanous:
LGTM

PR #1391, opened 2015-04-02
host/cli: Add support for bind mounting in `flynn-host run`

titanous:
This allows for arbitrary diagnostics/maintenance and is useful to extract data from clusters under some circumstances.


lmars:
LGTM

PR #1392, opened 2015-04-02
host/cli: Fallback to collecting container logs from disk

### LINKED ISSUE #1065: collect-debug-info should upload container logs
If flynn-host is not running, collect-debug-info should collect the on-disk logs.

titanous:
This ensures we get debug info even if flynn-host isn’t running.

Closes #1065


lmars:
LGTM

PR #1401, opened 2015-04-03
slugbuilder: Add static file buildpack

### LINKED ISSUE #1398: Add default static buildpack
It would be great to have a default static buildpack, so static apps can easily be deployed too.

titanous:
When a Staticfile file is present in the root of the repo, this buildpack is used to serve files using nginx.

Closes #1398

/cc @jvatic


jvatic:
LGTM

PR #1404, opened 2015-04-04
cli: fixed command line usages

arekkas:
Currently, it is these two usages don't result in the same output:

```
$ flynn --something a
$ flynn --something=a
```

This PR fixes this.

> Note, writing --input ARG (opposed to --input=ARG) is ambiguous, meaning it is not possible to tell whether ARG is option's argument or positional argument. In usage patterns this will be interpreted as option with argument only if option's description (covered below) for that option is provided. Otherwise it will be interpreted as separate option and positional argument.
[Source](http://docopt.org/)

*Notes:*
* This is the follow-up to #1153
* I wasn't able to test it on my environment, please check if nothing got broken

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -22,7 +22,7 @@ If run from a git repository, a 'flynn' remote will be created or replaced that
 allows deploying the application via git.
 
 Options:
-	-r, --remote <remote>  Name of git remote to create, empty string for none. [default: flynn]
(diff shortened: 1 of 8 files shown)


arekkas:
not sure why tests fail

titanous:
This looks good! One minor note: would you please amend the sign off in the commit message to have your name instead of your username? Thanks!

arekkas:
Sure :)

titanous:
Thanks!

PR #1407, opened 2015-04-05
bootstrap: Avoid divide by zero in host selection

titanous:
If there are no hosts running at this point, the host selection will panic with a divide by zero. Instead, if there are no hosts, return an error.


lmars:
LGTM

PR #1412, opened 2015-04-06
controller/scheduler: Sync job states at startup

### LINKED ISSUE #1406: Scheduler should remove jobs that are not running at start
This is a very specific subset of #686.

**Reproduction steps**

1. Start a single-node cluster.
1. Deploy an app with a single process.
1. Stop/start flynn-host.
1. Run `flynn ps`, and note that there are two jobs listed as running, even though only one is actually running.

One job was killed when shutting down the server, but this was not reflected in the controller database and it was never marked as down. When the server starts back up, a new job is started and added to the database. The "ghost" job prevents any subsequent deploys because the deployer sees it and refuses to complete the deploy without it going down.

**Solution**

A quick fix without resolving all of #686 is to scan and rectify the list of jobs that are marked as running vs actually running when the scheduler starts.

lmars:
This prevents jobs which are not running still appearing as up in the controller if the scheduler missed the event (e.g. because flynn-host was restarted).

Closes #1406.

### DIFF
--- a/controller/scheduler/main.go
+++ b/controller/scheduler/main.go
@@ -111,6 +111,8 @@ type clusterClient interface {
 }
 
 type controllerClient interface {
+	AppList() ([]*ct.App, error)
+	JobList(string) ([]*ct.Job, error)
 	GetRelease(releaseID string) (*ct.Release, error)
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

PR #1416, opened 2015-04-06
host: Copy search path from host resolv.conf

### LINKED ISSUE #1414: Get cedarish image to use search field of base server
If `/etc/resolv.conf` on the flynn-host has a search field, forward it through to the cedarish image.

titanous:
This parses and copies any search paths from the host `resolv.conf` into the container `resolv.conf`, allowing local domains to be resolved automatically.

Closes #1414


lmars:
LGTM

PR #1417, opened 2015-04-06
pkg/cluster: Retry requests with dial failures

lmars:
There is a small time window between a flynn-host leader being elected and the leader accepting sampi API requests, in which time `cluster.RegisterHost` will fail with:

```
dial tcp 10.76.22.3:1113: connection refused
```

In this case, and in general, the request should just be retried.


titanous:
LGTM

PR #1418, opened 2015-04-06
Set router sticky key in bootstrap

titanous:
It is now required, as not setting it will result in an all-zero key, which could be used by attackers to cause the router to connect to a specific address.


lmars:
LGTM

PR #1431, opened 2015-04-09
docs: Fix DATABASE_URL snippet on old versions of bash

titanous:
Old versions of bash (like the versions that ship with OS X) are not capable of sourcing a fd (like the one created during process substitution). This alternative that uses export should work with all versions of ZSH and Bash.


lmars:
LGTM

PR #1437, opened 2015-04-10
Prevented scaling process types that do not exist in the release

### LINKED ISSUE #1172: controller should not allow formation with process types that don't exist
The controller should not allow formations with process types that aren't in the release. The error message should be friendly enough that users of the CLI will understand the mistake easily.

jzila:
Blocked scaling nonexisting process types both in the controller and CLI. Tested both independently.

Fixes #1172

### DIFF
--- a/cli/scale.go
+++ b/cli/scale.go
@@ -82,6 +82,7 @@ func runScale(args *docopt.Args, client *controller.Client) error {
 	for k, v := range current {
 		processes[k] = v
 	}
+	invalid := make([]string, 0, len(release.Processes))
 	for _, arg := range typeCounts {
 		i := strings.IndexRune(arg, '=')
 		if i < 0 {
@@ -91,7 +92,15 @@ func runScale(args *docopt.Args, client *controller.Client) error {
 		if err != nil {
 			fmt.Println(commands["scale"].usage)
 		}
-		processes[arg[:i]] = val
+		processType := arg[:i]
+		if _, ok := release.Processes[processType]; ok {
+			processes[processType] = val
+		} else {
+			invalid = append(invalid, fmt.Sprintf("%q", processType))
+		}
+	}
+	if len(invalid) > 0 {
+		return errors.New(fmt.Sprintf("ERROR: Unknown process types: %s", strings.Join(invalid, ", ")))
 	}
 	formation.Processes = processes
 
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/flynn/flynn/controller/client"
 	tu "github.com/flynn/flynn/controller/testutils"
 	ct "github.com/flynn/flynn/controller/types"
+	hh "github.com/flynn/flynn/pkg/httphelper"
 	"github.com/flynn/flynn/pkg/postgres"
 	"github.com/flynn/flynn/pkg/random"
 	"github.com/flynn/flynn/pkg/testutils/postgres"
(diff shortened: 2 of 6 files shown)


titanous:
Thanks! This should have a unit test.

jzila:
K now to figure out how to write a unit test :P

cli/scale.go
```diff
@@ -91,7 +91,12 @@ func runScale(args *docopt.Args, client *controller.Client) error {
 		if err != nil {
 			fmt.Println(commands["scale"].usage)
 		}
-		processes[arg[:i]] = val
+		processType := arg[:i]
+		if _, ok := release.Processes[processType]; ok {
+			processes[processType] = val
+		} else {
+			fmt.Printf("No process type exists with name \"%s\".\n", processType)
```
lmars: I think this should collect the invalid process types and then exit if necessary, e.g.:

```
$ flynn scale web=1 workr=2 foo=3
ERROR: unknown process types: "workr", "foo"
```

Also, you should use `%q` rather than `\"%s\"` in the formatter string.

controller/controller_test.go
```diff
@@ -255,13 +256,25 @@ func (s *S) TestCreateRelease(c *C) {
...
+		// First create an empty release and attempt to scale a nonexistent process type. Will fail.
 		release := s.createTestRelease(c, &ct.Release{})
 		app := s.createTestApp(c, &ct.App{Name: fmt.Sprintf("create-formation-%d", i)})
-
 		in := &ct.Formation{ReleaseID: release.ID, AppID: app.ID, Processes: map[string]int{"web": 1}}
 		if useName {
 			in.AppID = app.Name
 		}
+		err := s.c.PutFormation(in)
+		c.Assert(hh.IsValidationError(err), Equals, true)
```
lmars: Is it worth checking the code of the validation error?

controller/controller_test.go
```diff
@@ -255,13 +256,25 @@ func (s *S) TestCreateRelease(c *C) {
...
+
+		// Now create a release with a valid process type and scale it successfully
+		release = s.createTestRelease(c, &ct.Release{
+			Processes: map[string]ct.ProcessType{"web": {}},
+		})
+
+		in = &ct.Formation{ReleaseID: release.ID, AppID: app.ID, Processes: map[string]int{"web": 1}}
```
lmars: Could just set `in.ReleaseID` here.

cli/scale.go
jzila: OK so do what I did in the controller? Also, I don't see a reason not to succeed with the valid process types. So in your above example, it could be:
```
$ flynn scale web=1 workr=2 foo=3
WARNING: unknown process types: "workr", "foo"
scaling web 0=>1
```

controller/controller_test.go
```diff
@@ -255,13 +256,25 @@ func (s *S) TestCreateRelease(c *C) {
...
 		if useName {
 			in.AppID = app.Name
 		}
+		err := s.c.PutFormation(in)
+		c.Assert(hh.IsValidationError(err), Equals, true)
+
+		// Now create a release with a valid process type and scale it successfully
+		release = s.createTestRelease(c, &ct.Release{
+			Processes: map[string]ct.ProcessType{"web": {}},
+		})
```
lmars: You could avoid creating multiple releases by creating the inital release with a `web` process type, and then calling `PutFormation` first with invalid processes (e.g. `map[string]int{"foo": 1}`), and then with valid processes.

controller/formation.go
```diff
@@ -52,11 +52,30 @@ func procsHstore(m map[string]int) hstore.Hstore {
...
 	return res
+}
 
+func (r *FormationRepo) validateFormProcs(f *ct.Formation) error {
+	release, err := r.releases.Get(f.ReleaseID)
+	if err != nil {
+		return err
+	}
+	rel := release.(*ct.Release)
+	invalidProcessTypes := make([]string, 0, len(f.Processes))
```
lmars: I would just call this `invalid`

test/test_cli.go
```diff
@@ -614,7 +614,9 @@ func (s *CLISuite) TestRelease(t *c.C) {
 	t.Assert(r.Env, c.DeepEquals, release.Env)
 	t.Assert(r.Processes, c.DeepEquals, release.Processes)
 
-	t.Assert(app.flynn("scale", "--no-wait", "env=1"), Succeeds)
+	flynnLog := app.flynn("scale", "--no-wait", "env=1", "foo=1")
```
lmars: This should be called `scaleCmd` or similar.

cli/scale.go
lmars: Well, if I run `flynn scale web=1 workr=2` I've obviously made a typo, my intention was to scale both the `web` and `worker` processes so I would like Flynn to tell me that and exit, leaving me to re-run the command correctly to do what I actually wanted. WDYT?

controller/controller_test.go
jzila: I don't set the code right now since that wasn't the pattern in other code I cribbed from. I could do that, however, if you prefer.

cli/scale.go
jzila: I think a case can be made for either solution. I suppose it's safer to go your route though.

jzila:
Blocked on https://github.com/flynn/flynn/pull/1455

lmars:
Thanks!

PR #1443, opened 2015-04-11
test: Remove Gradle buildpack test

titanous:
The deploy fails with:

    Downloading https://services.gradle.org/distributions/gradle-2.1-bin.zip

    Exception in thread "main" javax.net.ssl.SSLHandshakeException:
    java.security.cert.CertificateException: No subject alternative
    DNS name matching services.gradle.org found.


lmars:
LGTM

PR #1448, opened 2015-04-13
Fix installer log

### LINKED ISSUE #1395: Bug in `flynn install` logging
```text
INFO[04-02|14:07:50] The built-in dashboard can be accessed at http://dashboard.redacted.flynnhub.com with login token redacted LOG15_ERROR= LOG15_ERROR="Normalized odd number of arguments by adding nil"
```

titanous:
The second return value from `DashboardLoginMsg` (an error which shouldn’t occur) was causing the log15 context to be unbalanced.

Closes #1395

/cc @jvatic


jvatic:
LGTM

PR #1449, opened 2015-04-14
dashboard: Should detect SSL problems with CORS

### LINKED ISSUE #1411: SSL Certificate and CORS
When setting up Flynn, you're being asked to download and install the certificate. IIRC this only works on per-browser basis. Whenever another client (different browser instance) logs into the dashboard, he'll see a warning:

flynn1

Once added, he can sign in and proceed to the apps tab, but he won't see any:

flynn2

This is because the security exception applies only for this specific domain.

I suggest to add a warning to the dashboard which instructs the user to download the certificate if network calls are blocked due to insecure responses.

If anyone can hint me at the proper direction for this, I'd  be happy to create a PR.

arekkas:
**WORK IN PROGRESS**

Closes #1411

ping @jvatic

### DIFF
--- a/dashboard/app/lib/javascripts/dashboard/client.js
+++ b/dashboard/app/lib/javascripts/dashboard/client.js
@@ -97,6 +97,15 @@ Dashboard.Client = Marbles.Utils.createClass({
 		});
 	},
 
+	ping: function (endpoint, protocol) {
+		endpoint = window.location.host.replace("dashboard", endpoint);
+		endpoint = protocol + "://" + endpoint + "/ping";
+
+		return this.performRequest('GET', {
+			url: endpoint
+		});
+	},
+
 	getApps: function () {
 		return this.performControllerRequest('GET', {
 			url: "/apps"
--- a/dashboard/app/lib/javascripts/dashboard/main.js
+++ b/dashboard/app/lib/javascripts/dashboard/main.js
@@ -13,8 +13,20 @@ window.Dashboard = {
 	config: {},
 
 	waitForRouteHandler: Promise.resolve(),
+	errCertNotInstalled: new Error("HTTPS certificate is not trusted."),
+	errServiceUnavailable: new Error("Service is unavailable."),
 
 	run: function () {
+		var self = this, loadURL, resolveWaitForNav;
+		this.waitForNav = new Promise(function(resolve, reject) {
+			resolveWaitForNav = resolve;
+		});
+
+		loadURL = function() {
+			resolveWaitForNav();
+			Marbles.history.loadURL();
+		};
+
 		if ( Marbles.history && Marbles.history.started ) {
 			throw new Error("Marbles.history already started!");
 		}
@@ -38,11 +50,11 @@ window.Dashboard = {
 			dispatcher: this.Dispatcher,
 			trigger: false
 		});
-		if (window.location.protocol === "http:" && this.config.INSTALL_CERT) {
-			Marbles.history.navigate("installcert");
-			Marbles.history.loadURL();
+
+		if (this.config.INSTALL_CERT) {
+			this.__isCertInstalled().then(loadURL);
 		} else {
-			Marbles.history.loadURL();
+			loadURL();
 		}
 	},
 
@@ -63,6 +75,92 @@ window.Dashboard = {
 	__redirectToLogin: function () {
 		var redirectPath = Marbles.history.path ? '?redirect='+ encodeURIComponent(Marbles.history.path) : '';
 		Marbles.history.navigate('login'+ redirectPath);
+	},
+
+	__catchInsecurePingResponse: function(httpsArgs) {
+		var httpsXhr = httpsArgs[1], self = this,
+			handleSuccess, handleError;
+
+		handleSuccess = function (httpArgs) {
+			var httpXhr = httpArgs[1];
+			// https did not work but http did...something is wrong with the cert
+			self.Dispatcher.handleAppEvent({
+				name: "HTTPS_CERT_MISSING",
+				status: httpXhr.status
+			});
+			return Promise.reject(self.errCertNotInstalled);
+		};
+		handleError = function (httpArgs) {
+			var httpXhr = httpArgs[1];
+			if (httpArgs.length === 1) {
+				return httpArgs;
+			}
+
+			if (httpXhr.status === 0) {
+				// https is failing as well...service is unavailable
+				self.Dispatcher.handleAppEvent({
+					name: "SERVICE_UNAVAILABLE",
+					status: httpXhr.status
+				});
+				return Promise.reject(self.errServiceUnavailable);
+			}
+			// https did not work but http did without a network error
+			// => missing ssl exception for controller
+			self.Dispatcher.handleAppEvent({
+				name: "HTTPS_CERT_MISSING",
+				status: httpXhr.status
+			});
+			return Promise.reject(self.errCertNotInstalled);
+		};
+
+		if (httpsXhr.status === 0) {
+			// https is unavailable, let's see if http works
+			return self.client.ping("controller", "http").then(handleSuccess).catch(handleError);
+		}
+		// an error code other than 0
+		self.Dispatcher.handleAppEvent({
+			name: "SERVICE_UNAVAILABLE",
+			status: httpsXhr.status
+		});
+		return Promise.reject(this.errServiceUnavailable);
+	},
+
+	__catchSecurePingResponse: function(args) {
+		var xhr = args[1];
+		if (xhr.status === 0) {
+			// We were not able to access the controller due to a network error (ssl, timeout)
+			// In order to understand what's happening, we have to switch to http.
+			this.Dispatcher.handleAppEvent({
+				name: "CONTROLLER_UNREACHABLE_FROM_HTTPS",
+				status: xhr.status
+			});
+			return Promise.reject(new Error("CONTROLLER_UNREACHABLE_FROM_HTTPS"));
+		}
+
+		// an error code other than 0
+		this.Dispatcher.handleAppEvent({
+			name: "SERVICE_UNAVAILABLE",
+			status: xhr.status
+		});
+		return Promise.reject(this.errServiceUnavailable);
+	},
+
+	__successPingResponse: function(args) {
+		var xhr = args[1];
+		if (xhr.status === 200) {
+			window.location.href = window.location.href.replace("http:", "https:");
+		}
+	},
+
+	__isCertInstalled: function() {
+		var self = this;
+		if (window.location.protocol === "https:") {
+			return this.client.ping("controller", "https").catch(this.__catchSecurePingResponse.bind(this));
+		} else {
+			return this.client.ping("controller", "https")
+				.then(this.__successPingResponse.bind(this))
+				.catch(this.__catchInsecurePingResponse.bind(this));
+		}
 	},
 
 	__handleEvent: function (event) {
@@ -95,11 +193,7 @@ window.Dashboard = {
 	__handleAppEvent: function (event) {
 		switch (event.name) {
 			case "CONFIG_READY":
-				if (window.location.protocol === "http:" && this.config.INSTALL_CERT) {
-					window.location.href = window.location.href.replace("http", "https");
-				} else {
-					this.__handleConfigReady();
-				}
+				this.__handleConfigReady();
 			break;
 
 			case "AUTH_CHANGE":
@@ -110,12 +204,17 @@ window.Dashboard = {
 				this.__handleGithubAuthChange(event.authenticated);
 			break;
 
+			case "CONTROLLER_UNREACHABLE_FROM_HTTPS":
+				// Controller isn't accessible via https. Redirect to http and try again.
+				window.location.href = window.location.href.replace("https", "http");
+			break;
+
+			case "HTTPS_CERT_MISSING":
+				Marbles.history.navigate("installcert", {force: true});
+			break;
+
 			case "SERVICE_UNAVAILABLE":
-				if (window.location.protocol === "http:" && this.config.INSTALL_CERT) {
-					this.__handleConfigReady();
-				} else {
-					this.__handleServiceUnavailable(event.status);
-				}
+				this.__handleServiceUnavailable(event.status);
 			break;
 		}
 	},
@@ -132,12 +231,16 @@ window.Dashboard = {
 		this.__renderNavComponent();
 
 		if ( !authenticated && !this.__isLoginPath() ) {
-			var currentHandler = Marbles.history.getHandler();
+			var currentHandler = Marbles.history.getHandler(),
+				self = this;
 			if (currentHandler && currentHandler.opts.auth === false) {
 				// Don't redirect to login from page not requiring auth
 				return;
 			}
-			this.__redirectToLogin();
+
+			this.waitForNav.then(function() {
+				self.__redirectToLogin();
+			});
 		}
 	},
 
--- a/dashboard/app/lib/javascripts/dashboard/config.js
+++ b/dashboard/app/lib/javascripts/dashboard/config.js
@@ -88,7 +88,7 @@ Dashboard.config.fetch = function () {
 
 	Marbles.HTTP({
 		method: 'GET',
-		url: Dashboard.config.API_SERVER + "/config",
+		url: Dashboard.config.API_SERVER.replace(/^https?:/, window.location.protocol) + "/config",
 		middleware: [
 			Marbles.HTTP.Middleware.WithCredentials,
 			Marbles.HTTP.Middleware.SerializeJSON


arekkas:
Oh, I didn't see that you're using tabs instead of whitespaces for identation..I'll fix that

dashboard/app/lib/javascripts/dashboard/client.js
```diff
@@ -97,6 +97,25 @@ Dashboard.Client = Marbles.Utils.createClass({
...
+	ping: function (protocol) {
+		var controllerKey = (Dashboard.config.user || {}).controller_key,
+			endpoint;
+
+		console.log(controllerKey, protocol, this.endpoints, this);
+
+		protocol = protocol === undefined ? 'https' : protocol;
+		endpoint = this.endpoints.cluster_controller.substr(0, 5) === "https" ?
+			protocol + endpoint.substr(5) :
+			protocol + endpoint.substr(4);
```
jvatic: This won't work as `endpoint` isn't assigned a value yet. A regex is the easiest thing here, so `endpoint = this.endpoints.cluster_controller.replace(/^https?/, protocol)`.

jvatic:
`config.endpoints` will be undefined if the config fails to load. We should update [config.js](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/config.js#L18) to add a default endpoints object containing `cluster_controller` in this case.

jvatic:
Or actually, scrap that last comment. We should fetch the config using `window.location.protocol` so that config will succeed even when the cert isn't installed, and remove the [logic](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/main.js#L114-L116) that assumes `config` will fail when the cert needs installing and the [logic](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/main.js#L98-L100) that assumes successfully fetching the config means it's installed (and use the `ping` endpoint for this).

dashboard/app/lib/javascripts/dashboard/client.js
```diff
@@ -97,6 +97,25 @@ Dashboard.Client = Marbles.Utils.createClass({
...
+
+		return this.performRequest('GET', {
+			middleware: [
+				Marbles.HTTP.Middleware.BasicAuth("", controllerKey)
```
jvatic: We don't need to authenticate this request

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -38,6 +38,11 @@ window.Dashboard = {
 			dispatcher: this.Dispatcher,
 			trigger: false
 		});
+
+		this.client.ping().then(function (args) {
+			console.log(args)
+		});
```
jvatic: A better place for this is [in the `CONFIG_READY` event handler below](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/main.js#L97-L103).

arekkas:
> config.endpoints will be undefined if the config fails to load. We should update config.js to add a default endpoints object containing cluster_controller in this case.

Damn, that makes sense... I'll get on to it tomorrow :)

dashboard/app/lib/javascripts/dashboard/client.js
arekkas: is cluster_controller always https?

jvatic: [yes](https://github.com/flynn/flynn/blob/master/dashboard/api.go#L182), so no need even for the regex (`replace("https", protocol)` will work).

arekkas:
`window.location.host` equals `dashboard.dev.localflynn.com`, which the cert exception will be added for in the browser ([example](https://cloud.githubusercontent.com/assets/3372410/7004132/c4ee5906-dc66-11e4-952b-94790e5a3252.PNG)). Instead, we need to check against a different subdomain (e.g. the controller) to reconstruct the CORS use-case. My best guess here is to do a `windows.location.host.replace("dashboard", "controller")` but I'm pretty sure that that's not the way to go here.

I'm not sure what you mean by:
>We should fetch the config using window.location.protocol so that config will succeed even when the cert isn't installed

As far as I can tell, `Dashboard.config` is quite complicated and needs a https request to load all the necessary information. I'm not sure how I should fetch all that information from the

arekkas:
I think this approach doesn't work, we forgot that mixed content will never be loaded:

> Mixed Content: The page at 'https://dashboard.dev.localflynn.com/login' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://controller.dev.localflynn.com/ping'. This request has been blocked; the content must be served over HTTPS.

jvatic:
> I'm not sure what you mean by:
> > We should fetch the config using window.location.protocol so that config will succeed even when the cert isn't installed

`window.location.protocol` will be either `http:` or `https:`, so if we do something like `Dashboard.config.API_SERVER.replace(/^https?:/, window.location.protocol)` [here](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/config.js#L91) then the config request won't fail when the cert isn't installed.

> Mixed Content: The page at 'https://dashboard.dev.localflynn.com/login' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://controller.dev.localflynn.com/ping'. This request has been blocked; the content must be served over HTTPS.

This will only occur when the user has added a security exception for `https://dashboard.CLUSTER_DOMAIN`. As such we should be able to redirect to `http://dashboard.CLUSTER_DOMAIN` when `http://controller.CLUSTER_DOMAIN` fails with a `xhr.status === 0` (we should use `localStorage` to make sure this doesn't turn into an infinite loop for whatever reason though, e.g. browser extension interfering).

jvatic:
There's no way to get around the error page shown by the browser when the user visits `https://dashboard.CLUSTER_DOMAIN` before installing the cert.

arekkas:
> This will only occur when the user has added a security exception for https://dashboard.CLUSTER_DOMAIN. As such we should be able to redirect to http://dashboard.CLUSTER_DOMAIN when http://controller.CLUSTER_DOMAIN fails with a xhr.status === 0 (we should use localStorage to make sure this doesn't turn into an infinite loop for whatever reason though, e.g. browser extension interfering).

Yep, I got a little bit mixed up that's why I deleted the comment :) I've force pushed some changes. I'm currently trying to figure out how to catch an error response in marbles [here](https://github.com/flynn/flynn/pull/1449/files#diff-b93b6be3c2c45e4391edf9ae23796d3aR80), `then` doesn't seem to work.

Is that what you meant?

arekkas:
I didn't touch the config variable yet as I'm not fully aware what implications changes might have there...

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -38,6 +38,7 @@ window.Dashboard = {
 			dispatcher: this.Dispatcher,
 			trigger: false
 		});
+
```
arekkas: I'll remove that..

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +66,24 @@ window.Dashboard = {
...
+			if (a + b !== 200 + 200) {
+				self.dispatcher.handleAppEvent({
+					name: "HTTPS_UNAVAILABLE",
+					status: xhr.status
+				});
+			}
+		};
+
+		self.client.ping("dashboard", "https").then(function (httpArgs) {
+			self.client.ping("controller", "https").then(function (httpsArgs) {
```
arekkas: I'm not sure how to catch a network error here, `then` doesn't seem to work

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -125,6 +148,7 @@ window.Dashboard = {
 		if ( !started ) {
 			this.__started = true;
 			this.run();
+			this.__isCertInstalled();
```
arekkas: `__isCertInstalled()` must be called after run(), because we need the dispatcher as well as the client.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +66,24 @@ window.Dashboard = {
 		Marbles.history.navigate('login'+ redirectPath);
 	},
 
+	__isCertInstalled: function() {
+		var self = this, dispatchIfCertIsMissing = function(a, b) {
+			if (a + b !== 200 + 200) {
```
arekkas: bold statement, because we could get 200 + 404 or something else here which does not imply an ssl/network error.

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: You'll need to use `catch` here instead of `then` (see `performRequest` above).

arekkas: thanks :)

arekkas: that will make error detection much easier, `catch` only gets executed on network errors? Or on 404/500 as well?

dashboard/app/lib/javascripts/dashboard/main.js
arekkas: might be replaced by catch logic

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: [`xhr.status >= 200 && xhr.status < 400`](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/client.js#L48)

jvatic: anything not matched by that ^

arekkas: ok  thanks :)

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: The logic here need to be `a === 200 && b === 0` (`http` ping succeeded so we know the controller is up, `https` failed so we know the cert isn't installed/working).

arekkas: we can't check if the controller is up via http, because the browser is pointed at `https://dashboard.FLYNN_CLUSTER` and it's not possible to request an http resource from an https resource.

arekkas: the only thing we can do is check if we're on `https://dashboard.FLYNN_CLUSTER` and assume, that the cert isn't working when `https://controller.FLYNN_CLUSTER` returns `0`. But it might be a timeout or something else..

arekkas:
Instead of trying to guess if the user installed the certificate correctly, wouldn't it be probably simpler to show a small error message, when requests to the controller fail?

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: Right. We should only request the `http` ping endpoint then when `window.location.protocol` is http.

arekkas: Oh I didn't know that `http` redirects to `https` when the certificate is installed! This clarifies things :) I'll get onto it in the next days

jvatic: If `https` ping fails is we're at `https://dashboard.CLUSTER_DOMAIN` (i.e. `window.location.protocol === 'https:'`), I think it's reasonable to assume the cert isn't installed and we can redirect to `http://dashboard.CLUSTER_DOMAIN`.

jvatic: If both `http` and `https` ping fail then we should fire a `SERVICE_UNAVAILABLE` event.

jvatic:
I think it's a much better user experience if we can correctly deduce the state of the certificate to show the install page for that.

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: :smile:

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +66,65 @@ window.Dashboard = {
...
+		} else {
+			self.client.ping("controller", "https").catch(function (httpsArgs) {
+				var httpsXhr = httpsArgs[1];
+				if (httpsXhr.status === 0) {
+					// https is unavailable, let's see if http works
+					self.client.ping("controller", "http").catch(function (httpArgs) {
+						var httpXhr = httpArgs[1];
+						if (httpXhr.status === 0) {
+							// https is failing as well...service is unavailable
+							self.dispatcher.handleAppEvent({
```
arekkas: I'll clean up nesting here...just wanted to see if that's what you meant

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +66,65 @@ window.Dashboard = {
...
 
+	__isCertInstalled: function() {
+		var self = this;
+		if (window.location.protocol === "https:") {
+			self.client.ping("controller", "https").catch(function (args) {
+				var xhr = args[1];
+				if (xhr.status === 0) {
+					// We were not able to access the controller due to a network error (ssl, timeout)
+					// In order to understand what's happening, we have to switch to http.
+					self.dispatcher.handleAppEvent({
```
arekkas: I get until here, but `self.dispatcher` is not defined. How can I access the dispatcher?

jvatic: Use [`Dashboard.Dispatcher`](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/dispatcher.js#L5)

jvatic: In this case `self.Dispatcher` would also work

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: The logic looks right

arekkas: ok cool, I'll see if it works and clean it up. What do you guys prefer as maximum nesting depth?

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -111,11 +172,22 @@ window.Dashboard = {
...
+//					this.__handleConfigReady();
+//				} else {
 					this.__handleServiceUnavailable(event.status);
-				}
+//				}
+
+			case "CONTROLLER_UNAVAILABLE_FROM_HTTPS":
+				console.log("CONTROLLER_UNAVAILABLE_FROM_HTTPS", window.location.protocol);
+				// Controller isn't accessible via https. Redirect to http and try again.
+				window.location.href = window.location.href.replace("https", "http");
```
jvatic: We should add some logic here using `window.localStorage` to ensure this doesn't become an infinite redirect loop (should treat it as a `SERVICE_UNAVAILABLE` event if the redirect has already occurred in the last 2 seconds or so (play around with the timing to see what works best)).

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: If you move the functions passed to `catch` and `then` into another method on `self` it should be fine. Try to avoid anything more than 5 levels of indentation, I'm more concerned with large blocks of nested code where it's hard to see the outer control statement(s) at a glance.

arekkas: ok

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -111,11 +172,22 @@ window.Dashboard = {
...
+				break;
+
+			case "HTTPS_CERT_MISSING":
+				console.log("HTTPS_CERT_MISSING", window.location.protocol);
+				Marbles.history.navigate("installcert");
+				Marbles.history.loadURL();
```
jvatic: No need to call `loadURL` here as `run` has already been called (the `loadURL` calls in `run` are to load the initial route due to the `trigger: false` option passed to `Marbles.history.start`, though the one called directly after `navigate` isn't necessary as `trigger: false` only effects the initial route).

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -110,12 +177,17 @@ window.Dashboard = {
...
+				// Controller isn't accessible via https. Redirect to http and try again.
+				window.location.href = window.location.href.replace("https", "http");
+			break;
+
+			case "HTTPS_CERT_MISSING":
+				Marbles.history.navigate("installcert");
+			break;
+
 			case "SERVICE_UNAVAILABLE":
-				if (window.location.protocol === "http:" && this.config.INSTALL_CERT) {
```
arekkas: @jvatic I removed this as you suggested. I'm not 100% sure that this does not affect some other issue. Please keep than in mind for the code inspection

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: This should only be called if `this.config.INSTALL_CERT` is `true`.

arekkas:
I think we made it :) As far as I was able to cover the use cases everything worked. No guarantees though ;)

I haven't added the `window.localStorage` yet because I'm not sure if there is a case where this could be necessary. If you think that it is a good idea to have it as a failsafe I'll add it.

arekkas:
Ok, that was to early. There's still something not working. I'll take a closer look into it.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -95,11 +166,7 @@ window.Dashboard = {
 	__handleAppEvent: function (event) {
 		switch (event.name) {
 			case "CONFIG_READY":
-				if (window.location.protocol === "http:" && this.config.INSTALL_CERT) {
-					window.location.href = window.location.href.replace("http", "https");
```
arekkas: I'm not sure this should be removed @jvatic . When requesting `http://dashboard.FLYNN_CLUSTER` I always get redirected to `/installcert` although the certificate is installed correctly. When requesting `https://dashboard.FLYNN_CLUSTER` everything seems to be working fine.

arekkas: As far as I can tell this problem does not originate from the new logic (the events don't get fired) but rather from removing this line. I'm not 100% sure though.

arekkas: Scrap the last comment. I don't think this line is the problem.

jvatic:
@arekkas in answer to [your question in IRC](https://botbot.me/freenode/flynn/2015-04-16/?msg=36781192&page=2), `config.INSTALL_CERT` indicates weather there is a certificate to install. This allows running the dashboard in an http-only configuration for development.

arekkas:
Ah thanks :) I'm still working on the issues. The main problem is that in the original code the loading of the config triggered a SERVICE_UNAVAILABLE which indicated a missing certificate. Since the config can be loaded over http now, this doesn't happen any more - which leads to undesired side effects. I'll figure something out :)

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +65,87 @@ window.Dashboard = {
...
+				name: "SERVICE_UNAVAILABLE",
+				status: xhr.status
+			});
+		}
+	},
+
+	__isCertInstalled: function() {
+		var self = this;
+		if (window.location.protocol === "https:") {
+			this.client.ping("controller", "https").catch(this.__catchControllerResponseFromSecureScope(this));
```
jvatic: Use bind instead and shorten the name a bit, so `.catch(this.__catchSecurePingReponse.bind(this))`. Then you don't have to return a function from these methods.

arekkas: Ah right, I'm so used to go right now that I forgot that there's `bind()` in js :D

jvatic:
The [cert redirect logic in run](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/main.js#L39-L46) can be removed.

Looks like all that's left is getting the login redirect to respect the cert redirect.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +65,87 @@ window.Dashboard = {
...
+		} else {
+			this.client.ping("controller", "https")
+				.catch(this.__catchControllerResponseFromInsecureScope(this))
+				.then(function() {
+					// Dashboard.config.API_SERVER.replace(/^https?:/, window.location.protocol) + "/config"
+					self.client.ping("dashboard", "https").then(function() {
+						// We're on http but actually we should be on https because both dashboard as well as
+						// controller work.
+						window.location.href = window.location.href.replace("http", "https")
```
jvatic: This appears to be breaking things

arekkas:
There's currently no other place where the client is redirected to "/installcert". So if it should be removed from `run()`, where should I move it to?

jvatic:
The `HTTPS_CERT_MISSING` handler currently redirects there.

arekkas:
Oh wow, time to go to bed ;)

arekkas:
> Looks like all that's left is getting the login redirect to respect the cert redirect.

I didn't catch that part. What do you mean exactly?

https://github.com/arekkas/flynn/commit/d2565a2f7b1c23209d6b046fbd4e296f5c12e275 solves all use cases I had:

* Use the dashboard through HTTP.
* Use the dashboard through HTTPS whilst having "dashboard.FLYNN_CLUSTER" on the whitelist but "controller.FLYNN_CLUSTER" is not.
* Use the dashboard through HTTP whilst having "dashboard.FLYNN_CLUSTER" on the whitelist but "controller.FLYNN_CLUSTER" is not.
* Use the dashboard through HTTPS and having the certificate installed globally.
* Use the dashboard through HTTP and having the certificate installed globally.

If I didn't mess up all use cases worked :)

arekkas:
> Looks like all that's left is getting the login redirect to respect the cert redirect.

Ah got it, didn't know login works through HTTP as well

jvatic:
Yup, it does appear to be working, though it's currently a bit glitchy (login page flashes when redirecting to installcert and installcert page flashes before redirecting to HTTPS when cert is installed and navigating to the HTTP version).

To fix this we should be doing the cert check before loading `Marbles.history` in `run` (call after `history.start` and pass `trigger: false` option, make it return a promise and call `Marbles.history.loadURL()` once it completes).

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +60,90 @@ window.Dashboard = {
...
+				status: httpXhr.status
+			});
+		};
+
+		if (httpsXhr.status === 0) {
+			// https is unavailable, let's see if http works
+			self.client.ping("controller", "http").catch(handleError).then(handleSuccess);
+			return;
+		}
+		// We got something else than 0 and it's an error. This results in SERVICE_UNAVAILABLE
```
jvatic: `s/something else than 0 and it's an error/an error code other than 0/` and remove the sentence.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +60,90 @@ window.Dashboard = {
...
+		if (xhr.status === 0) {
+			// We were not able to access the controller due to a network error (ssl, timeout)
+			// In order to understand what's happening, we have to switch to http.
+			this.Dispatcher.handleAppEvent({
+				name: "CONTROLLER_UNREACHABLE_FROM_HTTPS",
+				status: xhr.status
+			});
+			return;
+		}
+		// We got something else than 0 and it's an error. This results in SERVICE_UNAVAILABLE
```
jvatic: ditto

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +60,90 @@ window.Dashboard = {
...
+	},
+
+	__isCertInstalled: function() {
+		var self = this;
+		if (window.location.protocol === "https:") {
+			this.client.ping("controller", "https").catch(this.__catchSecurePingResponse.bind(this));
+		} else {
+			this.client.ping("controller", "https")
+				.catch(this.__catchInsecurePingResponse.bind(this))
+				.then(function(args) {
```
jvatic: Move function into another method

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +60,90 @@ window.Dashboard = {
...
+					var xhr = args[1];
+					if (xhr.status !== 200) {
+						// If we don't make sure that the response was successful, we might end up in an
+						// infinite loop.
+						return;
+					}
+					// TODO this probably should be Dashboard.config.API_SERVER
+					self.client.ping("dashboard", "https").then(function() {
```
jvatic: Is this needed? I think it's safe to assume we can reach the dashboard over https if the controller is reachable.

arekkas:
I refactored it with promises, login seems to be flashing still though.

jvatic:
Looks like the indentation changed to spaces instead of tabs in main.js

arekkas:
Sorry, should be fixed now.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +72,115 @@ window.Dashboard = {
...
+		}
+		window.location.href = window.location.href.replace("http:", "https:");
+		resolve();
+	},
+
+	__isCertInstalled: function() {
+		var self = this;
+		return new Promise(function(resolve, reject) {
+			if (window.location.protocol === "https:") {
+				self.client.ping("controller", "https").catch(function(args) {
```
jvatic: This should be in a `return` statement, otherwise the parent promise will resolve before this is finished.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +72,115 @@ window.Dashboard = {
...
+					self.__catchSecurePingResponse(args, resolve);
+				}).then(function(args){
+					// For some reason then() gets called always?
+					if (args === undefined) {
+						return;
+					}
+					resolve();
+				});
+			} else {
+				self.client.ping("controller", "https").catch(function(args) {
```
jvatic: ditto

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +72,115 @@ window.Dashboard = {
...
+		window.location.href = window.location.href.replace("http:", "https:");
+		resolve();
+	},
+
+	__isCertInstalled: function() {
+		var self = this;
+		return new Promise(function(resolve, reject) {
+			if (window.location.protocol === "https:") {
+				self.client.ping("controller", "https").catch(function(args) {
+					self.__catchSecurePingResponse(args, resolve);
```
jvatic: ditto

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +72,115 @@ window.Dashboard = {
...
+		};
+
+		if (httpsXhr.status === 0) {
+			// https is unavailable, let's see if http works
+			self.client.ping("controller", "http").catch(function(args) {
+				handleError(args, resolve);
+			}).then(function(args){
+				handleSuccess(args, resolve)
+			});
+			return;
```
jvatic: Move this return in front of `self.client.ping` so the resulting promise is returned.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +72,115 @@ window.Dashboard = {
...
+				}).then(function(args){
+					// For some reason then() gets called always?
+					if (args === undefined) {
+						return;
+					}
+					resolve();
+				});
+			} else {
+				self.client.ping("controller", "https").catch(function(args) {
+					self.__catchInsecurePingResponse(args, resolve);
```
jvatic: ditto

dashboard/app/lib/javascripts/dashboard/main.js
arekkas: Ah, is that the reason for https://github.com/flynn/flynn/pull/1449/files#diff-b93b6be3c2c45e4391edf9ae23796d3aR164 ?

jvatic: Actually, I missed that you're using the `resolve`/`reject` of the parent promise.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +72,115 @@ window.Dashboard = {
...
+	},
+
+	__isCertInstalled: function() {
+		var self = this;
+		return new Promise(function(resolve, reject) {
+			if (window.location.protocol === "https:") {
+				self.client.ping("controller", "https").catch(function(args) {
+					self.__catchSecurePingResponse(args, resolve);
+				}).then(function(args){
+					// For some reason then() gets called always?
```
jvatic: The reason for this would be the above `catch`.

dashboard/app/lib/javascripts/dashboard/main.js
arekkas: So no refactor needed?

jvatic: No refactor necessary.

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: So this `then` should be moved before the `catch` and can just be `.then(resolve)`.

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: A refactor would remove the potential for calling `resolve` or `reject` multiple times though.

arekkas: Ok, I'll take a look at it. Never worked with Promises in JS before so it's still kinda magic ;)

jvatic: MDN has a good reference on them: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

arekkas: Ah I see, thanks. :)

arekkas:
Refactored it, works fine :)

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +71,87 @@ window.Dashboard = {
...
+			// => missing ssl exception for controller
+			self.Dispatcher.handleAppEvent({
+				name: "HTTPS_CERT_MISSING",
+				status: httpXhr.status
+			});
+		};
+
+		if (httpsXhr.status === 0) {
+			// https is unavailable, let's see if http works
+			self.client.ping("controller", "http").then(handleSuccess).catch(handleError);
```
jvatic: The resulting promise should be returned

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +71,87 @@ window.Dashboard = {
 		Marbles.history.navigate('login'+ redirectPath);
 	},
 
+	__catchInsecurePingResponse: function(httpsArgs) {
+		var httpsXhr = httpsArgs[1], self = this,
+			handleSuccess, handleError, p;
```
jvatic: Doesn't look like `p` is used

jvatic:
The login view still flashes when the entry point is https with security exception applied.

jvatic:
There shouldn't be a merge commit (rebase against master instead).

jvatic:
Other than that this is looking good :)

arekkas:
> The login view still flashes when the entry point is https with security exception applied.

Do you have an idea how to fix this?

jvatic:
@arekkas looks like you'll need to do the following to clean up the commits as there appears to be changes appended to the merge commit:

```sh
git remote add flynn https://github.com/flynn/flynn.git
git fetch flynn
git diff flynn/master > ~/diff.patch # any path outside the repo
git reset --hard flynn/master
git apply ~/diff.patch
git commit -a
```

arekkas:
That did it :)

jvatic:
Looks like you'll need to rebase on `flynn/master`

jvatic:
The issue with the login view flashing appears to be due to `/login` already being in the URL by the time `CONTROLLER_UNREACHABLE_FROM_HTTPS` is handled. This would be due to the combination of the `AUTH_CHANGE` event being handled, then `loadURL` being called without any other call to `navigate` in between. A solution would be to ensure `loadURL` isn't called for this code path (could use a `Promise.reject()` and remove the `.catch(loadURL)`).

arekkas:
> Looks like you'll need to rebase on flynn/master

```
git pull --rebase origin master
From https://github.com/flynn/flynn
 * branch            master     -> FETCH_HEAD
Current branch is up to date.
```

arekkas:
it's getting kinda messy here with rebase. Should I just open another PR?

jvatic:
Hmm, try updating your master branch then `git rebase master` from the `dashboard-app` branch.

arekkas:
```
$ git rebase master
First, rewinding head to replay your work on top of it...
Applying: dashboard: Should detect SSL problems with CORS
```

```
$ git push arekkas dashboard-app
Everything up-to-date
```

And I can't seem to be able to amend commits any more:

```
$ git commit --amend -a
[dashboard-app-rebase e7c7713] dashboard: Should detect SSL problems with CORS
 Date: Sun Apr 26 17:58:11 2015 +0200
 18 files changed, 269 insertions(+), 213 deletions(-)
$ git push arekkas dashboard-app
Everything up-to-date
```

jvatic:
It looks like you're making changes on the `dashboard-app-rebase` branch instead of `dashboard-app`

arekkas:
Ah right. I solved the flashing login screen by the way. Commit still looks..messy - sorry I've never came much further than using commit and push in git.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +69,87 @@ window.Dashboard = {
...
+		var xhr = args[1];
+		if (xhr.status === 0) {
+			// We were not able to access the controller due to a network error (ssl, timeout)
+			// In order to understand what's happening, we have to switch to http.
+			this.Dispatcher.handleAppEvent({
+				name: "CONTROLLER_UNREACHABLE_FROM_HTTPS",
+				status: xhr.status
+			});
+			console.log(this);
+			promise.reject();
```
jvatic: [`return Promise.reject(new Error("..."))`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject) should work (the name of the above event is fine for the error string).

jvatic:
The login view still flashes because `navigate` doesn't wait for `loadURL` to be called. To get around this we should add promise that resolves once the check cert routine is finished and make `__redirectToLogin` wait for it to resolve before calling `navigate`.

jvatic:
...so something like

```javascript
// in `run`
var resolveNavWait;
this.navWait = new Promise(function (resolve) {
  resolveNavWait = resolve;
});

var loadURL = function () {
  resolve();
  // ...
};
```

```javascript
// in __redirectToLogin
this.__navWait.then(function () {
  // original method body
});
```

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +69,85 @@ window.Dashboard = {
...
+			window.location.href = window.location.href.replace("http:", "https:");
+		}
+	},
+
+	__isCertInstalled: function() {
+		var self = this;
+		if (window.location.protocol === "https:") {
+			return this.client.ping("controller", "https").catch(function(args) {
+				self.__catchSecurePingResponse(args, this);
+			});
```
jvatic: `.catch(this.__catchSecurePingResponse.bind(this))`

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +69,85 @@ window.Dashboard = {
...
+			return self.client.ping("controller", "http").then(handleSuccess).catch(handleError);
+		}
+		// an error code other than 0
+		self.Dispatcher.handleAppEvent({
+			name: "SERVICE_UNAVAILABLE",
+			status: httpsXhr.status
+		});
+	},
+
+	__catchSecurePingResponse: function(args, promise) {
```
jvatic: `promise` argument not used

arekkas:
> The login view still flashes because navigate doesn't wait for loadURL to be called. To get around this we should add promise that resolves once the check cert routine is finished and make __redirectToLogin wait for it to resolve before calling navigate.

These changes produce very random results. On reloading the page I sometimes get redirected to `http://dashboard.dev.localflynn.com/login?redirect=installcert` and sometimes to `http://dashboard.dev.localflynn.com/installcert`

arekkas:
Ok so, when I'm pointing the browser at `http://dashboard.dev.localflynn.com/` I get redirected to `http://dashboard.dev.localflynn.com/login`. When pointing the browser at `http://dashboard.dev.localflynn.com/login` I get redirected to `http://dashboard.dev.localflynn.com/installcert`.

jvatic:
@arekkas can you push these changes (preferably in a separate commit for now)?

arekkas:
Done, I tried putting `waitForNav.then()` in different places but it always resulted in a similar behaviour. What's also problematic: redirectToLogin should be not be blocking on consecutive calls.

arekkas:
My best guess is, that previously `Marbles.history.navigate('login'+ redirectPath)` was called before `Marbles.history.navigate("installcert");`. The delayed call to redirectToLogin reverses that order and causes this behavior.

arekkas:
I think this would be solvable if something like that was possible:

```
Marbles.history.navigate('login'+ redirectPath, false); // state is not directly pushed to window.location
Marbles.history.go(); // window.location.href = 'login' + redirectPath
```

That way it wouldn't matter when we call `.navigate` and could do `.then(Marbles.history.go)`

arekkas:
Otherwise we'd have to delay the `AUTH_CHANGED` event to after `__isCertInstalled()` which is difficult, because `AUTH_CHANGED` get's fired from `this.config` which is in turn required for `__isCertInstalled()` to work.

arekkas:
Scrap the `history.go()` comment as it would be only for this specific case and we should rather fix this by refactoring the AUTH_CHANGED event.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -15,6 +15,16 @@ window.Dashboard = {
...
 
 	run: function () {
+		var self = this, loadURL, resolveWaitForNav;
+		this.waitForNav = new Promise(function(resolve, reject) {
+			resolveWaitForNav = resolve;
+		});
+
+		loadURL = function() {
+			Marbles.history.loadURL();
+			resolveWaitForNav();
```
jvatic: The issue appears to be that this is being called after `HTTPS_CERT_MISSING` event is handled.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +75,83 @@ window.Dashboard = {
...
+	__catchInsecurePingResponse: function(httpsArgs) {
+		var httpsXhr = httpsArgs[1], self = this,
+			handleSuccess, handleError;
+
+		handleSuccess = function (httpArgs) {
+			var httpXhr = httpArgs[1];
+			// https did not work but http did...something is wrong with the cert
+			self.Dispatcher.handleAppEvent({
+				name: "HTTPS_CERT_MISSING",
+				status: httpXhr.status
```
jvatic: If this were to return a `Promise.reject(new Error("..."))`, it should prevent `loadURL` from being called above.

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +75,83 @@ window.Dashboard = {
...
+			});
+		};
+		handleError = function (httpArgs) {
+			var httpXhr = httpArgs[1];
+			if (httpXhr.status === 0) {
+				// https is failing as well...service is unavailable
+				self.Dispatcher.handleAppEvent({
+					name: "SERVICE_UNAVAILABLE",
+					status: httpXhr.status
+				});
```
jvatic: There should be a return here, otherwise both events will fire

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +75,83 @@ window.Dashboard = {
...
+			}
+			// https did not work but http did without a network error
+			// => missing ssl exception for controller
+			self.Dispatcher.handleAppEvent({
+				name: "HTTPS_CERT_MISSING",
+				status: httpXhr.status
+			});
```
jvatic: Ditto for returning a promise reject

dashboard/app/lib/javascripts/dashboard/main.js
jvatic: Hmm, we'll also want to return a reject here so the service unavailable view doesn't redirect to login

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +75,83 @@ window.Dashboard = {
...
+
+		if (httpsXhr.status === 0) {
+			// https is unavailable, let's see if http works
+			return self.client.ping("controller", "http").then(handleSuccess).catch(handleError);
+		}
+		// an error code other than 0
+		self.Dispatcher.handleAppEvent({
+			name: "SERVICE_UNAVAILABLE",
+			status: httpsXhr.status
+		});
```
jvatic: Ditto for a returning a promise reject

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +75,83 @@ window.Dashboard = {
...
+				status: xhr.status
+			});
+			return Promise.reject(new Error("CONTROLLER_UNREACHABLE_FROM_HTTPS"));
+		}
+
+		// an error code other than 0
+		this.Dispatcher.handleAppEvent({
+			name: "SERVICE_UNAVAILABLE",
+			status: xhr.status
+		});
```
jvatic: Ditto for a promise reject

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -65,6 +75,91 @@ window.Dashboard = {
...
+			// https did not work but http did...something is wrong with the cert
+			self.Dispatcher.handleAppEvent({
+				name: "HTTPS_CERT_MISSING",
+				status: httpXhr.status
+			});
+			return Promise.reject(new Error('HTTPS certificate is not trusted.'));
+		};
+		handleError = function (httpArgs) {
+			var httpXhr = httpArgs[1];
+			if (httpXhr === undefined) {
```
jvatic: This block appears to be causing the issue of nothing rendering

arekkas: The problem is that handleError always get's called if handleSuccess got called, because handleSuccess *always* returns a Promise.reject(): `return self.client.ping("controller", "http").then(handleSuccess).catch(handleError);
`

jvatic: Change the if to `if (arguments.length === 1) { return httpArgs; }` as we can safely assume it's an error in this case

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -38,11 +48,11 @@ window.Dashboard = {
 			dispatcher: this.Dispatcher,
 			trigger: false
 		});
-		if (window.location.protocol === "http:" && this.config.INSTALL_CERT) {
-			Marbles.history.navigate("installcert");
-			Marbles.history.loadURL();
+
+		if (this.config.INSTALL_CERT) {
+			this.__isCertInstalled().then(loadURL);
```
jvatic: Try adding a `.catch` here

jvatic: Scrap that actually

dashboard/app/lib/javascripts/dashboard/main.js
```diff
@@ -110,12 +201,17 @@ window.Dashboard = {
 				this.__handleGithubAuthChange(event.authenticated);
 			break;
 
+			case "CONTROLLER_UNREACHABLE_FROM_HTTPS":
+				// Controller isn't accessible via https. Redirect to http and try again.
+				window.location.href = window.location.href.replace("https", "http");
+			break;
+
+			case "HTTPS_CERT_MISSING":
+				Marbles.history.navigate("installcert");
```
jvatic: `Marbles.history.navigate("installcert", {force: true});` so the handler is called even if the `window.location.href` already matches.

jvatic:
LGTM pending squash

arekkas:
:tada:

jvatic:
Thanks!

PR #1453, opened 2015-04-14
test: Dump logs in runner

lmars:
This is so that logs are dumped if the test binary exits abnormally (e.g. due to a `SIGQUIT` timeout).


titanous:
LGTM

PR #1455, opened 2015-04-15
test: Refactored tests to separate asserting success and asserting output containing text

jzila:
OutputContains also validated success, and certain calls from the CLI returned errors with a successful error code. This refactor allows us to separate checking output from checking success, so we can correctly return error codes from the CLI and also check error content.

Tag @lmars since he suggested the refactor in IRC.

### DIFF
--- a/cli/cluster.go
+++ b/cli/cluster.go
@@ -127,8 +127,7 @@ func runClusterDefault(args *docopt.Args) error {
 	}
 
 	if !config.SetDefault(name) {
-		log.Printf("Cluster %q not found.", name)
-		return nil
+		return fmt.Errorf("Cluster %q does not exist and cannot be set as default.", name)
 	}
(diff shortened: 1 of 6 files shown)


cli/cluster.go
```diff
@@ -127,8 +127,7 @@ func runClusterDefault(args *docopt.Args) error {
 	}
 
 	if !config.SetDefault(name) {
-		log.Printf("Cluster %q not found.", name)
-		return nil
+		return errors.New(fmt.Sprintf("Cluster %q does not exist and cannot be set as default.", name))
```
lmars: This should use `fmt.Errorf`

lmars:
@jzila please squash the commits and rebase on master (as CI should be more stable now).

jzila:
@lmars done

lmars:
Thanks!

Issue #1125, opened 2015-03-02
HostSuite.TestVolumePersistence failure


titanous:
```text
FAIL: test_host.go:165: HostSuite.TestVolumePersistence

test_host.go:208:
    ...
test_host.go:176:
    t.Assert(h.DestroyVolume(vol.ID), c.IsNil)
... value httphelper.JSONError = httphelper.JSONError{Code:"unknown_error", Message:"Something went wrong", Detail:json.RawMessage(nil)} ("unknown_error: Something went wrong")
```

https://ci.flynn.io/builds/20150302072334-e3c6b69f

titanous:
Another: https://ci.flynn.io/builds/20150302073328-bc796f9a

titanous:
https://ci.flynn.io/builds/20150308041919-660b24af

jvatic:
https://ci.flynn.io/builds/20150313234344-23cc0d7f

jvatic:
https://ci.flynn.io/builds/20150317223236-e8235721

jvatic:
https://ci.flynn.io/builds/20150319205453-fa9e3890

jvatic:
https://ci.flynn.io/builds/20150322135053-0f5e44e2

lmars:
The error here is that the volume being destroyed is in use, for example in [20150322135053-0f5e44e2](https://ci.flynn.io/builds/20150322135053-0f5e44e2):

```
t=2015-03-22T13:54:16+0000 lvl=info msg="request started" component=host req_id=4e77532a9af142cd87e161e6c03b2a12 method=DELETE path=/storage/volumes/7b0852b8441c414196b2a931e530f6d4 client_ip=10.69.33.3
t=2015-03-22T13:54:16+0000 lvl=dbug msg="zfs destroy -f flynn-default/7b0852b8441c414196b2a931e530f6d4" package=zfs
t=2015-03-22T13:54:18+0000 lvl=dbug msg="zfs destroy -f flynn-default/7b0852b8441c414196b2a931e530f6d4" package=zfs
t=2015-03-22T13:54:19+0000 lvl=dbug msg="zfs destroy -f flynn-default/7b0852b8441c414196b2a931e530f6d4" package=zfs
t=2015-03-22T13:54:20+0000 lvl=dbug msg="zfs destroy -f flynn-default/7b0852b8441c414196b2a931e530f6d4" package=zfs
[...]

PR #1462, opened 2015-04-16
test: Skip TestVolumePersistence

lmars:
This test fails quite regularly in CI, we know the fix, and this area of code (i.e. volume management) isn't currently being worked on, so skip for now until we fix the underlying issue.

Ref #1125.


titanous:
LGTM

PR #1470, opened 2015-04-20
controller: Auto resume job and deployment event streams

lmars:
Extracted from #1376.

### DIFF
--- a/pkg/httpclient/json.go
+++ b/pkg/httpclient/json.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/flynn/flynn/pkg/httphelper"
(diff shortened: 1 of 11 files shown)


lmars:
I've realised this can cause races. For example, I have modified `StreamJobEvents` to return just the underlying stream, with the first request executing in the `ResumingStream` goroutine, so events can be missed in code like:

```
stream := client.StreamJobEvents(...)
client.PutFormation(...)
```

I plan to ensure the first request happens before returning.

lmars:
Ok race fixed, please review.

pkg/httpclient/json.go
```diff
@@ -178,6 +179,17 @@ func (c *Client) Stream(method, path string, in, out interface{}) (stream.Stream
 	return c.StreamWithHeader(method, path, make(http.Header), in, out)
 }
 
+func (c *Client) ResumingStream(method, path string, ch interface{}) (stream.Stream, error) {
```
titanous: I wonder if `ReconnectingStream` is better? I don't have a strong opinion.

pkg/httpclient/stream.go
```diff
@@ -69,3 +79,64 @@ func Stream(res *http.Response, outputCh interface{}) stream.Stream {
...
+	Total: 10 * time.Second,
+	Delay: 100 * time.Millisecond,
+}
+
+func ResumingStream(connect func(int64) (*http.Response, error), outputCh interface{}) (stream.Stream, error) {
+	stream := stream.New()
+	outValue := reflect.ValueOf(outputCh)
+	stopChanValue := reflect.ValueOf(stream.StopCh)
+	firstErr := make(chan error)
+	var once sync.Once
```
titanous: Everything but `stream` and `firstErr` can be in the anonymous function below.

pkg/httpclient/json.go
lmars: I think I prefer `ResumingStream` to indicate it is resuming from the last ID, rather than just reconnecting. Will leave as is, unless your opinions strengthen.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #1471, opened 2015-04-22
installer: cluster management

jvatic:
screen shot 2015-04-22 at 11 39 20
screen shot 2015-04-22 at 11 40 00


installer/aws_cluster.go
```diff
@@ -0,0 +1,574 @@
...
+	}()
+}
+
+func (c *AWSCluster) Delete() {
+	c.cf = cloudformation.New(c.creds, c.Region, nil)
+	go func() {
+		stackEventsSince := time.Now()
+		c.cluster.setState("deleting")
+		if err := c.fetchStack(); err != StackNotFoundError {
+			fmt.Println(err)
```
titanous: This looks like a debug print?

installer/aws_cluster.go
```diff
@@ -0,0 +1,574 @@
...
+	c.ImageID = imageID
+	if err := c.saveField("ImageID", imageID); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *AWSCluster) fetchLatestEC2Images() ([]*release.EC2Image, error) {
+	client := &http.Client{}
+	resp, err := client.Get("https://dl.flynn.io/ec2/images.json")
```
titanous: No need, for the `client`, just do `http.Get`

installer/aws_cluster.go
```diff
@@ -0,0 +1,574 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.New(fmt.Sprintf("Failed to fetch list of flynn images: %s", resp.Status))
+	}
+	dec := json.NewDecoder(resp.Body)
+	manifest := release.EC2Manifest{}
+	err = dec.Decode(&manifest)
+	if err != nil {
```
titanous:
```go
if err := json.NewDecoder(resp.Body).Decode(&manifest); err != nil {
```

installer/aws_cluster.go
```diff
@@ -0,0 +1,574 @@
...
+	return nil
+}
+
+func (c *AWSCluster) fetchLatestEC2Images() ([]*release.EC2Image, error) {
+	client := &http.Client{}
+	resp, err := client.Get("https://dl.flynn.io/ec2/images.json")
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
```
titanous:
```go
defer resp.Body.Close()
```

installer/aws_cluster.go
titanous: s/resp/res

installer/events.go
```diff
@@ -0,0 +1,273 @@
...
+		cluster: c,
+	})
+	return res.Input
+}
+
+func (c *Cluster) sendEvent(event *Event) {
+	c.installer.SendEvent(event)
+}
+
+func (c *Cluster) SendInstallLogEvent(description string) {
```
titanous: This is a mouthful, perhaps shorten to `SendLog` or something like that?

installer/aws_cluster.go
```diff
@@ -0,0 +1,574 @@
...
+func (e StackEventSort) Less(i, j int) bool {
+	return e[j].Timestamp.After(e[i].Timestamp)
+}
+
+func (c *AWSCluster) waitForStackCompletion(action string, after time.Time) error {
+	stackID := aws.String(c.StackID)
+
+	actionCompleteSuffix := "_COMPLETE"
+	actionFailureSuffix := "_FAILED"
+	actionInProgressSuffix := "_IN_PROGRESS"
```
titanous: These should all be `const` (they can stay in the method body)

installer/aws_cluster.go
```diff
@@ -0,0 +1,574 @@
...
+	isComplete := false
+	isFailed := false
+
+	stackEvents := make([]cloudformation.StackEvent, 0)
```
titanous: Use `var` if you're not allocating space.

installer/cluster.go
```diff
@@ -0,0 +1,410 @@
...
+		fields = append(fields, c.ID)
+		for i, ip := range c.InstanceIPs {
+			insertStmt += fmt.Sprintf("INSERT INTO instances (ClusterID, IP) VALUES ($1, $%d);\n", i+2)
+			fields = append(fields, ip)
+		}
+		q := fmt.Sprintf(`
+    %s
+    `, insertStmt)
+		_, err = tx.Exec(q, fields...)
+	case "Domain":
```
titanous: This is weird, I think it's better served by separate methods to do `Domain` and `InstanceIPs`.

installer/cluster.go
```diff
@@ -0,0 +1,410 @@
...
+		return err
+	}
+
+	switch field {
+	case "InstanceIPs":
+		insertStmt := ""
+		fields := make([]interface{}, 0, len(c.InstanceIPs)+1)
+		fields = append(fields, c.ID)
+		for i, ip := range c.InstanceIPs {
+			insertStmt += fmt.Sprintf("INSERT INTO instances (ClusterID, IP) VALUES ($1, $%d);\n", i+2)
```
titanous: Why not multiple exec calls instead of appending strings?

installer/cluster.go
```diff
@@ -0,0 +1,410 @@
...
+
+	if _, err = tx.Exec(`DELETE FROM instances WHERE ClusterID == $1`, c.ID); err != nil {
+		return
+	}
+
+	if _, err = tx.Exec(`DELETE FROM clusters WHERE ID == $1`, c.ID); err != nil {
+		return
+	}
+
+	if _, err = tx.Exec(`DELETE FROM aws_clusters WHERE ClusterID == $1`, c.ID); err != nil {
```
titanous: Probably better to soft delete.

installer/http.go
```diff
@@ -365,87 +157,69 @@ func (api *httpAPI) InstallHandler(w http.ResponseWriter, req *http.Request, par
...
-	api.InstallerStackMtx.Lock()
-	defer api.InstallerStackMtx.Unlock()
-	id := params.ByName("id")
-	s := api.InstallerStacks[id]
-	if s == nil {
-		w.WriteHeader(404)
+func (api *httpAPI) DeleteCluster(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
+	if err := api.Installer.DeleteCluster(params.ByName("id")); err != nil {
+		if err == ClusterNotFoundError {
+			w.WriteHeader(404)
```
titanous: This should be an object not found error via `Error`

installer/installer.go
```diff
@@ -1,874 +1,263 @@
...
-		s.SendEvent(fmt.Sprintf("Using saved key pair (%s)", s.SSHKeyName))
-		return nil
-	}
-
-	s.SendEvent("Creating key pair")
-	keypair, err := sshkeygen.Generate()
+	list, err := ql.Compile(fmt.Sprintf(`
+		INSERT INTO clusters VALUES (%s);
+		INSERT INTO aws_clusters VALUES(%s);
+	`, strings.Join(clustersVStr, ", "), strings.Join(awsVStr, ", ")))
```
titanous: Just do two separate execs.

installer/installer.go
```diff
@@ -1,874 +1,263 @@
...
-	var value string
-	for _, o := range s.Stack.Outputs {
-		if *o.OutputKey == name {
-			value = *o.OutputValue
-			break
+func (i *Installer) ClusterDeleted(id string) {
+	i.clustersMtx.Lock()
+	clusters := make([]interface{}, 0, len(i.clusters))
+	for _, c := range i.clusters {
+		idValue := reflect.Indirect(reflect.ValueOf(c)).FieldByName("ID")
```
titanous: This is sketchy, can you use an interface to do it?

installer/schema.go
```diff
@@ -0,0 +1,114 @@
...
+	NumInstances        int64             `json:"num_instances"`
+	ControllerKey       string            `json:"controller_key,omitempty"`
+	ControllerPin       string            `json:"controller_pin,omitempty"`
+	DashboardLoginToken string            `json:"dashboard_login_token,omitempty"`
+	Domain              *Domain           `json:"domain" ql:"-"`
+	CACert              string            `json:"ca_cert"`
+	SSHKey              *sshkeygen.SSHKey `json:"-" ql:"-"`
+	SSHKeyName          string            `json:"ssh_key_name,omitempty"`
+	VpcCidr             string            `json:"vpc_cidr_block,omitempty"`
+	SubnetCidr          string            `json:"subnet_cidr_block,omitempty"`
```
titanous: s/Cidr/CIDR/

titanous: Ditto for the rest in this file.

jvatic:
@titanous comments addressed

jvatic:
One thing that needs addressing is the aws rate limit

installer/aws_cluster.go
```diff
@@ -0,0 +1,605 @@
...
+	const maxBackoff = 10 * time.Second
+	backoff := 1 * time.Second
+	for {
+		err := runRequest()
+		if err != nil && err.Error() == rateExceededErrStr {
+			time.Sleep(backoff)
+			backoff = backoff * 2
+			if backoff > maxBackoff {
+				backoff = maxBackoff
+			}
```
titanous: This should eventually time out.

installer/events.go
```diff
@@ -0,0 +1,273 @@
...
+    INSERT INTO %s VALUES(%s);
+	`, tableName, strings.Join(vStr, ", ")))
+	if err != nil {
+		return err
+	}
+	tx, err := i.db.Begin()
+	if err != nil {
+		return err
+	}
+	_, err = tx.Exec(list.String(), fields...)
```
titanous: Is this transaction required? There is only one statement.

titanous: Looks like there are a few other similar transactions that only have one statement.

jvatic: Database mutations require a transaction, but I think they are implicitly added so should be able to just call `db.Exec`.

installer/aws_cluster.go
jvatic: Testing this with around 6 parallel installs, the backoff rarely exceeded `8s` (mostly `2s` and `4s`). Does a timeout after 34 seconds (5 retries) sound reasonable?

titanous: Yeah, sounds good.

installer/events.go
jvatic: Actually, looks like the transaction is needed.

jvatic:
@titanous comments addressed

installer/aws_cluster.go
```diff
@@ -0,0 +1,609 @@
...
+	c.base.installer.dbMtx.Lock()
+	defer c.base.installer.dbMtx.Unlock()
+
+	tx, err := c.base.installer.db.Begin()
+	if err != nil {
+		return err
+	}
+	_, err = tx.Exec(fmt.Sprintf(`
+  UPDATE aws_clusters SET %s = $2 WHERE ClusterID == $1
+  `, field), c.ClusterID, value)
```
titanous: Single statement transaction?

jvatic: ql gives an error for any mutation attempted outside a transaction, would it be better to make that part of the exec string?

titanous:
LGTM (squash before merge)

PR #1474, opened 2015-04-22
Receiver now waits for default formation scale before returning

### LINKED ISSUE #1260: receiver: Initial push should wait for web service to start
The receiver creates an initial `web=1` formation for the initial push of code with a `web` process type (see [here](https://github.com/flynn/flynn/blob/master/receiver/flynn-receive.go#L134-L147)), but it does not wait for the web process to start.

This is different to the behaviour of subsequent pushes which call `controller.DeployAppRelease` and so do wait for services to start (because it [waits for the deploy to complete](https://github.com/flynn/flynn/blob/master/controller/client/client.go#L351-L391)).

This caused an error in [this CI build](https://s3.amazonaws.com/flynn-ci-logs/20150314164949-4e5f4132-build-d3812afe0a52f8dc8e0d745a6b427fd3fb27fe3e-2015-03-14-16-59-38.txt) when scaling an app down because an up event from the initial push was received, thus failing the scale as it was expecting only down events.

jzila:
Fixes #1260

### DIFF
--- a/receiver/flynn-receive.go
+++ b/receiver/flynn-receive.go
@@ -35,6 +35,7 @@ func init() {
 var typesPattern = regexp.MustCompile("types.* -> (.+)\n")
 
 const blobstoreURL = "http://blobstore.discoverd"
+const scaleTimeout = 20 * time.Second
 
 func main() {
 	client, err := controller.NewClient("", os.Getenv("CONTROLLER_KEY"))
@@ -137,11 +138,31 @@ func main() {
 			ReleaseID: release.ID,
 			Processes: map[string]int{"web": 1},
 		}
+
+		watcher, err := client.WatchJobEvents(app.ID, release.ID)
+		if err != nil {
+			log.Fatalln("Error streaming job events", err)
+			return
+		}
+		defer watcher.Close()
+
 		if err := client.PutFormation(formation); err != nil {
 			log.Fatalln("Error putting formation:", err)
 		}
+		fmt.Println("=====> Waiting for web job to start...")
 
-		fmt.Println("=====> Added default web=1 formation")
+		err = watcher.WaitFor(ct.JobEvents{"web": {"up": 1}}, scaleTimeout, func(e *ct.JobEvent) error {
+			switch e.Job.State {
+			case "up":
+				fmt.Println("=====> Default web formation scaled to 1")
+			case "down", "crashed":
+				return fmt.Errorf("Failed to scale web process type")
+			}
+			return nil
+		})
+		if err != nil {
+			log.Fatalln(err.Error())
+		}
 	}
 }
 
--- a/test/test_git_deploy.go
+++ b/test/test_git_deploy.go
@@ -130,15 +130,17 @@ func (s *GitDeploySuite) runBuildpackTestWithResponsePattern(t *c.C, name string
 		t.Assert(r.flynn("resource", "add", resource), Succeeds)
 	}
 
-	watcher, err := s.controllerClient(t).WatchJobEvents(name)
+	watcher, err := s.controllerClient(t).WatchJobEvents(name, "")
 	t.Assert(err, c.IsNil)
 	defer watcher.Close()
 
 	push := r.git("push", "flynn", "master")
 	t.Assert(push, SuccessfulOutputContains, "Creating release")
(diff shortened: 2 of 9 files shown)


receiver/flynn-receive.go
```diff
@@ -139,11 +140,48 @@ func main() {
...
+						log.Fatalln("Error streaming job events", err)
+					} else {
+						log.Fatalln("Error streaming job events")
+					}
+					break loop
+				}
+				// ignore one-off jobs or starting events
+				if e.Job.State == "starting" || e.Job.Type == "" {
+					continue loop
+				}
```
lmars: You should check that the event is for a job from the new release.

receiver/flynn-receive.go
```diff
@@ -139,11 +140,48 @@ func main() {
...
+				switch e.Job.State {
+				case "up":
+					fmt.Println("=====> Default web formation scaled to 1")
+				case "down", "crashed":
+					log.Fatalln("Failed to scale web process type")
+				}
+				break loop
+			case <-time.After(scaleTimeout):
+				log.Fatalln("timed out waiting for scale events")
```
lmars: I think something like the following is more informative:

```
timed out waiting for web job to start (waited 20s)
```

receiver/flynn-receive.go
```diff
@@ -139,11 +140,48 @@ func main() {
...
+		}
+		defer stream.Close()
+
 		if err := client.PutFormation(formation); err != nil {
 			log.Fatalln("Error putting formation:", err)
 		}
-
 		fmt.Println("=====> Added default web=1 formation")
+
+	loop:
```
lmars: I think it would be good to inform the user what is happening, e.g.:

```
=====> Waiting for web job to start...
```

jzila:
@lmars looks like we're good on this. LMK if the code looks good and I'll squash the commits down.

test/test_git_deploy.go
```diff
@@ -130,15 +130,19 @@ func (s *GitDeploySuite) runBuildpackTestWithResponsePattern(t *c.C, name string
...
 
 	push := r.git("push", "flynn", "master")
 	t.Assert(push, SuccessfulOutputContains, "Creating release")
 	t.Assert(push, SuccessfulOutputContains, "Application deployed")
-	t.Assert(push, SuccessfulOutputContains, "Added default web=1 formation")
+	t.Assert(push, SuccessfulOutputContains, "Default web formation scaled to 1")
 	t.Assert(push, SuccessfulOutputContains, "* [new branch]      master -> master")
+	t.Assert(push, c.Not(OutputContains), "timed out waiting for scale")
+	if len(resources) == 0 {
+		t.Assert(push, SuccessfulOutputContains, "=====> Default web formation scaled to 1")
```
lmars: This is a duplication of the check on line 140

lmars:
LGTM pending minor comment

test/test_git_deploy.go
jzila: Fixed the check on 140.

jzila:
@lmars minor comment addressed, and commits squashed. CI passed, ready for merge.

test/test_git_deploy.go
```diff
@@ -130,15 +130,19 @@ func (s *GitDeploySuite) runBuildpackTestWithResponsePattern(t *c.C, name string
...
 	push := r.git("push", "flynn", "master")
 	t.Assert(push, SuccessfulOutputContains, "Creating release")
 	t.Assert(push, SuccessfulOutputContains, "Application deployed")
-	t.Assert(push, SuccessfulOutputContains, "Added default web=1 formation")
+	t.Assert(push, SuccessfulOutputContains, "Waiting for web job to start...")
 	t.Assert(push, SuccessfulOutputContains, "* [new branch]      master -> master")
+	t.Assert(push, c.Not(OutputContains), "timed out waiting for scale")
+	if len(resources) == 0 {
+		t.Assert(push, SuccessfulOutputContains, "=====> Default web formation scaled to 1")
+	}
```
lmars: There is no longer a need for the `len(resources) == 0` constraint (since #1520), this assertion should be true for all the buildpacks.

jzila:
@lmars OK we're good here.

PR #1480, opened 2015-04-26
test: Timeout building Flynn after 30 minutes

lmars:
Building Flynn blocked a CI job. The VM was pegging the CPU and I couldn't get an SSH shell so unsure of the cause.


titanous:
LGTM

PR #1481, opened 2015-04-26
test: Increase build VM RAM to 4GB

lmars:
Building the CLI requires a large amount of RAM due to the embedded installer.


titanous:
LGTM

PR #1482, opened 2015-04-27
discoverd: Decrease heartbeat interval on error

### LINKED ISSUE #1369: discoverd should retry heartbeats
Currently the discoverd client doesn't attempt to retry heartbeats when they fail, it just waits until the next scheduled heartbeat. We should retry in case the failure was transient.

lmars:
Closes #1369.


titanous:
LGTM

PR #1495, opened 2015-05-04
Makefile: Don’t clean up if there is no reason to

titanous:
`xargs` runs the command once by default even if there is no input, which results in an error from `rm`, because no arguments are provided.


lmars:
LGTM

PR #1499, opened 2015-05-05
host: Stop setting the libvirt VCPU option

lmars:
The `VCPU` option has no effect on containers.

Under libvirt, container CPU restrictions should be specified using "cputune" options:

https://libvirt.org/formatdomain.html#elementsCPUTuning

I have just removed the VCPU option rather than replacing with appropriate CPU tune options as it is unclear how we will expose the parameters to users.


titanous:
LGTM

Issue #1496, opened 2015-05-04
Resource specification


lmars:
Job resource limits are currently either non-existent or hard-coded (e.g. [available memory](https://github.com/flynn/flynn/blob/master/host/libvirt_lxc_backend.go#L567), [available CPUs](https://github.com/flynn/flynn/blob/master/host/libvirt_lxc_backend.go#L568), [max fd and processes](https://github.com/flynn/flynn/blob/master/host/containerinit/init.go#L416-L419)).

Expanding on #1230, it should be easy for users to specify values for resource limits, including (but not limited to):

* available memory
* available swap
* available CPUs
* CPU time (either hard limit or relative share)
* block I/O time (either hard limit or weighted)
* max file descriptor
* max number of processes

As well as providing a hard upper limit, it would also be useful to provide similar values which can be used by the scheduler to decide where to place jobs based on available resources across the cluster (and also act as a soft limit for `rlimit` style limits).

With this in mind, I propose a structure similar to the [Kubernetes Resource Specification](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/resources.md#resource-specifications):

```go
type Spec struct {
	// Request is the amount of resource a job expects to consume, so the
	// job should only be placed on a host with at least this amount of
	// resource available, and once scheduled this amount of resource
	// should then be unavailable on the given host.
	Request int64

	// Limit is an upper limit on the amount of resource a job can consume,
	// the outcome of hitting this limit being implementation defined (e.g.
	// a system error, throttling, catchable / uncatchable signals etc.)
	Limit int64
}
```

There would be a hard-coded list of resources with default specs, e.g.:

```go
type Type string

const (
	TypeMemory  Type = "memory"
	TypeNumCPUs Type = "num_cpus"
	TypeMaxFD   Type = "max_fd"
	...
)

var Defaults = map[Type]Spec{
	TypeMemory:  {Request: 1*GB, Limit: 2*GB},
	TypeNumCPUs: {Request: 1,    Limit: 2},
	...
}
```

In the controller, resource specs would be declared at the individual process level, e.g.:

```go
type ProcessType struct {
	...
	Resources map[resource.Type]resource.Spec
}
```

and these would be transferred to the job, which the container runner would use to impose the limits:

```go
type Job struct {
	...
	Resources map[resource.Type]resource.Spec
}
```

Hosts would also expose their available resources to aid in scheduling decisions:

```go
type Host struct {
	...
	AvailableResources map[resource.Type]int64
[...]

PR #1500, opened 2015-05-05
all: Add resource specifications

lmars:
This implements what was proposed in #1496 except for hosts declaring available resources.

You will notice calls to `resource.SetDefaults` in the controller / host code which I am using to explicitly fill in the resource specifications (rather than blank ones meaning default) so that it is clear to the end user from job / release config what the limits are, defaults or not.

### DIFF
--- a/bootstrap/run_app_action.go
+++ b/bootstrap/run_app_action.go
@@ -9,6 +9,7 @@ import (
 
 	ct "github.com/flynn/flynn/controller/types"
 	"github.com/flynn/flynn/controller/utils"
+	hostresource "github.com/flynn/flynn/host/resource"
 	"github.com/flynn/flynn/host/types"
 	"github.com/flynn/flynn/pkg/cluster"
 	"github.com/flynn/flynn/pkg/random"
@@ -124,6 +125,7 @@ func (a *RunAppAction) Run(s *State) error {
 		for i := 0; i < count; i++ {
 			hostID := hosts[i%len(hosts)].ID
 			config := utils.JobConfig(a.ExpandedFormation, typ, hostID)
+			hostresource.SetDefaults(&config.Resources)
 			if a.ExpandedFormation.Release.Processes[typ].Data {
 				if err := utils.ProvisionVolume(cc, hostID, config); err != nil {
 					return err
--- a/controller/log_test.go
+++ b/controller/log_test.go
@@ -14,6 +14,7 @@ import (
 
 	ct "github.com/flynn/flynn/controller/types"
 	logaggc "github.com/flynn/flynn/logaggregator/client"
+	"github.com/flynn/flynn/pkg/typeconv"
 
 	. "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
 )
@@ -116,10 +117,6 @@ func (f *fakeLogAggregatorClient) GetLog(channelID string, options *logaggc.LogO
 	return pr, nil
 }
 
-func strPtr(s string) *string { return &s }
-
-func intPtr(i int) *int { return &i }
-
 func (s *S) TestGetAppLog(c *C) {
 	app := s.createTestApp(c, &ct.App{Name: "get-app-log-test"})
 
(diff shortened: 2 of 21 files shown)


bootstrap/run_app_action.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	ct "github.com/flynn/flynn/controller/types"
 	"github.com/flynn/flynn/controller/utils"
+	r "github.com/flynn/flynn/host/resource"
```
titanous: I'm not sold on calling this `r`, perhaps something a bit longer?

host/containerinit/init.go
```diff
@@ -399,29 +401,39 @@ func setupCommon(c *Config, log log15.Logger) error {
...
-
-func setupLimits(log log15.Logger) {
-	type rlimit struct {
-		resource int
-		value    uint64
-	}
+const (
+	RLIMIT_NPROC          = 6
+	defaultMaxFD    int64 = 10000
+	defaultMaxProcs int64 = 256
```
titanous: Since we have defaults higher in the stack, we probably don't need to duplicate them here and can leave them to default if not specified.

host/resource/resource.go
```diff
@@ -0,0 +1,54 @@
...
+	// the outcome of hitting this limit being implementation defined (e.g.
+	// a system error, throttling, catchable / uncatchable signals etc.)
+	Limit int64 `json:"limit"`
+}
+
+type Type string
+
+const (
+	TypeMemory   Type = "memory"
+	TypeMaxFD    Type = "max_fd"
```
titanous: Probably should be plural: `TypeMaxFDs`

host/resource/resource.go
```diff
@@ -0,0 +1,54 @@
...
+
+func Defaults() Resources {
+	r := make(Resources)
+	SetDefaults(&r)
+	return r
+}
+
+func SetDefaults(r *Resources) {
+	if *r == nil {
+		*r = make(Resources)
```
titanous: Add a `len(defaults)` capacity hint.

host/resource/resource.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	for typ, s := range defaults {
+		spec := (*r)[typ]
+		if spec.Limit == 0 {
+			spec.Limit = s.Limit
+		}
+		if spec.Request == 0 {
```
titanous: Zero seems like a valid value for some resource types. Does it need to be a pointer?

host/resource/resource.go
lmars: This limit is a wrapper around `RLIMIT_NOFILE` which limits the maximum file descriptor number, rather than the maximum number of file descriptors, hence the singular.

From the man page:

```
RLIMIT_NOFILE
	Specifies a value one greater than the maximum file descriptor number
	that can be opened by this process. Attempts (open(2), pipe(2), dup(2),
	etc.) to exceed this limit yield the error EMFILE.
```

For example:

```
$ ulimit -n
1024
$ echo 1023>/dev/null

$ echo 1024>/dev/null                                                     
bash: 1024: Bad file descriptor
$ ulimit -n 2048
$ echo 1024>/dev/null

$
```

Will add a comment.

lmars:
@titanous comments addressed

host/resource/resource.go
```diff
@@ -0,0 +1,63 @@
...
+	// Limit is an upper limit on the amount of resource a job can consume,
+	// the outcome of hitting this limit being implementation defined (e.g.
+	// a system error, throttling, catchable / uncatchable signals etc.)
+	Limit int64 `json:"limit"`
+}
+
+type Type string
+
+const (
+	// TypeMemory specifies the available memory inside a container.
```
titanous: `in bytes`

host/resource/resource.go
```diff
@@ -0,0 +1,63 @@
...
+	// Request, if set, is the amount of resource a job expects to consume,
+	// so the job should only be placed on a host with at least this amount
+	// of resource available, and once scheduled this amount of resource
+	// should then be unavailable on the given host.
+	Request *int64 `json:"request"`
+
+	// Limit is an upper limit on the amount of resource a job can consume,
+	// the outcome of hitting this limit being implementation defined (e.g.
+	// a system error, throttling, catchable / uncatchable signals etc.)
+	Limit int64 `json:"limit"`
```
titanous: I think zero is a valid value for `Limit` too?

lmars:
@titanous comments addressed (I introduced a `typeconv` package to help creating `*int64` values)

pkg/typeconv/typeconv.go
```diff
@@ -0,0 +1,5 @@
+package typeconv
+
+func IntPtr(i int) *int       { return &i }
+func Int64Ptr(i int64) *int64 { return &i }
+func StrPtr(s string) *string { return &s }
```
titanous: `StringPointer`

controller/log_test.go
```diff
@@ -12,10 +12,10 @@ import (
 	"sync"
 	"time"
 
+	. "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
 	ct "github.com/flynn/flynn/controller/types"
 	logaggc "github.com/flynn/flynn/logaggregator/client"
-
-	. "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
+	. "github.com/flynn/flynn/pkg/typeconv"
```
titanous: I don't think we should add more dot-imports other than gocheck. https://code.google.com/p/go-wiki/wiki/CodeReviewComments#Import_Dot

lmars:
@titanous comments addressed

titanous:
LGTM

Issue #1484, opened 2015-04-29
Scheduler V2


lmars:
The current scheduler implementation feels slightly hacked together and there are a number of open issues relating to it's short comings (#1182, #1434, #686, #129, #54, #1340).

We should refactor the scheduler to both fix these issues and make it more flexible to handle different scheduling policies (e.g. resource based, tag limiting etc.).

Here are the main responsibilities of the scheduler:

* Keep an accurate state of running jobs in memory (being handled in #1373), and ensure it matches the requested scale
* React to formation changes by adding / removing jobs from hosts
* Restart crashed jobs using an exponential backoff policy
* Update the `jobs_cache` DB table with job states (including ones scheduled to start in the future)

Some improvements to be made based on the current implementation:

* Support arbitrary constraints when running new jobs (e.g. available resources, host tags, data locality etc.)
* Actively keep running jobs in sync with formations, using a deterministic algorithm to determine which jobs to remove if necessary (so if multiple schedulers are scheduling, they will both start new jobs then race to stop the same duplicate job, thus minimising the impact of #1182)
* Actively keep `jobs_cache` table in sync with cluster state
* When the leader crashes, maintain the backoff policy from the crashed scheduler
* Simplify "omni" jobs by modifying formation process counts on host change so that they always match the actual number of jobs of that type (it is confusing that `scheduler=2` could mean "2 jobs per host" rather than just "2 jobs in the cluster")
* Make unit testing simpler (all scheduler testing is currently at the integration level)

/cc @titanous

heavenlyhash:
Re: "if multiple schedulers are scheduling", maybe it would be easier to maintain consistency if schedulers only take actions driven by consulting (and immediately atomically updating) the actual consensus system?  If schedulers become stateless outside of the consensus state, then it doesn't matter how many of them there are participating.  After all, as was wisely said in #1373, making it "appear that there is a consistent view of the cluster, when that is not actually true" is a source of complications.  It might be possible to engineer schedulers to have local state with inconsistent views of the world and anneal toward consistency anyway, but it sounds a lot harder.  Rather than guiding schedulers towards races in their work, the race can be shuffled to transactions in the consensus system, and as a bonus, become more monitorable.

lmars:
[...]

PR #1502, opened 2015-05-06
Refactored job event waiting to use a new JobWatcher abstraction

### LINKED ISSUE #1475: Refactor waiting for job events
There are a couple of places where we currently wait for a number of `up` and / or `down` job events, and the code is pretty similar:

* [`cli/scale.go`](https://github.com/flynn/flynn/blob/master/cli/scale.go#L120-L161)
* [`test/test_deployer.go`](https://github.com/flynn/flynn/blob/master/test/test_scheduler.go#L43-L73)

We will also need to do this same thing to close #1260 (see #1474), so I think it would be good to refactor to something like:

```go
watcher, err := client.WatchJobEvents(app.ID)
if err != nil {
	return err
}
defer watcher.Close()

// make some job changes

err := watcher.WaitFor(controller.JobEvents{
	"web":    {"up": 1},
	"worker": {"down": 2},
})
```

Both examples of the code print some information when a job event is received, so we could provide a callback to be fired on each event, or give the watcher a logger, or something else.

jzila:
Fixes #1475

Implemented an abstraction for waiting for job events instead of
re-implementing the channel loop each time we need to watch for jobs.

In instances where we need to multiplex job streams with other channels,
the original code is still in place.

### DIFF
--- a/controller/client/client.go
+++ b/controller/client/client.go
@@ -31,6 +31,79 @@ type Client struct {
 	*httpclient.Client
 }
 
+type JobWatcher struct {
+	events chan *ct.JobEvent
+	stream stream.Stream
+}
+
+func newJobWatcher(events chan *ct.JobEvent, stream stream.Stream) *JobWatcher {
+	w := &JobWatcher{
+		events: events,
+		stream: stream,
+	}
+	return w
+}
+
+func jobEventsEqual(expected, actual ct.JobEvents) bool {
+	for typ, events := range expected {
+		diff, ok := actual[typ]
+		if !ok {
+			if len(events) == 0 {
+				continue
+			}
+			return false
+		}
+		for state, count := range events {
+			actualCount, ok := diff[state]
+			if !ok || actualCount != count {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+func (w *JobWatcher) WaitFor(expected ct.JobEvents, timeout time.Duration, callback func(*ct.JobEvent) error) error {
+	actual := make(ct.JobEvents)
+	for {
+		select {
+		case e, ok := <-w.events:
+			if !ok {
+				if err := w.stream.Err(); err != nil {
+					return err
+				}
+				return fmt.Errorf("Event stream unexpectedly ended")
+			}
+			if _, ok := actual[e.Type]; !ok {
+				actual[e.Type] = make(map[string]int)
+			}
+			switch e.State {
+			case "crashed":
+				actual[e.Type]["down"] += 1
+			case "starting", "up", "down":
+				fallthrough
+			default:
+				actual[e.Type][e.State] += 1
+			}
+			if callback != nil {
+				err := callback(e)
+				if err != nil {
+					return err
+				}
+			}
+			if jobEventsEqual(expected, actual) {
+				return nil
+			}
+		case <-time.After(timeout):
+			return fmt.Errorf("Timed out waiting for job events. Waited %d seconds.\nexpected: %v\nactual: %v", timeout, expected, actual)
+		}
+	}
+}
+
+func (w *JobWatcher) Close() error {
+	return w.stream.Close()
+}
+
 // ErrNotFound is returned when a resource is not found (HTTP status 404).
 var ErrNotFound = errors.New("controller: resource not found")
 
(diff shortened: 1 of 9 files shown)


jzila:
The last 2 CI failures have been intermittent failures independent of this change.

controller/client/client.go
```diff
@@ -351,6 +435,38 @@ func (c *Client) StreamJobEvents(appID string, output chan *ct.JobEvent) (stream
...
+func (c *Client) WatchJobEvents(appID string) (*JobWatcher, error) {
+	events := make(chan *ct.JobEvent)
+
+	stream, err := c.StreamJobEvents(appID, events)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return newJobWatcher(events, stream), nil
```
lmars: I don't think there is need for the blank lines in this method

controller/client/client.go
```diff
@@ -351,6 +435,38 @@ func (c *Client) StreamJobEvents(appID string, output chan *ct.JobEvent) (stream
...
+}
+
+func (c *Client) ExpectedScalingEvents(actual, expected map[string]int, clusterSize int) ct.JobEvents {
+	events := make(ct.JobEvents)
```
lmars: Add a `len(expected)` capacity hint.

controller/client/client.go
```diff
@@ -351,6 +435,38 @@ func (c *Client) StreamJobEvents(appID string, output chan *ct.JobEvent) (stream
...
+	stream, err := c.StreamJobEvents(appID, events)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return newJobWatcher(events, stream), nil
+}
+
+func (c *Client) ExpectedScalingEvents(actual, expected map[string]int, clusterSize int) ct.JobEvents {
```
lmars: Does this need to be a method on the client? Given it just creates `ct.JobEvents`, maybe it should be in the types package?

controller/client/client.go
```diff
@@ -351,6 +435,38 @@ func (c *Client) StreamJobEvents(appID string, output chan *ct.JobEvent) (stream
...
+	events := make(ct.JobEvents)
+	for typ, count := range expected {
+		diff := count
+		val, ok := actual[typ]
+		if ok {
+			diff = count - val
+		}
+		if typ == "omni" {
+			diff *= clusterSize
+		}
```
lmars: This check is incorrect, the type being named "omni" does not indicate the type is an omni job (it just so happens that this is the case in the tests), the correct check is against the [process type's Omni field](https://github.com/flynn/flynn/blob/master/controller/types/types.go#L49).

Note this handling of omni jobs will most likely be simplified in the next implementation of the scheduler, see #1484.

controller/client/client.go
```diff
@@ -31,6 +32,89 @@ type Client struct {
 	*httpclient.Client
 }
 
+type JobWatcher struct {
+	events chan *ct.JobEvent
+	stream stream.Stream
+}
```
lmars: This should be in the types package.

controller/client/client.go
```diff
@@ -31,6 +32,89 @@ type Client struct {
...
+				err := callback(e)
+				if err != nil {
+					return err
+				}
+			}
+			if jobEventsEqual(expected, actual) {
+				return nil
+			}
+		case <-time.After(timeout):
+			return fmt.Errorf("Timed out waiting for job events. Waited %d seconds.\nexpected: %s\nactual: %s", timeout, toJson(expected), toJson(actual))
```
lmars: I think the default printing of the `map` type (i.e. `%v`) rather than converting to JSON should suffice in the error output, e.g.:

```
map[web:map[up:1] worker:map[down:3]]
```

test/test_cli.go
```diff
@@ -189,7 +196,9 @@ func (s *CLISuite) TestPs(t *c.C) {
 	for _, j := range jobs {
 		t.Assert(j, Matches, "echoer")
 	}
-	t.Assert(app.flynn("scale", "echoer=0"), Succeeds)
+	scale := app.flynn("scale", "echoer=0")
+	app.waitFor(ct.JobEvents{"echoer": {"down": 3}})
```
lmars: This check is a duplication of the fact that the scale command succeeds

test/test_cli.go
```diff
@@ -231,17 +242,18 @@ func (s *CLISuite) TestScale(t *c.C) {
 
 	// unchanged processes shouldn't appear in output
 	scale = app.flynn("scale", "echoer=3", "printer=0")
+	app.waitFor(ct.JobEvents{"printer": {"down": 2}})
 	t.Assert(scale, Succeeds)
 	t.Assert(scale, SuccessfulOutputContains, "printer: 2=>0")
-	t.Assert(scale, c.Not(SuccessfulOutputContains), "echoer")
+	t.Assert(scale, c.Not(OutputContains), "echoer")
```
lmars: Why the change in assertion?

test/test_cli.go
jzila: Ostensibly yes, but I've had TestScale fail without this check.

test/test_cli.go
jzila: SuccessfulOutputContains is analogous to `Success AND OutputContains`. This means that `c.Not(SuccessfulOutputContains)` is analogous to `NOT Success OR NOT OutputContains`, when I want to explicitly check that the output doesn't contain a specific string.

controller/client/client.go
jzila: Then I need to move all the JobWatcher receiving methods to the types package as well. From what I could tell, that package was avoiding having types that receive methods. Did I misunderstand? Is there a cleaner way to do this?

test/test_cli.go
lmars: Hmm is that an intermittent failure? It would be good to understand the issue, as this check seems redundant.

controller/client/client.go
lmars: Ok yes I missed that point, fine being here then.

jzila:
@lmars does your latest comment indicate that we should remove the TestScale watcher waits, or are we good to go?

lmars:
> does your latest comment indicate that we should remove the TestScale watcher waits

I am just curious to understand why you needed to change `TestPs` because `TestScale` was failing?

Other than that, please squash to a single commit (e.g. `controller,cli: Refactor waiting for job events`) and we are good to merge.

jzila:
Sorry, I made the same change in TestScale so I was confused what you were asking about. Yes, both `TestScale` and `TestPs` fail intermittently without the check.

lmars:
Thanks!

Issue #1434, opened 2015-04-10
Flynn is confused by how many jobs is running


kwando:
```
kwando at alice in ~/simple-app on master
$ flynn ps
ID                                      TYPE
flynn-0db072aa69dc4b19b1674fb591bdcfb9  web
flynn-c7f866569b83426697661971890b056f  web

kwando at alice in ~/simple-app on master
$ flynn scale web=1
requested scale equals current scale, nothing to do!
```

https://gist.github.com/1b3181a28f37ec647e14

titanous:
What version of flynn are you running (`flynn-host version` on the server).

kwando:
v20150407.0

titanous:
This is some variant of #686, did it happen after a restart?

kwando:
I think I did ```stop flynn-host``` and then ```start flynn-host```. This procedure almost always brake flynn, a full reboot of the box usually fixes the problems though.

temujin9:
Having similar issues. No restart involved, but it did happen after a rebuild of the application.

jzila:
Might be related to #1182

lmars:
We have made some changes to the CLI (#1505) which should make it easier to diagnose the issue here:

* `flynn ps` will now show the release a process is from
* `flynn release show` will show information on the current release

So using these two commands, you should be able to identify which process should not be running (as it will not be from the current release), and scale that release down (e.g. `flynn scale --release RELEASE_ID web=0`).

titanous:
This should be resolved by the new scheduler (#1914).

PR #1505, opened 2015-05-06
cli: Make releases more visible

lmars:
This adds the release ID to the `flynn ps` output, as well as adding `flynn release show` which together will help diagnosing rogue processes in #1434.

### DIFF
--- a/cli/release.go
+++ b/cli/release.go
@@ -5,6 +5,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"os"
+	"strings"
+	"text/tabwriter"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
 	"github.com/flynn/flynn/controller/client"
@@ -14,6 +17,7 @@ import (
 func init() {
 	register("release", runRelease, `
 usage: flynn release add [-t <type>] [-f <file>] <uri>
+       flynn release show [<id>]
 
 Manage app releases.
 
@@ -30,6 +34,10 @@ Commands:
 		configuration in a JSON format. It's primarily used for specifying the
 		release environment and processes (similar to a Procfile). It can take any
 		of the arguments the controller Release type can take.
+
+	show  show information about a release
+
+		Omit the ID to show information about the current release.
 
 Examples:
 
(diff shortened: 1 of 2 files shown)


cli/release.go
```diff
@@ -48,11 +56,21 @@ Examples:
...
 	$ flynn release add -f config.json https://registry.hub.docker.com?name=flynn/slugbuilder&id=15d72b7f573b
-	Created release f55fde802170.
+	Created release 427537e78be4417fae2e24d11bc993eb.
+
+	$ flynn release show
+	ID            427537e78be4417fae2e24d11bc993eb
+	Artifact      docker+https://registry.hub.docker.com?name=flynn/slugbuilder&id=15d72b7f573b
+	ProcessTypes  echo
+	CreatedAt     2015-05-06 21:58:12.751741 +0000 UTC
+	ENV[MY_VAR]   Hello World, this will be available in all process types.
```
titanous: The env is redundant with `flynn env`, isn't it?

cli/release.go
```diff
@@ -63,6 +81,37 @@ func runRelease(args *docopt.Args, client *controller.Client) error {
...
+	types := make([]string, 0, len(release.Processes))
+	for typ := range release.Processes {
+		types = append(types, typ)
+	}
+	w := tabwriter.NewWriter(os.Stdout, 1, 2, 2, ' ', 0)
+	defer w.Flush()
+	listRec(w, "ID", release.ID)
+	listRec(w, "Artifact", fmt.Sprintf("%s+%s", artifact.Type, artifact.URI))
+	listRec(w, "ProcessTypes", strings.Join(types, ","))
+	listRec(w, "CreatedAt", release.CreatedAt)
```
titanous: Perhaps use colons and spaces?

```text
ID:
Artifact:
Process Types:
Created At:
```

cli/release.go
lmars: Well, `flynn env` only shows the current release, whereas this can show historic releases, and you may want to see historic env vars (e.g. `SLUG_URL`).

Alternatively, we could add a `--release` flag to `flynn env`, but I think I prefer this.

cli/release.go
```diff
@@ -63,6 +81,37 @@ func runRelease(args *docopt.Args, client *controller.Client) error {
...
+	var err error
+	if args.String["<id>"] != "" {
+		release, err = client.GetRelease(args.String["<id>"])
+	} else {
+		release, err = client.GetAppRelease(mustApp())
+	}
+	if err != nil {
+		return err
+	}
+	artifact, err := client.GetArtifact(release.ArtifactID)
```
titanous: The artifact isn't required, so if there is no ID it shouldn't be looked up.

cli/release.go
```diff
@@ -63,6 +81,37 @@ func runRelease(args *docopt.Args, client *controller.Client) error {
...
+	}
+	types := make([]string, 0, len(release.Processes))
+	for typ := range release.Processes {
+		types = append(types, typ)
+	}
+	w := tabwriter.NewWriter(os.Stdout, 1, 2, 2, ' ', 0)
+	defer w.Flush()
+	listRec(w, "ID", release.ID)
+	listRec(w, "Artifact", fmt.Sprintf("%s+%s", artifact.Type, artifact.URI))
+	listRec(w, "ProcessTypes", strings.Join(types, ","))
```
titanous: Join with comma space.

cli/release.go
titanous: Okay, this is fine.

cli/release.go
lmars: Hmm so I always have shell filtering in the back of my mind, so the first column having no spaces and no extraneous colon helps in that regard.

That being said, I could probably be persuaded...

titanous: Hmm, seems like the tab delimiter should be easy to split on?

lmars: Well, I am using a space as the pad character (as we do elsewhere), so there are no tabs in the output.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #1506, opened 2015-05-08
installer: DigitalOcean

jvatic:
ping @titanous 

screen shot 2015-05-11 at 10 39 54
screen shot 2015-05-11 at 10 39 58
screen shot 2015-05-11 at 10 40 12
screen shot 2015-05-11 at 10 40 16
screen shot 2015-05-11 at 10 40 31
screen shot 2015-05-11 at 15 22 37

There are still a few pending TODOs on the UX side of things:

- [x] Handle SSE events in batches (prevents page from locking up)
- [x] Fix CSS for install log (long lines break the layout)
- [x] Fix DO configuration page when coming from creating the first DO credential
- [x] The "New" nav item should reflect the current target cloud selection in URL
- [x] Error handling for credential management (create / delete)
- [x] Reloading page


jvatic:
Found a few more issues:

- [ ] Things get rather sluggish once enough events have accumulated in the database, there are some optimizations to be made there
- [ ] Closing an SSE stream while events are being sent to it locks out all other streams until the installer is restarted

jvatic:
The event stream still has a lot of overhead (fetching events from ql takes around half a second). It previously took under 10ms. This needs optimizing.

jvatic:
Added a few indexes which appears to have sped things up.

jvatic:
Actually, it still gets slow when the number of events in the db is high enough...

titanous:
@jvatic how many events is high? Do we need to keep them in the db?

jvatic:
@titanous it's taking between 200-900ms with ~3000 events in the db (can occur within a single install). There are a number of n+1 queries, so I'll see how much of an improvement reducing those yields. We could remove install log events completely once the installation has finished. Caching events in memory would also speed things up.

installer/app/src/views/asset-paths.js.erb
```diff
@@ -16,6 +16,9 @@ var AssetPaths = <%= Yajl::Encoder.encode([
 	"windows-cert-5.png",
 	"windows-cert-6.png",
 	"windows-cert-7.png",
+
+  "aws-logo.jpg",
+  "digitalocean-logo.jpg",
```
titanous: Should these be PNGs? (or are they not available?)

installer/aws_cluster.go
```diff
@@ -25,10 +25,46 @@ var DisallowedEC2InstanceTypes = []string{"t1.micro", "t2.micro", "t2.small", "m
 var DefaultInstanceType = "m3.medium"
 var StackNotFoundError = errors.New("Stack does not exist")
 
+func (c *AWSCluster) Type() string {
+	const t = "aws"
+	return t
```
titanous:
```go
return "aws"
```

r esults in a constant string.

jvatic:
Updating ql to the latest version increased performance noticeably, the caching brings the overhead down to below 30ms for the initial load (e.g. page refresh) and below 1ms after that.

jvatic:
@titanous comments addressed

installer/persistence.go
```diff
@@ -76,3 +76,28 @@ func loadSSHKey(name string) (*sshkeygen.SSHKey, error) {
 	}
 	return key, nil
 }
+
+func loadSSHKeys() ([]*sshkeygen.SSHKey, error) {
```
jvatic: This isn't actually used and can be removed for now. It's use will be for matching any local key in the installer keys dir with one saved on the install target.

installer/digital_ocean_cluster.go
```diff
@@ -0,0 +1,470 @@
...
+			}
+			return err
+		}
+		return nil
+	}
+}
+
+func (c *DigitalOceanCluster) instanceStartFlynn(sshConfig *ssh.ClientConfig, ipAddress string) error {
+	c.base.SendLog(fmt.Sprintf("Starting flynn on %s", ipAddress))
+	cmd := fmt.Sprintf(`echo "%s" | base64 -d > startScript && chmod +x startScript && ./startScript && rm startScript`, c.startScript)
```
titanous: Looks like this could be piped directly into bash

installer/digital_ocean_cluster.go
```diff
@@ -0,0 +1,470 @@
...
+	c.base.SendLog("Configuring domain")
+	instanceIP := c.base.InstanceIPs[0]
+	dr, _, err := c.client.Domains.Create(&godo.DomainCreateRequest{
+		Name:      c.base.Domain.Name,
+		IPAddress: instanceIP,
+	})
+	if err != nil {
+		return err
+	}
+	domain := dr.Domain
```
titanous: `dr.Domain.Name`

installer/digital_ocean_cluster.go
```diff
@@ -0,0 +1,470 @@
...
+	}
+	for _, op := range ops {
+		for _, ipAddress := range instanceIPs {
+			go func(ipAddress string) {
+				err := op(sshConfig, ipAddress)
+				if err != nil {
+					errChan <- err
+					return
+				}
+				done <- struct{}{}
```
titanous: Could combine `done` into `errChan`.

installer/http.go
```diff
@@ -137,49 +130,76 @@ func (api *httpAPI) AssetManifest() (*assetManifest, error) {
 }
 
 func (api *httpAPI) LaunchCluster(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
-	var input *jsonInput
-	if err := httphelper.DecodeJSON(req, &input); err != nil {
+
+	var inputJSON bytes.Buffer
+	if _, err := io.Copy(&inputJSON, req.Body); err != nil {
```
titanous:
```go
inputJSON.ReadFrom(req.Body)
```

installer/http.go
```diff
@@ -137,49 +130,76 @@ func (api *httpAPI) AssetManifest() (*assetManifest, error) {
...
 
-	var creds aws.CredentialsProvider
-	var credentialID string
-	var credentialSecret string
-	if input.Creds.AccessKeyID != "" && input.Creds.SecretAccessKey != "" {
-		creds = aws.Creds(input.Creds.AccessKeyID, input.Creds.SecretAccessKey, "")
-		credentialID = input.Creds.AccessKeyID
-		credentialSecret = input.Creds.SecretAccessKey
+	decodeJSON := func(dst interface{}) error {
+		in := bytes.NewBuffer(inputJSON.Bytes())
```
titanous: `NewReader`

installer/http.go
```diff
@@ -137,49 +130,76 @@ func (api *httpAPI) AssetManifest() (*assetManifest, error) {
...
+		dec := json.NewDecoder(in)
+		return dec.Decode(dst)
```
titanous:
```go
return json.Unmarshal(inputJSON.Bytes(), in)
```

titanous:
The AWS logo is wrong (there's no .com anymore)

titanous:
Looks good!

A few notes:

- Default region should be nyc3 and 2gb droplets.
- Droplet types should be sorted by size.
- Nickname for credentials shouldn't be required (also call it "account name")
- Something in the JS caused Firefox Nightly to peg my CPU while streaming install events.
- We need to configure an iptables firewall so that only the public ports (80, 443, 22, 3000-3500) are accessible on each node.
- An old cluster in my db is causing this error on every page load:

```text
GetEventsSince Error finding cluster flynn-1430244425: sql: no rows in result set
```

jvatic:
Updated AWS logo

jvatic:
> Something in the JS caused Firefox Nightly to peg my CPU while streaming install events.

I've reproduced this and made a few changes which should help reduce it. I'll spend some more time tracking down the exact cause once everything else is taken care of.

installer/digital_ocean_cluster.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	c.startScript = startScript
+
+	sshConfig, err := c.base.sshConfig()
+	if err != nil {
+		return err
+	}
+
+	instanceIPs := c.base.InstanceIPs
+	errChan := make(chan error)
```
titanous: Add a buffer so that sends don't block and leak goroutines if there is an error.

jvatic:
- [ ] Install cert -> Dashboard login is broken

jvatic:
@titanous comments addressed

installer/digital_ocean_cluster.go
```diff
@@ -0,0 +1,474 @@
...
+			}
+			return err
+		}
+		sshConn.Close()
+		return nil
+	}
+}
+
+func (c *DigitalOceanCluster) instanceConfigureIPTables(sshConfig *ssh.ClientConfig, ipAddress string) error {
+	c.base.SendLog(fmt.Sprintf("Configuring ip tables on %s", ipAddress))
```
titanous: s/ip tables/iptables firewall/

titanous:
LGTM pending comment

PR #1509, opened 2015-05-09
isoupdate

joshuacox:
merely updating the iso url as the 14.04.1 is no longer available


titanous:
Thanks! Please change the commit message to say `util/packer: Update Ubuntu URL` or similar and add a [DCO sign-off](https://flynn.io/docs/contributing#developer%E2%80%99s-certificate-of-origin).

PR #1511, opened 2015-05-09
util/packer: Update Ubuntu URL

joshuacox:
Ok, I think I got it right this time

Signed-off-by: Josh Cox <josh@webhosting.coop>


titanous:
Indeed! (for future reference you can force push to update a PR)

LGTM.

PR #1512, opened 2015-05-09
util/packer: added qemu and virtualbox-iso builders

joshuacox:
I need the qemu builder as my main VM host is KVM based in my closet, I added the virtualbox-iso as I'm already downloading that particular iso, I just have to symlink it to my packer_cache, no need to download the ova.

Signed-off-by: Josh Cox <josh@webhosting.coop>

### DIFF
--- a/util/packer/ubuntu-14.04/template.json
+++ b/util/packer/ubuntu-14.04/template.json
@@ -9,6 +9,58 @@
   },
   "builders": [
     {
+      "type": "qemu",
+      "boot_command": [
+        "<esc><esc><enter><wait>",
+        "/install/vmlinuz noapic preseed/url=http://{{ .HTTPIP }}:{{ .HTTPPort }}/preseed.cfg ",
+        "debian-installer=en_US auto locale=en_US kbd-chooser/method=us ",
+        "hostname=flynn ",
+        "fb=false debconf/frontend=noninteractive ",
+        "keyboard-configuration/modelcode=SKIP keyboard-configuration/layout=USA keyboard-configuration/variant=USA console-setup/ask_detect=false ",
+        "initrd=/install/initrd.gz -- <enter>"
+      ],
+      "boot_wait": "5s",
+      "accelerator": "kvm",
+      "headless": "{{ user `headless` }}",
+      "http_directory": "http",
+      "iso_checksum": "8acd2f56bfcba2f7ac74a7e4a5e565ce68c024c38525c0285573e41c86ae90c0",
+      "iso_checksum_type": "sha256",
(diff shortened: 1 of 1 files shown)


lmars:
@joshuacox I can't say I'm keen on preseeding a vanilla ISO when there are [pre-built cloud images available](https://cloud-images.ubuntu.com/trusty/current/). It is *much* quicker to just modify a cloud image and export the resulting image than preseeding.

That being said, it doesn't look like Packer supports doing this currently which is a shame.

I also don't understand the logic behind adding the `virtualbox-iso` builder, downloading the OVA is quicker (well, based on your connection speed) than doing the preseed, can you clarify why you want to add that?

joshuacox:
I am already downloading the same ISO for another project, and as well it is the same iso that is used in the vmware build, so if one were to be building all builders, it would be inefficient to download the ova as well as the iso for all projects.  Instead download one iso and build for all builders using the same iso.  This makes sense to me, but given if you were just building a single vm the ova might be beneficial.  I, for one, am a proponent of preseeding I think that @titanous post on December 23rd should underscore the necessity for ground up builds:
https://titanous.com/posts/docker-insecurity
I have no reason to suspect the ubuntu ova, or even ova's in general, but if I'm building for production and dev I'd rather have all builds start with something that is built straight from the god given iso,  in this case this may or may not be the canonical ( I just love the pun here ) version.

lmars:
@joshuacox 

> it would be inefficient to download the ova as well as the iso for all projects.

The current OVA is 364M which on my machine takes ~40s to download, which is *much* quicker than preseeding an ISO, even when you already have the ISO downloaded (I haven't timed it, but preseeding takes at least 5 mins on my machine).

>  Instead download one iso and build for all builders using the same iso

If you were preseeding once, and then building for virtualbox / vmware / qemu from that single preseeded image, then yes this may be efficient, but that is not the case, each builder will do the preseed step individually.

> I have no reason to suspect the ubuntu ova, or even ova's in general, but if I'm building for production and dev I'd rather have all builds start with something that is built straight from the god given iso

You are downloading the ISO from Canonical, and the cloud images are also provided by Canonical, so I see zero security benefit in downloading a verifiable ISO and preseeding it, as apposed to downloading a verifiable cloud image from the same entity.

titanous:
Closing, as there has been no response from @joshuacox.

joshuacox:
sorry didn't mean to be non-responsive.  I was just making pull requests to contribute upstream anything I might have added to your previous work.  All my changes are available here if anyone wants them:
https://github.com/joshuacox/flycker
on the other points, not going to argue, sorry if I came off as trollish, puns have no business in github issues.

PR #1520, opened 2015-05-11
receiver: Default scale apps which have no previous scale

lmars:
This means that as well as apps with no previous release, apps with a release but which haven't been scaled (e.g. the result of `flynn create foo && flynn env set FOO=BAR`) also get a default web=1 formation.

The test change can be reverted once #1474 lands.


titanous:
LGTM

PR #1529, opened 2015-05-13
Installer: Fixed typo in AWS credentials dialogue

arekkas:
The installer showed two times `AWS_ACCESS_KEY_ID` instead of `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` for each the key and the secret


jvatic:
Thanks!

Issue #993, opened 2015-02-12
App deletion cleanup


titanous:
We should add a worker that does the following:

- Removes any routes from the router (#709)
- Delete provisioned resources

I think it makes sense to add another pgx worker type and make the deployer daemon into a generic worker that can do background jobs for the controller.

titanous:
Fixed by #1533

PR #1532, opened 2015-05-13
controller: Convert the deployer to a generic worker

lmars:
This will make it easier to handle more background jobs (e.g. #993).


titanous:
LGTM

PR #1533, opened 2015-05-14
App deletion cleanup

lmars:
Opening as an early PR for feedback.

Things that need to be done:

- [x] remove cached git repo (and implement test)
- [x] move cleanup from client into background worker

I would like some feedback on the following:

Once the cleanup is moved to the background, the client will need to wait for the job to finish so implies some kind of event stream, but I don't think adding yet another `xxx_events` db table with an SSE endpoint is the way forward. Ideally there would be a single "app events" stream including formation changes, deployments, job events and now app deletion events. The question then is should these events go into a single db table, or should they go to the logaggregator and clients stream from there instead?

/cc @titanous


titanous:
> Ideally there would be a single "app events" stream including formation changes, deployments, job events and now app deletion events. The question then is should these events go into a single db table, or should they go to the logaggregator and clients stream from there instead?

A DB table makes sense. The log aggregator does not provide durability guarantees.

controller/client/client.go
```diff
@@ -190,8 +190,37 @@ func (c *Client) UpdateApp(app *ct.App) error {
...
+	}
+	a.DeletedRoutes = routes
+
+	// remove resources
+	resources, err := c.AppResourceList(appID)
+	if err != nil {
+		return nil, err
+	}
+	for _, resource := range resources {
+		if err := c.DeleteResource(resource.ProviderID, resource.ID); err != nil {
```
titanous: Resources can be associated with multiple apps, only makes sense to destroy if this is the last associated app.

lmars:
@titanous comments addressed

controller/worker/app_deletion/app_deletion.go
```diff
@@ -0,0 +1,132 @@
...
+		if err != nil {
+			log.Error("error creating app cache delete request", "err", err)
+			return err
+		}
+		res, err := http.DefaultClient.Do(req)
+		if err != nil {
+			log.Error("error performing app cache delete request", "err", err)
+			return err
+		}
+		res.Body.Close()
```
titanous: Should check that the status code is 200

lmars:
@titanous comments addressed

titanous:
LGTM

PR #1537, opened 2015-05-15
controller: Move events into single app event stream

lmars:
This will make it easier to add more event types which can be streamed to clients (e.g. app deletion events in #1533), as well as removing duplication in the controller.

### DIFF
--- a/controller/events.go
+++ b/controller/events.go
@@ -1,0 +1,208 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
+	ct "github.com/flynn/flynn/controller/types"
+)
+
+// ErrEventBufferOverflow is returned to clients when the in-memory event
+// buffer is full due to clients not reading events quickly enough.
+var ErrEventBufferOverflow = errors.New("event stream buffer overflow")
+
+// eventBufferSize is the amount of events to buffer in memory.
+const eventBufferSize = 1000
+
+// EventSubscriber receives app events from the EventListener loop and maintains
+// it's own loop to forward those events to the Events channel.
+type EventSubscriber struct {
+	Events chan *ct.AppEvent
+	Err    error
+
+	l          *EventListener
+	queue      chan *ct.AppEvent
+	appID      string
+	objectType string
+	objectID   string
+
+	stop     chan struct{}
+	stopOnce sync.Once
+}
+
+// Notify filters the event based on it's type and objectID and then pushes
+// it to the event queue.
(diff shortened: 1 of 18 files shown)


controller/schema.go
```diff
@@ -42,6 +42,27 @@ func migrateDB(db *sql.DB) error {
 )`,
 		`CREATE UNIQUE INDEX ON apps (name) WHERE deleted_at IS NULL`,
 
+		`CREATE SEQUENCE app_event_ids`,
+		`CREATE TABLE app_events (
+    event_id   bigint      PRIMARY KEY DEFAULT nextval('app_event_ids'),
+    app_id     uuid        NOT NULL REFERENCES apps (app_id),
+    object_id  text        NOT NULL,
```
titanous: Does it make sense to add an `object` enum to go with `object_id`?

lmars: So do you mean rename the `type` column to `object` with an enum?

titanous: yeah, maybe `object_type`

lmars:
The issues in CI around too many postgres connections seems to be due to an event stream leak. I haven't been able to identify the source of the leak, but I have made multiple event streams only open a single DB connection: e5433c2

lmars:
@titanous comments addressed, please review

controller/events.go
```diff
@@ -0,0 +1,240 @@
...
+		return nil
+	}
+	for len(e.events) == 0 {
+		e.cond.Wait()
+		if e.closed {
+			return nil
+		}
+	}
+	event := e.events[0]
+	e.events = e.events[1:]
```
titanous: I think this will result in excessive copying, as the pointer will always be headed towards the end of the backing array. A better approach would be a ring buffer (which would allow specifying a limit on the buffer size easily) or using `container/list` which implements a linked list.

lmars:
@titanous comments addressed (updated to use a linked list)

controller/events.go
```diff
@@ -0,0 +1,245 @@
...
+// Notify filters the event based on it's type and objectID and then pushes
+// it to the event queue.
+func (e *EventSubscriber) Notify(event *ct.AppEvent) {
+	if e.objectType != "" && e.objectType != string(event.ObjectType) {
+		return
+	}
+	if e.objectID != "" && e.objectID != event.ObjectID {
+		return
+	}
+	e.queue.Push(event)
```
titanous: I think there should be a max queue depth, after which the connection is terminated for not keeping up. It's not critical, so drop a TODO in here.

titanous: After thinking about this a bit, it seems like a buffered channel is actually better, as it would reduce the amount of code used by a significant amount (use a `select` to send with a `default` case that closes the client connection with a "slow reader" error)

lmars:
@titanous comments addressed (switched to a buffered channel)

titanous:
LGTM

PR #1541, opened 2015-05-16
cli: Fix scaling processes down

lmars:
The previous check of `count < 0` was never true, meaning down events were not being waited for.

I have updated the tests to check all the expected events appear in the scale output.

@jzila FYI this was the cause of the intermittent failures in #1502.


titanous:
LGTM

PR #1543, opened 2015-05-16
script: Only remove Flynn if "yes" is sent on stdin

### LINKED ISSUE #1542: Ubuntu installation script assumes "yes" to prompt on end of input
Steps to reproduce:
1. Download installation script from https://dl.flynn.io/install-flynn, and run it with --clean or --remove.
2. Get asked for confirmation "Are you sure this is what you want?"
3. Press ctrl-d or don't have a terminal.

Expected result: script should probably assume "no" and exit if --yes parameter was not used.
Actual result: script proceeds with removal even without --yes parameter.

lmars:
Previous to this change, receiving EOF on stdin would not abort the removal, which is unexpected.

Fixes #1542.

### DIFF
--- a/script/install-flynn.tmpl
+++ b/script/install-flynn.tmpl
@@ -167,11 +167,14 @@ do_remove() {
     echo -n "(yes/no): "
     while read answer; do
       case "${answer}" in
-        yes) break ;;
-        no)  exit ;;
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

PR #1552, opened 2015-05-19
cli: Error if curl download of binary fails

### LINKED ISSUE #1550: Flynn CLI installation should make download errors more visible
If the `curl` command in the Flynn CLI installation aborts, the user can end up with a non-functioning binary (see #1548).

An error should be outputted when this happens.

titanous:
Closes #1550


lmars:
LGTM

PR #1553, opened 2015-05-19
cli: Add limit command for managing resource limits

lmars:


cli/limit.go
```diff
@@ -0,0 +1,142 @@
...
+	$ flynn limit
+	web:     max_fd=12000  max_procs=1024  memory=512MB
+	worker:  max_fd=10000  max_procs=256   memory=1GB
+
+	$ flynn limit set web memory=256MB
+	Created release b39fe25d0ea344b6b2af5cf4d6542a80
+
+	$ flynn limit
+	web:     max_fd=12000  max_procs=1024  memory=256MB
+	worker:  max_fd=10000  max_procs=256   memory=1GB
```
titanous: We need to drop max_procs, as it doesn't work (and won't until we sort some uid management stuff out)

lmars:
@titanous comments addressed, as well as assigning default limits to one-off jobs.

host/resource/resource.go
```diff
@@ -62,3 +65,43 @@ func SetDefaults(r *Resources) {
...
+		return strconv.FormatInt(limit, 10)
+	}
+}
+
+var byteUnits = []string{"B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"}
+
+func byteSize(limit int64) string {
+	i := 0
+	unit := 1024.0
+	size := float64(limit)
```
titanous: Why are we using floats?

lmars: So that a limit of say `1.5GB` is reported as so (rather than `1GB`)

titanous: Ah, got it.

titanous:
LGTM

host/resource/resource.go
```diff
@@ -62,3 +65,43 @@ func SetDefaults(r *Resources) {
...
+func FormatLimit(typ Type, limit int64) string {
+	switch typ {
+	case TypeMemory:
+		return byteSize(limit)
+	default:
+		return strconv.FormatInt(limit, 10)
+	}
+}
+
+var byteUnits = []string{"B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"}
```
danielsiders: I'd like for us to add support for Novabytes (1000^9 / 10^27) and Dexabytes (1000^10 / 10^30) to help promote the new names I've devised and support higher limits on more powerful nodes.

PR #1554, opened 2015-05-20
host: Remove default max procs limit

titanous:
The nprocs rlimit limits the process count of the user id of the process
it is applied to. This makes it useless currently, as most jobs end up
running as either root or nobody, resulting in errors like:

    fork: Resource temporarily unavailable

When there are more than nproc process running for a specific uid. In
practice, this can happen when scaling a single app to a dozen or so
instances.

The long-term solution is to make sure every job has a different uid,
but this is a complex undertaking, so for now we’ll just remove the
default limit.

### DIFF
--- a/host/resource/resource.go
+++ b/host/resource/resource.go
@@ -35,9 +35,8 @@ const (
 )
 
 var defaults = Resources{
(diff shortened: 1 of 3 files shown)


lmars:
LGTM

Issue #1568, opened 2015-06-03
cli: release add creates a release although the url is not pointing at a docker image


aeneasr:
```
$ flynn -a someapp release add -f manifest.json "https://registry.hub.docker.com?name=idonot&id=exist"
Created release 7d0b4a4467714bfabf8508dc0ba1b896.
```

```
$ flynn -a someapp release add -f manifest.json.json "http://google.com"
Created release c4f7fd57aab545dca71989520f7a7de1.
```

With manifest.json being:

``` json
{
  "processes": {
    "server": {
      "cmd": ["mongod"],
      "ports": [{
        "port": 27017,
        "proto": "tcp",
        "service": {
          "name": "mongo",
          "create": true,
          "check": { "type": "tcp" }
        }
      }]
    }
  }
}
```

arekkas:
The problem is not the cli but rather the host, who doesn't verify if the url is a valid docker image

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

Issue #509, opened 2014-11-25
Add support for tag to 'release add' and drop support in host


dhrp:
Currently the only way to add a docker image [release](https://flynn.io/docs/cli) is to give both the registry name, and the last image id.

e.g.
```
flynn release add -f config.json https://registry.hub.docker.com/flynn/slugbuilder?id=15d72b7f573b
```

it would be great if this command would also support the ?tag=latest type, that [pinkerton](https://github.com/flynn/flynn/tree/master/pinkerton) seems to support.

e.g.
```
flynn release add -f config.json https://registry.hub.docker.com/debian?tag=latest
```

arekkas:
I'd like to fix this in the host as well, could you point me at a direction?

titanous:
@arekkas https://github.com/flynn/flynn/blob/4ed1a7c57aa2ce9298c431510e8605422c4f4736/host/libvirt_lxc_backend.go#L403-L415

arekkas:
That was quick, thanks I'll take a look :)

arekkas:
@titanous should I simply validate `job.Artifact.URI` validate against `https://registry.hub.docker.com/[^/]+/[^\?]+?id=[a-z0-9]+`?

titanous:
@arekkas No, since we don't force users to use the Docker, Inc. registry. Just check that it's a valid url with `name` and `id` parameters.

arekkas:
Alright, a simple `url.Parse()` should do it then

arekkas:
It might be smart to add tests for this, if you point me at a direction maybe i can figure it out :)

lmars:
I am going to close this as users can now push Docker images directly to Flynn using `flynn docker push REF` (see #2857).

PR #1570, opened 2015-06-03
cli: release add is now able to resolve tags to ids.

arekkas:
This is the first part of #509 and also #1568 . If a tag is not resolvable,
the cli returns an error.

```
$ ./cli/bin/flynn -a mgotest release add -f ~/c.json "https://registry.hub.docker.com?name=mongo&tag=5.0"
registry: tag "5.0" not found

$ ./cli/bin/flynn -a mgotest release add -f ~/c.json "https://registry.hub.docker.com?name=mongo&tag=3.0"
Resolved tag 3.0 => https://registry.hub.docker.com?id=077c6501ee1803efe213a9031a7d7a9ab9798f28bbf57cca579eb90b58297806&name=mongo
Created release 1844f52971b34c5eaa128351670e71ac.

$ ./cli/bin/flynn -a mgotest scale server=1 --no-wait

$ ./cli/bin/flynn -a mgotest run mongo --host mongo.discoverd --port 27017
MongoDB shell version: 3.0.3
connecting to: mongo.discoverd:27017/test
Welcome to the MongoDB shell.
```

This PR does not solve that the host accepts `https://registry.hub.docker.com?name=mongo&tag=3.0` and `http://notadockerimage.com` mentioned in #1568 . However, when using the cli to resolve a tag, it has error detection:

```
$ ./cli/bin/flynn -a mgotest release add -f ~/c.json "https://registry.hub.docker.com?name=mongo&tag=5.0"
registry: tag "5.0" not found
```

### DIFF
--- a/cli/release.go
+++ b/cli/release.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/url"
 	"os"
 	"strings"
 	"text/tabwriter"
@@ -14,19 +15,21 @@ import (
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/pinkerton/registry"
 )
 
 func init() {
 	register("release", runRelease, `
 usage: flynn release
-       flynn release add [-t <type>] [-f <file>] <uri>
+       flynn release add [-t <type>] [-f <file>] [--no-validate] <uri>
        flynn release show [<id>]
 
 Manage app releases.
 
 Options:
-	-t <type>          type of the release. Currently only 'docker' is supported. [default: docker]
-	-f, --file=<file>  release configuration file
+	-t <type>              type of the release. Currently only 'docker' is supported. [default: docker]
+	-f, --file=<file>      release configuration file
+	--no-validate          don't validate the image URI
 
 Commands:
 	With no arguments, shows a list of releases associated with the app.
@@ -61,6 +64,7 @@ Examples:
 		}
 	}
 	$ flynn release add -f config.json https://registry.hub.docker.com?name=flynn/slugbuilder&id=15d72b7f573b
+	Validating Docker image URI.
 	Created release 427537e78be4417fae2e24d11bc993eb.
 
 	$ flynn release
@@ -152,10 +156,20 @@ func runReleaseAddDocker(args *docopt.Args, client *controller.Client) error {
 		}
 	}
 
+	uri := args.String["<uri>"]
 	artifact := &ct.Artifact{
 		Type: "docker",
-		URI:  args.String["<uri>"],
+		URI:  uri,
 	}
+
+	if !args.Bool["--no-validate"] {
+		canonicalURI, err := canonicalizeDockerURI(uri)
+		if err != nil {
+			return err
+		}
+		artifact.URI = canonicalURI
+	}
+
 	if err := client.CreateArtifact(artifact); err != nil {
 		return err
 	}
@@ -172,4 +186,28 @@ func runReleaseAddDocker(args *docopt.Args, client *controller.Client) error {
 	log.Printf("Created release %s.", release.ID)
 
 	return nil
+}
+
+func canonicalizeDockerURI(rawurl string) (string, error) {
+	log.Println("Validating Docker image URI.")
+	u, err := url.Parse(rawurl)
+	if err != nil {
+		return "", err
+	}
+	q := u.Query()
+	ref, err := registry.NewRef(rawurl)
+	if err != nil {
+		return "", err
+	}
+	image, err := registry.NewDockerSession(ref).GetImage()
+	if err != nil {
+		return "", err
+	}
+	q.Set("id", image.ID)
+	if tag := q.Get("tag"); tag != "" {
+		q.Del("tag")
+		log.Printf("Resolved tag %s => %s", tag, u.String())
+	}
+	u.RawQuery = q.Encode()
+	return u.String(), nil
 }
--- a/pinkerton/registry/docker.go
+++ b/pinkerton/registry/docker.go
@@ -138,6 +138,10 @@ func (s *dockerSession) get(path string, out interface{}) (*http.Response, error
 		if out != nil {
 			defer res.Body.Close()
 		}
+		if res.StatusCode == 404 {
+			err = fmt.Errorf("registry: image not found")
+			continue
+		}
 		if res.StatusCode != 200 {
 			err = fmt.Errorf("registry: unexpected status %d", res.StatusCode)
 			continue
--- a/test/test_cli.go
+++ b/test/test_cli.go
@@ -630,7 +630,7 @@ func (s *CLISuite) TestRelease(t *c.C) {
 	file.Close()
 
 	app := s.newCliTestApp(t)
-	t.Assert(app.flynn("release", "add", "-f", file.Name(), imageURIs["test-apps"]), Succeeds)
+	t.Assert(app.flynn("release", "add", "--validate=false", "-f", file.Name(), imageURIs["test-apps"]), Succeeds)
 
 	r, err := s.controller.GetAppRelease(app.name)
 	t.Assert(err, c.IsNil)


arekkas:
Resolving tags takes some time due to the network roundtrip. I wasn't able to solve that.

arekkas:
```
$ flynn -a someapp release add -f manifest.json.json "http://google.com"
Created release c4f7fd57aab545dca71989520f7a7de1.
```

could be solved in the cli by doing:

```go
	if ref, err := registry.NewRef(rawurl); err != nil {
		return "", err
	} else if image, err := registry.NewDockerSession(ref).GetImage(); err != nil {
		return "", err
	} else {
	        if q.Get("tag") != "" {
			q.Del("tag")
		}
		q.Set("id", image.ID)
		u.RawQuery = q.Encode()
		log.Printf("Resolved tag %s => %s", tag, u.String())
	}
```

arekkas:
That CI log looks..interesting

cli/release.go
```diff
@@ -128,9 +130,14 @@ func runReleaseAddDocker(args *docopt.Args, client *controller.Client) error {
 		}
 	}
 
+	ref, err := canonicalizeDockerURL(args.String["<uri>"])
```
lmars: Call this `uri`, and maybe rename the function to `canonicalDockerURI`

cli/release.go
```diff
@@ -149,3 +156,24 @@ func runReleaseAddDocker(args *docopt.Args, client *controller.Client) error {
...
+		if ref, err := registry.NewRef(rawurl); err != nil {
+			return "", err
+		} else if image, err := registry.NewDockerSession(ref).GetImage(); err != nil {
+			return "", err
+		} else {
+			q.Del("tag")
+			q.Set("id", image.ID)
+			u.RawQuery = q.Encode()
+			log.Printf("Resolved tag %s => %s", tag, u.String())
+		}
```
lmars: It is more idiomatic to not use `else`:

```go
ref, err := ...
if err != nil {
  return "", err
}
image, err := ...
if err != nil {
  return "", err
}
```

cli/release.go
```diff
@@ -149,3 +156,24 @@ func runReleaseAddDocker(args *docopt.Args, client *controller.Client) error {
...
+func canonicalizeDockerURL(rawurl string) (string, error) {
+	u, err := url.Parse(rawurl)
+	if err != nil {
+		return "", err
+	}
+	q := u.Query()
+	if tag := q.Get("tag"); tag != "" {
+		if ref, err := registry.NewRef(rawurl); err != nil {
+			return "", err
+		} else if image, err := registry.NewDockerSession(ref).GetImage(); err != nil {
```
lmars: We may as well fix #1568 by always doing this (rather than just when `tag` is set).

Given this takes a few HTTP requests to complete, it would be useful to print a message like `validating Docker image URI`, so the output is something like:

```
$ flynn release add https://registry.hub.docker.com?name=ubuntu&tag=foo
validating Docker image URI
ERROR: unknown tag "foo"

$ flynn release add https://registry.hub.docker.com?name=bar&tag=foo
validating Docker image URI
ERROR: unknown image "bar"

$ flynn release add https://example.com?name=bar&tag=foo
validating Docker image URI
ERROR: invalid registry "example.com"
```

cli/release.go
arekkas: Will do

cli/release.go
arekkas: Oh I didn't know that, always thought it might save a few LOCs

cli/release.go
arekkas: Hm, shouldn't validation be moved to the host first?

arekkas:
```json
$ ~/go/src/github.com/flynn/flynn/cli/bin/flynn -a mg release add -f ~/c.json "https://registry.hub.docker.com?name=mongo&id=1"
Validating Docker image URI.
registry: image not found
```

```json
$ ~/go/src/github.com/flynn/flynn/cli/bin/flynn -a mg release add -f c.json "https://registry.hub.docker.com?name=notexist&id=1234"
Validating Docker image URI.
registry: repo not found
```

``json
$  ~/go/src/github.com/flynn/flynn/cli/bin/flynn -a mg release add -f c.json "https://example.com?name=mongo&id=216d9a0e82646f77b31b78eeb0e26db5500930bbd6085d5d5c3844ec27c0ca50"
Validating Docker image URI.
Invalid registry "example.com"
```

**Comment by arekkas** [comment]
Not sure how to write tests for this, `TestRelease` seems like the place, but I don't really know where to start

File: cli/release.go [hunk 4]
```diff
@@ -149,3 +156,31 @@ func runReleaseAddDocker(args *docopt.Args, client *controller.Client) error {
...
 }
+
+func canonicalizeDockerURI(rawurl string) (string, error) {
+	log.Println("Validating Docker image URI.")
+	u, err := url.Parse(rawurl)
+	if err != nil {
+		return "", err
+	}
+	if u.Host != "registry.hub.docker.com" {
+		return "", fmt.Errorf(`Invalid registry "%s"`, u.Host)
```
lmars: This check will prevent people from using a private registry. An invalid registry should be determined by the return value of `GetImage()` (it will return an error if the registry does not return valid responses).

cli/release.go
```diff
@@ -149,3 +156,31 @@ func runReleaseAddDocker(args *docopt.Args, client *controller.Client) error {
...
+	if err != nil {
+		return "", err
+	}
+	q.Set("id", image.ID)
+	if tag := q.Get("tag"); tag != "" {
+		q.Del("tag")
+		log.Printf("Resolved tag %s => %s", tag, u.String())
+	}
+	u.RawQuery = q.Encode()
+	log.Println("Docker image URI is valid!")
```
lmars: I don't think this is a necessary log message, successfully creating the release after telling the user the URI is being validated seems enough:

```bash
Validating Docker image URI
Created release 427537e78be4417fae2e24d11bc993eb.
```

cli/release.go
```diff
@@ -149,3 +156,31 @@ func runReleaseAddDocker(args *docopt.Args, client *controller.Client) error {
 
 	return nil
 }
+
+func canonicalizeDockerURI(rawurl string) (string, error) {
+	log.Println("Validating Docker image URI.")
```
lmars: This log message needs adding to the example output.

arekkas: Do you mean `("Validating Docker image %s", rawurl)`?

lmars: Sorry, I meant the help text (i.e. [this](https://github.com/flynn/flynn/blob/master/cli/release.go#L63-L64)) needs updating to include this log message.

arekkas:
@lmars good for :eyes:

lmars:
@arekkas looks like `CLISuite.TestRelease` now fails because the image URI is invalid, we should support a `--validate=false` flag and use it [here](https://github.com/flynn/flynn/blob/master/test/test_cli.go#L633) (the test image does not need validating, it is already present on the host so will not be downloaded).

arekkas:
Should we maybe add a tests which checks if an invalid url fails successfully as well?

lmars:
@arekkas it wouldn't hurt to do so

arekkas:
Not sure why tests fail :| Wasn't able to get them working locally as well

arekkas:
I also wanted to add

```go
	t.Assert(app.flynn("release", "add", "-f", file.Name(), "http://invalid/image-uri"), c.Not(Succeeds))
```

[in this function](https://github.com/flynn/flynn/blob/master/test/test_cli.go#L606).

lmars:
@arekkas please rebase this on master to get CI running properly (we merged the host refactor recently which changed a lot of things)

arekkas:
Hey, sorry for the long response times but I'm in the middle of my exams. I'll probably fix it over the weekend. Cheers :)

arekkas:
Tests pass :palm_tree:

cli/release.go
```diff
@@ -14,19 +15,21 @@ import (
...
        flynn release show [<id>]
 
 Manage app releases.
 
 Options:
-	-t <type>          type of the release. Currently only 'docker' is supported. [default: docker]
-	-f, --file=<file>  release configuration file
+	-t <type>              type of the release. Currently only 'docker' is supported. [default: docker]
+	-f, --file=<file>      release configuration file
+	--validate=<validate>  should the image URI be validated? [default: true]
```
titanous: This should be `--no-validate` and a boolean, so no `=<validate>` is required. (will be available in `args.Bool`)

PR #1581, opened 2015-06-05
host: libvirt lxc backend now checks for name and id params in artifact uri

### LINKED ISSUE #509: Add support for tag to 'release add' and drop support in host

arekkas:
This commit drops support for using tags image uris. Image uris now require each
the name and the id parameter. Closes #509

I wrote this without being aware or able to see if this breaks anything. Please give it a very close look :)

### DIFF
--- a/host/libvirt_lxc_backend.go
+++ b/host/libvirt_lxc_backend.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net"
+	"net/url"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


titanous:
Were you able to successfully bootstrap in the dev VM with this patch?

lmars:
@arekkas @titanous I am not sure this change is necessary, I have found myself running tagged images via `flynn-host run` in the past, e.g.:

```
flynn-host run https://registry.hub.docker.com?name=ubuntu&tag=trusty bash
```

Should this validation not be in the controller where immutability is more of a concern?

titanous:
@lmars Yeah, that seems reasonable.

PR #1585, opened 2015-06-08
WIP: Scheduler v2

jzila:


lmars:
@jzila I see a lot of mutexes in use, but there are no goroutines (and I imagine it staying that way so there is just a single threaded loop), can we drop them?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/flynn/flynn/host/types"
+)
+
+type Scheduler struct {
+	utils.ControllerClient
+	utils.ClusterClient
+	log        log15.Logger
+	formations *Formations
+
+	jobs    *jobMap
```
lmars: It seems both formations and the scheduler are holding the job state in two different formats (formations are holding type -> jobID -> job, and the scheduler just jobID -> job), is this necessary?

controller/scheduler-v2/formation.go
```diff
@@ -0,0 +1,221 @@
...
+	return f
+}
+
+type Formation struct {
+	mtx       sync.Mutex
+	AppID     string
+	AppName   string
+	Release   *ct.Release
+	Artifact  *ct.Artifact
+	Processes map[string]int
```
lmars: This could probably just embed `ct.ExpandedFormation`

controller/scheduler-v2/formation.go
```diff
@@ -0,0 +1,221 @@
...
+		return nil, err
+	}
+	if len(hosts) == 0 {
+		return nil, fmt.Errorf("scheduler: no online hosts")
+	}
+
+	if hostID == "" {
+		var minCount int = math.MaxInt32
+		for _, host := range hosts {
+			jobs, err := f.listJobs(host, typ)
```
lmars: When starting new jobs, rather than listing all hosts and then querying each host for it's list of jobs every time, we should just look at the in-memory state (later we will add to the scheduler loop to make sure the in-memory state is kept accurate).

lmars: And thinking ahead slightly, `Formation.Rectify` should maybe just return what changes need to be made, and that triggers a "job request" case on the next loop. Later on the code which watches for jobs stopping / restart timers expiring will do the same thing, so all the code for actually starting jobs is in the same place:


```go
for {
	select {
	case <-s.stop:
		return
	case jobReq := <-s.jobReqs:
		s.handleJobReq(jobReq)
		continue
	default:
	}

	select {
	case <-s.stop:
		return
	case <-s.clusterSync:
		s.doClusterSync()
	case f := <-s.formationChange:
		s.handleFormationChange(f)
	case event := <-s.jobEvents:
		s.handleJobEvent(event)
	}
}
```

So any code which needs to trigger a job start / stop just sends to `s.jobReqs`, with the type being something like:

```go
type JobRequest struct {
	Type string
	HostID string
	Diff int
}
```

WDYT?

jzila:
@lmars please @mention me here when you're available so we could chat on IRC.

controller/scheduler-v2/formation.go
jzila: Agreed.

controller/scheduler-v2/scheduler.go
jzila: This is mostly modeled based on the existing scheduler, since when I started I didn't have a good enough mental model to understand what was going on. I think it's now in a place where I could clean it up.

controller/scheduler-v2/formation.go
jzila: Ooh I love this idea. I will work on refactoring to this approach.

controller/scheduler-v2/scheduler.go
lmars: Yep totally understandable, I think starting broadly from the current scheduler and then cleaning things up is fine.

The reason the current scheduler has the states in two formats is because handling formation changes and job restarts is handled separately, we should be aiming to unify that somewhat.

controller/scheduler-v2/formation.go
```diff
@@ -0,0 +1,234 @@
...
+	"math"
+
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/controller/utils"
+	"github.com/flynn/flynn/host/types"
+)
+
+type Formations struct {
+	formations map[utils.FormationKey]*Formation
+}
```
lmars: This could just be `type Formations map[utils.FormationKey]*Formation` (then use `make(Formations)` rather than `newFormations()`)

controller/scheduler-v2/formation.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	fs.formations[f.key()] = f
+	return f
+}
+
+func (fs *Formations) RectifyAll() error {
+	for _, f := range fs.formations {
+		err := f.Rectify()
+		if err != nil {
+			return err
```
lmars:
```go
if err := f.Rectify(); err != nil {
	return err
}
```

controller/scheduler-v2/formation.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (f *Formation) getHostMap(typ string) map[string]int {
+	hostMap := make(map[string]int)
+	for _, j := range f.jobs[typ] {
+		hostMap[j.HostID]++
+	}
+	return hostMap
+}
+
+func (f *Formation) jobType(job *host.Job) string {
```
lmars: This function doesn't seem to be used.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		log.Error("error getting host list", "err", err)
+		return err
+	}
+	log.Info(fmt.Sprintf("got %d hosts", len(hosts)))
+
+	for _, h := range hosts {
+		log = log.New("host_id", h.ID())
+		log.Info("getting jobs list")
+		activeJobs, err := h.ListJobs()
```
lmars: `err` here is local to the for loop, so will not be returned when the `if err != nil` check is made outside the loop. You should explicitly declare `activeJobs` and use `activeJobs, err = h.ListJobs()`

lmars:
@jzila ok I've had a good look through and I think we should make the following changes to what we have currently:

* Move code for starting / stopping jobs from `Formation` to `Scheduler` (it seems odd that job requests go `Formation` -> `Scheduler` -> `Formation`)
* Add `Formation.Update` which takes new processes as an argument, calculates the diff from current processes, stores the new processes, and returns the diff. This will be a very easy method to test, and means the scheduler can do the work of actually starting / stopping jobs based on the diff.
* Store running jobs in `Scheduler` rather than `Formation` in a `map[Formation]jobTypeMap`, update `Sync` to iterate through it and ensure each formation has the correct number of running jobs.

I have made some of these changes here: https://github.com/lmars/flynn/compare/jzila-scheduler-v2-rebased...scheduler-v2-rebased

I think the above changes will put us in a good place to implement more of the requirements in #1484, so let's refactor this PR, merge it in then start iterating in smaller PRs going forward.

PR #1590, opened 2015-06-12
test: Increase timeout for resource limits job

lmars:
This job intermittently times out in CI, and it looks as though it is just congestion on the host slowing it down.


titanous:
LGTM

PR #1591, opened 2015-06-13
test: Handle postgres states being skipped

lmars:
The postgres appliance [drains the service discovery channel before applying config](https://github.com/flynn/flynn/blob/master/appliance/postgresql/state/state.go#L255) to ensure it is always acting on current data, which means that some expected states may not happen (e.g. if a new async is started and another goes down, the state is not guaranteed to transition via a state with both asyncs present).

The tests have been updated to check that received states are expected and in-order, but some expected states can be skipped.


titanous:
LGTM

PR #1598, opened 2015-06-15
slugbuilder: Move back to upstream staticfile buildpack

titanous:
They fixed the compatibility issue we forked to work around


lmars:
LGTM

PR #1600, opened 2015-06-16
script: Use older Ubuntu Virtualbox image

titanous:
The 20150611 image is apparently corrupt, the SHA256 doesn't match what the server is serving up.


lmars:
LGTM

PR #1605, opened 2015-06-20
Improve viewing logs in CI

lmars:
The file in S3 is now a multipart document and the output of each command from `DumpLogs` gets added to a separate part with an appropriate filename.

I chose to use a single multipart file rather than uploading multiple files to S3 so that we can easily fallback to viewing the raw log in S3 if necessary (it will just include some extra multipart header lines which can be ignored).

Once a build has finished, rather than redirecting to the log in S3, I have added a new page which shows a list of generated files on the left, and the corresponding logs on the right. The logs are delivered to the browser via an SSE stream, one line at a time.

I have added a `-f` flag to `flynn-host ps` to format the output so that `DumpLogs` can create descriptive filenames (e.g. `blobstore-web-b4d8e10b-4ec7d4d73bce40d897395848c30db097.log`).

The UI could do with some improvements:

* Make the list of files "sticky", so clicking a file only scrolls the logs div, not the whole window
* When scrolling the logs, highlight the current file being viewed

@jvatic could you help with the improvements?

I will leave my custom CI running with these changes, you can see a log page here:

https://ci.flynn.io:8443/builds/20150620141337-8efc25bd

### DIFF
--- a/host/cli/ps.go
+++ b/host/cli/ps.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"sort"
 	"text/tabwriter"
+	"text/template"
 	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/docker/docker/pkg/units"
@@ -17,7 +18,7 @@ import (
 
 func init() {
 	Register("ps", runPs, `
-usage: flynn-host ps [-a|--all] [-q|--quiet]
+usage: flynn-host ps [-a|--all] [-q|--quiet] [-f <format>]
 
 List jobs`)
 }
@@ -35,6 +36,18 @@ func runPs(args *docopt.Args, client *cluster.Client) error {
 	}
 	if args.Bool["-q"] || args.Bool["--quiet"] {
 		for _, job := range jobs {
+			if format := args.String["<format>"]; format != "" {
+				tmpl, err := template.New("format").Funcs(template.FuncMap{
+					"metadata": func(key string) string { return job.Job.Metadata[key] },
+				}).Parse(format + "\n")
+				if err != nil {
+					return err
+				}
+				if err := tmpl.Execute(os.Stdout, job); err != nil {
+					return err
+				}
+				continue
+			}
 			fmt.Println(clusterJobID(job))
 		}
 		return nil
--- a/test/buildlog/log.go
+++ b/test/buildlog/log.go
@@ -1,0 +1,37 @@
+package buildlog
+
+import (
+	"io"
+	"mime/multipart"
+	"time"
+
+	"github.com/flynn/flynn/pkg/iotool"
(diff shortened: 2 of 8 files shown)


jvatic:
@lmars Implemented the UI improvements

host/cli/ps.go
```diff
@@ -35,6 +36,19 @@ func runPs(args *docopt.Args, client *cluster.Client) error {
...
+				tmpl, err := template.New("format").Funcs(template.FuncMap{
+					"metadata": func(key string) string { return job.Job.Metadata[key] },
+				}).Parse(format)
+				if err != nil {
+					return err
+				}
+				if err := tmpl.Execute(os.Stdout, job); err != nil {
+					return err
+				}
+				fmt.Println()
```
titanous: I prefer being more explicit: `os.Stdout.Write([]byte("\n"))`

host/cli/ps.go
```diff
@@ -35,6 +36,19 @@ func runPs(args *docopt.Args, client *cluster.Client) error {
 	}
 	if args.Bool["-q"] || args.Bool["--quiet"] {
 		for _, job := range jobs {
+			if format := args.String["<format>"]; format != "" {
+				tmpl, err := template.New("format").Funcs(template.FuncMap{
+					"metadata": func(key string) string { return job.Job.Metadata[key] },
+				}).Parse(format)
```
titanous: This parse should be done once, not once per job.

host/cli/ps.go
titanous: Also, we could just append a newline to the template before parsing.

test/types/types.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"io"
+	"mime/multipart"
+	"time"
+
+	"github.com/flynn/flynn/pkg/iotool"
+)
+
+type BuildLog struct {
```
titanous: This feels like it should just be a `buildlog` package or similar.

host/cli/ps.go
lmars: The metadata function is job specific, and the template fails to parse without the function defined. I could make the function accept the job (e.g. `{{ metadata . "flynn-controller.app_name" }}`) but it feels a bit awkward like that.

titanous: Ah, I see. That's fine.

lmars:
@titanous comments addressed, @jvatic thanks for the UI tweaks :smile:

titanous:
https://github.com/flynn/flynn/commit/d0a01ece687b84857e009a4dff776d2af261b35b has a typo in the message `scrollbale`

lmars:
@titanous comments addressed

titanous:
LGTM

PR #1609, opened 2015-06-21
test: Fix potential timeout in PostgresSuite

lmars:
Now that we can potentially skip received states (commit 3000177), we may have already received all the expected states when we receive a deployment event, so we need to make sure we do not wait for further states when we do not expect them.


titanous:
LGTM

Issue #1515, opened 2015-05-10
Build and release pipeline


lmars:
There are some limitations with the current build workflow, for example a lack of support for installing system dependencies (#409) and an inability to perform custom actions both pre, post and during a build.

Before proposing changes, I think it's a good idea to understand what requirements a typical app has of a prospective build & release pipeline.

As a concrete example, here is how I would like Flynn to build one of my Rails applications:
- install system libs at a specific version (e.g. `libssl 1.0.2a`)
- install specific Ruby version (only performed if requested version changes)
- bundle install gems (bundler cache enabled, so adding one gem doesn't re-bundle all gems)
- run unit / integration tests (this involves provisioning resources like a test DB for model tests, memcached for cache tests etc., and ability to skip the step completely for deploying content changes, urgent bug fixes etc.)
- compile assets

This would then create an artifact that I can feed into the following release pipeline to deploy changes to an environment (e.g. staging / production):
- run database migrations (this may involve putting the site into maintenance mode for data integrity)
- rebuild search indexes (conditionally based on whether search related files have changed)
- run cache invalidation / warming tasks
- start / stop relevant jobs (zero downtime)
- run acceptance tests (only in staging, not production)
- notify third-party APIs of deployment

Thoughts are more than welcome on what other steps could be involved.

Here are some suggested changes which could bring us closer to supporting such a workflow:
- Track changes to the root filesystem at each build step, with the final artifact being a union fs of each step, making it easy to use familiar tools such as `apt`, `rpm` etc. for system deps
- Provide a Nix installation to ease installing specific system libs and improve cacheability
- Provide a wrapper around `git push` which supports per-deploy variables (e.g. whether to go into maintenance mode, skip tests etc.)
- Add metadata to aid in determining what code changes have been made

temujin9:
https://github.com/ddollar/heroku-buildpack-apt handles the dependencies problem well. Unfortunately, maintenance appears to have lapsed; I've contacted the owner by email, about transferring the repo to one of our organizations so it can be properly maintained.

lmars:
@temujin9 whilst the apt buildpack does provide options in terms of managing system deps, I think things would be simpler and more flexible if the resulting slug contained any changes to the root filesystem (rather than just changes to `/app`), meaning we wouldn't need to fiddle with things like `LD_LIBRARY_PATH`, and would support arbitrary dependency management (e.g. `apt` / `rpm` / `nix` / `make` etc.)

temujin9:
[...]

PR #1617, opened 2015-06-24
gitreceived: adding support for submodules, by switching to full repos (instead of --bare)

### LINKED ISSUE #62: Submodule support
gitreceived doesn't currently support submodules, due to the complexity involved. See flynn-archive/gitreceived#12 for some discussion.

temujin9:
Switching to full repositories (rather than bare) so that "git checkout" and "git submodule update --init" can be performed prior to passing to the receiver.

(Per conversation with @lmars, we will probably refactor how gitreceived -> receiver -> slugbuilder works, so this went in the simplest-to-implement place. We will still need to address distributing keys into containers, separately.)

Closes #62.

### DIFF
--- a/gitreceived/gitreceived.go
+++ b/gitreceived/gitreceived.go
@@ -27,8 +27,14 @@ import (
 )
 
 const PrereceiveHookTmpl = `#!/bin/bash
-set -eo pipefail; while read oldrev newrev refname; do
-[[ $refname = "refs/heads/master" ]] && git archive $newrev | {{RECEIVER}} "$RECEIVE_REPO" "$newrev" | sed -$([[ $(uname) == "Darwin" ]] && echo l || echo u) "s/^/"$'\e[1G\e[K'"/"
+set -eo pipefail;
+git-archive-all() {
+	git checkout --quiet $1
+	git submodule --quiet update --init --recursive
+	tar --create --exclude-vcs .
+}
+while read oldrev newrev refname; do
+	[[ $refname = "refs/heads/master" ]] && git-archive-all $newrev | {{RECEIVER}} "$RECEIVE_REPO" "$newrev" | sed -u "s/^/"$'\e[1G\e[K'"/"
 done
 `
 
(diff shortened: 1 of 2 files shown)


temujin9:
(This addresses #62.)

titanous:
I'm fine with this approach, though it does need an integration test. I'll let @lmars review the shell script.

I think we should hold off on implementing this in slugbuilder, and instead do it as an optional pipeline step in #1515.

temujin9:
@titanous Can you get me pointed in the right direction on the integration tests? I'm not sure where to look.

titanous:
Sure, here's an example of testing the build cache: https://github.com/flynn/flynn/blob/bf34d2c7f8eaaa9765da00b1e310131fcd8c3b99/test/test_git_deploy.go#L52-L69

gitreceived/gitreceived.go
```diff
@@ -26,8 +26,14 @@ import (
 )
 
 const PrereceiveHookTmpl = `#!/bin/bash
-set -eo pipefail; while read oldrev newrev refname; do
-[[ $refname = "refs/heads/master" ]] && git archive $newrev | {{RECEIVER}} "$RECEIVE_REPO" "$newrev" | sed -$([[ $(uname) == "Darwin" ]] && echo l || echo u) "s/^/"$'\e[1G\e[K'"/"
+set -eo pipefail;
+git-archive-all() {
+	git checkout $1 > /dev/null
+	git submodule update --init --recursive > /dev/null
```
lmars: Either pass `--quiet` to both these commands or discard stderr too (i.e. `&>/dev/null`)

lmars:
@temujin9 I actually wrote a test to check this out:

```go
func (s *GitDeploySuite) TestGitSubmodules(t *c.C) {
	r := s.newGitRepo(t, "empty")
	t.Assert(r.git("submodule", "add", "https://github.com/flynn-examples/go-flynn-example.git"), Succeeds)
	t.Assert(r.git("commit", "-m", "Add Submodule"), Succeeds)
	t.Assert(r.flynn("create"), Succeeds)
	t.Assert(r.git("push", "flynn", "master"), Succeeds)
	t.Assert(r.flynn("run", "ls", "go-flynn-example"), SuccessfulOutputContains, "main.go")
}
```

temujin9:
The `--quiet` flag is now thrown, and @lmars' test works without error, so this should be ready to go.

titanous:
Awesome, thanks!

PR #1631, opened 2015-07-04
all: Always include host ID in job ID

### LINKED ISSUE #1238: host: layer0 Job IDs should include the host ID
Job IDs are currently UUIDs in layer0 state, but have the host ID prefixed at the layer1 level (by the controller in SQL select statements) so that layer1 components know which host to dial when performing certain operations (e.g. attaching).

We should do the host ID prefixing at the layer0 level before committing the job so that `job.ID` has the form `hostID-UUID` in all parts of the system.

lmars:
This means that job IDs can always be resolved to the host the job is running on.

Closes #1238


titanous:
LGTM

PR #1635, opened 2015-07-06
flannel: Use discoverd backend

lmars:
Switch flannel to use discoverd as the backend, which means that discoverd must start the HTTP server before flannel starts, but then start the DNS server after flannel has obtained a subnet lease and configured the network.

The flannel commit will need updating once https://github.com/flynn/flannel/pull/1 is merged.


discoverd/daemon.go
```diff
@@ -53,32 +56,46 @@ func main() {
...
+		var recursors []string
+		if *resolvers != "" {
+			recursors = strings.Split(*resolvers, ",")
+		}
+		if err := startDNSServer(state, *dnsAddr, recursors); err != nil {
+			log.Fatalf("Failed to start DNS server: %s", err)
+		}
+		log.Printf("discoverd listening for DNS on %s", *dnsAddr)
+	} else {
+		go func() {
```
titanous: This branch should be explicitly triggered by a flag.

flannel/runner.go
```diff
@@ -33,36 +34,40 @@ var networkConfigAttempts = attempt.Strategy{
...
-		_, err = client.Create(*etcdKey, data, 0)
-		if e, ok := err.(*etcd.EtcdError); ok && e.ErrorCode == 105 {
-			// Skip if the key exists
-			err = nil
-		}
-		return err
-	}); err != nil {
+	// create service and config if not present
+	client := discoverd.NewClientWithURL(status.Discoverd.URL)
+	if err := client.AddService(serviceName, nil); err != nil {
```
titanous: Looks like there could be a valid `ErrExists` response?

lmars:
@titanous comments addressed (I added a notify hook after configuring the DNS server so that the host doesn't start containers before the DNS server is up)

discoverd/daemon.go
```diff
@@ -53,44 +58,97 @@ func main() {
...
+			status, err := waitForHostNetwork()
+			if err != nil {
+				log.Fatal(err)
+			}
+			ip, _, err := net.ParseCIDR(status.Network.Subnet)
+			if err != nil {
+				log.Fatal(err)
+			}
+			addr := ip.String() + ":53"
```
titanous: Use `net.JoinHostPort`

discoverd/daemon.go
```diff
@@ -53,44 +58,97 @@ func main() {
...
+}
+
+type Status struct {
+	URL string `json:"url"`
+	DNS string `json:"dns"`
+}
+
+var (
+	status    Status
+	statusMtx sync.Mutex
```
titanous: This should be scoped to a struct only available in the `func main` scope.

titanous:
LGTM pending comments.

PR #1636, opened 2015-07-06
Controller sse firehose

jvatic:
#737

### DIFF
--- a/controller/app.go
+++ b/controller/app.go
@@ -44,9 +44,14 @@ var appNamePattern = regexp.MustCompile(`^[a-z\d]+(-[a-z\d]+)*$`)
 
 func (r *AppRepo) Add(data interface{}) error {
 	app := data.(*ct.App)
+	tx, err := r.db.Begin()
+	if err != nil {
+		return err
+	}
 	if app.Name == "" {
 		var nameID uint32
-		if err := r.db.QueryRow("SELECT nextval('name_ids')").Scan(&nameID); err != nil {
+		if err := tx.QueryRow("SELECT nextval('name_ids')").Scan(&nameID); err != nil {
+			tx.Rollback()
 			return err
 		}
 		app.Name = name.Get(nameID)
@@ -61,20 +66,33 @@ func (r *AppRepo) Add(data interface{}) error {
 		app.Strategy = "all-at-once"
 	}
 	meta := metaToHstore(app.Meta)
-	if err := r.db.QueryRow("INSERT INTO apps (app_id, name, meta, strategy) VALUES ($1, $2, $3, $4) RETURNING created_at, updated_at", app.ID, app.Name, meta, app.Strategy).Scan(&app.CreatedAt, &app.UpdatedAt); err != nil {
+	if err := tx.QueryRow("INSERT INTO apps (app_id, name, meta, strategy) VALUES ($1, $2, $3, $4) RETURNING created_at, updated_at", app.ID, app.Name, meta, app.Strategy).Scan(&app.CreatedAt, &app.UpdatedAt); err != nil {
+		tx.Rollback()
 		if postgres.IsUniquenessError(err, "apps_name_idx") {
 			return httphelper.ObjectExistsErr(fmt.Sprintf("application %q already exists", app.Name))
 		}
 		return err
 	}
 	app.ID = postgres.CleanUUID(app.ID)
+
+	if err := createEvent(tx.Exec, &ct.Event{
+		AppID:      app.ID,
+		ObjectID:   app.ID,
+		ObjectType: ct.EventTypeApp,
+	}, app); err != nil {
+		tx.Rollback()
+		return err
+	}
+	if err := tx.Commit(); err != nil {
+		return err
+	}
+
 	if !app.System() && r.defaultDomain != "" {
 		route := (&router.HTTPRoute{
 			Domain:  fmt.Sprintf("%s.%s", app.Name, r.defaultDomain),
 			Service: app.Name + "-web",
 		}).ToRoute()
-		route.ParentRef = routeParentRef(app.ID)
-		if err := r.router.CreateRoute(route); err != nil {
+		if err := createRoute(r.db, r.router, app.ID, route); err != nil {
 			log.Printf("Error creating default route for %s: %s", app.Name, err)
 		}
 	}
@@ -84,9 +102,13 @@ func (r *AppRepo) Add(data interface{}) error {
 func scanApp(s postgres.Scanner) (*ct.App, error) {
 	app := &ct.App{}
 	var meta hstore.Hstore
-	err := s.Scan(&app.ID, &app.Name, &meta, &app.Strategy, &app.CreatedAt, &app.UpdatedAt)
+	var releaseID *string
+	err := s.Scan(&app.ID, &app.Name, &meta, &app.Strategy, &releaseID, &app.CreatedAt, &app.UpdatedAt)
 	if err == sql.ErrNoRows {
 		err = ErrNotFound
+	}
+	if releaseID != nil {
+		app.ReleaseID = postgres.CleanUUID(*releaseID)
 	}
 	if len(meta.Map) > 0 {
 		app.Meta = make(map[string]string, len(meta.Map))
@@ -105,7 +127,7 @@ type rowQueryer interface {
 }
 
 func selectApp(db rowQueryer, id string, update bool) (*ct.App, error) {
-	query := "SELECT app_id, name, meta, strategy, created_at, updated_at FROM apps WHERE deleted_at IS NULL AND "
+	query := "SELECT app_id, name, meta, strategy, release_id, created_at, updated_at FROM apps WHERE deleted_at IS NULL AND "
 	var suffix string
 	if update {
 		suffix = " FOR UPDATE"
@@ -142,7 +164,8 @@ func (r *AppRepo) Update(id string, data map[string]interface{}) (interface{}, e
 				tx.Rollback()
 				return nil, fmt.Errorf("controller: expected string, got %T", v)
 			}
-			if _, err := tx.Exec("UPDATE apps SET strategy = $2, updated_at = now() WHERE app_id = $1", app.ID, strategy); err != nil {
+			app.Strategy = strategy
+			if _, err := tx.Exec("UPDATE apps SET strategy = $2, updated_at = now() WHERE app_id = $1", app.ID, app.Strategy); err != nil {
 				tx.Rollback()
 				return nil, err
 			}
@@ -171,11 +194,20 @@ func (r *AppRepo) Update(id string, data map[string]interface{}) (interface{}, e
 		}
 	}
 
+	if err := createEvent(tx.Exec, &ct.Event{
+		AppID:      app.ID,
+		ObjectID:   app.ID,
+		ObjectType: ct.EventTypeApp,
+	}, app); err != nil {
+		tx.Rollback()
+		return nil, err
+	}
+
 	return app, tx.Commit()
 }
 
 func (r *AppRepo) List() (interface{}, error) {
-	rows, err := r.db.Query("SELECT app_id, name, meta, strategy, created_at, updated_at FROM apps WHERE deleted_at IS NULL ORDER BY created_at DESC")
+	rows, err := r.db.Query("SELECT app_id, name, meta, strategy, release_id, created_at, updated_at FROM apps WHERE deleted_at IS NULL ORDER BY created_at DESC")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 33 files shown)


controller/app.go
```diff
@@ -191,69 +209,39 @@ func (r *AppRepo) List() (interface{}, error) {
...
-	row := r.db.QueryRow("SELECT r.release_id, r.artifact_id, r.data, r.created_at FROM apps a JOIN releases r USING (release_id) WHERE a.app_id = $1", id)
-	return scanRelease(row)
+	e.App = app
+	data, err := json.Marshal(e)
+	if err != nil {
+		fmt.Printf("AppRepo.createEvent error (json.Marshal): %#v\n", err)
+		return err
+	}
+	query := "INSERT INTO events (app_id, object_id, object_type, data) VALUES ($1, $2, $3, $4)"
+	if err := r.db.Exec(query, appID, appID, string(ct.EventTypeApp), data); err != nil {
```
lmars: Given these events all follow DB actions, I wonder whether we should be creating the events using DB triggers (e.g. [like this](https://github.com/flynn/flynn/blob/master/controller/schema.go#L80))?

jvatic:
The postgres `to_json` function doesn't format time fields correctly, going back to creating events from within Go

\\cc @lmars

jvatic:
Using postgres 9.4 for unit tests fixes the formatting of time fields, however there are a few other factors which make creating events in the Go code preferable over db triggers when JSON encoding in involved:

- Some fields have different names in the db vs JSON
- Some fields should be omitted when null

jvatic:
ping @lmars, @titanous this is ready for a review pass

controller/app.go
```diff
@@ -171,11 +176,17 @@ func (r *AppRepo) Update(id string, data map[string]interface{}) (interface{}, e
 		}
 	}
 
-	return app, tx.Commit()
+	if err := tx.Commit(); err != nil {
+		return nil, err
+	}
+
+	r.createAppEvent(app.ID)
```
lmars: Should we not be creating the app events inside the transaction, and throwing an error if it fails?

controller/client/client.go
```diff
@@ -486,9 +486,9 @@ outer:
 
 // StreamJobEvents streams job events to the output channel.
 func (c *Client) StreamJobEvents(appID string, output chan *ct.JobEvent) (stream.Stream, error) {
-	appEvents := make(chan *ct.AppEvent)
-	go convertAppEvents(appEvents, output)
-	return c.ResumingStream("GET", fmt.Sprintf("/apps/%s/events?object_type=%s", appID, ct.EventTypeJob), appEvents)
+	appEvents := make(chan *ct.Event)
+	go convertEvents(appEvents, output)
+	return c.ResumingStream("GET", fmt.Sprintf("/apps/%s/events?object_types=%s", appID, ct.EventTypeJob), appEvents)
```
lmars: We should probably just move to a single path for events and just return `StreamEvents({AppID: appID, ObjectTypes: [ct.EventTypeJob]}` here

controller/events.go
```diff
@@ -1,16 +1,199 @@
 package main
 
 import (
+	"database/sql"
```
lmars: This import should be `github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-sql`

controller/events.go
```diff
@@ -177,6 +370,13 @@ func (e *EventListener) Notify(event *ct.AppEvent) {
 			sub.Notify(event)
 		}
 	}
+	if event.AppID != "" {
```
lmars: Should this be inverted?

controller/events_test.go
```diff
@@ -78,3 +81,273 @@ func (s *S) TestAppEvents(c *C) {
...
+			var eventApp *ct.App
+			c.Assert(json.Unmarshal(e.Data, &eventApp), IsNil)
+			c.Assert(e.AppID, Equals, app.ID)
+			c.Assert(e.ObjectType, Equals, ct.EventTypeApp)
+			c.Assert(e.ObjectID, Equals, app.ID)
+			c.Assert(eventApp, NotNil)
+			c.Assert(eventApp.ID, Equals, app.ID)
+			fn(eventApp)
+		case <-time.After(time.Second):
+			c.Fatal(fmt.Sprintf("Timed out waiting for event %d", i))
```
lmars: Use `c.Fatalf`

controller/events_test.go
```diff
@@ -78,3 +81,273 @@ func (s *S) TestAppEvents(c *C) {
...
+			}
+			var eventApp *ct.App
+			c.Assert(json.Unmarshal(e.Data, &eventApp), IsNil)
+			c.Assert(e.AppID, Equals, app.ID)
+			c.Assert(e.ObjectType, Equals, ct.EventTypeApp)
+			c.Assert(e.ObjectID, Equals, app.ID)
+			c.Assert(eventApp, NotNil)
+			c.Assert(eventApp.ID, Equals, app.ID)
+			fn(eventApp)
+		case <-time.After(time.Second):
```
lmars: We should probably up this timeout to 10s, CI can get slow sometimes.

controller/formation.go
```diff
@@ -138,9 +141,29 @@ func (r *FormationRepo) Remove(appID, releaseID string) error {
...
+func (r *FormationRepo) createScaleEvent(formation *ct.Formation) {
+	logErr := func(err error) {
+		log.Printf("Error creating scale event: %s\n", err)
+	}
+	data, err := json.Marshal(formation.Processes)
+	if err != nil {
+		logErr(err)
+		return
+	}
+	objectID := strings.Join([]string{formation.AppID, formation.ReleaseID}, ":")
```
lmars: I'd probably just do `formation.AppID + ":" + formation.ReleaseID`

controller/key.go
```diff
@@ -82,3 +92,25 @@ func (r *KeyRepo) List() (interface{}, error) {
...
+		return
+	}
+	query := "INSERT INTO events (object_id, object_type, data) VALUES ($1, $2, $3)"
+	if err := r.db.Exec(query, key.ID, string(ct.EventTypeKey), data); err != nil {
+		logErr(err)
+	}
+}
+
+func (r *KeyRepo) createKeyDeletionEvent(keyID string) {
+	query := "INSERT INTO events (object_id, object_type) VALUES ($1, $2)"
```
lmars: Any reason not to include the data of the deleted key?

controller/release.go
```diff
@@ -71,12 +72,33 @@ func (r *ReleaseRepo) Add(data interface{}) error {
...
+	}
+	data, err := json.Marshal(release)
+	if err != nil {
+		logErr(err)
+		return
+	}
+	query := "INSERT INTO events (object_id, object_type, data) VALUES ($1, $2, $3)"
+	if err := r.db.Exec(query, release.ID, string(ct.EventTypeRelease), data); err != nil {
+		logErr(err)
+	}
```
lmars: The event creation methods are all pretty similar, could refactor to a more generic:

```go
func createEvent(objectID string, objectType ct.EventType, v interface{})
```

controller/resource.go
```diff
@@ -173,7 +179,35 @@ func (r *ResourceRepo) Remove(id string) error {
...
+	logErr := func(err error) {
+		log.Printf("Error creating resource event: %s\n", err)
+	}
+	data, err := json.Marshal(resource)
+	if err != nil {
+		logErr(err)
+		return
+	}
+	objectID := strings.Join([]string{resource.ProviderID, resource.ID}, ":")
+	query := "INSERT INTO events (object_id, object_type, data) VALUES ($1, $2, $3)"
```
lmars: Resources can be associated with many apps, so does it make sense to insert this event for all the associated apps?

controller/events.go
jvatic: Nope, this is so subscribers not filtering by app get the event. Added a comment on the next line.

jvatic:
@lmars comments addressed

controller/controller.go
```diff
@@ -305,6 +318,26 @@ func (c *controllerAPI) getRoute(ctx context.Context) (*router.Route, error) {
...
+	if err != nil {
+		return err
+	}
+	if appID == "" {
+		query := "INSERT INTO events (object_id, object_type, data) VALUES ($1, $2, $3)"
+		if _, err := dbExec(query, objectID, string(objectType), encodedData); err != nil {
+			return err
+		}
+
+	} else {
```
lmars: Return the `err` in the `if` branch and drop the `else`

lmars:
This needs a rebase, ensuring we maintain the fixes from #1667 and #1678.

controller/events.go
lmars: Not sure I follow. Let's say I subscribe to artifact changes whose events don't have `AppID` set, my subscriber will be in `subscribers[""]`, and when an artifact event comes along, `event.AppID` will be `""` so I won't get notified due to this condition?

controller/app.go
```diff
@@ -73,21 +79,29 @@ func (r *AppRepo) Add(data interface{}) error {
 			Domain:  fmt.Sprintf("%s.%s", app.Name, r.defaultDomain),
 			Service: app.Name + "-web",
 		}).ToRoute()
-		route.ParentRef = routeParentRef(app.ID)
-		if err := r.router.CreateRoute(route); err != nil {
+		if err := createRoute(r.db, r.router, app.ID, route); err != nil {
```
lmars: We should move this logic to after committing the transaction, so if the transaction fails we don't have a dangling route.

controller/routes.go
```diff
@@ -5,11 +5,32 @@ import (
...
+		return err
+	}
+	if err := rc.CreateRoute(route); err != nil {
+		return err
+	}
+	tx, err := db.Begin()
+	if err != nil {
+		return err
+	}
+	if err := createEvent(tx.Exec, appID, route.ID, ct.EventTypeRoute, route); err != nil {
```
lmars: Does this need to be in a transaction?

controller/routes.go
```diff
@@ -65,5 +80,21 @@ func (c *controllerAPI) DeleteRoute(ctx context.Context, w http.ResponseWriter,
 		respondWithError(w, err)
 		return
 	}
+	tx, err := c.appRepo.db.Begin()
+	if err != nil {
+		respondWithError(w, err)
+		return
+	}
+	appID := c.getApp(ctx).ID
+	if err := createEvent(tx.Exec, appID, route.ID, ct.EventTypeRouteDeletion, route); err != nil {
```
lmars: Does this need to be in a transaction?

controller/events.go
jvatic: It's to handle the case where the subscriber is in `subscribers[""]` and `event.AppID` is set, e.g. subscribing to all app events but not specifying an object ID.

controller/routes.go
jvatic: `tx.Exec` returns `(sql.Result, error)` and `db.Exec` just returns `error`, the transaction allows `createEvent` to be simpler. I suppose `db.Exec` could be wrapped in a function matching the `tx.Exec` signature.

controller/routes.go
jvatic: Ditto

controller/app.go
jvatic: Done

jvatic:
Route events should be streamed from the router

jvatic:
@lmars comments addressed

I'm refactoring the route events to be streamed from the router.

jvatic:
@lmars ready for another review pass

router/api.go
```diff
@@ -152,3 +155,45 @@ func deleteRoute(params martini.Params, router *Router, r render.Render) {
...
+	tcpEvents := make(chan *router.Event)
+	sseEvents := make(chan *router.StreamEvent)
+	go r.HTTP.Watch(httpEvents)
+	go r.TCP.Watch(tcpEvents)
+	defer r.HTTP.Unwatch(httpEvents)
+	defer r.TCP.Unwatch(tcpEvents)
+	go func() {
+		for {
+			select {
+			case e, ok := <-httpEvents:
```
jvatic: Don't need the select

router/api.go
```diff
@@ -152,3 +155,45 @@ func deleteRoute(params martini.Params, router *Router, r render.Render) {
...
+					Route: e.Route,
+					Error: e.Error,
+				}
+			}
+		}
+	}()
+	go func() {
+		for {
+			select {
+			case e, ok := <-tcpEvents:
```
jvatic: Ditto

router/api.go
```diff
@@ -152,3 +155,45 @@ func deleteRoute(params martini.Params, router *Router, r render.Render) {
...
+				if !ok {
+					return
+				}
+				sseEvents <- &router.StreamEvent{
+					Event: e.Event,
+					Route: e.Route,
+					Error: e.Error,
+				}
```
jvatic: Pull this out into a func as this is identical to above

controller/controller.go
```diff
@@ -105,6 +115,42 @@ func main() {
...
+	if err != nil {
+		return err
+	}
+	go func() {
+		for {
+			e, ok := <-events
+			if !ok {
+				return
+			}
+			route := e.Route
```
jvatic: `route.ID = postgres.CleanUUID(route.ID)`

controller/controller.go
```diff
@@ -105,6 +115,42 @@ func main() {
...
+			var appID string
+			if strings.HasPrefix(route.ParentRef, routeParentRefPrefix) {
+				appID = strings.TrimPrefix(route.ParentRef, routeParentRefPrefix)
+			}
+			eventType := ct.EventTypeRoute
+			if e.Event == "remove" {
+				eventType = ct.EventTypeRouteDeletion
+			}
+			if err := createEvent(wrapDBExec(db.Exec), appID, route.ID, eventType, route); err != nil {
```
jvatic: This should probably set `events.unique_id`

controller/controller.go
```diff
@@ -305,6 +364,21 @@ func (c *controllerAPI) getRoute(ctx context.Context) (*router.Route, error) {
 	return route, err
 }
 
+func createEvent(dbExec func(string, ...interface{}) (sql.Result, error), appID, objectID string, objectType ct.EventType, data interface{}) error {
```
jvatic: Refactor to `func createEvent(dbExec func(string, ...interface{}) (sql.Result, error), e *ct.Event, data interface{}) error`

jvatic:
There should be an integration test for route events

controller/controller.go
jvatic: Done

jvatic:
ping @lmars for review

controller/schema.go
```diff
@@ -42,27 +42,32 @@ func migrateDB(db *sql.DB) error {
...
-		`CREATE FUNCTION notify_app_event() RETURNS TRIGGER AS $$
+
+		`CREATE UNIQUE INDEX ON events (unique_id)`,
+		`CREATE FUNCTION notify_event() RETURNS TRIGGER AS $$
     BEGIN
-	PERFORM pg_notify('app_events', NEW.event_id || ':' || NEW.app_id);
+  IF NEW.app_id IS NOT NULL THEN
+    PERFORM pg_notify('events', NEW.event_id || ':' || NEW.app_id);
+  ELSE
+    PERFORM pg_notify('events', NEW.event_id || '');
```
titanous: Is the append to an empty string necessary?

controller/artifact.go
```diff
@@ -27,17 +27,38 @@ func (r *ArtifactRepo) Add(data interface{}) error {
...
 	}
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
 	a.ID = postgres.CleanUUID(a.ID)
-	return err
+	if err := createEvent(tx.Exec, &ct.Event{
+		ObjectID:   a.ID,
+		ObjectType: ct.EventTypeArtifact,
```
titanous: Looks like this event shouldn't be triggered if we hit the uniqueness error.

controller/schema.go
jvatic: `event_id` is an int, the append converts it to a string which is necessary. This should probably use a more obvious method of conversion though.

controller/resource.go
```diff
@@ -58,7 +58,16 @@ func (rr *ResourceRepo) Add(r *ct.Resource) error {
 		}
 		r.Apps[i] = postgres.CleanUUID(r.Apps[i])
 	}
-	r.ID = postgres.CleanUUID(r.ID)
+	for _, appID := range r.Apps {
+		if err := createEvent(tx.Exec, &ct.Event{
+			AppID:      appID,
+			ObjectID:   r.ProviderID + ":" + r.ID,
```
titanous: The resource ID should be unique.

controller/schema.go
titanous: Yeah, use a type coercion thing.

jvatic:
@titanous comments addressed

titanous:
The commits need some body to describe a bit about what's going on, but other than that this LGTM. :shipit:

PR #1638, opened 2015-07-07
test: Redirect old build logs to S3

lmars:
I've added versions to builds to determine whether they are compatible with the scrolling log viewer.

I plan to manually migrate the CI database so that all builds before `20150626200913-1e351d94` are v1, and all after are v2.


titanous:
LGTM

PR #1639, opened 2015-07-07
host: Remove legacy volume support

### LINKED ISSUE #1168: Remove legacy volume support
There is a broken path somewhere in flynn-host that causes the legacy volume path to be empty, putting volumes in the root of the filesystem.

titanous:
It has been fully replaced by the new volume system.

Closes #1168


lmars:
LGTM

PR #1640, opened 2015-07-07
test: Fix job ID parsing in DumpLogs

lmars:
app names can contain dashes, so just take the last two strings from the split, with the assumption that the job ID only contains one dash.


titanous:
LGTM

PR #1641, opened 2015-07-08
host: Clean up mounts that leak into libvirt_lxc

### LINKED ISSUE #1125: HostSuite.TestVolumePersistence failure

titanous:
When a mount namespace is created, it inherits all of the mounts from its parent. If they are not `umount(2)`'ed, they will remain and prevent devices from being cleaned up. This currently causes problems when destroying ZFS volumes and cleaning up temporary container rootfs mountpoints.

The libvirt-lxc container system creates two mount namespaces for each container: one in the `libvirt_lxc` supervisor process, and one in the process that execs the actual container process. The latter cleans up the mounts that leak in, but the former does not. This is tracked upstream at https://bugzilla.redhat.com/show_bug.cgi?id=1067489

We work around this issue by adding a small `flynn-nsumount` executable that can reach into an arbitrary mount namespace (using a pid) and call `umount(2)` against a list of paths. This executable is called immediately after launching each libvirt-lxc container to clean up any leaked mountpoints that are known to cause trouble with flynn.

Note that this is not a complete fix, just a temporary band-aid. The following issues remain:

- Any devices such as USB drives will not be automatically cleaned
  up, as we only unmount paths that we know we control. This could
  probably be solved, but the hope is that this workaround will not
  be required for very long.
- There is a non-zero time period between creating the namespace and
  unmounting the leaked mounts, which could cause a race with
  processes that are destroying mounts. In practice this probably
  won't happen and can be worked around by retrying requests.

Closes #1125

### DIFF
--- a/host/cli/download.go
+++ b/host/cli/download.go
@@ -75,7 +75,7 @@ func runDownload(args *docopt.Args) error {
 	if err := os.MkdirAll(args.String["--bin-dir"], 0755); err != nil {
 		return fmt.Errorf("error creating bin dir: %s", err)
 	}
-	for _, path := range []string{"/flynn-linux-amd64", "/flynn-init"} {
+	for _, path := range []string{"/flynn-linux-amd64", "/flynn-init", "/flynn-nsumount"} {
 		dst, err := downloadGzippedFile(client, path, args.String["--bin-dir"])
 		if err != nil {
 			return err
--- a/host/libvirt_lxc_backend.go
+++ b/host/libvirt_lxc_backend.go
@@ -706,6 +706,10 @@ func (c *libvirtContainer) watch(ready chan<- error) error {
 		if err := c.cleanupMounts(c.domain.ID); err != nil {
 			g.Log(grohl.Data{"at": "cleanup_mounts", "status": "error", "err": err})
 		}
+
+		// The bind mounts are copied when we spin up the container, we don't
+		// need them in the root mount namespace any more.
+		c.unbindMounts()
 		g.Log(grohl.Data{"at": "cleanup_mounts", "status": "finish"})
 	}()
 
@@ -793,8 +797,8 @@ func (c *libvirtContainer) watch(ready chan<- error) error {
 	return nil
 }
 
-func (c *libvirtContainer) cleanup() error {
-	g := grohl.NewContext(grohl.Data{"backend": "libvirt-lxc", "fn": "cleanup", "job.id": c.job.ID})
+func (c *libvirtContainer) unbindMounts() {
+	g := grohl.NewContext(grohl.Data{"backend": "libvirt-lxc", "fn": "unbind_mounts", "job.id": c.job.ID})
 	g.Log(grohl.Data{"at": "start"})
 
 	if err := syscall.Unmount(filepath.Join(c.RootPath, ".containerinit"), 0); err != nil {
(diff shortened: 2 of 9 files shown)


host/cli/download.go
```diff
@@ -75,7 +75,7 @@ func runDownload(args *docopt.Args) error {
 	if err := os.MkdirAll(args.String["--bin-dir"], 0755); err != nil {
 		return fmt.Errorf("error creating bin dir: %s", err)
 	}
-	for _, path := range []string{"/flynn-linux-amd64", "/flynn-init"} {
+	for _, path := range []string{"/flynn-linux-amd64", "/flynn-init", "/nsumount"} {
```
lmars: The path should be `/flynn-nsumount`

lmars:
LGTM pending comment

host/libvirt_lxc_backend.go
```diff
@@ -612,6 +626,27 @@ func (l *LibvirtLXCBackend) openLog(id string) *logbuf.Log {
...
+	args := make([]string, 1, len(list)+1)
+	args[0] = strconv.Itoa(pid)
+	for _, m := range list {
+		if strings.HasPrefix(m.Mountpoint, imageRoot) || strings.HasPrefix(m.Mountpoint, flynnRoot) {
+			args = append(args, m.Mountpoint)
+		}
+	}
+
+	if err := exec.Command(c.l.UmountPath, args...).Run(); err != nil {
+		return fmt.Errorf("host: error running nsumount %s: %s", pid, err)
```
lmars: The pid format should be `%d`

lmars:
@titanous CI build [20150708185342-5617bb11](https://ci.flynn.io/builds/20150708185342-5617bb11) failed to start two blobstore jobs in the release cluster, with the following errors in the host logs:

```
now=2015-07-08T18:59:25+0000 err="host: error running nsumount %!s(int=4713): exit status 1" at=cleanup_mounts app=host job.id=df56904b-eeeaca59951e4cceb05372b6142b0b7c backend=libvirt-lxc fn=run status=error

now=2015-07-08T18:59:25+0000 app=host fn=run job.id=df56904b-951251a16962410d9f94f509bddcf6e9 backend=libvirt-lxc at=cleanup_mounts status=error err="host: error running nsumount %!s(int=4631): exit status 1"
```

Not very informative, we should probably log the output of the command.

titanous:
Updated to attempt to unmount every path, print all errors, and not blow up if there are errors (since this is a best effort thing anyway).

titanous:
I moved the unmount code and added some more cleanup (fixing a major pinkerton cleanup bug along the way). I believe this is ready to merge.

host/libvirt_lxc_backend.go
```diff
@@ -665,6 +698,18 @@ func (c *libvirtContainer) watch(ready chan<- error) error {
...
 
+	go func() {
+		// Workaround for mounts leaking into the libvirt_lxc supervisor process,
+		// see https://github.com/flynn/flynn/issues/1125 for details. Remove
+		// nsumount from the tree when deleting.
+		g.Log(grohl.Data{"at": "cleanup_mounts", "status": "start"})
+		if err := c.cleanupMounts(c.domain.ID); err != nil {
+			g.Log(grohl.Data{"at": "cleanup_mounts", "status": "error", "err": err})
+		}
+		c.unbindMounts()
```
lmars: It took me a minute to realise why we are immediately unbinding all the mounts, so I think this could do with a comment that by this point, the mount namespace has been cloned and these are no longer necessary in the root namespace.

lmars:
LGTM pending comment

PR #1648, opened 2015-07-09
scheduler-v2: Formation Sync updates

jzila:
TODO: still need to refactor tests to handle new formations.

### DIFF
--- a/controller/scheduler-v2/formation.go
+++ b/controller/scheduler-v2/formation.go
@@ -52,17 +52,9 @@ func (f *Formation) Update(procs map[string]int) map[string]int {
 	return diff
 }
 
-type formationCounts map[utils.FormationKey]map[string]int
+type formationJobs map[utils.FormationKey]map[string]int
 
-func NewFormationCounts(fs Formations) formationCounts {
-	fc := make(formationCounts)
-	for fKey := range fs {
-		fc[fKey] = make(map[string]int)
-	}
-	return fc
-}
-
-func (fc formationCounts) AddJob(j *Job) {
+func (fc formationJobs) AddJob(j *Job) {
 	key := j.Formation.key()
 	f, ok := fc[key]
 	if !ok {
--- a/controller/scheduler-v2/formation.go
+++ b/controller/scheduler-v2/formation.go
@@ -51,3 +51,23 @@ func (f *Formation) Update(procs map[string]int) map[string]int {
 	f.Processes = procs
 	return diff
 }
+
+type formationCounts map[utils.FormationKey]map[string]int
+
+func NewFormationCounts(fs Formations) formationCounts {
+	fc := make(formationCounts)
+	for fKey := range fs {
+		fc[fKey] = make(map[string]int)
+	}
+	return fc
+}
+
+func (fc formationCounts) AddJob(j *Job) {
+	key := j.Formation.key()
+	f, ok := fc[key]
+	if !ok {
+		fc[key] = make(map[string]int)
+		f = fc[key]
+	}
+	f[j.Type]++
+}
--- a/controller/scheduler-v2/formation.go
+++ b/controller/scheduler-v2/formation.go
@@ -52,14 +52,23 @@ func (f *Formation) Update(procs map[string]int) map[string]int {
 	return diff
 }
 
-type formationProcesses map[utils.FormationKey]map[string]int
+type formationJobs map[utils.FormationKey]map[string][]*Job
 
-func (fc formationProcesses) AddJob(j *Job) {
+func (fc formationJobs) AddJob(j *Job) {
 	key := j.Formation.key()
-	f, ok := fc[key]
+	_, ok := fc[key]
 	if !ok {
-		fc[key] = make(map[string]int)
-		f = fc[key]
+		fc[key] = make(map[string][]*Job)
 	}
-	f[j.Type]++
+	fc[key][j.Type] = append(fc[key][j.Type], j)
+}
+
+func (fc formationJobs) GetProcesses(key utils.FormationKey) map[string]int {
+	procs := make(map[string]int)
+
+	for typ, jobs := range fc[key] {
+		procs[typ] = len(jobs)
+	}
+
+	return procs
 }
--- a/controller/scheduler-v2/formation.go
+++ b/controller/scheduler-v2/formation.go
@@ -54,6 +54,14 @@ func (f *Formation) Update(procs map[string]int) map[string]int {
 
 type formationJobs map[utils.FormationKey]map[string][]*Job
 
+func NewFormationJobs(jobs map[string]*Job) formationJobs {
+	fj := make(formationJobs)
+	for _, job := range jobs {
+		fj.AddJob(job)
+	}
+	return fj
+}
+
 func (fc formationJobs) AddJob(j *Job) {
 	key := j.Formation.key()
 	_, ok := fc[key]
--- a/controller/scheduler-v2/formation.go
+++ b/controller/scheduler-v2/formation.go
@@ -52,9 +52,9 @@ func (f *Formation) Update(procs map[string]int) map[string]int {
 	return diff
 }
 
-type formationJobs map[utils.FormationKey]map[string]int
+type formationProcesses map[utils.FormationKey]map[string]int
 
-func (fc formationJobs) AddJob(j *Job) {
+func (fc formationProcesses) AddJob(j *Job) {
 	key := j.Formation.key()
 	f, ok := fc[key]
 	if !ok {
--- a/controller/scheduler-v2/scheduler.go
+++ b/controller/scheduler-v2/scheduler.go
@@ -62,6 +62,9 @@ func (s *Scheduler) Run() error {
 	log.Info("starting scheduler loop")
 	defer log.Info("exiting scheduler loop")
 
+	jobSync := time.Tick(30 * time.Second)
+	formationSync := time.Tick(time.Minute)
+
 	for {
 		// first, check if we should stop or process pending job events
 		select {
@@ -73,16 +76,18 @@ func (s *Scheduler) Run() error {
 		default:
 		}
 
-		log.Info("starting cluster sync")
-		if err := s.Sync(); err != nil {
-			log.Error("error performing cluster sync", "err", err)
-			continue
-		}
-
-		log.Info("starting watching events")
+		// next, handle sync and formation change events
 		select {
 		case <-s.stop:
 			return nil
+		case <-jobSync:
+			log.Info("starting cluster sync")
+			if err := s.SyncJobs(); err != nil {
+				log.Error("error performing cluster sync", "err", err)
+				continue
+			}
+		case <-formationSync:
+			s.SyncFormations()
 		case fc := <-s.formationChange:
 			if err := s.FormationChange(fc); err != nil {
 				log.Error("error performing formation change", "err", err)
@@ -94,7 +99,7 @@ func (s *Scheduler) Run() error {
 	return nil
 }
 
-func (s *Scheduler) Sync() (err error) {
+func (s *Scheduler) SyncJobs() (err error) {
 	log := s.log.New("fn", "Sync")
 
 	defer func() {
@@ -115,7 +120,7 @@ func (s *Scheduler) Sync() (err error) {
 	for k, v := range s.jobs {
 		unSyncedJobs[k] = v
 	}
-	fc := NewFormationCounts(s.formations)
+	fc := make(formationJobs)
 	for _, h := range hosts {
 		log = log.New("host.id", h.ID())
 		log.Info("getting jobs list")
@@ -194,7 +199,7 @@ func (s *Scheduler) SyncFormations() (err error) {
 	return err
 }
 
-func (s *Scheduler) rectifyFormations(fc formationCounts) (err error) {
+func (s *Scheduler) rectifyFormations(fc formationJobs) (err error) {
 	for fKey, formation := range s.formations {
 		for typ, count := range formation.Processes {
 			diff := count - fc[fKey][typ]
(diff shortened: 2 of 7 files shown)


lmars:
It's probably worth adding a test for new formations to avoid regressions.

jzila:
Yeah I thought so too. I'm implementing the formation rectification and will add some tests as part of that.

jzila:
Temporarily closing PR...

jzila:
@lmars this should be ready for review with fixes for the [first 3 items](https://github.com/flynn/flynn/issues/1484#issuecomment-119734001)

controller/scheduler-v2/formation.go
```diff
@@ -51,3 +51,23 @@ func (f *Formation) Update(procs map[string]int) map[string]int {
 	f.Processes = procs
 	return diff
 }
+
+type formationCounts map[utils.FormationKey]map[string]int
```
lmars: I think this would be better named `formationJobs`

controller/scheduler-v2/formation.go
```diff
@@ -51,3 +51,23 @@ func (f *Formation) Update(procs map[string]int) map[string]int {
...
+
+func NewFormationCounts(fs Formations) formationCounts {
+	fc := make(formationCounts)
+	for fKey := range fs {
+		fc[fKey] = make(map[string]int)
```
lmars: Do we need to create all these maps upfront, given they are lazy initialised in `AddJob`?

controller/scheduler-v2/scheduler.go
```diff
@@ -101,6 +101,8 @@ func (s *Scheduler) Sync() (err error) {
 		s.sendEvent(NewEvent(EventTypeClusterSync, err, nil))
 	}()
 
+	s.SyncFormations()
```
lmars: I think we should be triggering `SyncFormations` less often than `Sync` (and maybe call the latter `SyncJobs`), e.g. in `Scheduler.Run`:

```go
jobSync := time.Tick(30 * time.Second)
formationSync := time.Tick(time.Minute)

for {
	select {
	case <-jobSync:
		s.SyncJobs()
	case <-formationSync:
		s.SyncFormations()
	default:
	}

	// code for starting jobs
}
```

controller/scheduler-v2/scheduler.go
```diff
@@ -145,12 +154,69 @@ func (s *Scheduler) Sync() (err error) {
...
+			}
+			delete(fc[fKey], typ)
+		}
+		if len(fc[fKey]) == 0 {
+			delete(fc, fKey)
+		} else {
+			err = fmt.Errorf("Unknown job type in formation %v", fc[fKey])
+		}
+	}
+	return err
```
lmars: The rectifying of formations after `SyncFormations` needs to update the in-memory formations so that we have the latest expected process counts in memory, so this should probably just trigger a `FormationChange` for each formation retrieved in the sync.

controller/scheduler-v2/scheduler.go
```diff
@@ -145,12 +154,69 @@ func (s *Scheduler) Sync() (err error) {
...
+				j := NewJob(f, jobType, h.ID(), job.ID)
+				s.AddJob(j, appName, utils.JobMetaFromMetadata(job.Metadata))
+				fc.AddJob(j)
 			}
 		}
 	}
-	// TODO rectify formations
-	return
+	for jobID := range unSyncedJobs {
+		delete(s.jobs, jobID)
```
lmars: I'm not really sure what this is achieving?

controller/scheduler-v2/scheduler.go
```diff
@@ -297,6 +364,7 @@ func (s *Scheduler) stopJob(req *JobRequest) (err error) {
 		}
 		s.sendEvent(NewEvent(EventTypeJobStop, err, nil))
 	}()
+	//FIXME: HostID and JobID are most likely empty right now
```
lmars: If `JobID` is not set, this should find all the formation jobs of the given type and deterministically pick one to stop (e.g. the one created most recently)

controller/scheduler-v2/scheduler.go
jzila: Ensures that s.jobs doesn't have orphaned jobs that no longer exist.

lmars: Ah ok, maybe rename `unSyncedJobs` to `inactiveJobs` or `stoppedJobs`?

jzila:
@lmars ready for review.

controller/scheduler-v2/scheduler.go
```diff
@@ -62,63 +69,92 @@ func (s *Scheduler) Run() error {
...
-		select {
-		case <-s.stop:
-			return nil
-		case fc := <-s.formationChange:
-			if err := s.FormationChange(fc); err != nil {
-				log.Error("error performing formation change", "err", err)
-				continue
+		handled := false
+		for _, chHandler := range chanList {
+			handled = s.handleChan(chHandler)
```
lmars: I think we should be explicit here to keep things as readable as possible:

```go
jobSync := time.Tick(30 * time.Second)
formationSync := time.Tick(time.Minute)

select {
case <-s.stop:
	return
case <-rectifyJobs:
	s.RectifyJobs()
default:
}

select {
case <-jobSync:
	s.SyncJobs()
case <-formationSync:
	s.SyncFormations()
case ef := <-formationChange:
	s.HandleFormationChange(ef)
case req := <-jobRequests:
	s.HandleJobRequest(req)
}
```

WDYT?

controller/scheduler-v2/scheduler.go
```diff
@@ -62,63 +69,92 @@ func (s *Scheduler) Run() error {
...
+			if handled {
+				break
 			}
-		case <-time.After(time.Second):
+		}
+		if !handled {
+			<-time.After(time.Second)
```
lmars: Theres no need for this sleep as the select will just block until there is something to do.

controller/scheduler-v2/scheduler.go
```diff
@@ -62,63 +69,92 @@ func (s *Scheduler) Run() error {
...
 
 	defer func() {
 		s.sendEvent(NewEvent(EventTypeClusterSync, err, nil))
+		if err != nil {
+			s.formationSync <- struct{}{}
+		}
 	}()
 
-	log.Info("getting host list")
+	drainChannel(s.jobSync)
```
lmars: I guess you are draining the channel to prevent lags in the main loop leading to two job syncs happening one after the other, which we could solve by only triggering a job sync 30s after a previous one finishes, e.g.:

```go
jobSyncTimer := time.NewTimer(30*time.Second)
select {
case <-jobSyncTimer.C:
	s.SyncJobs()
	jobSyncTimer.Reset(30*time.Second)
}
```

controller/scheduler-v2/scheduler.go
```diff
@@ -62,63 +69,92 @@ func (s *Scheduler) Run() error {
...
+			<-time.After(time.Second)
 		}
 	}
 	return nil
 }
 
-func (s *Scheduler) Sync() (err error) {
-	log := s.log.New("fn", "Sync")
+func (s *Scheduler) SyncJobs() (err error) {
+	fLog := s.log.New("fn", "Sync")
```
lmars: Was there a particular reason for the variable name change?

controller/scheduler-v2/scheduler.go
jzila: I tried that; `select case` doesn't prioritize, and I need events to be handled in a specific priority. The way I made it does that: each event is handled explicitly in a priority order. I also had it as follows, but it wasn't as readable, I think.
```go
select {
case <-s.rectifyJobs:
    ...
    continue
default:
}
select {
case <- jobSync:
    ...
    continue
default:
}
...etc
```

controller/scheduler-v2/scheduler.go
jzila: If we have multiple selects, then we need to give each an equal chance every interval to evaluate.

controller/scheduler-v2/scheduler.go
jzila: Other events can trigger the jobSync channel.

controller/scheduler-v2/scheduler.go
jzila: Lower down in the function there's a `log = log.New(...)`, which was recursing the parameters. By making the variables explicit, we no longer have that problem.

controller/scheduler-v2/scheduler.go
```diff
@@ -135,88 +171,102 @@ func (s *Scheduler) Sync() (err error) {
...
 					continue
 				}
 
 				log.Info("getting formation")
-				f, err := s.getFormation(appID, appName, releaseID)
-				if err != nil {
-					log.Error("error getting formation", "err", err)
+				f := s.formations.Get(appID, releaseID)
+				if f == nil {
+					err = fmt.Errorf("No such formation obtained")
```
lmars: This should probably just be a flag which triggers the formation sync:

```go
unknownFormations := false
defer func() {
	if unknownFormations {
		s.formationSync <- struct{}{}
	}
}()
```

and thinking about this, we probably want to trigger another job sync straight after the triggered formation sync, and before we handle the formation change for the new formation, which complicates the loop somewhat.

This is quite an edge case: two schedulers running jobs, one sees a new formation, starts a job, the other syncs with the cluster before seeing the new formation. Given there is a chance that this can happen, we shouldn't ignore it, but I wonder whether we should just add the job to the in-memory map and move on?

controller/scheduler-v2/scheduler.go
```diff
@@ -135,88 +171,102 @@ func (s *Scheduler) Sync() (err error) {
...
-			App:       &ct.App{ID: appID, Name: appName},
-			Release:   release,
-			Artifact:  artifact,
-			Processes: formation.Processes,
-		})
-		log.Info("adding formation")
-		f = s.formations.Add(f)
 	}
-	return f, nil
+	return err
```
lmars: This should trigger a rectify.

controller/scheduler-v2/scheduler.go
jzila: Definitely. An edge case I missed, of which I'm sure there will be more.

controller/scheduler-v2/scheduler.go
jzila: Yep.

controller/scheduler-v2/scheduler.go
lmars: In fact we should just lookup the formation from the controller.

jzila: I actually prefer your first solution: trigger the formation sync.

controller/scheduler-v2/scheduler.go
lmars: Why do we need to prioritise?

controller/scheduler-v2/scheduler.go
lmars: I still don't follow, why would we want to block the main loop for a whole second?

controller/scheduler-v2/scheduler.go
lmars: Sync jobs is just to avoid getting out of sync with the cluster (under normal operation we will rely on watching jobs from each host to maintain the in-memory state), so I don't think anything else should be triggering this.

controller/scheduler-v2/scheduler.go
lmars: Sure, but if we trigger the formation sync, then we have to also trigger another job sync afterwards so that we have this job in-memory, which seems like a lot of work when we could just do a lookup of the formation from the controller and move on.

controller/scheduler-v2/scheduler.go
```diff
@@ -390,6 +525,66 @@ func (s *Scheduler) Jobs() map[string]*Job {
...
+
+func drainChannel(ch chan interface{}) {
+	for {
+		select {
+		case <-ch:
+		default:
+			return
+		}
+	}
+}
```
lmars: Rather than draining channels, we should use channels with a buffer length of 1 and do non-blocking writes like:

```go
func triggerJobSync() {
	select {
	case jobSync <- struct{}{}
	default:
	}
}
```

controller/scheduler-v2/scheduler.go
jzila: 1) Several events can conflict. E.g. jobSync reveals jobs with unknown formations, triggers formation sync. Formation sync pulls down the new formations, which triggers a jobSync and a rectify. If the rectify happens before the jobSync, it will launch a new job instead of syncing to the existing one.
2) Simplifies testing.

controller/scheduler-v2/scheduler.go
jzila: I guess we don't

controller/scheduler-v2/scheduler.go
jzila: As above, a jobSync can result in jobs with unknown formations, requiring a formation sync and then another job sync.

controller/scheduler-v2/scheduler.go
jzila: Yep, thanks.

controller/scheduler-v2/scheduler.go
jzila: OK I'm fine with that. And that might remove the need to prioritize the events.

controller/scheduler-v2/scheduler.go
```diff
@@ -62,63 +69,71 @@ func (s *Scheduler) Run() error {
...
-		default:
-		}
-
-		log.Info("starting cluster sync")
-		if err := s.Sync(); err != nil {
-			log.Error("error performing cluster sync", "err", err)
-			continue
+	jobTicker := time.Tick(30 * time.Second)
+	formationTicker := time.Tick(time.Minute)
+	formationTimeout := time.After(time.Second)
```
lmars: Eventually we are going to stream formation changes from the controller, and the stream first sends all current formations, so we don't need this initial trigger.

jzila: We need this for testing at the very least. I can change it once we implement streaming.

controller/scheduler-v2/scheduler.go
```diff
@@ -62,63 +69,71 @@ func (s *Scheduler) Run() error {
...
+	go func() {
+		for {
+			select {
+			case <-jobTicker:
+				triggerChan(s.jobSync)
+			case <-formationTicker:
+				triggerChan(s.formationSync)
+			case <-formationTimeout:
+				triggerChan(s.formationSync)
+			}
```
lmars: We should just select on the ticker channels in the main loop and drop this goroutine.

jzila: I use the channels in the tests to trigger them manually.

PR #1658, opened 2015-07-13
installer: Fix building Azure authorize URL

### LINKED ISSUE #1655: Not getting a call back to local host after creating Azure app with oauth token endpoint
The flynn installer is not being called after setting up the Azure application and providing the local host address, application id and token endpoint, this was working two releases ago I am now on v20150710.0
The error should be recreatable by doing an Azure install using the installer.

jvatic:
Closes #1655


titanous:
LGTM

PR #1667, opened 2015-07-17
Concurrency fixes

titanous:
Fix races and other concurrency bugs found while updating to Go 1.5.

### DIFF
--- a/controller/events.go
+++ b/controller/events.go
@@ -21,8 +21,9 @@ const eventBufferSize = 1000
 // EventSubscriber receives app events from the EventListener loop and maintains
 // it's own loop to forward those events to the Events channel.
 type EventSubscriber struct {
-	Events chan *ct.AppEvent
-	Err    error
+	Events  chan *ct.AppEvent
+	Err     error
+	errOnce sync.Once
 
 	l          *EventListener
 	queue      chan *ct.AppEvent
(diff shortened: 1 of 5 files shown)


controller/events.go
```diff
@@ -170,9 +170,8 @@ func (e *EventListener) Listen() error {
 // Notify notifies all sbscribers of the given event.
 func (e *EventListener) Notify(event *ct.AppEvent) {
 	e.subMtx.RLock()
-	subscribers := e.subscribers
-	e.subMtx.RUnlock()
-	if subs, ok := subscribers[event.AppID]; ok {
+	defer e.subMtx.RUnlock()
```
lmars: Locking the mutex for the duration of `sub.Notify` can cause issues because it can call [sub.CloseWithError](https://github.com/flynn/flynn/blob/master/controller/events.go#L49), which ends up calling [l.Unsubscribe](https://github.com/flynn/flynn/blob/master/controller/events.go#L68), which then [tries to lock the mutex](https://github.com/flynn/flynn/blob/master/controller/events.go#L123).

controller/events.go
```diff
@@ -198,8 +198,8 @@ func (e *EventListener) IsClosed() bool {
 func (e *EventListener) CloseWithError(err error) {
 	e.SetClosed()
 	e.subMtx.RLock()
+	defer e.subMtx.RUnlock()
 	subscribers := e.subscribers
-	e.subMtx.RUnlock()
 	for _, subs := range subscribers {
 		for sub := range subs {
 			sub.CloseWithError(err)
```
lmars: Should this now be in a goroutine?

lmars:
LGTM pending comment

PR #1671, opened 2015-07-21
Various fixes

lmars:
A collection of small fixes I have made whilst working on flynn-host zero-downtime updates.

PR #1672, opened 2015-07-21
host: Support zero-downtime updates

lmars:
Opening for early feedback on the approach.

All but the last 3 commits are in #1671 so will disappear once that is merged.

### DIFF
--- a/host/cli/update.go
+++ b/host/cli/update.go
@@ -2,12 +2,15 @@ package cli
 
 import (
 	"bytes"
+	"compress/gzip"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"sync"
+	"syscall"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/docker/docker/pkg/term"
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
@@ -16,17 +19,22 @@ import (
 	"github.com/flynn/flynn/pinkerton/layer"
 	"github.com/flynn/flynn/pkg/cluster"
 	"github.com/flynn/flynn/pkg/exec"
+	"github.com/flynn/flynn/pkg/tufutil"
 )
 
 func init() {
 	Register("update", runUpdate, `
-usage: flynn-host update [--driver=<name>] [--root=<path>] [--repository=<uri>] [--tuf-db=<path>]
+usage: flynn-host update [options]
 
 Options:
   -d --driver=<name>       image storage driver [default: aufs]
   -r --root=<path>         image storage root [default: /var/lib/docker]
   -u --repository=<uri>    image repository URI [default: https://dl.flynn.io/tuf]
   -t --tuf-db=<path>       local TUF file [default: /etc/flynn/tuf.db]
+  -b --bin-dir=<dir>       directory to download binaries to [default: /usr/local/bin]
+  -c --config-dir=<dir>    directory to download config files to [default: /etc/flynn]
+  --is-latest              internal flag (skip updating local tuf DB and re-execing latest binary)
+  --is-tempfile            internal flag (binary is a temp file which requires removal)
 
 Update Flynn components`)
 }
@@ -48,10 +56,13 @@ func runUpdate(args *docopt.Args) error {
 	}
 	client := tuf.NewClient(local, remote)
 
-	log.Info("updating TUF data")
-	if _, err := client.Update(); err != nil && !tuf.IsLatestSnapshot(err) {
-		log.Error("error updating TUF client", "err", err)
-		return err
+	if !args.Bool["--is-latest"] {
+		return updateAndExecLatest(client, log)
+	}
+
+	// unlink the current binary if it is a temp file
+	if args.Bool["--is-tempfile"] {
+		os.Remove(os.Args[0])
 	}
 
 	// read the TUF db so we can pass it to hosts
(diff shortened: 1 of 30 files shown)


lmars:
@titanous ready for a review (apologies for the large diff)

host/cli/update.go
```diff
@@ -16,17 +19,21 @@ import (
...
+usage: flynn-host update [options]
 
 Options:
   -d --driver=<name>       image storage driver [default: aufs]
   -r --root=<path>         image storage root [default: /var/lib/docker]
   -u --repository=<uri>    image repository URI [default: https://dl.flynn.io/tuf]
   -t --tuf-db=<path>       local TUF file [default: /etc/flynn/tuf.db]
+  -b --bin-dir=<dir>       directory to download binaries to [default: /usr/local/bin]
+  -c --config-dir=<dir>    directory to download config files to [default: /etc/flynn]
+  --is-latest              internal flag (skip updating local tuf DB and re-execing latest binary)
```
titanous: It looks like if this flag is specified, it will delete the binary that is being run from. This is not obvious, perhaps use a separate flag to cause the removal?

host/cli/update.go
```diff
@@ -148,3 +219,54 @@ func runUpdate(args *docopt.Args) error {
 	log.Info("update complete")
 	return nil
 }
+
+// updateAndExecLatest updates the tuf DB, downloads the latest flynn-host
+// binary to a temp file and execs it.
+//
+// Latest snapshot errors are ignored because, even though we may have the
+// latest snapshot, the cluster may not be fully up to date (a previous update
+// may of failed).
```
titanous: s/of/have/

host/host.go
```diff
@@ -221,6 +250,54 @@ func runDaemon(args *docopt.Args) {
...
+	if fdEnv := os.Getenv("FLYNN_CONTROL_FD"); fdEnv != "" {
+		controlFD, err = strconv.Atoi(fdEnv)
+		if err != nil {
+			shutdown.Fatal(err)
+		}
+		msg := make([]byte, len(ControlMsgResume))
+		if _, err := syscall.Read(controlFD, msg); err != nil {
+			shutdown.Fatalf("error reading from parent control socket: %s", err)
+		}
+		if !reflect.DeepEqual(msg, ControlMsgResume) {
```
titanous: Use `bytes.Equal`

host/http.go
```diff
@@ -21,18 +30,45 @@ import (
 	"github.com/flynn/flynn/pkg/httphelper"
 	"github.com/flynn/flynn/pkg/shutdown"
 	"github.com/flynn/flynn/pkg/sse"
+	"github.com/flynn/flynn/pkg/version"
+)
+
+var (
+	// ControlMsgResume is sent via a control socket from a parent daemon
+	// to it's child to request that the child start serving requests.
```
titanous: s/it's/its/

host/http.go
```diff
@@ -21,18 +30,45 @@ import (
...
+	ControlMsgResume = []byte{0}
+
+	// ControlMsgOK is sent via a control socket from a child daemon to
+	// it's parent to indicate it has received a resume request and is now
```
titanous: s/it's/its/

host/http.go
```diff
@@ -157,6 +187,47 @@ func (h *jobAPI) PullImages(w http.ResponseWriter, r *http.Request, ps httproute
...
+
+	local, err := tuf.FileLocalStore(tufDB)
+	if err != nil {
+		httphelper.Error(w, err)
+		return
+	}
+	opts := &tuf.HTTPRemoteOptions{
+		UserAgent: fmt.Sprintf("flynn-host/%s %s-%s pull", version.String(), runtime.GOOS, runtime.GOARCH),
+	}
+	remote, err := tuf.HTTPRemoteStore(r.URL.Query().Get("repository"), opts)
```
titanous: Parse the Query once and save it in a local variable.

host/http.go
```diff
@@ -248,33 +344,231 @@ func (h *jobAPI) RegisterRoutes(r *httprouter.Router) error {
 	r.DELETE("/host/jobs/:id", h.StopJob)
 	r.PUT("/host/jobs/:id/signal/:signal", h.SignalJob)
 	r.POST("/host/pull-images", h.PullImages)
+	r.POST("/host/pull-binaries-config", h.PullBinariesAndConfig)
```
titanous: Not sold on this endpoint, perhaps `/pull/images` and `/pull/binaries`?

host/http.go
```diff
@@ -248,33 +344,231 @@ func (h *jobAPI) RegisterRoutes(r *httprouter.Router) error {
...
+// An outline of the process:
+//
+// * parent receives a request to exec a new daemon
+//
+// * parent creates a control socket pair (via socketpair(2))
+//
+// * parent starts a child process, passing the API listener as FD 3, and a
+//   control socket as FD 4
+//
+// * parent closes it's API listener FD and state DBs
```
titanous: s/it's/its

host/http.go
```diff
@@ -248,33 +344,231 @@ func (h *jobAPI) RegisterRoutes(r *httprouter.Router) error {
...
+
+// setEnv sets the given environment variables for the command, ensuring they
+// are only set once.
+func setEnv(cmd *exec.Cmd, envs map[string]string) {
+	env := os.Environ()
+	cmd.Env = make([]string, 0, len(env)+len(envs))
+outer:
+	for _, e := range env {
+		for k := range envs {
+			if strings.HasPrefix(e, k) {
```
titanous: `k+"="` otherwise this will catch prefixes.

host/http.go
```diff
@@ -248,33 +344,231 @@ func (h *jobAPI) RegisterRoutes(r *httprouter.Router) error {
...
+// ControlMsgOK response
+func (c *Child) Resume() error {
+	if _, err := syscall.Write(c.sock, ControlMsgResume); err != nil {
+		return err
+	}
+	msg := make([]byte, len(ControlMsgOK))
+	if _, err := syscall.Read(c.sock, msg); err != nil {
+		return err
+	}
+	if !reflect.DeepEqual(msg, ControlMsgOK) {
```
titanous: use `bytes.Equal`

host/types/types.go
```diff
@@ -270,7 +270,14 @@ type DiscoverdConfig struct {
...
+	PID       int              `json:"pid"`
 	URL       string           `json:"url"`
 	Discoverd *DiscoverdConfig `json:"discoverd,omitempty"`
 	Network   *NetworkConfig   `json:"network,omitempty"`
 }
+
+type Command struct {
+	Path string   `json:"path"`
+	Args []string `json:"args"`
+	Pid  int      `json:"pid"`
```
titanous: s/Pid/PID

pkg/cluster/host.go
```diff
@@ -171,3 +171,15 @@ func (c *Host) PullImages(repository, driver, root string, tufDB io.Reader, ch c
 	path := fmt.Sprintf("/host/pull-images?repository=%s&driver=%s&root=%s", repository, driver, root)
 	return c.c.StreamWithHeader("POST", path, header, tufDB, ch)
 }
+
+// PullBinariesAndConfig pulls binaries and config from a TUF repository using the local TUF file in tufDB
+func (c *Host) PullBinariesAndConfig(repository, binDir, configDir string, tufDB io.Reader) (map[string]string, error) {
+	path := fmt.Sprintf("/host/pull-binaries-config?repository=%s&bin-dir=%s&config-dir=%s", repository, binDir, configDir)
```
titanous: I'd prefer using `url.Query` to build the query part of the URL.

lmars:
@titanous comments addressed (I've added a [discoverd shutdown fix](https://github.com/flynn/flynn/commit/ce370765c434d1c717ed7d7d4d3710156304a9c6) too)

titanous:
Has this been tested to determine what happens to log streams (to disk and logaggregator) during/after the update? (no duplicate or dropped lines)

lmars:
@titanous comments addressed, I've added a lot of small but related fixes, let me know if you would find them easier to review as separate PRs, but most have descriptions in the commit message.

host/containerinit/init.go
```diff
@@ -726,12 +732,15 @@ func containerInitApp(c *Config, logFile *os.File) error {
 // This code is run INSIDE the container and is responsible for setting
 // up the environment before running the actual process
 func Main() {
-	logRd, logWr, err := os.Pipe()
+	pair, err := syscall.Socketpair(syscall.AF_UNIX, syscall.SOCK_STREAM, 0)
 	if err != nil {
 		os.Exit(70)
 	}
+	logR := os.NewFile(uintptr(pair[0]), "log")
+	logW := os.NewFile(uintptr(pair[1]), "log")
```
titanous: This syscall and file init should be wrapped up in a helper function.

host/http.go
```diff
@@ -244,33 +415,293 @@ func (h *jobAPI) RegisterRoutes(r *httprouter.Router) error {
...
+		if err != nil {
+			log.Error("error recreating HTTP listener", "err", err)
+			return err
+		}
+		h.listener = l
+
+		log.Info("reopening state databases")
+		if err := h.OpenDBs(); err != nil {
+			log.Error("reerror opening state databases", "err", err)
+			return err
```
titanous: If this or any of the above fails, we should probably exit so that the process supervisor can attempt a restart, right?

host/http.go
```diff
@@ -244,33 +415,293 @@ func (h *jobAPI) RegisterRoutes(r *httprouter.Router) error {
...
+//   socket, followed by any partial log buffers.
+//
+// * child receives resume request, opens state DBs, seeds the log followers
+//   with the partial buffers and starts serving API requests
+//
+// * child signals parent it is now serving requests by sending "ok" message to
+//   control socket
+//
+// * parent sends response to client and shuts down seconds later
+//
```
titanous: All of the update stuff should be moved to a separate file for clarity.

host/state.go
```diff
@@ -194,7 +209,59 @@ func (s *State) initializePersistence() {
...
+//
+// ErrDBClosed is returned if the DB is already closed so API requests will
+// fail before any actions are performed.
+func (s *State) Acquire() error {
+	s.dbCond.L.Lock()
+	defer s.dbCond.L.Unlock()
+	if s.stateDB == nil {
+		return ErrDBClosed
+	}
+	s.dbUsers += 1
```
titanous: `s.dbUsers++`

host/state.go
```diff
@@ -194,7 +209,59 @@ func (s *State) initializePersistence() {
...
+	return nil
+}
+
+// Release releases the state by decrementing s.dbUsers, broadcasting the
+// condition variable if no users are left to wake CloseDB.
+func (s *State) Release() {
+	s.dbCond.L.Lock()
+	defer s.dbCond.L.Unlock()
+	s.dbUsers -= 1
```
titanous: `s.dbUsers--`

host/types/types.go
```diff
@@ -275,7 +275,18 @@ type DiscoverdConfig struct {
...
+
+type Command struct {
+	Path string   `json:"path"`
+	Args []string `json:"args"`
+	PID  int      `json:"pid"`
+}
+
+type LogBuffers map[string]LogBuffer
+
+type LogBuffer map[string][]byte
```
titanous: This will encode the data to base64 in the JOSN, perhaps it should be a `string`?

test/test_host.go
```diff
@@ -290,3 +297,95 @@ func (s *HostSuite) TestResourceLimits(t *c.C) {
...
+		syscall.Wait4(cmd.Process.Pid, nil, 0, nil)
+		close(done)
+	}()
+	select {
+	case <-done:
+	case <-time.After(15 * time.Second):
+		t.Fatal("timed out waiting for flynn-host daemon to exit")
+	}
+
+	// client.GetStatus intermittently returns io.EOF right after the update. We
```
titanous: This EOF might come from the HTTP transport if it doesn't detect that the connection is closed before the request is made.

lmars: There should be no closing of the connection though right?

titanous: Copying the listener over just copies the listener, all established connections will be closed when the old process exits.

lmars: By this point, the old daemon should not be accepting requests, so this request should be accepted by the new daemon. Are you claiming that the old process may still be accepting requests after the update has finished?

titanous: The HTTP client transport will re-use any available TCP connections to send the HTTP request, it seems likely that there is one available that was established with the old process, when the client attempts to use it, it may encounter an EOF if it hasn't been cleaned up yet.

titanous:
What's left before we can merge this?

lmars:
@titanous the only thing left is dealing with the possibility of getting EOF from the host API due to established connections not being closed (i.e. [this discussion](https://github.com/flynn/flynn/pull/1672#discussion-diff-38484679)), but since the scheduler is now more robust (it will retry starting jobs on a host if it gets this error), I don't think this should prevent the PR from being merged.

titanous:
LGTM!

Issue #1669, opened 2015-07-19
Issues installing on ubuntu/openstack


adrianmoya:
I'm following the install instructions for ubuntu, and I'm having an issues when trying to bootstrap Layer 1:

I'm getting 

`{"s":3,"t":1437329964384,"m":"t=2015-07-19T18:19:24+0000 lvl=info msg=resuming app=containerinit fn=containerInitApp\nt=2015-07-19T18:19:24+0000 lvl=info msg=\"setting up the container\" app=containerinit fn=containerInitApp\nt=2015-07-19T18:19:24+0000 lvl=eror msg=\"error setting up the container\" app=containerinit fn=containerInitApp err=\"invalid argument\"\nt=2015-07-19T18:19:24+0000 lvl=info msg=\"changing state\" app=containerinit fn=changeState state=failed err=\"invalid argument\" exitStatus=-1\nt=2015-07-19T18:19:24+0000 lvl=info msg=\"sending state change\" app=containerinit fn=StreamState state=failed err=\"invalid argument\" exitStatus=-1\n"}`

when running 

`sudo CLUSTER_DOMAIN=mydomain.com flynn-host bootstrap --min-hosts 3 --discovery https://discovery.flynn.io/clusters/xxxx`

Any ideas on how to proceed?

Debug info: https://gist.github.com/anonymous/3d22f2b1f16695a2b219

titanous:
Do you have the output from the bootstrap command handy?

titanous:
Also, it looks like you are running `v20150706.0` which is an older version, the current version is `v20150718.0`.

adrianmoya:
Output is
`sudo: unable to resolve host shcloud-demo-8d5395d4-f549-4048-9341-30eea1a818f0
18:19:22.550962 check online-hosts
18:19:23.065085 require-env require-env
18:19:23.079672 run-app etcd
18:19:24.442204 run-app etcd error: bootstrap: host error while launching job: "invalid argument"
`

Let me start fresh with the latest version.

adrianmoya:
No luck with the new installation. This is the procedure I'm using:

1. Created an instance with ubuntu 14.04.
2. Upgraded packages and installed flynn using the online script.
3. Created an snapshot of this instance.
4. Launched 2 instances of the snapshot.
5. Init discovery on the 3 hosts (first with init-discovery, other two with the discovery token from 1)
6. Started flynn-host on the 3 nodes.
7. Tried to bootstrap Layer 1:

```
[...]

PR #1673, opened 2015-07-21
host: Remove host ID from container hostnames

titanous:
Including the host ID in the container hostname can result in the name being longer than 64 characters, which is the maximum length that `sethostname(2)` will take.

Refs #1669


lmars:
LGTM

PR #1680, opened 2015-07-28
controller/scheduler: Handle no available hosts when starting jobs

lmars:
The v2 scheduler will handle this better by retrying the host lookup: https://github.com/flynn/flynn/issues/1484#issuecomment-125691878


titanous:
LGTM

PR #1685, opened 2015-08-03
util/packer: Explicitly install linux-image-generic-lts-vivid

titanous:
This should resolve this complaint from the AWS builder:

    linux-generic-lts-vivid : Depends: linux-image-generic-lts-vivid (= 3.19.0.25.12) but it is not going to be installed


lmars:
LGTM

Issue #1684, opened 2015-07-31
Update expiring snapshot.json in timestamp cron


dominion525:
Hi, when I try install ,  I get this error:

```
===> 16:55:10.511 running "/usr/local/bin/flynn-host download --repository https://dl.flynn.io/tuf --tuf-db /etc/flynn/tuf.db --config-dir /etc/flynn --bin-dir /usr/local/bin"
16:55:12.729837 host.go:122: tuf: failed to decode snapshot.json: expired at 2015-07-31 06:04:24 +0000 UTC
===> 16:55:12.731 ERROR: failed to run "/usr/local/bin/flynn-host download --repository https://dl.flynn.io/tuf --tuf-db /etc/flynn/tuf.db --config-dir /etc/flynn --bin-dir /usr/local/bin", exit status 1
```

it's same case this issue :  https://github.com/flynn/flynn/issues/1564

I want to ask you if you need work on the server side.

titanous:
Sorry about that! This should be fixed now. I'll leave this open until we fix the script to update the snapshot.json correctly.

PR #1690, opened 2015-08-04
script: Add release-timestamp

lmars:
Includes a fix for #1684 by running `tuf snapshot` if the snapshot has expired.

### DIFF
--- a/script/release-timestamp
+++ b/script/release-timestamp
@@ -1,0 +1,76 @@
+#!/bin/bash
+#
+# A script to update the timestamp of a Flynn TUF repository
+
+set -eo pipefail
+
+ROOT="$(cd "$(dirname "$0")/.." && pwd)"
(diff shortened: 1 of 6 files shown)


script/release-timestamp
```diff
@@ -0,0 +1,75 @@
...
+    exit 1
+  fi
+
+  info "downloading existing TUF metadata"
+  download_tuf_metadata "${tuf_dir}" "${bucket}"
+
+  cd "${tuf_dir}"
+
+  info "checking snapshot expires"
+  if metadata_expired "repository/snapshot.json"; then
```
titanous: We should check if it expires in the next N, where N is the next two or three check intervals. If it's expired, we're too late.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #1694, opened 2015-08-04
Scheduler v2: Implemented host, job event, and formation streaming.

jzila:
Also implemented goroutine for handling PutJob queue.

### DIFF
--- a/controller/jobs.go
+++ b/controller/jobs.go
@@ -16,6 +16,7 @@ import (
 	"github.com/flynn/flynn/controller/schema"
 	ct "github.com/flynn/flynn/controller/types"
 	"github.com/flynn/flynn/controller/utils"
+	"github.com/flynn/flynn/discoverd/client"
 	"github.com/flynn/flynn/host/resource"
 	"github.com/flynn/flynn/host/types"
 	"github.com/flynn/flynn/pkg/cluster"
@@ -145,9 +146,9 @@ func (c clusterClientWrapper) Hosts() ([]utils.HostClient, error) {
 	return res, nil
 }
 
-func (c clusterClientWrapper) StreamHosts(ch chan utils.HostClient) (stream.Stream, error) {
-	hostChan := make(chan *cluster.Host)
-	stream, err := c.Client.StreamHosts(hostChan)
+func (c clusterClientWrapper) StreamHostEvents(ch chan *discoverd.Event) (stream.Stream, error) {
+	hostChan := make(chan *discoverd.Event)
+	stream, err := c.Client.StreamHostEvents(hostChan)
 	if err != nil {
 		return nil, err
 	}
@@ -168,7 +169,7 @@ func (c clusterClientWrapper) StreamHosts(ch chan utils.HostClient) (stream.Stre
 }
 
 type clusterStream struct {
-	ch           chan utils.HostClient
+	ch           chan *discoverd.Event
 	parentStream stream.Stream
 }
 
--- a/controller/jobs.go
+++ b/controller/jobs.go
@@ -148,23 +148,23 @@ func (c clusterClientWrapper) Hosts() ([]utils.HostClient, error) {
 func (c clusterClientWrapper) StreamHosts(ch chan utils.HostClient) (stream.Stream, error) {
 	hostChan := make(chan *cluster.Host)
 	stream, err := c.Client.StreamHosts(hostChan)
-	if err == nil {
-		go func() {
-			for {
-				select {
-				case h, ok := <-hostChan:
-					if !ok {
-						return
-					}
-					ch <- h
-				}
-			}
-		}()
+	if err != nil {
+		return nil, err
 	}
+	go func() {
+		for {
+			h, ok := <-hostChan
+			if !ok {
+				close(ch)
+				return
+			}
+			ch <- h
+		}
+	}()
 	return &clusterStream{
 		ch:           ch,
 		parentStream: stream,
-	}, err
+	}, nil
 }
 
 type clusterStream struct {
@@ -173,7 +173,6 @@ type clusterStream struct {
 }
 
 func (cs *clusterStream) Close() error {
-	close(cs.ch)
 	return cs.parentStream.Close()
 }
 
--- a/controller/jobs.go
+++ b/controller/jobs.go
@@ -245,6 +245,9 @@ func (c *controllerAPI) KillJob(ctx context.Context, w http.ResponseWriter, req 
 	}
 
 	if err = client.StopJob(jobID); err != nil {
+		if _, ok := err.(ct.NotFoundError); ok {
+			err = ErrNotFound
+		}
 		respondWithError(w, err)
 		return
 	}
--- a/controller/jobs.go
+++ b/controller/jobs.go
@@ -23,6 +23,7 @@ import (
 	"github.com/flynn/flynn/pkg/httphelper"
 	"github.com/flynn/flynn/pkg/postgres"
 	"github.com/flynn/flynn/pkg/random"
+	"github.com/flynn/flynn/pkg/stream"
 )
 
 /* SSE Logger */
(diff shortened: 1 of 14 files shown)


jzila:
@lmars ready for review

controller/jobs.go
```diff
@@ -153,6 +154,42 @@ func (c clusterClientWrapper) Hosts() ([]utils.HostClient, error) {
...
+				select {
+				case h, ok := <-hostChan:
+					if !ok {
+						return
+					}
+					ch <- h
+				}
+			}
+		}()
+	}
```
lmars: Reduce the indentation:

```go
if err != nil {
	return nil, err
}
go func() {
	// ...
}()
```

controller/jobs.go
```diff
@@ -153,6 +154,42 @@ func (c clusterClientWrapper) Hosts() ([]utils.HostClient, error) {
...
+	if err == nil {
+		go func() {
+			for {
+				select {
+				case h, ok := <-hostChan:
+					if !ok {
+						return
+					}
+					ch <- h
+				}
```
lmars: No need for the `select`

controller/jobs.go
```diff
@@ -153,6 +154,42 @@ func (c clusterClientWrapper) Hosts() ([]utils.HostClient, error) {
...
+	}, err
+}
+
+type clusterStream struct {
+	ch           chan utils.HostClient
+	parentStream stream.Stream
+}
+
+func (cs *clusterStream) Close() error {
+	close(cs.ch)
```
lmars: Defer the close of this channel in the goroutine which is writing to it so that we don't cause a panic writing to a closed channel.

controller/scheduler-v2/scheduler.go
```diff
@@ -69,62 +80,91 @@ func (s *Scheduler) Run() error {
...
+			continue
+		case he := <-s.hostEvents:
+			s.handleHostEvent(he)
+			continue
 		case ef := <-s.formationChange:
-			s.formationChangeHandler(ef)
-		case <-s.formationSync:
+			s.FormationChange(ef)
+			continue
+		case <-time.After(50 * time.Millisecond):
```
lmars: Why the 50ms wait?

controller/scheduler-v2/scheduler.go
```diff
@@ -69,62 +80,91 @@ func (s *Scheduler) Run() error {
...
+			s.followHost(h)
+			continue
+		case he := <-s.hostEvents:
+			s.handleHostEvent(he)
+			continue
 		case ef := <-s.formationChange:
-			s.formationChangeHandler(ef)
-		case <-s.formationSync:
+			s.FormationChange(ef)
+			continue
```
lmars: We need to handle these channels closing.

We already do stream reconnects (with a [10s timeout](https://github.com/flynn/flynn/blob/master/pkg/httpclient/stream.go#L83-L86)), but I think we should also have goroutines in the scheduler reconnecting the streams in case things go down for a long time.

controller/scheduler-v2/scheduler.go
```diff
@@ -69,62 +80,91 @@ func (s *Scheduler) Run() error {
...
 
-func (s *Scheduler) SyncJobs() (err error) {
-	fLog := s.log.New("fn", "Sync")
+func (s *Scheduler) SyncHosts() (err error) {
+	log := s.log.New("fn", "SyncHosts")
 
-	defer func() {
-		s.sendEvent(NewEvent(EventTypeClusterSync, err, nil))
-	}()
+	defer s.sendEvent(NewEvent(EventTypeHostSync, err, nil))
```
lmars: This needs to be enclosed in an anonymous function so that `err` is evaluated when the defer is called (otherwise it will always be `nil`).

controller/scheduler-v2/scheduler.go
```diff
@@ -209,9 +212,19 @@ func (s *Scheduler) SyncFormations() (err error) {
 }
 
 func (s *Scheduler) RectifyJobs() (err error) {
-	log := s.log.New("fn", "RectifyJobs")
+	s.leaderMtx.RLock()
+	defer s.leaderMtx.RUnlock()
```
lmars: We should handle the leader demotion using a channel in the main loop and avoid this mutex.

controller/scheduler-v2/scheduler.go
```diff
@@ -290,6 +296,110 @@ func (s *Scheduler) HandleJobRequest(req *JobRequest) (err error) {
 	return
 }
 
+func (s *Scheduler) RunPutJobs() {
+	strategy := attempt.Strategy{Delay: 100 * time.Millisecond, Total: time.Minute}
+	for {
+		select {
+		case job, ok := <-s.putJobs:
+			if !ok {
+				return
```
lmars: Does this channel ever close?

controller/scheduler-v2/scheduler.go
```diff
@@ -290,6 +296,110 @@ func (s *Scheduler) HandleJobRequest(req *JobRequest) (err error) {
 	return
 }
 
+func (s *Scheduler) RunPutJobs() {
+	strategy := attempt.Strategy{Delay: 100 * time.Millisecond, Total: time.Minute}
+	for {
+		select {
```
lmars: Drop the select

controller/scheduler-v2/scheduler.go
```diff
@@ -290,6 +296,110 @@ func (s *Scheduler) HandleJobRequest(req *JobRequest) (err error) {
...
+	}
+}
+
+func (s *Scheduler) followHost(h utils.HostClient) {
+	log := s.log.New("fn", "followHost")
+	_, ok := s.hosts[h.ID()]
+	if !ok {
+		log.Info("Following host", "host.id", h.ID())
+		s.hosts[h.ID()] = h
+		h.StreamEvents("all", s.hostEvents)
```
lmars: We should handle this stream closing and reconnect.

controller/scheduler-v2/scheduler.go
```diff
@@ -488,44 +576,47 @@ func (s *Scheduler) Unsubscribe(events chan Event) {
...
+func tickChannel(ch chan interface{}, d time.Duration) {
+	ticker := time.Tick(d)
+
+	go func() {
+		for {
+			select {
+			case <-ticker:
+				triggerChan(ch)
+			}
+		}
```
lmars:
```go
for range ticker {
	triggerChan(ch)
}
```

controller/scheduler-v2/scheduler.go
jzila: so we're not spinning continuously. Since I've split out the `select` into two, either the `for` loop spins hot, or we put in a wait somewhere.

controller/scheduler-v2/scheduler.go
jzila: Understood.

controller/scheduler-v2/scheduler.go
jzila: Haha yeah...not sure what I was thinking here...

controller/scheduler-v2/scheduler.go
jzila: Good catch; it should be closed in `Stop()`

controller/scheduler-v2/scheduler.go
lmars: Ok I get that, but the issue with where this is at the moment is if a formation change event comes in and 10 jobs need to be started, we are going to sleep 50ms between starting each one (and maybe even longer as we receive host events for those jobs), so if this is needed it should be at the end of the loop when we know there is no work to do.

Also, given I asked the question, I would put a comment next to it (arbitrary sleeps always look suspicious).

jzila:
@lmars I'm ready for final review on this PR

lmars:
@jzila I am going to merge this and open a PR with all changes so I can review as a whole.

PR #1699, opened 2015-08-07
Add health status endpoints to all services

titanous:
This PR adds health status endpoints at `/.well-known/status` to all services, including a few that didn't have HTTP servers previously. The reasoning for this is that we will want to add more diagnostic endpoints in the future including metrics, debugging, profiling, etc.

The only service that does not have an endpoint is flannel, which will be added in a separate PR to the flannel repo.

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -1,0 +1,91 @@
+package status
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+/*
+
+	GET /.well-known/status
+	Accept: application/json
+
+	{
+	  "data": {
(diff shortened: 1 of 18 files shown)


pkg/status/status.go
```diff
@@ -0,0 +1,91 @@
+package status
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+/*
+
+	GET /.well-known/health
```
lmars: This should be `/.well-known/status`

lmars:
LGTM pending minor comment

Issue #1719, opened 2015-08-15
Build Vagrant boxes for VMware and QEMU


josephglanville:
The current VM release pipeline only builds Virtualbox and AWS Vagrant boxes.
I propose this is expanded to cover additionally VMware and QEMU.
This can be accomplished by modifying the way the build process works.

Specifically the changes need are:
- Install VMware Player on the build server
- Install QEMU on build server
- Change build process to download the Ubuntu Cloud OVA
- Use ovftool to create a VMX file for use with a new Packer builder that will go from VMX -> box
- Convert the VMDK inside the OVA to QCOW2 and use it with a new QEMU builder to produce the libvirt compatible box.

The proposed build process looks like this:

```
+-------+      +-------+      +----------+
|  OVA  +----->+  VMX  +----->+  VMware  |
+---+---+      +-------+      +----------+
    |                                     
    |          +-------+      +----------+
    +--------->+ QCOW2 +----->+   QEMU   |
    |          +-------+      +----------+
    |                                     
    |                         +----------+
    +------------------------>+VirtualBox|
                              +----------+
```

josephglanville:
Further investigation has revealed that using the OVA as a source isn't really a viable option due to it's dependence on cloud-init.

Instead a slightly more complex but more hygienic approach is warranted.
That is building base boxes from ISO for each provider and then doing the build on-top of those.

This will be done as a 2 stage process so that the base image will only need to be build once per provider, per version of the source ISO.

For consistency I think it's wise to do this for VirtualBox as well despite it being possible to skip due to the existing conversion from box -> OVA -> box.

PR #1718, opened 2015-08-15
Vagrant + Packer improvements

josephglanville:
Few changes to get up and running using Packer + VMWare Fusion.

* Ubuntu 14.04.3 has been released, as such Ubuntu 14.04.2 is no longer available on release mirrors.
* Ignore the Packer build directories
* Configure the VMWare Fusion provider in Vagrantfile to match Virtualbox

PR #1720, opened 2015-08-15
VM release improvements

josephglanville:
Builds on #1718 to address #1719 

Still somewhat WIP and requires more testing and probably work to make it conform to bash styleguide.
But the general idea is there.

Still to go is QEMU support and verification that the publishing side is not bug riddled.

### DIFF
--- a/script/release-vm-images
+++ b/script/release-vm-images
@@ -22,9 +22,13 @@
 #   sudo dpkg -i /tmp/s3cmd.deb
 #   rm /tmp/s3cmd.deb
 #
+# - Install aws CLI tool
+#   sudo apt-get install awscli
+#
 # - Set AWS keys
 #   export AWS_ACCESS_KEY_ID=XXXXXXXXXXXXXXXXXXXX
 #   export AWS_SECRET_ACCESS_KEY=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+#   export AWS_DEFAULT_REGION=us-east-1
 
 set -eo pipefail
 
@@ -35,7 +39,6 @@ source "${ROOT}/script/lib/util.sh"
 
 UBUNTU_BOX_URL="https://cloud-images.ubuntu.com/vagrant/trusty/20150610/trusty-server-cloudimg-amd64-vagrant-disk1.box"
 UBUNTU_BOX_SHA="5174904851577f6423444fe477b3fae01944f8592fabd48f0ba7291d7e645d99"
-UBUNTU_USEAST_AMI="ami-c135f3aa" # us-east-1 | trusty | 14.04 LTS | amd64 | hvm:ebs-ssd | 20150629 | hvm
 
 usage() {
   cat <<USAGE >&2
@@ -86,6 +89,9 @@ main() {
 
   check_aws_keys
 
+  # Assume us-east-1 if not set.
+  AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-us-east-1}
+
   local version=$1
   bucket=${bucket:-"flynn"}
   dir="${dir:-$(mktemp -d)}"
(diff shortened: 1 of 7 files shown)


josephglanville:
@lmars if you could take a look at this that would be great.

I am probably not going to get time to look at libvirt/qemu based stuff until the weekend so maybe do that as separate PR.

josephglanville:
While I am here there is another problem with this currently. It stores the stage1 artifacts in the same temporary directory as the final build products, meaning they are erased if the upload is successful.

How does adding another argument that takes a path to be used as a cache directory for the stage1 VM images sound?

script/release-vm-images
```diff
@@ -44,19 +48,20 @@ usage: $0 [options] VERSION
 OPTIONS:
   -h            Show this message
   -k            Keep release directory
-  -b BUCKET     The S3 bucket to upload vagrant images to [default: `flynn`]
-  -d DOMAIN     The CloudFront domain [default: `dl.flynn.io`]
+  -b BUCKET     The S3 bucket to upload vagrant images to [default: 'flynn']
+  -d DOMAIN     The CloudFront domain [default: 'dl.flynn.io']
   -r DIR        Resume the release using DIR
+  -c CACHE      Directory to cache base VM images [default: '/tmp']
```
lmars: It doesn't look like you are actually setting the default value of `cache`, I would expect to see the following somewhere

```bash
cache="${cache:-"/tmp"}"
```

script/release-vm-images
```diff
@@ -86,46 +97,89 @@ main() {
 
   check_aws_keys
 
+  # Assume us-east-1 if not set.
+  AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-us-east-1}
```
lmars: This should be a flag like the other optional variables.

script/release-vm-images
```diff
@@ -86,46 +97,89 @@ main() {
...
-  if [[ -z "${box}" ]] || [[ -z "${amis}" ]]; then
-    local ubuntu_box="/tmp/ubuntu-${UBUNTU_BOX_SHA}.box"
-    if [[ ! -f "${ubuntu_box}" ]]; then
-      info "fetching base Ubuntu box"
-      curl "${UBUNTU_BOX_URL}" > "${ubuntu_box}"
+  info "checking for stage1 targets"
+  local stage1_imgs=(${dir}/base_*_*)
+  local stage1_targets=()
+  for builder in ${builders[@]}; do
+    if [[ ! "${stage1_imgs[@]}" =~ "base_${builder}_${UBUNTU_ISO_SHA}" ]]; then
```
lmars: I would drop the `stage1_imgs` variable and use `test -f`:


```bash
if [[ ! -f "${dir}/base_${builder}_${UBUNTU_ISO_SHA}" ]]; then
  stage1_targets+=("${builder}-iso")
fi
```

script/release-vm-images
```diff
@@ -86,46 +97,89 @@ main() {
...
+      stage1_targets+=("${builder}-iso")
     fi
+  done
 
-    info "validating base Ubuntu box"
-    if [[ "$(sha256 "${ubuntu_box}")" != "${UBUNTU_BOX_SHA}" ]]; then
-      rm -f "${ubuntu_box}"
-      fail "base Ubuntu box has incorrect checksum"
+  if [ ${#stage1_targets[@]} -eq 0 ]; then
```
lmars: Use `[[`

script/release-vm-images
```diff
@@ -86,46 +97,89 @@ main() {
...
+    popd >/dev/null
+  fi
+
+  local amis="$(cat "${dir}/amis.txt" 2>/dev/null)"
+
+  info "checking for stage2 targets"
+  local boxes=(${dir}/flynn-base_*_*.box)
+  local stage2_targets=()
+  for builder in ${builders[@]}; do
+    if [[ ! "${boxes[@]}" =~ "flynn-base_${version}_${builder}.box" ]]; then
```
lmars: Again I would just use `test -f`

script/release-vm-images
josephglanville: Good catch.

Initialised it to "/tmp" rather than creating the unset local.

script/release-vm-images
josephglanville: Done.

script/release-vm-images
josephglanville: Yeah, that is much nicer.

script/release-vm-images
josephglanville: Done.

script/release-vm-images
josephglanville: Done.

josephglanville:
@lmars I have addressed comments and added another trivial change to configure the amount of memory/vcpus the VMs use. My build server isn't as chunky so I wanted to scale them down a little heh.

titanous:
Closing until this is revisited.

PR #1729, opened 2015-08-20
Vagrantfile: Adding DeleteOnTermination to root volume for AWS

temujin9:
By default, the root volumes stick around after the Vagrant goes away. This is not a good default if builder instances are cycled regularly.


titanous:
The commit needs a DCO sign-off, but other than that, LGTM.

temujin9:
Derp. I always forget the DCO . . . fixed.

PR #1735, opened 2015-08-21
controller: Fix app lookup when streaming events

lmars:
The Events handler is not wrapped in appLookup, so there will never be an app in the context.

The appID is also passed as a query string rather than a param, so we can't use appLookup.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -331,13 +331,6 @@ func (c *controllerAPI) getApp(ctx context.Context) *ct.App {
 	return ctx.Value("app").(*ct.App)
 }
 
-func (c *controllerAPI) maybeGetApp(ctx context.Context) *ct.App {
-	if app, ok := ctx.Value("app").(*ct.App); ok {
-		return app
-	}
-	return nil
-}
-
 func (c *controllerAPI) getRelease(ctx context.Context) (*ct.Release, error) {
 	params, _ := ctxhelper.ParamsFromContext(ctx)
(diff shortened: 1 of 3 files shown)


jvatic:
The [TestRouteEvents](https://github.com/flynn/flynn/blob/0f5b7ad81e91c2662fbba018168a91efb3ca256e/test/test_controller.go#L382-L436) test is failing due to the stream being opened before the app is created. As previous to this fix that worked the same as not passing in the app id, simply removing it from the test seems like a reasonable option.

jvatic:
LGTM pending comment

titanous:
Is there a test that can be added to verify the fix?

lmars:
@jvatic 

> TestRouteEvents test is failing due to the stream being opened before the app is created

Maybe we should just not do the app lookup? I think it's reasonable to want to stream events for an app before you create it to avoid races (although that can be avoided by passing `past=true`), wdyt?

@titanous 

> Is there a test that can be added to verify the fix?

I will write a test (create two apps, stream events for one, check I don't get events for the other).

jvatic:
@lmars 

> Maybe we should just not do the app lookup? I think it's reasonable to want to stream events for an app before you create it to avoid races (although that can be avoided by passing `past=true`), wdyt?

If we don't do the app lookup we'll need to make sure the app name can be used in place of the id. I think using `past=true` is the simpler option.

lmars:
@jvatic @titanous comments addressed

titanous:
LGTM

jvatic:
LGTM

PR #1736, opened 2015-08-21
demo: Also set memory and vcpus for vmware_fusion provider

josephglanville:
Small quality of life improvement for VMware Fusion users.


lmars:
Thanks!

PR #1737, opened 2015-08-21
controller: Wait indefinitely for router to come up before streaming events

titanous:
If we don’t do this, a bootstrap may be slower than the timeout for the subsequent stream request, causing the controller to crash and the bootstrap to fail.

Reported by @temujin9 in IRC.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -125,6 +125,21 @@ func wrapDBExec(dbExec func(string, ...interface{}) error) func(string, ...inter
 }
 
 func streamRouterEvents(rc routerc.Client, db *postgres.DB, doneCh chan struct{}) error {
+	// wait for router to come up
+	{
+		events := make(chan *discoverd.Event)
+		stream, err := discoverd.NewService("router-api").Watch(events)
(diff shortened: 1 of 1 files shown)


jvatic:
LGTM

controller/controller.go
```diff
@@ -125,6 +125,21 @@ func wrapDBExec(dbExec func(string, ...interface{}) error) func(string, ...inter
...
 
 func streamRouterEvents(rc routerc.Client, db *postgres.DB, doneCh chan struct{}) error {
+	// wait for router to come up
+	{
+		events := make(chan *discoverd.Event)
+		stream, err := discoverd.NewService("router-api").Watch(events)
+		if err != nil {
+			return err
+		}
+		for e := range events {
```
lmars: I know this is merged, but we should probably return an error if `events` gets closed (rather than panicking as it will do now).

titanous: Where will it panic?

lmars: I have no idea what I was talking about :trollface:

PR #1739, opened 2015-08-22
appliance/postgresql: Add PGRouting extension to PostgreSQL appliance

josephglanville:
This enables additional GIS workloads including importing and working with the Open Street Map dataset.

### DIFF
--- a/appliance/postgresql/Dockerfile
+++ b/appliance/postgresql/Dockerfile
@@ -17,7 +17,8 @@ RUN apt-get update &&\
       postgresql-contrib-9.4 \
       postgresql-9.4-pgextwlist \
       postgresql-9.4-plv8 \
-      postgresql-9.4-postgis &&\
+      postgresql-9.4-postgis \
+      postgresql-9.4-pgrouting &&\
     apt-get clean &&\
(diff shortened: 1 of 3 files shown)


josephglanville:
On the topic of extensions, as they are so hard to customise would it make sense to just install all the extensions available in the pgdg repo?

titanous:
How many other extensions are there?

josephglanville:
These are the ones built and packaged by pgdg:
debversion
ip4r
orafce
pgespresso
pgextwlist
pgfincore
pgmp
pgpool2
pgq3
pgrouting
plproxy
plr
plv8
postgis
prefix
preprepare
python-multicorn
python3-multicorn
reorg
repmgr
slony

Most of them are related to replication and backup tools so maybe not super useful but installing them is probably no harm. They may come in handy if you want to setup trigger based replication to/from an external database though. 

Then there is stuff like ip4r, which are IP address range types, and python-multicorn which is an extension for building foreign data wrappers.

I am not too fussed either way, just figured it would be useful because inevitably someone will want one of these things.

titanous:
I'm going to merge this for now, let's see if more requests come up before moving forward with others.

PR #1741, opened 2015-08-22
Add syslog structured data encoding/parsing

titanous:
This is a prerequisite for work I'm doing on logaggregator replication. I also put in a fix for a bug that I found while fuzzing the main parser. Both parsers have been extensively fuzzed with go-fuzz.


lmars:
LGTM (brownie points for using go-fuzz :bowtie:)

PR #1743, opened 2015-08-23
host: Fix attaching to exited jobs with no logs

lmars:
Previously, attaching to a job which had exited and had `DisableLog` set would immediately close the request rather than sending an Attach protocol message.

This change means `AttachSuccess` is sent, followed by `io.EOF` to indicate there is no output.


lmars:
FYI this fixes CI always doing log fallback due to an error trying to fetch slugbuilder logs.

titanous:
LGTM

PR #1747, opened 2015-08-25
cli: List deployments

### LINKED ISSUE #1176: Deploys should be visible via the CLI
I should be able to see current and past deploys via the CLI, and be able to stop stuck deploys, e.g.

```
$ flynn deploy list
ID        Status   Started      Finished
312e09ee  running  5 mins ago
b1aad0f7  success  11 mins ago  10 mins ago
e95004dd  failed   22 mins ago  21 mins ago

$ flynn deploy stop 312e09ee
Deploy 312e09ee stopped
```

josephglanville:
List deployments along with their status, start and finish times.

Closes #1176

### DIFF
--- a/cli/deployment.go
+++ b/cli/deployment.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
+	"github.com/flynn/flynn/controller/client"
+)
+
+func init() {
+	register("deployment", runDeployments, `
+usage: flynn deployment
+
+Manage app deployments
+
+Commands:
+    With no arguments, shows a list of deployments
+
+Examples:
+
+	$ flynn deployment
+	ID                                STATUS    CREATED             FINISHED
+	37a63fb05fe946f18f11f741aed74d60  running   4 seconds ago
+	51cbf2bba1204e94b1d847ae0122c647  complete  16 seconds ago      14 seconds ago
+	12875d153f5c4c6cb64e263c4b422e8c  failed    About a minute ago  About a minute ago
+	21d4a8174a4240a0b1dcb6303f40cad5  complete  4 minutes ago       4 minutes ago
+`)
+}
+
+func runDeployments(args *docopt.Args, client *controller.Client) error {
+	deployments, err := client.DeploymentList(mustApp())
+	if err != nil {
+		return err
+	}
+
+	w := tabWriter()
+	defer w.Flush()
+
+	listRec(w, "ID", "STATUS", "CREATED", "FINISHED")
+	for _, d := range deployments {
+		listRec(w, d.ID, d.Status, humanTime(d.CreatedAt), humanTime(d.FinishedAt))
+	}
+	return nil
+}
--- a/cli/main.go
+++ b/cli/main.go
@@ -39,28 +39,29 @@ Options:
 	-h, --help
 
 Commands:
-	help      show usage for a specific command
-	install   install flynn
-	cluster   manage clusters
-	create    create an app
-	delete    delete an app
-	apps      list apps
-	ps        list jobs
-	kill      kill a job
-	log       get app log
-	scale     change formation
-	run       run a job
-	env       manage env variables
-	meta      manage app metadata
-	route     manage routes
-	pg        manage postgres database
-	provider  manage resource providers
-	remote    manage git remotes
-	resource  provision a new resource
-	release   add a docker image release
-	export    export app data
-	import    create app from exported data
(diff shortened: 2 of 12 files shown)


cli/deployment.go
```diff
@@ -0,0 +1,42 @@
...
+usage: flynn deploy
+
+Manage app deployments
+
+Commands:
+    With no arguments, shows a list of deployments
+
+Examples:
+
+	$ flynn deploy
```
titanous: I think we should probably call this `flynn deployment` to avoid the confusion of an action vs a namespace.

cli/main.go
```diff
@@ -277,6 +280,14 @@ func tabWriter() *tabwriter.Writer {
 	return tabwriter.NewWriter(os.Stdout, 1, 2, 2, ' ', 0)
 }
 
+func humanTime(ts *time.Time) string {
+	var fmted string
+	if !ts.IsZero() {
```
titanous: I believe this should be inverted, the exceptional path should be indented, so `if ts.IsZero() { return "" }`

cli/deployment.go
josephglanville: Yep, that was my original thought. I think I changed it to fit the initial issue but deployment does make more sense to me too.

cli/main.go
josephglanville: Good call.

titanous:
This is great!

It looks like we need a test for the singleton get endpoint, there is already a `controller/deployment_test.go` file that it could be added to.

josephglanville:
Add tests for List and Get controller endpoints.

controller/deployment.go
```diff
@@ -81,18 +94,58 @@ func (r *DeploymentRepo) Add(data interface{}) error {
...
+                     processes, d.created_at, d.finished_at
+              FROM deployments d
+              LEFT JOIN deployment_events e1 ON d.deployment_id = e1.object_id::uuid
+              LEFT OUTER JOIN deployment_events e2 ON (d.deployment_id = e2.object_id::uuid AND e1.created_at < e2.created_at)
+              WHERE e2.created_at IS NULL AND d.app_id = $1 ORDER BY d.created_at DESC`
+	rows, err := r.db.Query(query, appID)
+	if err != nil {
+		return nil, err
+	}
+	deployments := []*ct.Deployment{}
```
titanous: For zero values we typically declare like this to avoid an extra allocation:

```go
var deployments []*ct.Deployment
```

controller/deployment_test.go
```diff
@@ -88,3 +89,57 @@ func (s *S) TestStreamDeployment(c *C) {
...
+	d, err = s.c.CreateDeployment(app.ID, newRelease.ID)
+
+	// test we can retrieve it
+	deployment, err := s.c.GetDeployment(d.ID)
+	c.Assert(err, IsNil)
+	c.Assert(deployment.ID, Equals, d.ID)
+	c.Assert(deployment.AppID, Equals, app.ID)
+	c.Assert(deployment.OldReleaseID, Equals, release.ID)
+	c.Assert(deployment.NewReleaseID, Equals, newRelease.ID)
+	c.Assert(reflect.DeepEqual(deployment.Processes, map[string]int{"web": 1}), Equals, true)
```
titanous: It looks like this doesn't assert the status? (ditto below)

controller/deployment.go
josephglanville: Cool, will keep that in mind.

titanous:
Would you please rebase this on the latest master? I'd like to see a test run before merging to master.

lmars:
`CLISuite.TestDeploy` is hanging when pushing via git, and we pinned it down to the fact that the gitreceive app is being deployed shortly after bootstrap, and the router does not close the client git connection.

Although [a deployment is created](https://github.com/flynn/flynn/blob/master/bootstrap/deploy_app_action.go#L101) for gitreceive during bootstrap, [this code](https://github.com/flynn/flynn/blob/master/controller/deployment.go#L53-L60) on master prevents the deployment from actually being enqueued and subsequently worked off. That code [has been changed on this branch](https://github.com/josephglanville/flynn/blob/deployments_cli/controller/deployment.go#L53-L65), so that could be the reason gitreceive is now being deployed after bootstrap.

We also need to determine why the client git connection does not get closed when gitreceive dies.

josephglanville:
The cause of the hang is not related to this patch set, it occurs even on master if a deploy happens early enough after bootstrap as demonstrated in #1804.

josephglanville:
ping @titanous this should be good to go now.

titanous:
LGTM.

PR #1754, opened 2015-08-26
cli: Implement resource removal

### LINKED ISSUE #1413: Remove/destroy resources
We need an API endpoint for remove and destroy resources (e. g. PostgreSQL db), something like

```sh
$ flynn resource
ID                                Provider ID                       Provider Name
3eb13704c1e74697a5fb542bdee189e6  91718ec7103f4525bb77ffabf15e355b  postgres

$ flynn resource remove 3eb13704c1e74697a5fb542bdee189e6
Resource 3eb13704c1e74697a5fb542bdee189e6 (postgres provider) removed from app.
```

josephglanville:
Allow removal of resources via CLI

Closes #1413

### DIFF
--- a/cli/resource.go
+++ b/cli/resource.go
@@ -12,19 +12,24 @@ func init() {
 	register("resource", runResource, `
 usage: flynn resource
        flynn resource add <provider>
+       flynn resource remove <provider> <resource>
 
 Manage resources for the app.
 
 Commands:
-	With no arguments, shows a list of resources.
+       With no arguments, shows a list of resources.
 
-	add   provisions a new resource for the app using <provider>.
+       add     provisions a new resource for the app using <provider>.
+       remove  removes the existing <resource> provided by <provider>.
 `)
 }
 
 func runResource(args *docopt.Args, client *controller.Client) error {
 	if args.Bool["add"] {
 		return runResourceAdd(args, client)
+	}
+	if args.Bool["remove"] {
+		return runResourceRemove(args, client)
 	}
 
 	resources, err := client.AppResourceList(mustApp())
@@ -69,6 +74,39 @@ func runResourceAdd(args *docopt.Args, client *controller.Client) error {
 	}
 
 	log.Printf("Created resource %s and release %s.", res.ID, releaseID)
+
+	return nil
+}
+
+func runResourceRemove(args *docopt.Args, client *controller.Client) error {
+	provider := args.String["<provider>"]
+	resource := args.String["<resource>"]
+
+	res, err := client.DeleteResource(provider, resource)
+	if err != nil {
+		return err
+	}
+
+	release, err := client.GetAppRelease(mustApp())
+	if err != nil {
+		return err
+	}
+
+	// Unset all the keys associated with this resource
+	env := make(map[string]*string)
+	for k := range res.Env {
+		// Only unset the key if it hasn't been modified
+		if release.Env[k] == res.Env[k] {
+			env[k] = nil
+		}
+	}
+
+	releaseID, err := setEnv(client, "", env)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("Deleted resource %s, created release %s.", res.ID, releaseID)
 
 	return nil
 }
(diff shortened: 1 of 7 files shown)


cli/resource.go
```diff
@@ -12,13 +12,15 @@ func init() {
 	register("resource", runResource, `
 usage: flynn resource
        flynn resource add <provider>
+	   flynn resource remove <provider> <resource>
```
lmars: This whitespace looks off

cli/resource.go
```diff
@@ -12,13 +12,15 @@ func init() {
...
 
 Manage resources for the app.
 
 Commands:
 	With no arguments, shows a list of resources.
 
-	add   provisions a new resource for the app using <provider>.
+	add    provisions a new resource for the app using <provider>.
+	remove removes the existing <resource> provided by <provider>.
```
lmars: Usage should include multiple spaces (at least two) between commands and their description

cli/resource.go
```diff
@@ -72,3 +77,33 @@ func runResourceAdd(args *docopt.Args, client *controller.Client) error {
...
+func runResourceRemove(args *docopt.Args, client *controller.Client) error {
+	provider := args.String["<provider>"]
+	resource := args.String["<resource>"]
+
+	res, err := client.GetResource(provider, resource)
+	if err != nil {
+		return err
+	}
+
+	err = client.DeleteResource(provider, resource)
```
lmars: Would it make sense for `DeleteResource` to return the resource which was deleted, thus avoiding the `GetResource` above?

josephglanville: Yeah good call.

cli/resource.go
josephglanville: Yeah, bad editor config. Done.

cli/resource.go
josephglanville: Done.

cli/resource.go
```diff
@@ -72,3 +77,28 @@ func runResourceAdd(args *docopt.Args, client *controller.Client) error {
...
+
+	res, err := client.DeleteResource(provider, resource)
+	if err != nil {
+		return err
+	}
+
+	// Unset all the keys associated with this resource
+	env := make(map[string]*string)
+	for k := range res.Env {
+		env[k] = nil
```
titanous: What do you think about comparing the values too? It's totally possible that they were changed since the resource was added. (long-term we need to come up with a more robust way of handling resource config)

test/test_cli.go
```diff
@@ -462,6 +462,21 @@ func (s *CLISuite) TestResourceList(t *c.C) {
...
+	add := app.flynn("resource", "add", "postgres")
+	t.Assert(add, Succeeds)
+	t.Assert(app.flynn("resource").Output, Matches, `postgres`)
+	id := strings.Split(add.Output, " ")[2]
+
+	remove := app.flynn("resource", "remove", "postgres", id)
+	t.Assert(remove, Succeeds)
+
+	t.Assert(app.flynn("resource").Output, c.Not(Matches), `postgres`)
+}
```
titanous: This should test the environment too.

cli/resource.go
josephglanville: Yeah, I am not sure.
There is somewhat of an assumption variables introduced by the resource are associated with it and if they were to be modified it should have been done to modify the behaviour of the application specific to that resource.
That said is there a means of comparing it? The original value of the variables is only known at resource provisioning time (I think).

test/test_cli.go
josephglanville: Indeed, good call.

cli/resource.go
titanous: Are the values not available in `res.Env`?

josephglanville: Oh. Facepalm. Easy enough to do then. I will change it so it unsets only when the variable hasn't changed since provisioning.

josephglanville:
Changed removal to only delete modified env vars.
Added assertions that test this behaviour.

titanous:
LGTM!

Issue #1758, opened 2015-08-26
util/packer,installer: Reintroduce eu-central-1 AWS region


lmars:
We previously [removed eu-central-1 support](https://github.com/flynn/flynn/commit/c56d0dc00fcf0718b03ef0e7bcca20de6a766da5) due to mitchellh/packer#1646, but that issue is now resolved, so we can support it again.

Necessary changes:
* update AWS packer builder: https://github.com/flynn/flynn/blob/v20150826.0/util/packer/ubuntu-14.04/template.json#L60-L68
* allow it to be selected in the installer: https://github.com/flynn/flynn/blob/v20150826.0/installer/app/src/views/aws-region-picker.js.jsx

PR #1759, opened 2015-08-26
release: Restore eu-central-1

josephglanville:
Now that Packer has migrated away from goamz v4 signatures are used and eu-central-1 can be supported.

Note: Requires packer version of 0.8.0 or greater.

Closes: #1758


titanous:
LGTM

Issue #1757, opened 2015-08-26
Second rails push with ruby-2.2.2 fails


titanous:
With an empty Rails app and `ruby "2.2.2"` specified in the `Gemfile`, the second push fails with this error:

```text
-----> Building cranes-cuddle-belleville...
-----> Ruby app detected
remote: cp: failed to access 'vendor/heroku': Permission denied
-----> Compiling Ruby/Rails
remote: /usr/lib/ruby/1.9.1/fileutils.rb:247:in `mkdir': Permission denied - vendor/ruby-2.2.2 (Errno::EACCES)
remote: 	from /usr/lib/ruby/1.9.1/fileutils.rb:247:in `fu_mkdir'
remote: 	from /usr/lib/ruby/1.9.1/fileutils.rb:221:in `block (2 levels) in mkdir_p'
remote: 	from /usr/lib/ruby/1.9.1/fileutils.rb:219:in `reverse_each'
remote: 	from /usr/lib/ruby/1.9.1/fileutils.rb:219:in `block in mkdir_p'
remote: 	from /usr/lib/ruby/1.9.1/fileutils.rb:205:in `each'
remote: 	from /usr/lib/ruby/1.9.1/fileutils.rb:205:in `mkdir_p'
remote: 	from /tmp/buildpacks/000003_heroku-buildpack-ruby/lib/language_pack/ruby.rb:318:in `block in install_ruby'
remote: 	from /tmp/buildpacks/000003_heroku-buildpack-ruby/lib/language_pack/instrument.rb:18:in `block (2 levels) in instrument'
remote: 	from /tmp/buildpacks/000003_heroku-buildpack-ruby/lib/language_pack/instrument.rb:40:in `yield_with_block_depth'
remote: 	from /tmp/buildpacks/000003_heroku-buildpack-ruby/lib/language_pack/instrument.rb:17:in `block in instrument'
remote: 	from /usr/lib/ruby/1.9.1/benchmark.rb:295:in `realtime'
remote: 	from /tmp/buildpacks/000003_heroku-buildpack-ruby/lib/language_pack/instrument.rb:16:in `instrument'
remote: 	from /tmp/buildpacks/000003_heroku-buildpack-ruby/lib/language_pack/base.rb:48:in `instrument'
remote: 	from /tmp/buildpacks/000003_heroku-buildpack-ruby/lib/language_pack/base.rb:44:in `instrument'
[...]

PR #1764, opened 2015-08-26
Fix git cache directory permissions issues

titanous:
The change in #1617 to checkout a git working directory instead of using `git-archiv`e caused a permissions handling bug that mangled permissions and omitted the +x flag from some directories when the working directory was restored from the repo cache.

This was fixed by stripping the (redundant) checked out working directory when creating the cache. A test was added to verify that the permissions bug does not return.

refs #1757


lmars:
LGTM

PR #1768, opened 2015-08-27
Dashboard sse firehose

jvatic:

PR #1769, opened 2015-08-27
controller,cli,test: Use Job in place of JobEvent

jvatic:
This makes the events endpoint more useful for apps such as the dashboard which need job meta for identification (e.g. in the case of taffy jobs).

Dependency of #1768


titanous:
LGTM

Issue #1744, opened 2015-08-23
Switch git from SSH to HTTPS 


titanous:
I believe we should change gitreceived to use the Smart HTTP(S) transport instead of SSH. This has a number of benefits, and basically no downsides.

For starters, here are some of the benefits:

1. No more `flynn key add` step (potentially requiring SSH key generation). We can preconfigure/generate credentials of many different types (including TLS client certificates) automatically without user knowledge.
1. The git push/pull streams will go through the router's HTTP stack, which means that technically only port 443 needs to be exposed to get useful Flynn functionality. (this will also be helpful for those that can't use "weird" protocols or need to proxy all network traffic)
1. We can easily and securely pre-pin the CA certificate using git's configuration options, this is way easier than trying to wrangle SSH configs.
1. It will be trivial to add all sorts of alternative auth mechanisms using `gitcredentials(7)` and the associated configuration options. This should allow two-factor auth, macaroons, and arbitrary SSO integration. Doing this with SSH would be annoying to say the least.
1. SSH configurations are very user-centric, programmatic management of them is hard for a variety of reasons, Git's HTTP support is the opposite.
1. Less crypto is in play, we won't have to audit the SSH library, only the TLS library which we already use for everything else.
1. Less code should be required to handle the push, for example: https://gitlab.com/gitlab-org/gitlab-git-http-server/blob/master/main.go
1. We already use HTTP(S) everywhere else, this is one less protocol we need to be able to understand and debug.

I can't really think of any real downsides, I do think that we should drop SSH support right away, since there's no reason to keep two protocols in play.

Any thoughts @lmars @nobert @temujin9?

lmars:
SGTM

temujin9:
+1, though I'll need #1647 at the same time (as my current "maintenance mode" relies on port blocking).

nobert:
I do support this, but don't we still need to use SSH for bootstrapping Flynn on platforms that don't support booting from custom images?

titanous:
@nobert Yeah, but that's a completely separate thing. We use SSH in the installer, but the key management is completely invisible to the user and we don't use OpenSSH as the client.

PR #1773, opened 2015-08-28
Git over HTTP

titanous:
This changes gitreceived to use the Smart HTTP git protocol instead of SSH, which removes the need for users to manage SSH keys.

The CLI now configures two global git options that pin the CA certificate for the git remote and use a credential helper built into the CLI to get the password. In the future this helper will be used to support per-user auth, etc.

Port 2222 is no longer used and is removed from all configs/docs.

The CLI cluster config is simplified, but remains temporarily backwards-compatible with SSH gitreceive, we'll remove the backwards compatibility when few users have old versions deployed.

There is also some cleanup, moving things around and removing the keys endpoint from the controller.

refs #1744


controller/utils/utils.go
```diff
@@ -71,3 +78,28 @@ type HostClient interface {
...
+var AppNamePattern = regexp.MustCompile(`^[a-z\d]+(-[a-z\d]+)*$`)
+
+func ParseBasicAuth(h http.Header) (username, password string, err error) {
+	s := strings.SplitN(h.Get("Authorization"), " ", 2)
+
+	if len(s) != 2 {
+		return "", "", errors.New("failed to parse authentication string ")
+	}
+	if s[0] != "Basic" {
+		return "", "", fmt.Errorf("authorization scheme is %v, not Basic ", s[0])
```
lmars: Drop the trailing whitespace from this and the error above.

gitreceive/server.go
```diff
@@ -0,0 +1,373 @@
...
+
+func (h *gitHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var g gitService
+
+	if r.URL.Path == status.Path {
+		status.HealthyHandler.ServeHTTP(w, r)
+		return
+	}
+
+	log.Printf("%s %q", r.Method, r.URL)
```
lmars: Should we wrap the handler in `httphelper.NewRequestLogger` to standardise our logging? Or at least use log15?

titanous: Yeah, I'll fix up the logging, this was derived from an existing codebase.

gitreceive/server.go
```diff
@@ -0,0 +1,373 @@
...
+
+	// The client request body may have been gzipped.
+	if r.Header.Get("Content-Encoding") == "gzip" {
+		body, err = gzip.NewReader(r.Body)
+		if err != nil {
+			fail500(w, "handlePostRPC", err)
+			return
+		}
+	} else {
+		body = r.Body
```
lmars: Could set this above and avoid the `else`.

gitreceive/server.go
```diff
@@ -0,0 +1,373 @@
...
+		return
+	}
+	defer cleanUpProcessGroup(cmd) // Ensure brute force subprocess clean-up
+
+	// Write the client request body to Git's standard input
+	if _, err := io.Copy(stdin, body); err != nil {
+		fail500(w, "handlePostRPC write to subprocess", err)
+		return
+	}
+	stdin.Close()
```
lmars: This close was deferred above, do we need to do it explicitly here?

gitreceive/server.go
```diff
@@ -0,0 +1,373 @@
...
+	w.WriteHeader(200) // Don't bother with HTTP 500 from this point on, just return
+	if err := pktLine(w, fmt.Sprintf("# service=%s\n", rpc)); err != nil {
+		logContext("handleGetInfoRefs response", err)
+		return
+	}
+	if err := pktFlush(w); err != nil {
+		logContext("handleGetInfoRefs response", err)
+		return
+	}
+	if _, err := io.Copy(w, pipe); err != nil {
```
lmars: Should this use a `newFlushWriter` like `handlePostRPC`?

gitreceive/server.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	process := cmd.Process
+	if process != nil && process.Pid > 0 {
+		// Send SIGTERM to the process group of cmd
+		syscall.Kill(-process.Pid, syscall.SIGTERM)
+	}
+
+	// reap our child process
+	cmd.Wait()
```
lmars: Do we need to explicitly wait for this (i.e. should we do it in a goroutine?)

lmars:
Looks great! LGTM pending comments.

gitreceive/server.go
titanous: It doesn't need it, there is no build output from this endpoint.

titanous:
Comments addressed.

PR #1775, opened 2015-08-28
controller: Create app_release event instead of app event when setting app release

jvatic:
This makes watching for app releases much easier (i.e. not having to hold onto all release events until an app event containing the release id is received).

Dependency of #1768

### DIFF
--- a/controller/app.go
+++ b/controller/app.go
@@ -228,6 +228,11 @@ func (r *AppRepo) SetRelease(app *ct.App, releaseID string) error {
 	if err != nil {
 		return err
 	}
+	var release *ct.Release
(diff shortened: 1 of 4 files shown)


controller/app.go
```diff
@@ -224,6 +224,14 @@ func (r *AppRepo) List() (interface{}, error) {
...
 
 func (r *AppRepo) SetRelease(app *ct.App, releaseID string) error {
+	var release *ct.Release
+	{
+		var err error
+		row := r.db.QueryRow("SELECT release_id, artifact_id, data, created_at FROM releases WHERE release_id = $1 AND deleted_at IS NULL", releaseID)
+		if release, err = scanRelease(row); err != nil {
+			return err
+		}
+	}
```
titanous: This should be in the same transaction as below.

jvatic:
@titanous comment addressed

titanous:
LGTM

PR #1777, opened 2015-08-28
cli: Add providers via CLI

josephglanville:
This is a stepping stone to full custom provider management.


titanous:
LGTM, thanks!

PR #1749, opened 2015-08-25
util/cedarish: Add postgres client libraries

titanous:
For upstream parity: https://github.com/heroku/stack-images/commit/ac4f62dd1b3417f1c035e1e634cc7f97ed65dcbe

PR #1782, opened 2015-08-29
Revert "util/cedarish: Add postgres client libraries"

titanous:
Reverts flynn/flynn#1749

Something about this is broken, the package doesn't appear to exist.


lmars:
LGTM

PR #1783, opened 2015-08-29
test: Use `flynn cluster add` before running tests

lmars:
This is because the CLI now does more than just encode the config into `~/.flynnrc` (it also creates CA cert files)


titanous:
LGTM

PR #1784, opened 2015-08-30
Standardise on hyphenated UUIDS

### LINKED ISSUE #995: inconsistency in UUID formatting
in #959, @titanous brought up the point that Flynn strips dashes out of UUIDs at the presentation layer:

> In the controller we strip the dashes from UUIDs before they hit the presentation layer. I can't remember the rationale for it, and I'm open to reversing that decision, but that should happen on another ticket.

### LINKED ISSUE #1796: controller route id inconsistency
Route events coming through the controller remove the hyphens from the ids, but all other endpoints return the hyphenated route id as returned by the router.

#1784 fixes this by returning hyphenated ids everywhere.

josephglanville:
Closes #995
Closes #1796

### DIFF
--- a/router/http_test.go
+++ b/router/http_test.go
@@ -20,6 +20,8 @@ import (
 	"github.com/flynn/flynn/router/types"
 )
 
+const UUIDRegex = "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
+
 var httpClient = newHTTPClient("example.com")
 
 // borrowed from net/http/httptest/server.go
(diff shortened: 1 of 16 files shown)


router/http_test.go
```diff
@@ -279,7 +279,9 @@ func httpHeaderTestHandler(c *C, ip, port string) http.Handler {
 		c.Assert(req.Header["X-Forwarded-Proto"][0], Equals, "http")
 		c.Assert(len(req.Header["X-Request-Start"][0]), Equals, 13)
 		c.Assert(req.Header["X-Forwarded-For"][0], Equals, ip)
-		c.Assert(len(req.Header["X-Request-Id"][0]), Equals, 32)
+		// size of UUID + the 4 hyphens
+		// maybe better to actually parse the UUID back to bytes here?
```
josephglanville: I kind of feel like we should parse the UUID here for completeness rather than just counting bytes, thoughts?

titanous: You could use `fmt.Scan*` to avoid writing a parser

josephglanville: Decided to just use the regex matcher here, simple and ensure it's valid.

titanous:
The commit message needs an `all: ` prefix.

josephglanville:
ping @titanous

titanous:
LGTM. Thanks!

PR #1785, opened 2015-08-31
cli: Add --quiet flag to `flynn release`

### LINKED ISSUE #1664: `flynn scale` should list scales for all releases
Prior releases may be left scaled up, if certain conditions occur. In order to maintain in that event, it would be helpful to see a list of all (non-zero) scales, not just the one for the current release.

lmars:
This makes it easier to use this command in a pipeline, for example listing the scale of all releases:

```
$ flynn release -q | xargs -L1 flynn scale -r
```

Closes #1664.

### DIFF
--- a/cli/release.go
+++ b/cli/release.go
@@ -18,13 +18,14 @@ import (
 
 func init() {
 	register("release", runRelease, `
-usage: flynn release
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

PR #1786, opened 2015-08-31
cli: flynn remote add

### LINKED ISSUE #1762: cli: Add command to (re)create the git remote for an existing app
Users should be able to just add the git remote to a clone of an application that has already been created in Flynn via a simple CLI command.

josephglanville:
CLI for managing git remotes for flynn applications

Closes #1762

### DIFF
--- a/cli/common.go
+++ b/cli/common.go
@@ -1,6 +1,9 @@
 package main
 
-import "fmt"
+import (
+	"fmt"
+	"log"
+)
 
 func promptYesNo(msg string) (result bool) {
 	fmt.Print(msg)
@@ -17,4 +20,21 @@ func promptYesNo(msg string) (result bool) {
 			fmt.Print("Please type 'yes' or 'no': ")
 		}
 	}
+}
+
+func promptReplaceRemote(remote string) (bool, error) {
+	remotes, err := gitRemoteNames()
+	if err != nil {
+		return false, err
+	}
+	for _, r := range remotes {
+		if r == remote {
+			fmt.Println("There is already a git remote called", remote)
+			if !promptYesNo("Are you sure you want to replace it?") {
+				log.Println("The remote was not created. Please declare the desired local git remote name with --remote flag.")
+				return false, nil
+			}
+		}
+	}
+	return true, nil
 }
--- a/cli/remote.go
+++ b/cli/remote.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"log"
+	"os/exec"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
+	"github.com/flynn/flynn/controller/client"
+)
+
+func init() {
+	register("remote", runRemote, `
+usage: flynn remote add [<remote>] [-y]
+
+Create a git remote that allows deploying the application via git.
+If a name for the remote is not provided 'flynn' will be used.
+
+Options:
+	-y, --yes              Skip the confirmation prompt if the git remote already exists.
+
+Examples:
+
+	$ flynn -a turkeys-stupefy-perry remote add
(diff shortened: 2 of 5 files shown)


cli/remote.go
```diff
@@ -0,0 +1,68 @@
...
+	"log"
+	"os/exec"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
+	"github.com/flynn/flynn/controller/client"
+)
+
+func init() {
+	register("remote", runRemote, `
+usage: flynn remote <name> [-r <remote>] [-y]
```
titanous: 1. I think this should be `flynn remote add`
2. So far we always specify the app name with `flynn -a <app> remote add`, which doesn't exactly roll off the tongue and has positional issues, but we should probably support it.
3. It seems like the primary argument of this command should be the name of the remote, without having to specify `-r`.

Thoughts?

josephglanville: All of that sounds good.

Will do that and add tests. :)

cli/remote.go
```diff
@@ -0,0 +1,70 @@
...
+Create a git remote that allows deploying the application via git.
+If a name for the remote is not provided 'flynn' will be used.
+
+Options:
+	-y, --yes              Skip the confirmation prompt if the git remote already exists.
+
+Examples:
+
+	$ flynn remote turkeys-stupefy-perry
+	Created remote flynn with url https://git.dev.localflynn.com/turkeys-stupefy-perry.git
```
titanous: This example looks wrong.

cli/remote.go
```diff
@@ -0,0 +1,70 @@
...
+		remote = "flynn"
+	}
+
+	if !inGitRepo() {
+		log.Print("Must be executed within a git repository.")
+		return nil
+	}
+
+	if remote == "" {
+		log.Print("Remote name should not be empty")
```
titanous: This doesn't look possible.

cli/remote.go
```diff
@@ -0,0 +1,70 @@
...
+
+	for _, r := range remotes {
+		if r == remote {
+			fmt.Println("There is already a git remote called", remote)
+			if !promptYesNo("Are you sure you want to replace it?") {
+				log.Println("The remote was not created. Please, declare the desired local git remote name with --remote flag.")
+				return nil
+			}
+		}
+	}
```
titanous: Do we already have logic for this in `flynn create`? Can it be shared?

josephglanville:
@titanous comments addressed and tests added.

cli/remote.go
```diff
@@ -0,0 +1,59 @@
...
+
+Create a git remote that allows deploying the application via git.
+If a name for the remote is not provided 'flynn' will be used.
+
+Options:
+	-y, --yes              Skip the confirmation prompt if the git remote already exists.
+
+Examples:
+
+	$ flynn remote add turkeys-stupefy-perry
```
titanous: I think this remote name should be `staging` or similar and reflected in the line below, currently it looks like an app name (maybe move that to demonstrate the `-a` flag).

josephglanville: Done.

titanous:
It looks like the command needs to be added to the list in `main.go`

cli/common.go
```diff
@@ -18,3 +21,20 @@ func promptYesNo(msg string) (result bool) {
...
+func promptReplaceRemote(remote string) (bool, error) {
+	remotes, err := gitRemoteNames()
+	if err != nil {
+		return false, err
+	}
+	for _, r := range remotes {
+		if r == remote {
+			fmt.Println("There is already a git remote called", remote)
+			if !promptYesNo("Are you sure you want to replace it?") {
+				log.Println("The remote was not created. Please, declare the desired local git remote name with --remote flag.")
```
titanous: Remove the comma after "Please".

titanous:
LGTM pending my last two comments.

PR #1788, opened 2015-08-31
pinkerton: Remove unused storage drivers

titanous:
We currently only use AUFS, no need to compile the rest of the drivers.


lmars:
LGTM

PR #1792, opened 2015-08-31
Postgres deployment fixes

### LINKED ISSUE #1661: When updating I get the error "pg cluster in unhealthy state"
pg cluster in unhealthy state (too few syncs)

https://gist.github.com/3266345d084810a50b51

lmars:
Closes #1661.

### DIFF
--- a/controller/worker/deployment/postgres.go
+++ b/controller/worker/deployment/postgres.go
@@ -48,6 +48,10 @@ func (d *DeployJob) deployPostgres() (err error) {
 	if 2+len(state.Async) != d.Processes["postgres"] {
 		return loggedErr(fmt.Sprintf("pg cluster in unhealthy state (too few asyncs)"))
 	}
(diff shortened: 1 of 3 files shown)


controller/worker/deployment/postgres.go
```diff
@@ -48,6 +49,10 @@ func (d *DeployJob) deployPostgres() (err error) {
 	if 2+len(state.Async) != d.Processes["postgres"] {
 		return loggedErr(fmt.Sprintf("pg cluster in unhealthy state (too few asyncs)"))
 	}
+	if reflect.DeepEqual(d.newReleaseState, d.Processes) {
```
titanous: Don't use `reflect` if you don't have to. In this case it's a bit more code, but we can have a type-safe comparison.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #1798, opened 2015-09-01
controller: Handle lack of transaction in DeploymentRepo.Add

lmars:


josephglanville:
How hard is it to fix the TODO rather than add this workaround?

lmars:
@josephglanville good point, `DeploymentRepo.Add` has access to the `pgx.ConnPool`, so I will use that to create the transaction for the rest of the method.

lmars:
@josephglanville turns out switching to `*pgx.Tx` isn't a straight forward switch, getting the following error:

```
Cannot encode *string into oid 2950 - *string must implement Encoder or be converted to a string
```

I'm sure we could fix this up, but I am tempted to merge this and switch to pgx separately.

josephglanville:
Yeah, it definitely looks that way for now. Was worth a shot though. :)

controller/deployment.go
```diff
@@ -70,11 +70,17 @@ func (r *DeploymentRepo) Add(data interface{}) error {
...
-	}); err != nil {
+
+	// Delete the deployment if we get ErrDeadConn because the client will
+	// retry on that error, but client retries will fail due to the
+	// deployment already existing in the database.
+	//
+	// TODO: wrap all of this in a transaction and drop the delete once we
+	//       move to pgx
+	job := &que.Job{Type: "deployment", Args: args}
+	if err := r.q.Enqueue(job); err == pgx.ErrDeadConn {
```
titanous: Why just `ErrDeadConn`? If we're hackily emulating a transaction rollback, should any error trigger this?

lmars:
@titanous comments addressed

controller/deployment.go
```diff
@@ -70,12 +70,15 @@ func (r *DeploymentRepo) Add(data interface{}) error {
...
-		return err
+
+	// Delete the deployment if we get an error so the client can retry.
+	//
+	// TODO: wrap all of this in a transaction and drop the delete once we
+	//       move to pgx
+	job := &que.Job{Type: "deployment", Args: args}
+	if err := r.q.Enqueue(job); err != nil {
+		logger.New("fn", "DeploymentRepo.Add").Warn("error enqueueing deployment job", "err", err)
+		return r.db.Exec("DELETE FROM deployments WHERE deployment_id = $1", d.ID)
```
titanous: Eat the delete error, return the original error.

titanous:
LGTM pending comment, but we should really get a transaction in here, this is bad.

PR #1806, opened 2015-09-02
bootstrap: Don't deploy apps

lmars:
Since we are creating initial releases, there is no need to deploy them, we just need to set the app release.


lmars:
Closing as this was not causing issues, and is more inline with what the receiver does.

PR #1807, opened 2015-09-02
test: Don't deploy gitreceive concurrently

lmars:
`TestPrivateSSHKeyClone` deploys gitreceive (via `flynn env set`) so can be disruptive to other tests.


titanous:
LGTM

PR #1812, opened 2015-09-03
router: Report conflict on duplicate route

### LINKED ISSUE #708: Router and controller should report sanely on duplicate route
Currently the router sends a 500 when trying to create a duplicate route, and that passes through the controller all the way to the CLI, resulting in a cryptic error. They should return success with a warning (on an identical route) or a clear failure (on a duplication that is not identical, such as a route already aimed to another app).

josephglanville:
If the router receives a request to add a duplicate route it will now respond with a 409 Conflict.

The resultant error on the command line now looks like this:

```
$ flynn route add http example.com
conflict: Duplicate route
$ echo $?
1
```

Closes #708

### DIFF
--- a/router/api.go
+++ b/router/api.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"log"
 	"net/http"
 	"os"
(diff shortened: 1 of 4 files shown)


titanous:
LGTM.

PR #1813, opened 2015-09-03
cli: Allow selecting port for TCP routes

### LINKED ISSUE #1524: router: Support binding TCP services to specific ports
I should be able to add a TCP route which binds to a specific port, for example:

```
$ flynn route add tcp --service mail-server --port 25
```

josephglanville:
Allows users to specify the port they wish to allocate.
If the port is take it will return a duplicate route error.

Closes #1524


titanous:
LGTM

temujin9:
Whoot! I was just looking at my backlog this morning, and wondering whether I needed to ask y'all about this . . .

PR #1818, opened 2015-09-03
scheduler-v2: Pre master review

lmars:
A PR to review the scheduler-v2 implementation, ensuring it addresses issues raised in #1484.

This is a PR into `scheduler-v2-base`, with the hope that once this is merged, we can PR to master and get more peer review.

/cc @jzila


controller/jobs.go
```diff
@@ -144,9 +146,39 @@ func (c clusterClientWrapper) Hosts() ([]utils.HostClient, error) {
 	return res, nil
 }
 
-type clusterClient interface {
-	Host(string) (utils.HostClient, error)
-	Hosts() ([]utils.HostClient, error)
+func (c clusterClientWrapper) StreamHostEvents(ch chan *discoverd.Event) (stream.Stream, error) {
+	hostChan := make(chan *discoverd.Event)
+	stream, err := c.Client.StreamHostEvents(hostChan)
```
lmars: Can we not just return `c.Client.StreamHostEvents(ch)`?

controller/jobs_test.go
```diff
@@ -94,7 +94,8 @@ func (s *S) TestKillJob(c *C) {
 	hc := tu.NewFakeHostClient(hostID)
 	s.cc.AddHost(hc)
 
-	c.Assert(s.c.DeleteJob(app.ID, jobID), IsNil)
+	err := s.c.DeleteJob(app.ID, jobID)
+	c.Assert(err, ErrorMatches, "controller: resource not found")
```
lmars: `c.Assert(err, Equals, controller.ErrNotFound)`

controller/scheduler-v2/formation.go
```diff
@@ -0,0 +1,188 @@
...
+type Formation struct {
+	*ct.ExpandedFormation
+}
+
+func NewFormation(ef *ct.ExpandedFormation) *Formation {
+	return &Formation{ef}
+}
+
+func (f *Formation) key() utils.FormationKey {
+	return utils.FormationKey{f.App.ID, f.Release.ID}
```
lmars: Given the type is from another package, use explicit field labels:

```go
utils.FormationKey{AppID: xxx, ReleaseID: xxx}
```

controller/jobs.go
jzila: Oh yes. When this was returning `utils.HostClient`s, then this wrapper was necessary. It might not be necessary any more.

controller/scheduler-v2/formation.go
```diff
@@ -0,0 +1,188 @@
...
+	_, ok := fc[key]
+	if !ok {
+		fc[key] = make(map[string][]*Job)
+	}
+	fc[key][j.Type] = append(fc[key][j.Type], j)
+}
+
+type pendingJobs map[utils.FormationKey]map[string]map[string]int
+
+func CopyPendingJobs(pending pendingJobs) pendingJobs {
```
lmars: Make this `func(p pendingJobs) Clone() pendingJobs`

controller/scheduler-v2/formation.go
```diff
@@ -0,0 +1,188 @@
...
+	for key, form := range pending {
+		copied[key] = make(map[string]map[string]int)
+		for typ, hosts := range form {
+			copied[key][typ] = make(map[string]int)
+			for hostID, numJobs := range hosts {
+				copied[key][typ][hostID] = numJobs
+			}
+		}
+	}
+	return copied
```
lmars: The calls to `make` in this function should include capacities.

controller/scheduler-v2/formation.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func MergePendingJobs(pending1, pending2 pendingJobs) pendingJobs {
```
lmars: Does this need to return a copy? How about `func(p pendingJobs) Update(other pendingJobs)`?

controller/scheduler-v2/formation_test.go
```diff
@@ -0,0 +1,135 @@
+package main
+
+import (
+	. "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
+	. "github.com/flynn/flynn/controller/testutils"
+	ct "github.com/flynn/flynn/controller/types"
+	utils "github.com/flynn/flynn/controller/utils"
```
lmars: The package is already called `utils`

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,836 @@
...
+	formationChange chan *ct.ExpandedFormation
+	jobRequests     chan *JobRequest
+	putJobs         chan *ct.Job
+}
+
+func NewScheduler(cluster utils.ClusterClient, cc utils.ControllerClient) *Scheduler {
+	return &Scheduler{
+		ControllerClient: cc,
+		ClusterClient:    cluster,
+		isLeader:         false,
```
lmars: `false` is the zero value of `bool`, so no need to set it

jzila: Is it harmful to be explicit?

jzila:
Bug: requesting to stop multiple jobs can request to stop the same job twice.
```
t=2015-09-04T20:48:53+0000 lvl=dbug msg="Updating processes" component=scheduler fn=RectifyJobs formation.processes=map[echoer:3] cluster.processes="map[echoer:3 printer:2]"
t=2015-09-04T20:48:53+0000 lvl=info msg="sending event to listeners" component=scheduler event.type=rectify-jobs event.error=nil listeners.count=0
t=2015-09-04T20:48:53+0000 lvl=info msg="requested job stop from host" component=scheduler fn=stopJob host.id=flynn job.type=printer job.id=flynn-631118a2-1cfb-461b-900a-0acc629c351f
t=2015-09-04T20:48:53+0000 lvl=info msg="sending event to listeners" component=scheduler event.type=request-job event.error=nil listeners.count=0
t=2015-09-04T20:48:53+0000 lvl=info msg="requested job stop from host" component=scheduler fn=stopJob host.id=flynn job.type=printer job.id=flynn-631118a2-1cfb-461b-900a-0acc629c351f
t=2015-09-04T20:48:53+0000 lvl=info msg="sending event to listeners" component=scheduler event.type=request-job event.error=nil listeners.count=0
```

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,836 @@
...
+		schedulerProcs := schedulerFormation.Processes
+		clusterProcs := fj.GetProcesses(fKey)
+
+		if eq := reflect.DeepEqual(clusterProcs, schedulerProcs); !eq {
+			log.Info("Updating processes", "formation.processes", schedulerProcs, "cluster.processes", clusterProcs)
+			schedulerFormation.Processes = clusterProcs
+
+			diff := schedulerFormation.Update(schedulerProcs)
+			s.sendDiffRequests(schedulerFormation, diff)
+		}
```
lmars: This needs to ignore jobs with no type (otherwise known as "one-off jobs", like what `flynn run bash` starts) as they are currently being killed shortly after starting.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,848 @@
...
+
+	listeners map[chan Event]struct{}
+	listenMtx sync.RWMutex
+
+	stop     chan interface{}
+	stopOnce sync.Once
+
+	rectifyJobs     chan interface{}
+	syncJobs        chan interface{}
+	syncFormations  chan interface{}
```
titanous: Use `chan struct{}` for sentinel channels.

lmars:
@jzila the main scheduler loop should not block on network calls (i.e. controller / postgres / host clients), instead writing to channels which other goroutines can receive from and do the necessary calls. I am currently working on this.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,848 @@
...
+		}
+
+		// Finally, handle triggering cluster changes
+		select {
+		case <-s.rectifyJobs:
+			err = s.RectifyJobs()
+		case <-time.After(50 * time.Millisecond):
+			// block so that
+			//	1) mutate events are given a chance to happen
+			//  2) we don't spin hot
```
titanous: It probably makes sense to copy all of the unique channel receives from above into this select so that we don't need a timer.

lmars:
@jzila I have merged in your fixes, I am currently working on improving the logs.

lmars:
@jzila I have quite a few log changes I intend to push soon, let me know if you are working on things as it may cause you some rebase pain.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1041 @@
...
+	log.Info("handling formation change")
+	_, err = s.changeFormation(ef)
+	if err != nil {
+		log.Error("error handling formation change", "err", err)
+		return
+	}
+
+	// Trigger sync jobs in case we've ignored an existing job because we
+	// didn't know about the formation.
+	s.triggerSyncJobs()
```
lmars: When we get a job event, we [ask the controller for the formation if we don't already have it](https://github.com/flynn/flynn/blob/scheduler-v2-review/controller/scheduler-v2/scheduler.go#L580-L592), so do we need to trigger a job sync after each formation change, rather than just a rectify?

My main thinking here is that `SyncJobs` does quite a lot of work ([looks up all the hosts](https://github.com/flynn/flynn/blob/scheduler-v2-review/controller/scheduler-v2/scheduler.go#L212), [gets the jobs from each host](https://github.com/flynn/flynn/blob/scheduler-v2-review/controller/scheduler-v2/scheduler.go#L223), [saves the jobs](https://github.com/flynn/flynn/blob/scheduler-v2-review/controller/scheduler-v2/scheduler.go#L596) which leads to both [handleJobStart](https://github.com/flynn/flynn/blob/scheduler-v2-review/controller/scheduler-v2/scheduler.go#L814) and [putJobs](https://github.com/flynn/flynn/blob/scheduler-v2-review/controller/scheduler-v2/scheduler.go#L820) being called for every active job).

jzila: Yep I think you're right. This is an artifact of a previous iteration that didn't have the concerns separated out as well as they are now.

jzila: I've fixed that in my branch.

lmars:
@jzila FYI I have refactored the tests (5565f09) and I am now verifying we have enough test coverage.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1199 @@
...
+		hostLog.Info(fmt.Sprintf("getting jobs for host %s", h.ID()))
+		activeJobs, err := h.ListJobs()
+		if err != nil {
+			hostLog.Error("error getting jobs list", "err", err)
+			continue
+		}
+		hostLog.Info(fmt.Sprintf("got %d active job(s) for host %s", len(activeJobs), h.ID()))
+
+		for _, job := range activeJobs {
+			s.handleActiveJob(&job)
```
lmars: This should only be called if the job is either new or it's state has changed (currently this leads to all active jobs being persisted to the controller every 30s, and the list of active jobs increases as time goes on).

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1199 @@
...
+	case EventTypeLeaderChange:
+		isLeader, _ := data.(bool)
+		return &LeaderChangeEvent{Event: &DefaultEvent{err: err, typ: typ}, IsLeader: isLeader}
+	default:
+		return &DefaultEvent{err: err, typ: typ}
+	}
+}
+
+func getBackoffPeriod() time.Duration {
+	backoffPeriod := time.Second
```
lmars: This should default to 10 minutes.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1199 @@
...
+	}
+	return jobs
+}
+
+func (s *Scheduler) scheduleJobStart(job *Job) error {
+	log := fnLogger()
+	if !s.isLeader {
+		return errors.New("this scheduler is not the leader")
+	}
+	if job.startedAt.Before(time.Now().Add(-s.backoffPeriod)) {
```
lmars: It seems that `startedAt` is the time when the `Job` struct was created, but the logic here should be referring to when the job actually started on the host.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1199 @@
...
+		log.Info("resetting job restarts", "backoffPeriod", s.backoffPeriod, "job.startedAt", job.startedAt)
+		job.restarts = 0
+	}
+	backoff := s.getBackoffDuration(job.restarts)
+	if backoff == s.maxBackoffPeriod {
+		log.Warn("reached maximum backoff period, allowing job to stay crashed")
+		return errors.New("maximum backoff period reached")
+	}
```
lmars: The backoff period should be capped, but we should continue to restart the job at that capped backoff period (there is no reason to leave a job we believe should be running as not running).

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1199 @@
...
+		stop:             make(chan struct{}),
+		rectify:          make(chan struct{}, 1),
+		syncJobs:         make(chan struct{}, 1),
+		syncFormations:   make(chan struct{}, 1),
+		formationEvents:  make(chan *ct.ExpandedFormation, eventBufferSize),
+		hostEvents:       make(chan *discoverd.Event, eventBufferSize),
+		jobRequests:      make(chan *JobRequest, eventBufferSize),
+		putJobs:          make(chan *ct.Job, eventBufferSize),
+	}
+	s.maxBackoffPeriod = 10 * s.backoffPeriod
```
lmars: The backoff periods increase in powers of 2, so it probably makes sense to cap at a particular power (e.g 32).

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1199 @@
...
+	return nil
+}
+
+func (s *Scheduler) getBackoffDuration(restarts uint) time.Duration {
+	// Overflow guard
+	if restarts > 30 {
+		return s.maxBackoffPeriod
+	}
+	// s.backoffPeriod * 2^(restarts - 1)
+	delay := s.backoffPeriod * time.Duration((1<<restarts)>>1)
```
lmars: The comment isn't accurate in the case `restarts=0`

lmars:
@jzila left a few comments related to job backoffs and `SyncJobs`, let me know in IRC if you start working on them, as I may fix them later today.

controller/scheduler-v2/scheduler.go
jzila: The only way to know if the state has changed is to `handleActiveJob` :P

controller/scheduler-v2/scheduler.go
jzila: K, although does that matter? Flynn sets `BACKOFF_PERIOD` anyway.

controller/scheduler-v2/scheduler.go
jzila: You're right, we should be setting `startedAt` to the value that comes from `ActiveJob` in `HandleActiveJob`

controller/scheduler-v2/scheduler.go
jzila: Rectify will restart it past this point. Unless you think we should continue explicitly restarting it?

controller/scheduler-v2/scheduler.go
jzila: K, will fix comment.

controller/scheduler-v2/scheduler.go
lmars: Ok, but this still needs fixing right? I think `SyncJobs` should be explicit in what it is doing (i.e. just making sure that `s.jobs` and the controller are in sync with the cluster), so I feel like things would be clearer if this were something like:

```go
for _, job := range jobs {
  if j, ok := s.jobs[job.ID]; ok && j.State != job.State {
    // the state has changed from what we know, treat it like an event
    s.HandleJobEvent(job)
  }
}
```

I've seen hosts with thousands of active jobs, so potentially doing thousands of unnecessary `PutJob` calls every 30s feels wrong.

controller/scheduler-v2/scheduler.go
lmars: `BACKOFF_PERIOD` is not set explicitly to 10m anywhere, we rely on the default value, and only override it in tests.

controller/scheduler-v2/scheduler.go
lmars: Ah ok, in that case why would we want to start it again straight away (via rectify) rather than continuing at the maximum backoff? Lets stick closely to the [Heroku policy](https://devcenter.heroku.com/articles/dynos#dyno-crash-restart-policy), which is what our policy is based on.

controller/scheduler-v2/scheduler.go
lmars: This would also have the benefit of not logging excessively.

controller/scheduler-v2/scheduler.go
jzila: Will do.

controller/scheduler-v2/scheduler.go
jzila: K understood

controller/scheduler-v2/scheduler.go
jzila: OK I'll wait for you to merge #1864 and then fix it.

lmars:
In the [latest build failure](https://ci.flynn.io/builds/20150913151356-a9a745ad), the main loop of the scheduler leader (job `89d4f3ae-7ac8112a-bd2a-485f-8023-5256e382aabe`) got blocked, meaning it stopped reacting to events.

The last action I can see in the logs is stopping a job due to a formation change:

```
t=2015-09-13T15:26:42+0000 lvl=info msg="rectifying formation" component=scheduler fn=Rectify app.id=427c6ab2-5bed-4b07-864e-dbf8e085ad26 release.id=5953a97b-56a8-466e-a51b-0ad453f08e36
t=2015-09-13T15:26:42+0000 lvl=info msg="existing formation in incorrect state" component=scheduler fn=Rectify app.id=427c6ab2-5bed-4b07-864e-dbf8e085ad26 release.id=5953a97b-56a8-466e-a51b-0ad453f08e36 e
xpected=map[resources:1] actual=map[resources:2] diff=map[resources:-1]
t=2015-09-13T15:26:42+0000 lvl=info msg="requesting removal of 1 job(s) of type resources" component=scheduler fn=sendDiffRequests app.id=427c6ab2-5bed-4b07-864e-dbf8e085ad26 release.id=5953a97b-56a8-466e
-a51b-0ad453f08e36
t=2015-09-13T15:26:42+0000 lvl=info msg="handling job request" component=scheduler fn=HandleJobRequest req.id= req.type=down
t=2015-09-13T15:26:42+0000 lvl=info msg="stopping job" component=scheduler fn=stopJob req.host.id= req.job.id= job.type=resources
t=2015-09-13T15:26:42+0000 lvl=info msg="selected job for termination" component=scheduler fn=stopJob req.host.id= req.job.id= job.type=resources job.id=697e7980-33e63547-9450-4dfd-9b7d-872510f65b1c job.h
ost.id=697e7980
t=2015-09-13T15:26:42+0000 lvl=info msg="getting host client" component=scheduler fn=stopJob req.host.id= req.job.id= job.type=resources
t=2015-09-13T15:26:42+0000 lvl=info msg="requesting host to stop job" component=scheduler fn=stopJob req.host.id= req.job.id= job.type=resources
```

Everything after that is logged from other goroutines (i.e. `RunPutJobs`, `triggerSyncJobs` or `triggerSyncFormations`) but no actions are being taken.

Looking in the host logs of `697e7980`, the `StopJob` request is logged as started, but is indeed blocked (there is no log of its completion):

```
t=2015-09-13T15:26:42+0000 lvl=info msg="request started" req_id=9a4db7a4-94c3-4ef6-a2c9-fa05420a1eec component=host method=DELETE path=/host/jobs/697e7980-33e63547-9450-4dfd-9b7d-872510f65b1c client_ip=10.69.10.4
```

We should investigate why the host is not returning a response, but this highlights the fact that the scheduler is vulnerable to blocking network requests bringing everything to a halt, something I think we really need to fix (see my [earlier comment](https://github.com/flynn/flynn/pull/1818#issuecomment-138135473)).

/cc @jzila

controller/Tupfile
```diff
@@ -1,6 +1,6 @@
 include_rules
 : |> !go |> bin/flynn-controller
-: |> !go ./scheduler |> bin/flynn-scheduler
+: |> !go ./scheduler-v2 |> bin/flynn-scheduler
```
lmars: We should remove the old scheduler code

controller/jobs_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	. "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
+	controller "github.com/flynn/flynn/controller/client"
```
lmars: This package is already called `controller`

controller/scheduler-v2/job.go
```diff
@@ -0,0 +1,179 @@
...
+
+const (
+	JobRequestTypeUp   JobRequestType = "up"
+	JobRequestTypeDown JobRequestType = "down"
+	JobStateStarting                  = "starting"
+	JobStateRunning                   = "running"
+	JobStateStopped                   = "stopped"
+	JobStateCrashed                   = "crashed"
+	JobStateRequesting                = "requesting"
+	JobStateNew                       = "new"
```
lmars: These should have a type of `JobState`

controller/scheduler-v2/job.go
```diff
@@ -0,0 +1,179 @@
...
+
+func NewJob(f *Formation, appID, releaseID, typ, hostID, id string) *Job {
+	return &Job{
+		Type:      typ,
+		AppID:     appID,
+		ReleaseID: releaseID,
+		HostID:    hostID,
+		JobID:     id,
+		Formation: f,
+		startedAt: time.Now(),
```
lmars: Does the logic rely on this being set, even though the job has not actually started? If not we should probably leave it as the zero value until the job starts.

controller/scheduler-v2/job.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (j *Job) Clone() *Job {
+	// Shallow copy
+	cloned := *j
+	return &cloned
+}
+
+func (j *Job) IsScheduled() bool {
+	return j.state != JobStateStopped
```
lmars: This seems to either be named wrong, or the logic is wrong. `IsScheduled` to me sounds like it should indicate that the job is scheduled to start by the scheduler, but is not an active job in the cluster, but this seems to be returning something else? Renaming / commenting would be useful.

controller/scheduler-v2/job.go
```diff
@@ -0,0 +1,179 @@
...
+			counts[j.HostID]++
+		}
+	}
+	return counts
+}
+
+func (js Jobs) GetProcesses(key utils.FormationKey) Processes {
+	procs := make(Processes)
+	for _, j := range js {
+		if j.IsScheduled() && j.Formation != nil && j.Formation.key() == key {
```
lmars: There are quite a few `j.Formation != nil` checks, do we expect jobs with no formation? If so, do they belong in `s.jobs`?

controller/scheduler-v2/job.go
```diff
@@ -0,0 +1,179 @@
...
+			procs[j.Type]++
+		}
+	}
+	return procs
+}
+
+func (js Jobs) AddJob(j *Job) *Job {
+	j = j.Clone()
```
lmars: Why do we have to copy this?

controller/scheduler-v2/job.go
```diff
@@ -0,0 +1,179 @@
...
+	return j
+}
+
+func (js Jobs) IsJobInState(id string, state JobState) bool {
+	j, ok := js[id]
+	if ok {
+		return j.state == state
+	}
+	return false
+}
```
lmars:
```go
if j, ok := js[id]; ok && j.state == state {
  return true
}
return false
```

controller/scheduler-v2/job.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (js Jobs) SetState(id string, state JobState) error {
+	j, ok := js[id]
+	if ok {
```
lmars: `if j, ok := js[id]; ok { ... }`

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+func fnLogger(ctx ...interface{}) log15.Logger {
+	pc := make([]uintptr, 10) // at least 1 entry needed
+	runtime.Callers(2, pc)
+	f := runtime.FuncForPC(pc[0])
+	name := f.Name()
+	name = name[strings.LastIndex(name, ".")+1:]
+	params := []interface{}{"fn", name}
+	params = append(params, ctx...)
+	return logger.New(params...)
+}
```
lmars: Is this necessary? It adds overhead to pretty much every method, and its purpose seems only to avoid writing `"fn", "func-name"` when creating a logger?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+		stop:             make(chan struct{}),
+		rectify:          make(chan struct{}, 1),
+		syncJobs:         make(chan struct{}, 1),
+		syncFormations:   make(chan struct{}, 1),
+		formationEvents:  make(chan *ct.ExpandedFormation, eventBufferSize),
+		hostEvents:       make(chan *discoverd.Event, eventBufferSize),
+		jobRequests:      make(chan *JobRequest, eventBufferSize),
+		putJobs:          make(chan *ct.Job, eventBufferSize),
+	}
+	return s
```
lmars: No need for the variable assignment

controller/scheduler-v2/job.go
titanous:
```go
j, ok := js[id]
return ok && j.state == state
```

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+}
+
+func main() {
+	log := fnLogger()
+
+	log.Info("creating cluster and controller clients")
+	hc := &http.Client{Timeout: 5 * time.Second}
+	cc := cluster.NewClientWithHTTP(nil, hc)
+	clusterClient := utils.ClusterClientWrapper(cc)
+	controllerClient, err := controller.NewClient("", os.Getenv("AUTH_KEY"))
```
lmars: We should use an explicit `http.Client` with a low timeout when creating the controller client too.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+			s.SyncFormations()
+		case <-s.syncJobs:
+			s.SyncJobs()
+		}
+	}
+	return nil
+}
+
+func (s *Scheduler) SyncJobs() {
+	defer s.sendEvent(NewEvent(EventTypeClusterSync, nil, nil))
```
lmars: Could let `sendEvent` call `NewEvent` to reduce noise.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+		if err != nil {
+			hostLog.Error("error getting jobs list", "err", err)
+			continue
+		}
+		hostLog.Info(fmt.Sprintf("got %d active job(s) for host %s", len(activeJobs), h.ID()))
+
+		for _, job := range activeJobs {
+			s.handleActiveJob(&job)
+			if j, ok := s.jobs[job.Job.ID]; ok {
+				newJobs[j.JobID] = s.jobs[j.JobID]
```
lmars: It seems this is copying all active jobs which already exist in `s.jobs` into `newJobs`, which seems odd. Either rename the variable (`knownJobs`?) or fix the logic to add unknown jobs into `newJobs`.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+			}
+		}
+	}
+
+	for id, j := range s.jobs {
+		if _, ok := newJobs[id]; !ok && j.IsRunning() {
+			s.handleJobStop(j)
+		}
```
lmars: What is this trying to catch? It seems as though it is marking all jobs in `s.jobs` as stopped which no longer appear as an active job from any host in the cluster, which will only happen if a host which was previously running jobs no longer appears in `cluster.Hosts`. Is this the intention?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+
+	for key, formation := range s.formations {
+		formationLog := log.New("app.id", key.AppID, "release.id", key.ReleaseID)
+		formationLog.Info("rectifying formation")
+
+		expected := formation.GetProcesses()
+		actual := s.jobs.GetProcesses(key)
+
+		if expected.Equals(actual) {
+			formationLog.Info("formation in correct state", "expected", expected, "actual", actual)
```
lmars: We should drop this log message (`Rectify` happens every time we get a job event, so this will get logged for each formation each time)

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+			continue
+		}
+		// TODO: make this formation.Diff(actual)?
```
lmars: We should either make this change or drop the TODO

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+	var err error
+	defer func() {
+		s.sendEvent(NewEvent(EventTypeFormationChange, err, nil))
+	}()
+
+	log := fnLogger()
+	if ef == nil || ef.App == nil || ef.Release == nil {
+		log.Error("invalid formation")
+		return
+	}
```
lmars: We should either drop this check (I don't think it's necessary), or comment why it is necessary.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+
+	log = log.New("app.id", ef.App.ID, "release.id", ef.Release.ID)
+	log.Info("handling formation change")
+	_, err = s.changeFormation(ef)
+	if err != nil {
+		log.Error("error handling formation change", "err", err)
+		return
+	}
+
+	s.triggerRectify()
```
lmars: Rather than triggering `Rectify` (which iterates through all formations, unnecessary here), how about we just call something like `RectifyFormation(ef)` (which will maybe call `sendDiffRequests`)?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+			log.Error("error handling job request", "err", err)
+		}
+		s.sendEvent(NewEvent(EventTypeJobRequest, err, req))
+	}()
+
+	log.Info("handling job request")
+	switch req.RequestType {
+	case JobRequestTypeUp:
+		// startJob sets the HostID on the request if successful
+		// otherwise it remains blank
```
lmars: I don't understand this comment, why does a job request need `HostID`? Doesn't each request lead to `findHost` being called?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+func (s *Scheduler) RunPutJobs() {
+	log := fnLogger()
+	log.Info("starting job persistence loop")
+	strategy := attempt.Strategy{Delay: 100 * time.Millisecond, Total: time.Minute}
+	for {
+		job, ok := <-s.putJobs
+		if !ok {
+			log.Info("stopping job persistence loop")
+			return
+		}
```
lmars: This could just be:

```go
for job := range s.putJobs {
  ...
}
log.Info("stopping job persistence loop")
```

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+
+func (s *Scheduler) HandleLeaderChange(isLeader bool) {
+	log := fnLogger()
+	s.isLeader = isLeader
+	if isLeader {
+		log.Info("handling leader promotion")
+		s.triggerRectify()
+	} else {
+		log.Info("handling leader demotion")
+		// TODO: stop job restart timers
```
lmars: We should either fix this or open an issue to fix it post merge.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+}
+
+func (s *Scheduler) sendDiffRequests(f *Formation, diff Processes) {
+	log := fnLogger("app.id", f.App.ID, "release.id", f.Release.ID)
+	for typ, n := range diff {
+		if n > 0 {
+			log.Info(fmt.Sprintf("requesting %d new job(s) of type %s", n, typ))
+			for i := 0; i < n; i++ {
+				req := NewJobRequest(f, JobRequestTypeUp, typ, "", random.UUID())
+				req.state = JobStateRequesting
```
lmars: Why does a `JobRequest` need a state? It is a request to start a job, so isn't it always `JobStateRequesting`?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+				s.jobs.AddJob(req.Job)
+				s.HandleJobRequest(req)
```
lmars: Given the request is being explicitly handled immediately (rather than being sent to `s.jobRequests`), is there any need for the request? Could this not just be something like `startJob(NewJob(...))`?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+
+	s.hostStreams[h.ID()] = stream
+
+	s.triggerSyncFormations()
+
+	go func() {
+		for e := range events {
+			s.jobEvents <- e
+		}
+		// TODO: reconnect this stream unless unfollowHost was called
```
lmars: We should either fix this, or add an anti-entropy method to make sure we are following all hosts (this will get disconnected when flynn-host is updated for example). A post merge issue is fine.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+		return
+	}
+
+	log.Info("unfollowing host")
+	for jobID, job := range s.jobs {
+		if job.HostID == id {
+			log.Info("removing job", "job.id", jobID)
+			s.handleJobStop(job)
+		}
+	}
```
lmars: We should only do this if the host has been offline for a certain amount of time, so if a host temporarily goes down (e.g. during a flynn-host update), we don't go wild re-scheduling all its jobs straight away.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+
+func (s *Scheduler) startJob(req *JobRequest) (err error) {
+	log := fnLogger("job.type", req.Type)
+	log.Info("starting job", "job.restarts", req.restarts, "request.attempts", req.attempts)
+	// Copy on Write
+	s.handleJobStop(req.Job)
+	newReq := req.Clone()
+	newReq.HostID = ""
+	newReq.JobID = random.UUID()
+	newReq.state = JobStateRequesting
```
lmars: Why does it need copying?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+	defer func() {
+		if err == nil || newReq.attempts < maxJobAttempts {
```
lmars: This should probably log a message when it gives up trying to start a job

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+			return errors.New(e)
+		}
+	}
+	job = s.handleJobStop(job)
+	if job.HostID != "" {
+		log.Info("getting host client")
+		host, err := s.Host(job.HostID)
+		if err != nil {
+			log.Error("error getting host client", "err", err)
+			s.unfollowHost(job.HostID)
```
lmars: Is this necessary?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+	hosts, err := s.getHosts()
+	if err != nil {
+		log.Error("error getting host list", "err", err)
+		return nil, err
+	}
+	if len(hosts) == 0 {
+		e := "no hosts found"
+		log.Error(e)
+		return nil, errors.New(e)
+	}
```
lmars: This should be handled in `getHosts`

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+
+	if hostID != "" {
+		log.Info("using explicit host", "host.id", hostID)
+		return s.Host(hostID)
+	}
```
lmars: Under what circumstances is `startJob` provided with an explicit host ID? Should we not be picking a host each time?

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+			hostID = host.ID()
+		}
+	}
+	log.Info(fmt.Sprintf("using host with least %s jobs", typ), "host.id", hostID)
+	return s.Host(hostID)
+}
+
+func (s *Scheduler) getHosts() ([]utils.HostClient, error) {
+	hosts, err := s.Hosts()
+
```
lmars: Drop this whitespace

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Ensure that we're only following hosts that we can discover
```
lmars: Is it necessary to do this every time we call `getHosts`? We don't expect much host churn in a cluster.

controller/scheduler-v2/scheduler.go
```diff
@@ -0,0 +1,1190 @@
...
+			s.handleJobCrash(job)
+		} else {
+			s.handleJobStop(job)
+		}
+	}
+	if !s.jobs.IsJobInState(job.JobID, job.state) {
+		log := fnLogger("job.id", job.JobID, "app.id", job.AppID, "app.name", appName, "release.id", job.ReleaseID, "job.type", job.Type, "job.status", status)
+		log.Info("queuing job for persistence")
+		s.putJobs <- controllerJobFromSchedulerJob(job, jobState(status), metadata)
+		return true
```
lmars: Shouldn't this also set the job's state, given it knows it is incorrect?

test/test_scheduler.go
```diff
@@ -263,10 +264,17 @@ func (s *SchedulerSuite) TestOmniJobs(t *c.C) {
 }
 
 func (s *SchedulerSuite) TestJobRestartBackoffPolicy(t *c.C) {
+	// To run this test on local, set BACKOFF_PERIOD on the flynn host machine
```
lmars: We should set `BACKOFF_PERIOD` in `script/run-integration-tests`

controller/scheduler-v2/job.go
lmars: Or that :smile:

lmars:
@jzila lots of minor comments and a few bigger ones, I think some of them can just be opened as separate issues to fix post merge rather than blocking things.

I also think we should squash to a small number of commits (maybe even one?).

controller/jobs_test.go
jzila: I think explicitly identifying a package whose name is different from its path is a good idea.

controller/scheduler-v2/job.go
jzila: `IsScheduled` is supposed to mean that the scheduler knows about it and is handling it.

controller/scheduler-v2/job.go
jzila: Hmm. I _think_ that this shouldn't happen any more, but I'm not sure. I could remove the checks and see?

controller/scheduler-v2/job.go
jzila: I did this to remove some bugs with job IDs getting changed in a job that was in `s.jobs`. I think copy-on-write is a good idea anyway, since mutating something referenced elsewhere is risky. I can probably remove the clone since I think the path where it was needed no longer exists.

controller/scheduler-v2/scheduler.go
jzila: I wish Go had macros like `__func__`. I added this because grepping and searching for function names became arduous because of the string getting doubled up. If you think the overhead is to much, we can add the names back in, but I do think this results in much cleaner code.

controller/scheduler-v2/scheduler.go
jzila: yes

controller/scheduler-v2/scheduler.go
titanous: Please benchmark the code, I expect this is pretty expensive.

controller/scheduler-v2/scheduler.go
jzila: This is necessary. We receive formations with empty app/release from the controller all the time (maybe one-off jobs?). We receive nil formations too.

controller/scheduler-v2/scheduler.go
jzila: A `JobRequest` is really just a temporary `Job`. This is less verbose than writing `reg.Job.state`.

controller/scheduler-v2/scheduler.go
jzila: `startJob` acts on `JobRequest`s anyway. Could make it call that instead of `HandleJobRequest` though. Not sure what we'd gain by doing that...

controller/scheduler-v2/scheduler.go
jzila: We need to do this anyway to make the scheduler async.

controller/scheduler-v2/scheduler.go
jzila: Post-merge issue. Will conflict with the async refactor anyway.

controller/scheduler-v2/scheduler.go
jzila: https://github.com/flynn/flynn/pull/1818#discussion_r39888498

controller/scheduler-v2/scheduler.go
jzila: It won't be once we implement SyncHosts and correctly handle intermittent host disconnecting. I'll add a TODO.

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: post-merge issue

controller/scheduler-v2/scheduler.go
jzila: This comment is stale, from when I was manually handling state in `s.pendingStarts`

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: I'll drop the TODO.

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: I'll rename it.

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/job.go
jzila: :+1:

controller/scheduler-v2/job.go
jzila: :+1:

controller/scheduler-v2/job.go
jzila: There's no benefit to this ever being the zero value. It's only used in the scheduler for timing crashed job restarts. In the event that we miss a start event but receive a crash, it's helpful to know when this was "started" so that we don't wait a full 320 seconds for the first restart.

controller/scheduler-v2/job.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: :+1:

controller/scheduler-v2/scheduler.go
jzila: Probably not, but this will get fixed anyway once we move to async. Post-merge issue.

controller/scheduler-v2/scheduler.go
jzila: It already did. This is a retroactive check. One of the reasons why copy-on-write is nice is that you can rely on `job` not mutating while handling it.

jzila:
@lmars I'll fix the minor ones and open issues for the major ones. I responded to each comment for both me and you to communicate what I'll be fixing.

As for squashing, I'll handle that once there is no feedback left to implement. I'll squash every set of consecutive commits by one person into 1 commit, so they'll end up alternating.

controller/jobs_test.go
jzila: Additional point: `vim-go` can't follow definitions if the package name is different from its path leaf. By naming it, `vim-go` is able to resolve definitions.

controller/scheduler-v2/job.go
jzila: I've changed it to `IsStopped` and reversed all the conditions of each call to `IsScheduled`

controller/scheduler-v2/job.go
jzila: I've moved the checks into its own function.

controller/scheduler-v2/scheduler.go
jzila: `HandleJobRequest` also checks whether the current scheduler is the leader, and I'd like to minimize the number of places I have to put that check.

controller/scheduler-v2/scheduler.go
jzila: I stand corrected. Looks like none of the recent CI builds have this error any more since you implemented the formation sentinel.

test/test_scheduler.go
jzila: :+1:

lmars:
@jzila latest fixes merged, please open issues for the points we are not handling in this PR.

I will run run the tests a few times and comment with any failures.

controller/scheduler/scheduler.go
```diff
@@ -0,0 +1,1216 @@
...
+	} else {
+		var ok bool
+		job, ok = s.jobs[req.JobID]
+		if !ok {
+			e := "unknown job"
+			log.Error(e)
+			return errors.New(e)
+		}
+	}
+	s.jobs.SetState(job.JobID, JobStateStopped)
```
lmars: Setting the job's state to `JobStateStopped` here means that when we get a job event indicating the job is down, we don't persist it to the controller because the job is already in the stopped state, so it seems only `SyncJobs` will cause stopped events to be persisted.

jzila: OK I'll introduce a new state: `PendingStop`

controller/scheduler/scheduler.go
```diff
@@ -0,0 +1,1216 @@
...
+	switch status {
+	case host.StatusStarting:
+		s.handleJobEvent(job, JobStateStarting)
+	case host.StatusRunning:
+		s.handleJobEvent(job, JobStateRunning)
+	default:
+		if job.IsStopped() {
+			s.handleJobEvent(job, JobStateStopped)
+		} else {
+			s.handleJobCrash(job)
```
lmars: We need to check whether the job should actually be running before we treat it as crashed and restart it (i.e. by checking the formation), because another scheduler may have stopped the job just before we became leader.

PR #1822, opened 2015-09-04
all: Use "official" busybox image

titanous:
It is kept up-to-date, ours is unmaintained.


titanous:
Looks like flannel is currently not a static binary, I'll land go1.5 before this, as it includes some fixes that will make building flannel easier.

PR #1833, opened 2015-09-05
controller: Move hstore to jsonb

josephglanville:
This removes usage of hstore as it's overly complicated for our usecase and provides negligible performance benefit. It also makes migration to pgx easier as we would have needed to port to pgx hstore API.


controller/app.go
```diff
@@ -108,11 +110,8 @@ func scanApp(s postgres.Scanner) (*ct.App, error) {
 	if releaseID != nil {
 		app.ReleaseID = *releaseID
 	}
-	if len(meta.Map) > 0 {
-		app.Meta = make(map[string]string, len(meta.Map))
-		for k, v := range meta.Map {
-			app.Meta[k] = v.String
-		}
+	if len(meta) > 0 {
+		err = json.Unmarshal(meta, &app.Meta)
```
titanous: This nukes the previous error, add a `err != nil` check above.

controller/deployment.go
```diff
@@ -151,13 +149,7 @@ func scanDeployment(s postgres.Scanner) (*ct.Deployment, error) {
...
 		d.Status = *status
 	}
-	d.Processes = make(map[string]int, len(procs.Map))
-	for k, v := range procs.Map {
-		n, _ := strconv.Atoi(v.String)
-		if n > 0 {
-			d.Processes[k] = n
-		}
-	}
+	err = json.Unmarshal(procs, &d.Processes)
```
titanous: Ditto on the `err` overwrite.

controller/resource.go
```diff
@@ -87,15 +83,14 @@ func split(s string, sep string) []string {
...
 	var appIDs string
 	err := s.Scan(&r.ID, &r.ProviderID, &r.ExternalID, &env, &appIDs, &r.CreatedAt)
 	if err == sql.ErrNoRows {
 		err = ErrNotFound
 	}
-	r.Env = make(map[string]string, len(env.Map))
-	for k, v := range env.Map {
-		r.Env[k] = v.String
+	if len(env) > 0 {
+		err = json.Unmarshal(env, &r.Env)
```
titanous: Another `err` bites the dust.

josephglanville:
err clobbering fixed up.

controller/app.go
```diff
@@ -99,20 +101,17 @@ func (r *AppRepo) Add(data interface{}) error {
 
 func scanApp(s postgres.Scanner) (*ct.App, error) {
 	app := &ct.App{}
-	var meta hstore.Hstore
+	var meta []byte
 	var releaseID *string
 	err := s.Scan(&app.ID, &app.Name, &meta, &app.Strategy, &releaseID, &app.CreatedAt, &app.UpdatedAt)
 	if err == sql.ErrNoRows {
-		err = ErrNotFound
+		return nil, ErrNotFound
```
titanous: This isn't the only error that can be returned from `Scan`, probably should add an `else if err != nil`

titanous:
LGTM.

PR #1838, opened 2015-09-08
Fixed some test failures and stopped jobs

jzila:


controller/scheduler-v2/scheduler.go
```diff
@@ -339,7 +337,13 @@ func (s *Scheduler) HandleFormationChange(ef *ct.ExpandedFormation) {
...
 	}()
 
-	log := logger.New("fn", "HandleFormationChange", "app.id", ef.App.ID, "release.id", ef.Release.ID)
+	log := logger.New("fn", "HandleFormationChange")
+	if ef.App != nil {
+		log = log.New("app.id", ef.App.ID)
+	}
+	if ef.Release != nil {
+		log = log.New("release.id", ef.Release.ID)
+	}
```
lmars: These should always be set, and if they aren't, whatever calls `HandleFormationChange` should drop them (I suspect it is the sentinel value)

jzila: HandleFormationChange is being called in the scheduler loop, which is supposed to do nothing except pass through the received structs.

lmars: Ok, we should explicitly ignore the sentinel value here then.

jzila: In `s.changeFormation` we do the same check as well. What is the sentinel value?

lmars: When `StreamFormations` is called, it first sends all the formations updated since the `since` argument, so if you get disconnected, you can reconnect and get the missed changes. To tell when you are up-to-date, a sentinel value is sent (i.e. [here](https://github.com/flynn/flynn/blob/master/controller/formation.go#L285))

PR #1841, opened 2015-09-08
Added vagrant user to libvirtd group so it has permissions to kill discoverd

jzila:

### DIFF
--- a/util/packer/ubuntu-14.04/scripts/install.sh
+++ b/util/packer/ubuntu-14.04/scripts/install.sh
@@ -138,7 +138,8 @@ enable_cgroups() {
 create_groups() {
   groupadd docker
   groupadd fuse || true
-  usermod -a -G docker,fuse "${SUDO_USER}"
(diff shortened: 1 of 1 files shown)


lmars:
We should do this in the [packer install script](https://github.com/flynn/flynn/blob/master/util/packer/ubuntu-14.04/scripts/install.sh#L138-L142).

util/packer/ubuntu-14.04/scripts/install.sh
```diff
@@ -138,7 +138,7 @@ enable_cgroups() {
 create_groups() {
   groupadd docker
   groupadd fuse || true
-  usermod -a -G docker,fuse "${SUDO_USER}"
+  usermod -a -G docker,fuse,libvirtd "${SUDO_USER}"
```
lmars: This runs before installing libvirt, so we should probably create the group first like the other ones.

jzila:
comments addressed

PR #1842, opened 2015-09-09
Scheduler V2 Bugfixes

jzila:


controller/scheduler-v2/scheduler.go
```diff
@@ -712,7 +730,7 @@ func (s *Scheduler) stopJob(req *JobRequest) (err error) {
 		return err
 	}
 
-	log.Info("requesting host to stop job")
+	log.Info("requesting host to stop job", "job.id", job.JobID)
```
lmars: `job.id` is already in the log

jzila: hmm, don't know how it was missing in the log output...

controller/scheduler-v2/scheduler.go
```diff
@@ -446,11 +453,9 @@ func (s *Scheduler) HandleFormationChange(ef *ct.ExpandedFormation) {
...
 	log := logger.New("fn", "HandleFormationChange")
-	if ef.App != nil {
-		log = log.New("app.id", ef.App.ID)
-	}
-	if ef.Release != nil {
-		log = log.New("release.id", ef.Release.ID)
+	if ef == nil || ef.App == nil || ef.Release == nil {
+		log.Error("invalid formation")
+		return
 	}
```
lmars: This should no longer be necessary as the sentinel is [explicitly ignored in streamFormationEvents](https://github.com/flynn/flynn/blob/scheduler-v2-review/controller/scheduler-v2/scheduler.go#L182-L185), so if we are getting a nil formation it's a bug.

controller/scheduler-v2/scheduler.go
```diff
@@ -461,7 +466,7 @@ func (s *Scheduler) HandleFormationChange(ef *ct.ExpandedFormation) {
 
 	// Trigger sync jobs in case we've ignored an existing job because we
 	// didn't know about the formation.
-	s.triggerSyncJobs()
+	s.triggerRectify()
```
lmars: The comment above no longer seems necessary

controller/scheduler-v2/scheduler.go
jzila: I still feel like having it is better than not having it, so it can at least appear in our logs instead of panicking.

lmars: :+1:

jzila:
comments addresssed

Issue #1340, opened 2015-03-23
cap exponential backoff period


aep:
backoff period should be capped, so that it doesn't become infinitely long, when the service has been failing for a long time.

It might fail due to external issues, and re-deploying is not the correct answer.

I suggest a default of 10 * BACKOFF_PERIOD and making it configurable through the env.

PR #1843, opened 2015-09-09
Scheduler V2 Bugfixes

jzila:


jzila:
The ControllerSuite and SchedulerSuite tests run successfully on my dev machine; not sure what's going on in CI.

jzila:
Starting to look a lot better: https://ci.flynn.io/builds/20150909070756-14914cf0

controller/scheduler-v2/scheduler.go
```diff
@@ -498,7 +498,7 @@ func (s *Scheduler) HandleJobRequest(req *JobRequest) {
 func (s *Scheduler) RunPutJobs() {
 	log := logger.New("fn", "RunPutJobs")
 	log.Info("starting job persistence loop")
-	strategy := attempt.Strategy{Delay: 100 * time.Millisecond, Total: time.Minute}
+	strategy := attempt.Strategy{Delay: 100 * time.Millisecond, Total: time.Second}
```
lmars: The controller client [uses a retry dialer](https://github.com/flynn/flynn/blob/scheduler-v2-review/controller/client/client.go#L134) which use a [1s connect timeout and 30s retry timeout](https://github.com/flynn/flynn/blob/scheduler-v2-review/pkg/dialer/retry.go#L10-L20), so having a 1s strategy here probably doesn't make sense.

It is actually quite important for these calls to eventually succeed (see #686), so maybe we need a pool of goroutines which are calling `PutJob`? We need to make sure they are not delivered out of order for any given job though (see [this DB constraint](https://github.com/flynn/flynn/blob/master/controller/schema.go#L132-L140)).

controller/scheduler-v2/job.go
```diff
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	ct "github.com/flynn/flynn/controller/types"
+	host "github.com/flynn/flynn/host/types"
```
lmars: This package is already called `host`

controller/scheduler-v2/scheduler.go
```diff
@@ -913,18 +919,18 @@ func (s *Scheduler) Unsubscribe(events chan Event) {
...
 		fallthrough
 	case host.StatusRunning:
 		s.handleJobStart(job)
-		controllerState = "up"
 	default:
 		delete(s.jobs, job.JobID)
 		delete(s.stoppedJobs, job.JobID)
 	}
-	s.putJobs <- controllerJobFromSchedulerJob(job, controllerState, metadata)
+	log.Info("Queuing job for persistence")
```
lmars: Log messages should not be capitalised

jzila:
@lmars please disregard the last commmit: I figured out a better approach to achieve the same goal.

controller/scheduler-v2/scheduler.go
```diff
@@ -18,25 +20,40 @@ import (
...
-	serviceName         = "controller-scheduler"
+	eventBufferSize    int           = 1000
+	serviceName                      = "controller-scheduler"
+	maxJobAttempts     uint          = 30
+	jobAttemptInterval time.Duration = 500 * time.Millisecond
 )
 
 var logger = log15.New("component", "scheduler")
 
+func fnLogger(ctx ...interface{}) log15.Logger {
```
jzila: I added this to simplify searching (both in the editor and grep). This way function names only show up once.

controller/scheduler-v2/scheduler.go
```diff
@@ -84,10 +101,12 @@ func NewScheduler(cluster utils.ClusterClient, cc utils.ControllerClient) *Sched
 		jobRequests:      make(chan *JobRequest, eventBufferSize),
 		putJobs:          make(chan *ct.Job, eventBufferSize),
 	}
+	s.maxBackoffPeriod = 10 * s.backoffPeriod
```
jzila: Fix for #1340

controller/scheduler-v2/scheduler.go
```diff
@@ -736,26 +763,51 @@ func (s *Scheduler) updateFormation(f *ct.Formation) (*Formation, error) {
...
+	newReq := &JobRequest{
+		RequestType: req.RequestType,
+		Job:         &job,
+	}
 	defer func() {
+		s.pendingStarts.AddJob(newReq.Job)
 		if err != nil {
-			s.scheduleJobRequest(req)
 			log.Error("error starting job", "err", err)
+			if newReq.attempts < maxJobAttempts {
```
jzila: Limit number of times a failed job request can try to start. Rely on `Rectify` to handle the rest.

controller/scheduler-v2/scheduler.go
```diff
@@ -936,49 +995,73 @@ func (s *Scheduler) Jobs() map[string]*Job {
...
 	log.Info("adding job to in-memory state")
 	_, ok := s.jobs[job.JobID]
 	if s.isLeader && !ok && s.pendingStarts.HasStarts(job) {
 		s.pendingStarts.RemoveJob(job)
 	}
 	s.jobs[job.JobID] = job
+	delete(s.stoppedJobs, job.JobID)
+}
+
+func (s *Scheduler) handleJobCrash(job *Job) {
```
jzila: Implemented restarting of crashed jobs.

controller/scheduler-v2/scheduler_test.go
```diff
@@ -282,13 +282,13 @@ func (TestSuite) TestExponentialBackoffNoHosts(c *C) {
 			c.Fatalf("expected JobRequestEvent, got %T", event)
 		}
 		c.Assert(e.Request, NotNil)
-		c.Assert(e.Request.restarts, Equals, expected)
 	}
 
 	// wait for the formation to cascade to the scheduler
 	s.waitRectify()
-	waitRestarts(1*time.Second+50*time.Millisecond, 2)
-	waitRestarts(2*time.Second+50*time.Millisecond, 3)
```
jzila: Switch to linear restart attempts for job requests (exponential for crashed job restarts)

controller/scheduler-v2/scheduler.go
```diff
@@ -936,49 +995,73 @@ func (s *Scheduler) Jobs() map[string]*Job {
...
+	log := fnLogger()
+	if req.startedAt.Before(time.Now().Add(-s.backoffPeriod)) {
+		log.Info("resetting job restarts", "backoffPeriod", s.backoffPeriod, "job.startedAt", req.startedAt)
+		req.restarts = 0
 	}
-}
-
-func (s *Scheduler) scheduleJobRequest(req *JobRequest) {
 	backoff := s.getBackoffDuration(req.restarts)
+	if backoff == s.maxBackoffPeriod {
```
jzila: Don't use exponential backoff past the max period (`10 * s.backoffPeriod`). After that, the attempt will fail and we'll have to rely on `Rectify`.

lmars:
@jzila the unit tests are failing with:

```
controller/scheduler/main.go:939: too many arguments in call to utils.JobConfig
```

jzila:
Ah, sry about that. I wonder why the incremental `make` on my machine didn't pick that up...

lmars:
@jzila we are no longer building scheduler v1 on this branch, we should probably just delete it

jzila:
Looks like when a release is added to the `controller` app, the scheduler correctly stops itself upon which the controller starts a new one. However, we don't notify the controller with a final `PutJob`. In fact, I'm unclear as to how the controller is starting a new scheduler given that it isn't notified about it via PutJob until the next scheduler comes up.

Maybe we should add in a `SIGTERM` handler, upon which we write to the `s.stop` channel instead of terminating right away.

lmars:
@jzila 

> when a release is added to the controller app, the scheduler correctly stops itself upon which the controller starts a new one

This is incorrect. A new controller release leads to a one-by-one deployment, so the deployer incrementally scales up the new release and scales down the old release. Only the schedulers are actually starting / stopping jobs, and when the current leader is scaled down, one of the new schedulers should take the reign and continue reacting to the scaling being done by the deployer.

> we don't notify the controller with a final PutJob

True, but the other schedulers should see the leader go down and call PutJob themselves. This actually highlights the fact that when the scheduler starts, it needs to wait until it is watching all the hosts *before* registering with discoverd, so the deployer doesn't scale things too fast and lead to no schedulers watching all hosts.

> Maybe we should add in a SIGTERM handler, upon which we write to the s.stop channel instead of terminating right away.

Makes sense, we can do this easily using `shutdown.BeforeExit`.

lmars:
@jzila the latest build is showing ~100 schedulers running on host `a0a72333` :scream:

I am looking into it, and I may make some code changes. If I do, I will open a PR with my changes on top of yours (I can't push to this PR).

lmars:
@jzila I am going to merge this then I have a few fixes I will PR separately

PR #1845, opened 2015-09-09
discoverd: fix proxy peer port

benbjohnson:
## Overview

This pull request updates the discoverd DNS store proxy to use the HTTP port instead of the Raft port.


titanous:
LGTM

PR #1848, opened 2015-09-09
test: Move host removal wait into test/helper.go

lmars:
This makes the error more visible, and means the same discoverd client is being used to for both adding and removing hosts.


titanous:
LGTM

PR #1849, opened 2015-09-10
router: Add logging to main function

lmars:
This is to help diagnose intermittent `TestOmniJobs` failures, but is also useful in general.


titanous:
LGTM

PR #1850, opened 2015-09-10
scheduler: Simplify main loop select

lmars:
There doesn't seem to be a reason to prioritise certain channels, and this removes the need for a sleep in the loop, as it will block on all channels if there is nothing to do.

/cc @jzila

### DIFF
--- a/controller/Tupfile
+++ b/controller/Tupfile
@@ -1,6 +1,6 @@
 include_rules
 : |> !go |> bin/flynn-controller
-: |> !go ./scheduler |> bin/flynn-scheduler
+: |> !go ./scheduler-v2 |> bin/flynn-scheduler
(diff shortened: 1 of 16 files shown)


lmars:
@jzila lots of failures in the latest build. If there is an explicit need to prioritise the channels, it may be worth making it clearer why in the comments. Happy to drop this change if the next build fails as badly.

jzila:
@lmars yeah I was about to say. I'll add comments to explain why the priority is necessary.

jzila:
Ultimately it comes down to this: if the scheduler decides to rectify or handle job requests before handling events from the hosts/cluster/controller, then it can cause behavior that thrashes (e.g. creating extra jobs, deleting extra jobs, etc).

lmars:
Closing as this was fixed in #1843

PR #1858, opened 2015-09-11
test: Expose list of failures in CI

lmars:
This turns the "failure" label in CI into a button which once clicked shows a list of the failures from the build.


titanous:
Neat! LGTM.

PR #1864, opened 2015-09-13
Scheduler V2 Fixes

lmars:
The main change here is making sure the scheduler is in sync with hosts and formations before starting the main loop.

I have also changed the startup so that the scheduler only registers in service discovery once it is in sync
with the hosts so that a deployment doesn't kill old schedulers before the new schedulers are watching events, which could lead to events being missed. This meant doing discoverd related things in `Run`, so I moved the functionality into an interface to avoid running a discoverd server in the tests.

### DIFF
--- a/controller/scheduler-v2/scheduler.go
+++ b/controller/scheduler-v2/scheduler.go
@@ -562,14 +562,12 @@ func (s *Scheduler) sendDiffRequests(f *Formation, diff Processes) {
 }
 
 func (s *Scheduler) followHost(h utils.HostClient) {
-	_, ok := s.hostStreams[h.ID()]
-	if ok {
+	if _, ok := s.hostStreams[h.ID()]; ok {
 		return
 	}
 
 	log := fnLogger("host.id", h.ID())
 	log.Info("streaming job events")
-	// TODO: reconnect this stream, stopping only if unfollowHost is called
 	events := make(chan *host.Event)
 	stream, err := h.StreamEvents("all", events)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


controller/scheduler-v2/scheduler.go
```diff
@@ -191,15 +139,29 @@ func (s *Scheduler) streamFormationEvents() error {
...
 	go func() {
+		var isCurrent bool
 	outer:
 		for {
 			for formation := range events {
+				// an empty formation indicates we now have the
+				// current list of formations.
 				if formation.App == nil {
-					// sentinel
+					if !isCurrent {
```
jzila: Why is this necessary for the formation list? The scheduler won't try to stop anything it doesn't have a formation for.

jzila: nvmind, probably better to have this than not.

controller/scheduler-v2/scheduler.go
```diff
@@ -568,33 +554,41 @@ func (s *Scheduler) sendDiffRequests(f *Formation, diff Processes) {
...
+	log.Info("getting active jobs")
+	jobs, err := h.ListJobs()
+	if err != nil {
+		log.Error("error getting active jobs", "err", err)
+		return
+	}
+	log.Info(fmt.Sprintf("got %d active job(s) for host %s", len(jobs), h.ID()))
+
+	for _, job := range jobs {
+		s.handleActiveJob(&job)
```
jzila: This will happen already as part of StreamEvents.

jzila:
LGTM other than my comment on calling `handleActiveJob` in `followHost`

controller/scheduler-v2/scheduler.go
lmars: `StreamEvents` does not return the current state like other streams do, so this makes sure we have the current state when we start following a host (rather than relying on the next trigger of `SyncJobs`).

controller/scheduler-v2/scheduler.go
lmars: I think ensuring the scheduler is in sync before taking (or not taking) any action is a good idea.

PR #1866, opened 2015-09-13
logaggregator improvements

titanous:
This repackages the useful refactorings from the replication branch, minus the replication code, which will not be used (I'm working on an alternative approach to HA).


titanous:
Superceded by #1966

PR #1867, opened 2015-09-13
scheduler: Implemented PR fixes

jzila:

### DIFF
--- a/controller/scheduler-v2/scheduler.go
+++ b/controller/scheduler-v2/scheduler.go
@@ -52,8 +52,7 @@ type Scheduler struct {
 	discoverd Discoverd
 	isLeader  bool
 
-	backoffPeriod    time.Duration
-	maxBackoffPeriod time.Duration
+	backoffPeriod time.Duration
 
 	formations  Formations
 	hostStreams map[string]stream.Stream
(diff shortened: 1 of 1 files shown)


controller/scheduler-v2/scheduler.go
```diff
@@ -96,7 +96,7 @@ func NewScheduler(cluster utils.ClusterClient, cc utils.ControllerClient, disc D
 		jobRequests:      make(chan *JobRequest, eventBufferSize),
 		putJobs:          make(chan *ct.Job, eventBufferSize),
 	}
-	s.maxBackoffPeriod = 10 * s.backoffPeriod
+	s.maxBackoffPeriod = 32 * s.backoffPeriod
```
lmars: Looking a bit closer, we already have an `if restarts > 30` check in `getBackoffDuration`, so we might as well just unify the two restrictions, e.g.:

```go
func (s *Scheduler) getBackoffDuration(restarts uint) time.Duration {
  multiplier := 32 // max multiplier
  if restarts < 6 {
    multiplier := (1<<restarts)>>1
  }
  return s.backoffPeriod * time.Duration(multiplier)
}
```

Issue #1696, opened 2015-08-07
Dashboard: blank screen/no login token after initially importing ephemeral certificate


nobert:
In Chrome (and possibly other browsers) on OSX 10.10 (and possibly other versions), after running `make dashboard` for the first time and importing+trusting the ephemeral certificate, the user is thrown to a blank screen. Reloading the page gives the user the Login token prompt.

lmars:
This just happened to me:

* Chrome 44.0.2403.155 (64-bit)
* OSX 10.10.3
* flynn-host dev build (commit 9d2ad95)

lmars:
I should add that I did not provide the login token in the URL due to #1722 (and the fact that my dashboard is publicly accessible).

PR #1878, opened 2015-09-16
dashboard: Fix issue causing blank page instead of login view

jvatic:
This fixes cases where a blank page is rendered instead of the login view.

#1696


lmars:
LGTM

Issue #1881, opened 2015-09-16
Dashboard: Add service flow fails after github authentication


nobert:
Browser: Chrome stable
OS: OSX 10.10.5

After initial login and github authentication flow, the dashboard immediately aborts when launching an app, before the progress frame opens. The app does launch successfully, but the dashboard flow will always abort. Any other actions (e.g. delete app) will also abort. Everything works inside of a new tab, which indicates that a state has failed somehow.

jvatic:
Steps to reproduce:

1) Ensure session is unauthenticated (logged out of dashboard)

2) Login

3) Without reloading the page, attempt to deploy an app

It's due to the events endpoint trying to connect while unauthenticated and not reconnecting after authenticating.

PR #1883, opened 2015-09-16
dashboard: Ensure event stream is connected

jvatic:
Fixes issue where event stream would try to connected before app is authenticated and didn't retry after authenticating.

#1881


titanous:
LGTM

PR #1891, opened 2015-09-19
Support running Flynn clusters on a single machine

lmars:
This is made possible by creating virtual network interfaces for each node and binding all host network services (i.e. flynn-host, discoverd, flannel and router) to those interfaces. 

My main motivator here is being able to test updating a discoverd cluster on my local machine, but this is also pretty useful in general.

Boot a 5 node cluster:

```
$ script/bootstrap-flynn --size 5
```

or run integration tests against a 5 node cluster:

```
$ script/run-integration-tests --size 5
```

It works fine for me on an EC2 instance, but I would appreciate other people testing this out in their own environments.

/cc @titanous @jvatic @jzila @benbjohnson

### DIFF
--- a/bootstrap/manifest_template.json
+++ b/bootstrap/manifest_template.json
@@ -12,7 +12,8 @@
     "action": "run-app",
     "release": {
       "env": {
-        "DISCOVERD": "none"
+        "DISCOVERD": "none",
+        "BACKEND": "{{ getenv \"FLANNEL_BACKEND\" }}"
       },
       "processes": {
         "app": {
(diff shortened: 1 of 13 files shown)


lmars:
Oh I forgot to mention, the only downside here is that when you boot a multi-node cluster, `flynn-host` commands won't work by default (as they try to connect to discoverd at `127.0.0.1`), so you need to:

```
export DISCOVERD="198.51.100.1:1111"
```

titanous:
:balloon: :tada: :heart_eyes:

jzila:
@lmars could you put the export into the Vagrantfile?

jvatic:
:+1:

lmars:
@jzila the only issue with adding it to the Vagrantfile is that booting a single node cluster does not create the virtual interfaces, instead listening on all interfaces (so it is accessible from the host in case of Vagrant).

We could also start single node clusters using a virtual interface and always access at `198.51.100.1`, but then you can't access it from the host (well, not without doing more work like using iptables).

PR #1894, opened 2015-09-20
Update discoverd

lmars:
Currently a WIP to get discoverd deployments working.

The strategy I have taken is as follows:

* Bind the raft address to an allocated port (forcing to `1110` in bootstrap) so that multiple processes can participate in raft on the same host
* Only set `DISCOVERD_PEERS` at bootstrap, instead looking up peers using the `discoverd-raft` service if the cluster is already running
* Call `AddPeer` if the advertise address is not a current peer, but another peer is running on the same host (so assume we are going to replace it)
* Use `SO_REUSEPORT` for the HTTP  / DNS listeners so multiple processes can handle requests

The deployments are not currently working as the new `discoverd-raft` services don't seem to appear, even though the jobs are running.

@benbjohnson I would appreciate any thoughts on this workflow before I start digging into the raft code.


benbjohnson:
> Use `SO_REUSEPORT` for the HTTP / DNS listeners so multiple processes can handle requests.

What's the benefit of using `SO_REUSEPORT` in this case? Why would two `discoverd` processes share the same port?

lmars:
@benbjohnson 

> What's the benefit of using SO_REUSEPORT in this case? Why would two discoverd processes share the same port?

In general when doing a deployment, we start up the new processes before killing the old ones, so there will be a period when there are two discoverd processes on a single host. We can't use different ports for the APIs as all the jobs are pointing at ports `1111` and `53`.

We could have a different deployment strategy which kills current processes before starting new ones, but that would cause API errors, and we want this process to have as little impact as possible on running apps.

benbjohnson:
It seems like `SO_REUSEPORT` could still drop the connections on the closing process. Would something like [`goagain`](https://github.com/rcrowley/goagain) be a better choice?

lmars:
@benbjohnson 

> It seems like SO_REUSEPORT could still drop the connections on the closing process

DNS requests are the main thing I am worried about because that is the most likely API apps will use to discover services, so being UDP this shouldn't be an issue.

As for HTTP requests, yes there is a chance that connections will close, but we will recommend apps use using client libraries with reconnect logic to handle this.

> Would something like goagain be a better choice?

Using fd sharing like goagain (which by the way is what we are doing for the `flynn-host` update, see #1672) would mean the new discoverd would have to be a child of the current discoverd, so we would have to inject the new binary / supporting files from the new image into the current image, which is much more complicated than just running the new image and binding on the same port.

This also has the same connection drop issues (see [this discussion](https://github.com/flynn/flynn/pull/1672/files#diff-16dbe4698fb02ae8788cf368f61d0c00R374)).

lmars:
@benbjohnson FYI I am having issues tearing down the old processes (see https://github.com/hashicorp/raft/issues/62), let me know if you have any thoughts.

benbjohnson:
@lmars We deferred membership changes in `discoverd` in order to get something deployed. Right now I think it's limited to a fixed peer set because it uses `SetPeers()` instead of using `AddPeer()`. I'm not sure that a node can currently be removed from the membership until we change that.

lmars:
@benbjohnson 

> Right now I think it's limited to a fixed peer set because it uses SetPeers() instead of using AddPeer()

Apologies I should've give more background as to where I am up to with this.

I have updated things so that each node registers a `discoverd-raft` service (either [using the existing cluster](https://github.com/flynn/flynn/blob/update-discoverd/discoverd/main.go#L107-L112) if it exists, or [once everything is running](https://github.com/flynn/flynn/blob/update-discoverd/discoverd/main.go#L209-L213) for initial peers), and the leader will add those as peers when they come up ([see this WIP code](https://github.com/flynn/flynn/blob/update-discoverd/discoverd/main.go#L116-L137)).

Adding peers works fine, but as soon as the leader [unregisters itself](https://github.com/flynn/flynn/blob/update-discoverd/discoverd/main.go#L217-L221) (which leads to [RemovePeer](https://github.com/flynn/flynn/blob/update-discoverd/discoverd/main.go#L128-L134) getting called), the cluster does not recover.

I plan to refactor this code but I have pushed just to give a sense of the direction I am heading in.

titanous:
@lmars I think we need to be careful about dynamic cluster membership, I'm not convinced that using discoverd to manage discoverd peers is a good idea. We should start with the explicit membership in the `DISCOVERD_PEERS` environment variable and have an explicit API call (called via a `flynn-host` command) to add a node to the consensus cluster.

All that being said, gracefully stepping down during an update seems like a good idea.

lmars:
@titanous 

> We should start with the explicit membership in the DISCOVERD_PEERS environment variable and have an explicit API call (called via a flynn-host command) to add a node to the consensus cluster.

I agree in terms of adding extra hosts to a cluster, but here I am working on replacing existing discoverd processes with updated ones during a deployment, and trying to make this transparent to the rest of the cluster, in which case I don't think `flynn-host` needs to be involved?

titanous:
@lmars Agreed.

lmars:
Closing as I am going to use a different approach which doesn't involve calling `AddPeer` / `RemovePeer`

PR #1895, opened 2015-09-20
Implemented PR feedback from general Scheduler V2 Review

jzila:
Implemented all the minor feedback. Additionally, factored out formation rectification
so only a minimally needed amount of work occurs. This is to address https://github.com/flynn/flynn/pull/1818#discussion_r39878345.


controller/scheduler/discoverd.go
```diff
@@ -59,6 +59,11 @@ func (d *discoverdWrapper) Register() (bool, error) {
 	outer:
 		for {
 			for leader := range leaders {
+				if leader == nil {
+					// This should never happen, but it has happened, so a nil check is necessary
+					log.Error("Received nil leader event")
```
lmars: downcase the `R`

controller/scheduler/scheduler.go
```diff
@@ -722,51 +734,64 @@ func (s *Scheduler) changeFormation(ef *ct.ExpandedFormation) (*Formation, error
...
 			f.Processes = ef.Processes
 		}
 	}
+	s.triggerRectify(f.key())
 	return f, nil
 }
 
+func (s *Scheduler) triggerRectify(key utils.FormationKey) {
+	logger.Info("triggering rectify", "key", key)
+	s.formations.TriggerRectify(key)
```
lmars: Does this need to go via `s.formations` and have all the generated case statements? Can we not make `s.rectify` a `chan *Formation`, with `nil` meaning rectify all formations?

controller/scheduler/scheduler.go
```diff
@@ -722,51 +734,64 @@ func (s *Scheduler) changeFormation(ef *ct.ExpandedFormation) (*Formation, error
...
+}
+
 func (s *Scheduler) updateFormation(f *ct.Formation) (*Formation, error) {
 	ef, err := utils.ExpandFormation(s, f)
 	if err != nil {
 		return nil, err
 	}
-	return s.changeFormation(ef)
+	form, err := s.changeFormation(ef)
+	return form, err
```
lmars: Is this change necessary?

controller/scheduler/scheduler.go
jzila: no, I want the advantage of being able to rectify each formation once or not at all.

jzila: That is, s.rectify responding to `chan *Formation` has a chance of rectifying multiple times per formation for no reason...thanks to the priority, all events leading up to the "first" rectify will already have happened, so the "first" rectify will be current.

controller/scheduler/scheduler.go
jzila: No, I originally had it returning more parameters but now it's the same.

controller/scheduler/scheduler.go
lmars: Ok so I am not really a fan of the reflection and a channel per formation just to avoid the chance of a loop iteration triggering the same rectify multiple times, as the duplicates will effectively be no-ops anyway right? The loop should be fast enough that the chance of having duplicates is low. WDYT?

lmars: Thinking about it, another option is to have a buffer wrapper around the rectify channel, and have `s.rectify` receive a "batch" of formations, e.g.:

```go
batch := map[utils.FormationKey]struct{}
var out chan map[utils.FormationKey]struct{}
for {
  select {
  case key := <-in:
    // add key to batch
    batch[key] = struct{}{}
    // switch `out` so we try and send the batch on next loop
    out = s.rectify
  case out <- batch:
    // batch sent, reset
    batch = map[utils.FormationKey]struct{}
    out = nil
  }
}
```

jzila: YES! I like this.

jzila: Well, I'd do it differently:

```go
func (s *Scheduler) triggerRectify(key utils.FormationKey) {
    s.rectifyBatch[key] = struct{}{}
    select {
        case s.rectify <- struct{}{}:
        default:
    }
}

func (s *Scheduler) HandleRectify() {
    for key := range s.rectifyBatch {
        s.rectifyFormation(key)
    }
}
```

PR #1898, opened 2015-09-22
dashboard: List app history (release and scale)

jvatic:
The next step is adding a revert button to each item.

### DIFF
--- a/controller/client/client.go
+++ b/controller/client/client.go
@@ -528,10 +528,7 @@ type StreamEventsOptions struct {
 }
 
 func (c *Client) StreamEvents(opts StreamEventsOptions, output chan *ct.Event) (stream.Stream, error) {
-	path, err := url.Parse("/events")
-	if err != nil {
-		return nil, err
-	}
+	path, _ := url.Parse("/events")
 	q := path.Query()
 	if opts.AppID != "" {
 		q.Set("app_id", opts.AppID)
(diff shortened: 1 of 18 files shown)


jvatic:
There's a few intermittently failing tests due to failed port allocation, e.g. `router: tcp sync error: error binding to port (check if another service is listening on 127.0.0.1:45000): listen tcp4 127.0.0.1:45000: bind: address already in use`.

jvatic:
ping @titanous for review

titanous:
This is great! I think the list should have similar styling to the source history and icons for each event type.

jvatic:
Added icons and more detail to scale events (now shows the delta).

controller/formation.go
```diff
@@ -74,6 +74,14 @@ func (r *FormationRepo) Add(f *ct.Formation) error {
 	if err != nil {
 		return err
 	}
+	scale := &ct.Scale{
+		Processes: f.Processes,
+		ReleaseID: f.ReleaseID,
+	}
+	prevFormation, _ := scanFormation(tx.QueryRow("SELECT app_id, release_id, processes, created_at, updated_at FROM formations WHERE app_id = $1 AND release_id = $2", f.AppID, f.ReleaseID))
```
jvatic: This should use `r.Get(appID, releaseID)`

controller/formation.go
```diff
@@ -143,6 +151,13 @@ func (r *FormationRepo) Remove(appID, releaseID string) error {
 	if err != nil {
 		return err
 	}
+	scale := &ct.Scale{
+		ReleaseID: releaseID,
+	}
+	prevFormation, _ := scanFormation(tx.QueryRow("SELECT app_id, release_id, processes, created_at, updated_at FROM formations WHERE app_id = $1 AND release_id = $2", appID, releaseID))
```
jvatic: Ditto

controller/client/client.go
```diff
@@ -548,10 +549,62 @@ func (c *Client) StreamEvents(opts StreamEventsOptions, output chan *ct.Event) (
...
+	SinceID     *int64
+	Count       int
+}
+
+func (c *Client) ListEvents(opts ListEventsOptions) ([]*ct.Event, error) {
+	var events []*ct.Event
+	path, err := url.Parse("/events")
+	if err != nil {
+		return nil, err
+	}
```
titanous: This should never error.

titanous:
- [ ] The UI should not show the scale events associated with deploying a new release (currently there is a down+up shown)
- [ ] The log should scroll to the top when a new event is added

jvatic:
- [ ] Fix bug causing scale UI to disappear when saving ENV

jvatic:
@titanous comments addressed

titanous:
Looks like there is still an event that is not getting hidden:

1. Scale app to 3 instances.
1. Set environment variable.
1. Note that there is an extra `web: 3` event before the new release event and after the scale up event.

controller/types/types.go
```diff
@@ -217,6 +217,17 @@ type Event struct {
 	CreatedAt  *time.Time      `json:"created_at,omitempty"`
 }
 
+type Scale struct {
+	PrevProcesses map[string]int `json:"prev_processes,omitempty"`
+	Processes     map[string]int `json:"processes"`
+	ReleaseID     string         `json:"release_id"`
```
jvatic: The json key should be `release` to be consistent with other types.

jvatic: `App.ReleaseID` should also be updated.

jvatic:
Filtered out all initial scale events (no prev release).

controller/client/client.go
```diff
@@ -548,8 +588,18 @@ func (c *Client) StreamEvents(opts StreamEventsOptions, output chan *ct.Event) (
...
 		q.Set("object_id", opts.ObjectID)
 	}
+	if opts.Count > 0 {
+		q.Set("count", strconv.Itoa(opts.Count))
+	}
 	path.RawQuery = q.Encode()
-	return c.ResumingStream("GET", path.String(), output)
+	h := make(http.Header)
+	h.Set("Accept", "application/json")
+	res, err := c.RawReq("GET", path.String(), h, nil, &events)
```
titanous: Why `RawReq` over `Get`?

titanous:
LGTM pending comment.

controller/client/client.go
jvatic: So the header can be set.

PR #1903, opened 2015-09-25
all: Update Go to 1.4.3

titanous:
Also tweaked the CLI OS/architecture pairs to match the available builds:

- darwin 386 was removed
- freebsd amd64 was added


lmars:
LGTM

PR #1911, opened 2015-09-28
scheduler: Instead of directly assigning s.isLeader, use HandleLeaderChange to populate logs correctly

jzila:

### DIFF
--- a/controller/scheduler/scheduler.go
+++ b/controller/scheduler/scheduler.go
@@ -256,11 +256,11 @@ func (s *Scheduler) Run() error {
 		return err
 	}
 
-	var err error
-	s.isLeader, err = s.discoverd.Register()
+	isLeader, err := s.discoverd.Register()
 	if err != nil {
 		return err
 	}
+	s.HandleLeaderChange(isLeader)
 	leaderCh := s.discoverd.LeaderCh()
 
 	if err := s.streamFormationEvents(); err != nil {


controller/scheduler/scheduler.go
```diff
@@ -257,10 +257,12 @@ func (s *Scheduler) Run() error {
 	}
 
 	var err error
-	s.isLeader, err = s.discoverd.Register()
+	var isLeader bool
+	isLeader, err = s.discoverd.Register()
```
lmars: Just use `:=`

jzila: Already done

controller/scheduler/scheduler.go
```diff
@@ -256,11 +256,11 @@ func (s *Scheduler) Run() error {
 		return err
 	}
 
-	var err error
-	s.isLeader, err = s.discoverd.Register()
+	isLeader, err := s.discoverd.Register()
 	if err != nil {
 		return err
 	}
+	s.HandleLeaderChange(isLeader)
```
lmars: This has the side effect of calling [rectifyAll](https://github.com/jzila/flynn/blob/scheduler-v2/controller/scheduler/scheduler.go#L521), should we not just leave it as it was and log a message here?

jzila: Nope, we want `rectifyAll` to be called if it's the new leader.

jzila: Oh I see the formations will won't be populated yet. Hmm.

jzila: OK so right now the `rectifyAll` call will be a harmless no-op. I guess we'll let the formation stream worry about what to rectify.

lmars:
LGTM

PR #1912, opened 2015-09-29
scheduler: To prevent host.StopJob from blocking, put it in a goroutine

jzila:


controller/scheduler/scheduler.go
```diff
@@ -849,10 +849,11 @@ func (s *Scheduler) stopJob(req *JobRequest) (err error) {
 		}
 
 		log.Info("requesting host to stop job")
-		if err := host.StopJob(job.JobID); err != nil {
-			log.Error("error requesting host to stop job", "err", err)
-			return err
-		}
+		go func() {
```
lmars: Lets add a comment as to why we put this in a goroutine.

PR #1914, opened 2015-09-29
Implement Scheduler V2

### LINKED ISSUE #1484: Scheduler V2

lmars:
Closes #1484


controller/scheduler/job.go
```diff
@@ -0,0 +1,181 @@
...
+	JobStateCrashed    JobState       = "crashed"
+	JobStateRequesting JobState       = "requesting"
+	JobStateNew        JobState       = "new"
+)
+
+type JobRequest struct {
+	*Job
+	RequestType JobRequestType
+	attempts    uint
+}
```
lmars: The need for `JobRequest` (rather than just using `Job`) should be documented

controller/scheduler/job.go
```diff
@@ -0,0 +1,181 @@
...
+	return r.Job.Formation.Release.Processes[r.Type].Data
+}
+
+func (r *JobRequest) Clone() *JobRequest {
+	return &JobRequest{
+		Job:         r.Job.Clone(),
+		RequestType: r.RequestType,
+		attempts:    r.attempts,
+	}
+}
```
lmars: The reason why both `JobRequest` and `Job` need to be copied should be documented

controller/scheduler/scheduler.go
```diff
@@ -0,0 +1,1239 @@
...
+
+const (
+	eventBufferSize    int           = 1000
+	maxJobAttempts     uint          = 30
+	jobAttemptInterval time.Duration = 500 * time.Millisecond
+)
+
+var logger = log15.New("component", "scheduler")
+
+func fnLogger(ctx ...interface{}) log15.Logger {
```
lmars: This methods purpose should be documented

lmars:
@titanous please review

controller/scheduler/formation.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/controller/utils"
+)
+
+type Formations map[utils.FormationKey]*Formation
+
+func (fs Formations) Get(appID, releaseID string) *Formation {
+	if form, ok := fs[utils.FormationKey{AppID: appID, ReleaseID: releaseID}]; ok {
+		return form
```
titanous: Checking `ok` is not necessary, can just return the result of the map lookup.

controller/scheduler/scheduler.go
```diff
@@ -0,0 +1,1239 @@
...
+	pc := make([]uintptr, 10) // at least 1 entry needed
+	runtime.Callers(2, pc)
+	f := runtime.FuncForPC(pc[0])
+	name := f.Name()
+	name = name[strings.LastIndex(name, ".")+1:]
+	params := []interface{}{"fn", name}
+	params = append(params, ctx...)
+	return logger.New(params...)
+}
```
titanous: Was this function benchmarked?

controller/scheduler/scheduler.go
```diff
@@ -0,0 +1,1239 @@
...
+		if job.IsSchedulable() {
+			s.triggerRectify(job.Formation.key())
+		}
+		return true
+	}
+	return false
+}
+
+func (s *Scheduler) Jobs() map[string]*Job {
+	jobs := make(map[string]*Job)
```
titanous: This should have a size hint of `len(s.jobs)`

lmars:
@titanous @jzila [this commit](https://github.com/flynn/flynn/commit/d2709ef720c6ce5cbda5d27c27e65149f890c7de) should hopefully fix the status app deploy failure in `TestReleaseImages`

titanous:
This LGTM but we should get rid of the runtime magic in `fnLogger`.

jzila:
K I'll finish the final feedback today.

PR #1932, opened 2015-10-03
dashboard,taffy: Don't create intermediate release for taffy deploys

jvatic:
```
Usage: taffy <app> <repo> <branch> <rev> [-e <var>=<val>]... [-m <key>=<val>]...
```


jvatic:
This is ready for review

taffy/main.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
+	"github.com/flynn/flynn/pkg/version"
+)
+
+func init() {
+	log.SetFlags(0)
+}
+
+const DEST = "/tmp/app"
```
titanous: We don't use all caps constants in Go, maybe `DestPath`?

taffy/main.go
```diff
@@ -0,0 +1,111 @@
...
+			args = append(args, fmt.Sprintf("%s=%s", k, v))
+		}
+	}
+	cmd := exec.Command("/bin/flynn-receiver", args...)
+	cmd.Stdin = &buf
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := gitArchive(rev, &buf); err != nil {
+		return err
+	}
```
titanous: This should be run in a goroutine with an `io.Pipe` instead of buffering the entire tarball in memory.

jvatic:
@titanous comments addressed

titanous:
LGTM

PR #1941, opened 2015-10-06
dashboard,installer: Use ESLint instead of JSHint, fix formatting

jvatic:
[ESLint](http://eslint.org) does a much better job and is more configurable than [JSHint](http://jshint.com/).

ESLint is now integrated into the assetbuilder so the build will fail if dashboard or installer js/jsx files have linting errors.


titanous:
Please split the whitespace changes into a separate commit.

titanous:
LGTM

PR #1947, opened 2015-10-06
script: Use tags to determine release and tag before releasing

### LINKED ISSUE #1321: Release should upload tag before starting
The release script should upload the tag before starting, and not continue unless it is successful (#1320). The iteration number should be derived from the tags downloaded from GitHub during the git clone to avoid collisions.

### LINKED ISSUE #1803: release with tag but no manifest breaks version calculation
If a release fails after tagging but before the manifest is pushed, a subsequent release on the same day will result in this error:

```text
===> 18:55:50.261 building flynn
fatal: tag 'v20150901.0' already exists
```

Instead of using the manifest, we should use the git tags to figure out the iteration number.

josephglanville:
Closes #1803 
Closes #1321

### DIFF
--- a/script/release-flynn
+++ b/script/release-flynn
@@ -116,14 +116,15 @@ main() {
     manifest="$(new_release_manifest)"
   fi
 
-  info "checking for existing release"
-  if jq --exit-status ".versions[] | select(.commit==\"${commit}\")" <<< "${manifest}"; then
-    fail "commit ${commit} has already been released"
-  fi
+  info "getting previous release tag"
+  local previous=$(git tag --list "v*" --sort "v:refname" | tail -n 1)
 
   info "determining version"
-  local version="$(next_release_version <<< "${manifest}")"
+  local version="$(next_release_version ${previous})"
   info "using version ${version}"
+
+  info "tagging release"
+  tag_release "${commit}" "${version}"
 
   info "releasing components"
   "${ROOT}/script/release-components" \
@@ -133,9 +134,6 @@ main() {
     -t "${tuf_dir}" \
     "${commit}" \
     "${version}"
-
-  info "tagging release"
-  tag_release "${commit}" "${version}"
 
   info "releasing vm images"
   "${ROOT}/script/release-vm-images" \
--- a/script/lib/release.sh
+++ b/script/lib/release.sh
@@ -5,10 +5,17 @@ new_release_manifest() {
     echo '{"versions":[]}'
 }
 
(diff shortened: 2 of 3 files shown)


script/release-flynn
```diff
@@ -116,15 +116,16 @@ main() {
     manifest="$(new_release_manifest)"
   fi
 
-  info "checking for existing release"
-  if jq --exit-status ".versions[] | select(.commit==\"${commit}\")" <<< "${manifest}"; then
-    fail "commit ${commit} has already been released"
-  fi
+  info "getting previous release tag"
+  local previous=$(git tag | tail -n 1)
```
lmars: This should be explicit about which tags it is listing (i.e. `git tag --list 'v*'`)

lmars:
LGTM pending comment

script/release-flynn
lmars: Hmm thinking about it, the ordering will break if we have more than 10 releases in a day (which could happen if say we have some intermittent issue that is blocking releases).

We can either zero pad the iteration (`printf "%02d" "${iteration}"`) or just count matching tags (`git tag --list "v${date}*" | wc -l`). I prefer the latter.

josephglanville: Good call, I think the latter is the way to go too.

josephglanville:
Turns out git-tag has a --sort option that can sort by x.y.z style version rather than lexicographic, which is a simpler solution to the 10 versions problem.

script/release-flynn
```diff
@@ -116,15 +116,16 @@ main() {
     manifest="$(new_release_manifest)"
   fi
 
-  info "checking for existing release"
-  if jq --exit-status ".versions[] | select(.commit==\"${commit}\")" <<< "${manifest}"; then
-    fail "commit ${commit} has already been released"
-  fi
+  info "getting previous release tag"
+  local previous=$(git tag --list v* --sort "v:refname" | tail -n 1)
```
lmars: You need to quote `v*` to avoid bash expanding it to a list of files:

```
# unquoted
$ bash -x -c 'git tag --list v*' | wc -l
+ git tag --list version.json
       0

# quoted
$ bash -x -c 'git tag --list "v*"' | wc -l
+ git tag --list 'v*'
     203
```

lmars:
@josephglanville +1 for `--sort` option, wasn't aware of that.

lmars:
P.S it wouldn't hurt to have a test which checks `.9 -> .10`

josephglanville:
Comments addressed.

PR #1949, opened 2015-10-07
script: Clear apt lists to ensure clean slate

josephglanville:
The idea here is that the failures we are seeing indicate that the trusty release is either untrusted or not correctly downloaded, by clearing the existing state we eliminate some sources of uncertainty.
It's unlikely this will fix the problem but it makes it easier to reason about what could be wrong.


titanous:
LGTM

PR #1952, opened 2015-10-07
script: create local git tag in tag_release rather than in release-components

josephglanville:
We already create the tag in `release-flynn` so we shouldn't try retag it with the same tag here, it will fail.

### DIFF
--- a/script/release-components
+++ b/script/release-components
@@ -119,7 +119,6 @@ main() {
 
   info "building flynn"
   git checkout --force --quiet $commit
-  git tag "v${version}"
 
   make release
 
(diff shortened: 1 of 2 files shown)


script/release-components
```diff
@@ -119,7 +119,6 @@ main() {
 
   info "building flynn"
   git checkout --force --quiet $commit
-  git tag "v${version}"
```
lmars: We tag via the GitHub API, so do we either need to pull the tag or create the local tag in `tag_release`?

lmars: Alternatively, use `git tag --force`

josephglanville: Ahh I see. I think we should create the tag locally then in `tag_release`

PR #1954, opened 2015-10-07
cli,test: Add --all option to scale command

jvatic:
#1664

### DIFF
--- a/cli/scale.go
+++ b/cli/scale.go
@@ -23,11 +23,19 @@ Ommitting the arguments will show the current scale.
 Options:
 	-n, --no-wait            don't wait for the scaling events to happen
 	-r, --release=<release>  id of release to scale (defaults to current app release)
+	-a, --all                show non-zero formations from all releases (only works when listing formations, can't be combined with --release)
 
 Example:
 
 	$ flynn scale
 	web=4 worker=2
+
+	$ flynn scale --all
+	496d6e74-9db9-4cff-bcce-a3b44015907a (current)
+	web=1 worker=2
+
+	632cd907-85ab-4e53-90d0-84635650ec9a
+	web=2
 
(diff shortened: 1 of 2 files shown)


cli/scale.go
```diff
@@ -23,12 +23,20 @@ Ommitting the arguments will show the current scale.
...
 
 	$ flynn scale
 	web=4 worker=2
 
+	$ flynn scale --all
+	496d6e74-9db9-4cff-bcce-a3b44015907a
+	web=1 worker=2
+
+	632cd907-85ab-4e53-90d0-84635650ec9a
+	web=2
```
titanous: The current release (if any) should be marked.

cli/scale.go
```diff
@@ -144,6 +159,64 @@ func runScale(args *docopt.Args, client *controller.Client) error {
...
+
+	formations := make([]*ct.Formation, len(releases))
+	for i, r := range releases {
+		formation, err := client.GetFormation(app, r.ID)
+		if err == controller.ErrNotFound {
+			formation = &ct.Formation{
+				AppID:     app,
+				ReleaseID: r.ID,
+				Processes: make(map[string]int),
+			}
```
titanous: Can we `append` with allocated capacity instead of putting blank releases in?

jvatic:
@titanous comment addressed, also added skipping releases where nothing to show

cli/scale.go
jvatic: That will make iteration below more difficult

titanous:
LGTM

cli/scale.go
jvatic: I suppose we could make `formations` a `map[string]*ct.Formation` where the release id is the lookup key

titanous: Yeah, that's a good idea.

PR #1955, opened 2015-10-07
host/cli: Add timeout option for bootstrap

### LINKED ISSUE #1938: Make timeout configurable for bootstrap
Currently bootstrap's timeout is hard-coded to 30 seconds: https://github.com/flynn/flynn/blob/master/bootstrap/bootstrap.go#L205

That can run into trouble with race conditions; nodes from AWS may not have all finished installing by the time bootstrap is run. It would be good if timeout could be configured, to avoid it in similar environments.

josephglanville:
Closes #1938

### DIFF
--- a/host/cli/bootstrap.go
+++ b/host/cli/bootstrap.go
@@ -20,6 +20,7 @@ usage: flynn-host bootstrap [options] [<manifest>]
 
 Options:
   -n, --min-hosts=MIN  minimum number of hosts required to be online
+  -t, --timeout=SECS   seconds to wait for hosts to come online [default: 30]
   --json               format log output as json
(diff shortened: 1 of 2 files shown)


host/cli/bootstrap.go
```diff
@@ -20,6 +20,7 @@ usage: flynn-host bootstrap [options] [<manifest>]
 
 Options:
   -n, --min-hosts=MIN  minimum number of hosts required to be online
+  -t, --timeout=SEC    seconds to wait for hosts to come online [default: 30]
```
titanous: s/SEC/SECS/

titanous:
LGTM pending comment.

PR #1957, opened 2015-10-08
Migrate cluster domain

### LINKED ISSUE #1816: Add easy procedure for migrating CLUSTER_DOMAIN
Users shouldn't be stuck with the `CLUSTER_DOMAIN` after bootstrap.

jvatic:
- [x] Test / Add integration test
- [x] Expose via cli command

Closes #1816

### DIFF
--- a/cli/cluster.go
+++ b/cli/cluster.go
@@ -1,15 +1,19 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"log"
 	"strings"
+	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
 	cfg "github.com/flynn/flynn/cli/config"
 	"github.com/flynn/flynn/controller/client"
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/pkg/shutdown"
 )
 
 func init() {
@@ -18,6 +22,7 @@ usage: flynn cluster
        flynn cluster add [-f] [-d] [-g <githost>] [--git-url <giturl>] [-p <tlspin>] <cluster-name> <domain> <key>
        flynn cluster remove <cluster-name>
        flynn cluster default [<cluster-name>]
+       flynn cluster migrate-domain <domain>
 
 Manage clusters in the ~/.flynnrc configuration file.
 
@@ -39,6 +44,11 @@ Examples:
 
 	$ flynn cluster add -p KGCENkp53YF5OvOKkZIry71+czFRkSw2ZdMszZ/0ljs= default dev.localflynn.com e09dc5301d72be755a3d666f617c4600
 	Cluster "default" added.
+
+	$ flynn cluster migrate-domain new.example.com
+	Migrate cluster domain from "example.com" to "new.example.com"? (yes/no): yes
+	Migrating cluster domain (this can take up to 2m0s)...
+	Changed cluster domain from "example.com" to "new.example.com"
 `)
 }
 
@@ -53,6 +63,8 @@ func runCluster(args *docopt.Args) error {
 		return runClusterRemove(args)
 	} else if args.Bool["default"] {
 		return runClusterDefault(args)
+	} else if args.Bool["migrate-domain"] {
+		return runClusterMigrateDomain(args)
 	}
 
 	w := tabWriter()
@@ -194,4 +206,65 @@ func runClusterDefault(args *docopt.Args) error {
 
 	log.Printf("%q is now the default cluster.", name)
 	return nil
+}
+
+func runClusterMigrateDomain(args *docopt.Args) error {
+	client, err := getClusterClient()
+	if err != nil {
+		shutdown.Fatal(err)
+	}
+
+	dm := &ct.DomainMigration{
+		Domain: args.String["<domain>"],
+	}
+
+	release, err := client.GetAppRelease("controller")
+	if err != nil {
+		return err
+	}
+	dm.OldDomain = release.Env["DEFAULT_ROUTE_DOMAIN"]
+
+	if !promptYesNo(fmt.Sprintf("Migrate cluster domain from %q to %q?", dm.OldDomain, dm.Domain)) {
+		fmt.Println("Aborted")
+		return nil
+	}
+
+	maxDuration := 2 * time.Minute
+	fmt.Printf("Migrating cluster domain (this can take up to %s)...\n", maxDuration)
+
+	events := make(chan *ct.Event)
+	stream, err := client.StreamEvents(controller.StreamEventsOptions{
+		ObjectTypes: []ct.EventType{ct.EventTypeDomainMigration},
+	}, events)
+	if err != nil {
+		return nil
+	}
+	defer stream.Close()
+
+	if err := client.PutDomain(dm); err != nil {
+		return err
+	}
+
+	timeout := time.After(maxDuration)
+	for {
+		select {
+		case event, ok := <-events:
+			if !ok {
+				return stream.Err()
+			}
+			var e *ct.DomainMigrationEvent
+			if err := json.Unmarshal(event.Data, &e); err != nil {
+				return err
+			}
+			if e.Error != "" {
+				fmt.Println(e.Error)
+			}
+			if e.DomainMigration.FinishedAt != nil {
+				fmt.Printf("Changed cluster domain from %q to %q\n", dm.OldDomain, dm.Domain)
+				return nil
+			}
+		case <-timeout:
+			return errors.New("timed out waiting for domain migration to complete")
+		}
+	}
 }
--- a/controller/domain.go
+++ b/controller/domain.go
@@ -1,0 +1,95 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/que-go"
+	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/net/context"
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/pkg/httphelper"
+	"github.com/flynn/flynn/pkg/postgres"
+	"github.com/flynn/flynn/pkg/tlscert"
+)
+
+type DomainMigrationRepo struct {
+	db *postgres.DB
+}
+
+func NewDomainMigrationRepo(db *postgres.DB) *DomainMigrationRepo {
+	return &DomainMigrationRepo{db: db}
+}
+
+func (repo *DomainMigrationRepo) Add(dm *ct.DomainMigration) error {
+	tx, err := repo.db.Begin()
+	if err != nil {
+		return err
+	}
+	if err := tx.QueryRow("domain_migration_insert", dm.OldDomain, dm.Domain, dm.OldTLSCert, dm.TLSCert).Scan(&dm.ID, &dm.CreatedAt); err != nil {
+		tx.Rollback()
+		return err
+	}
+	if err := createEvent(tx.Exec, &ct.Event{
+		ObjectID:   dm.ID,
+		ObjectType: ct.EventTypeDomainMigration,
+	}, ct.DomainMigrationEvent{
+		DomainMigration: dm,
+	}); err != nil {
+		tx.Rollback()
+		return err
+	}
+	return tx.Commit()
+}
+
+func (c *controllerAPI) MigrateDomain(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	var dm *ct.DomainMigration
+	if err := json.NewDecoder(req.Body).Decode(&dm); err != nil {
+		respondWithError(w, err)
+		return
+	}
+	defaultRouteDomain := os.Getenv("DEFAULT_ROUTE_DOMAIN")
+	if dm.OldDomain != defaultRouteDomain {
+		respondWithError(w, ct.ValidationError{
+			Message: fmt.Sprintf(`Can't migrate from "%s" when currently using "%s"`, dm.OldDomain, defaultRouteDomain),
+		})
+		return
+	}
+
+	app, err := c.appRepo.Get("router")
+	if err != nil {
+		respondWithError(w, err)
+		return
+	}
+	release, err := c.appRepo.GetRelease(app.(*ct.App).ID)
+	if err != nil {
+		respondWithError(w, err)
+		return
+	}
+	dm.OldTLSCert = &tlscert.Cert{
+		Cert:       release.Env["TLSCERT"],
+		PrivateKey: release.Env["TLSKEY"],
+	}
+
+	if err := c.domainMigrationRepo.Add(dm); err != nil {
+		respondWithError(w, err)
+		return
+	}
+
+	args, err := json.Marshal(dm)
+	if err != nil {
+		respondWithError(w, err)
+		return
+	}
+
+	if err := c.que.Enqueue(&que.Job{
+		Type: "domain_migration",
+		Args: args,
+	}); err != nil {
+		respondWithError(w, err)
+		return
+	}
+
+	httphelper.JSON(w, 200, &dm)
+}
--- a/controller/types/types.go
+++ b/controller/types/types.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/flynn/flynn/host/resource"
 	"github.com/flynn/flynn/host/types"
+	"github.com/flynn/flynn/pkg/tlscert"
 	"github.com/flynn/flynn/router/types"
 )
 
@@ -96,6 +97,16 @@ type Job struct {
 	Meta      map[string]string `json:"meta,omitempty"`
 	CreatedAt *time.Time        `json:"created_at,omitempty"`
 	UpdatedAt *time.Time        `json:"updated_at,omitempty"`
+}
+
+type DomainMigration struct {
+	ID         string        `json:"id"`
+	OldTLSCert *tlscert.Cert `json:"old_tls_cert,omitempty"`
+	TLSCert    *tlscert.Cert `json:"tls_cert,omitempty"`
+	OldDomain  string        `json:"old_domain"`
+	Domain     string        `json:"domain"`
+	CreatedAt  *time.Time    `json:"created_at,omitempty"`
+	FinishedAt *time.Time    `json:"finished_at,omitempty"`
 }
 
 func (e *Job) IsDown() bool {
@@ -216,6 +227,7 @@ const (
 	EventTypeKeyDeletion      EventType = "key_deletion"
 	EventTypeRoute            EventType = "route"
 	EventTypeRouteDeletion    EventType = "route_deletion"
+	EventTypeDomainMigration  EventType = "domain_migration"
 )
 
 type Event struct {
@@ -248,4 +260,9 @@ type AppDeletion struct {
 type AppDeletionEvent struct {
 	AppDeletion *AppDeletion `json:"app_deletion"`
 	Error       string       `json:"error"`
+}
+
+type DomainMigrationEvent struct {
+	DomainMigration *DomainMigration `json:"domain_migration"`
+	Error           string           `json:"error,omitempty"`
 }
--- a/controller/types/types.go
+++ b/controller/types/types.go
@@ -11,6 +11,8 @@ import (
 	"github.com/flynn/flynn/router/types"
 )
 
+const RouteParentRefPrefix = "controller/apps/"
+
 type ExpandedFormation struct {
 	App       *App           `json:"app,omitempty"`
 	Release   *Release       `json:"release,omitempty"`
--- a/controller/worker/domain_migration/domain_migration.go
+++ b/controller/worker/domain_migration/domain_migration.go
@@ -1,0 +1,388 @@
+package domain_migration
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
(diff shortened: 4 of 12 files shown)


controller/domain.go
```diff
@@ -0,0 +1,73 @@
...
+type DomainMigrationRepo struct {
+	db *postgres.DB
+}
+
+func NewDomainMigrationRepo(db *postgres.DB) *DomainMigrationRepo {
+	return &DomainMigrationRepo{db: db}
+}
+
+func (repo *DomainMigrationRepo) Add(dm *ct.DomainMigration) error {
+	return repo.db.QueryRow("INSERT INTO domain_migrations (old_domain, new_domain, old_tls_cert) VALUES ($1, $2) RETURNING migration_id, created_at", dm.OldDomain, dm.NewDomain, dm.OldTLSCert).Scan(&dm.ID, &dm.CreatedAt)
```
lmars: Looks like this is missing a `$3` to insert `old_tls_cert`

jvatic:
- [ ] Expose via cli command

controller/domain.go
```diff
@@ -0,0 +1,93 @@
...
+func NewDomainMigrationRepo(db *postgres.DB) *DomainMigrationRepo {
+	return &DomainMigrationRepo{db: db}
+}
+
+func (repo *DomainMigrationRepo) Add(dm *ct.DomainMigration) error {
+	tx, err := repo.db.Begin()
+	if err != nil {
+		return err
+	}
+	if err := tx.QueryRow("INSERT INTO domain_migrations (old_domain, new_domain, old_tls_cert) VALUES ($1, $2, $3) RETURNING migration_id, created_at", dm.OldDomain, dm.NewDomain, dm.OldTLSCert).Scan(&dm.ID, &dm.CreatedAt); err != nil {
```
jvatic: Should also insert `dm.TLSCert` to allow client to bypass one being generated

controller/types/types.go
```diff
@@ -96,6 +96,16 @@ type Job struct {
 	UpdatedAt *time.Time        `json:"updated_at,omitempty"`
 }
 
+type DomainMigration struct {
+	ID         string     `json:"id"`
+	OldTLSCert []byte     `json:"old_tls_cert,omitempty"`
+	TLSCert    []byte     `json:"tls_cert,omitempty"`
+	OldDomain  string     `json:"old_domain"`
+	NewDomain  string     `json:"new_domain"`
```
jvatic: Drop the `New` prefix

cli/domain.go
```diff
@@ -0,0 +1,61 @@
...
+usage: flynn domain set <domain>
+
+Migrate to a new domain.
+
+Examples:
+
+	$ flynn domain set 127.0.0.1.xip.io
+	Success.
+
+	$ flynn domain set 127.0.0.1.xip.io --ca /path/to/ca.cert
```
titanous: What about `flynn cluster domain` or something like that? It seems weird to have a top-level command and we can add `flynn cluster upgrade`, etc.

titanous: Maybe even `flynn cluster migrate-domain`

titanous: The `--ca` flag looks unused.

jvatic: Yeah, `flynn cluster domain` seems better.

titanous: Also, the command should prompt for confirmation (showing the old domain) before continuing.

cli/domain.go
```diff
@@ -0,0 +1,61 @@
...
+		var e *ct.DomainMigrationEvent
+		if err := json.Unmarshal(event.Data, &e); err != nil {
+			fmt.Printf("ERROR parsing event: %s\n", err)
+			continue
+		}
+		if e.Error != "" {
+			fmt.Printf("ERROR: %s\n", e.Error)
+		}
+		if e.DomainMigration.FinishedAt != nil {
+			fmt.Println("Success.")
```
titanous: How about `Changed cluster domain from example.com to new.example.com`

controller/domain.go
```diff
@@ -0,0 +1,88 @@
...
+	return tx.Commit()
+}
+
+func (c *controllerAPI) MigrateDomain(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	var dm *ct.DomainMigration
+	if err := json.NewDecoder(req.Body).Decode(&dm); err != nil {
+		respondWithError(w, err)
+		return
+	}
+	dm.OldDomain = os.Getenv("DEFAULT_ROUTE_DOMAIN")
```
titanous: This should be specified in the request.

titanous: And then matched against the local env var for confirmation.

cli/domain.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil
+	}
+	defer stream.Close()
+
+	if err := client.PutDomain(dm); err != nil {
+		return err
+	}
+
+	for {
+		event := <-events
```
lmars:
```go
for event := range events {
  ...
}
```

lmars: In fact, we should probably use `select` and add a timeout case.

controller/domain.go
```diff
@@ -0,0 +1,88 @@
...
+func NewDomainMigrationRepo(db *postgres.DB) *DomainMigrationRepo {
+	return &DomainMigrationRepo{db: db}
+}
+
+func (repo *DomainMigrationRepo) Add(dm *ct.DomainMigration) error {
+	tx, err := repo.db.Begin()
+	if err != nil {
+		return err
+	}
+	if err := tx.QueryRow("INSERT INTO domain_migrations (old_domain, domain, old_tls_cert, tls_cert) VALUES ($1, $2, $3, $4) RETURNING migration_id, created_at", dm.OldDomain, dm.Domain, dm.OldTLSCert, dm.TLSCert).Scan(&dm.ID, &dm.CreatedAt); err != nil {
```
lmars: This should use a [prepared query](https://github.com/flynn/flynn/blob/v20151013.0/controller/schema/queries.go).

test/test_domain_migration.go
```diff
@@ -0,0 +1,101 @@
...
+
+	prevRouterRelease, err := client.GetAppRelease("router")
+	t.Assert(err, c.IsNil)
+
+	err = client.PutDomain(dm)
+	t.Assert(err, c.IsNil)
+
+	// created
+	var event ct.DomainMigrationEvent
+	e := <-events
```
lmars: Wrap this in a timeout

test/test_domain_migration.go
```diff
@@ -0,0 +1,101 @@
...
+	t.Assert(event.DomainMigration.OldDomain, c.Equals, dm.OldDomain)
+	t.Assert(event.DomainMigration.Domain, c.Equals, dm.Domain)
+	t.Assert(event.DomainMigration.TLSCert, c.IsNil)
+	t.Assert(event.DomainMigration.OldTLSCert, c.Not(c.IsNil))
+	t.Assert(event.DomainMigration.CreatedAt, c.Not(c.IsNil))
+	t.Assert(event.DomainMigration.CreatedAt.Equal(*dm.CreatedAt), c.Equals, true)
+	t.Assert(event.DomainMigration.FinishedAt, c.IsNil)
+
+	// complete
+	e = <-events
```
lmars: Wrap this in a timeout

test/test_domain_migration.go
```diff
@@ -0,0 +1,101 @@
...
+
+	res, err = (&http.Client{}).Get(fmt.Sprintf("http://dashboard.%s/ping", newDomain))
+	t.Assert(err, c.IsNil)
+	t.Assert(res.StatusCode, c.Equals, 200, c.Commentf("failed to ping dashboard"))
+
+	return dm
+}
+
+func (s *DomainMigrationSuite) TestDomainMigration(t *c.C) {
+	dm := s.migrateDomain(t, fmt.Sprintf("%s.xip.io", routerIP))
```
lmars: It's probably worth adding a comment as to why we are using a `xip.io` domain.

test/test_domain_migration.go
```diff
@@ -0,0 +1,101 @@
...
+	t.Assert(dashboardRelease.Env["DEFAULT_ROUTE_DOMAIN"], c.Equals, newDomain)
+	t.Assert(dashboardRelease.Env["CONTROLLER_DOMAIN"], c.Equals, fmt.Sprintf("controller.%s", newDomain))
+	t.Assert(dashboardRelease.Env["URL"], c.Equals, fmt.Sprintf("dashboard.%s", newDomain))
+	t.Assert(dashboardRelease.Env["CA_CERT"], c.Equals, cert.CACert)
+
+	res, err := (&http.Client{}).Get(fmt.Sprintf("http://controller.%s/ping", newDomain))
+	t.Assert(err, c.IsNil)
+	t.Assert(res.StatusCode, c.Equals, 200, c.Commentf("failed to ping controller"))
+
+	res, err = (&http.Client{}).Get(fmt.Sprintf("http://dashboard.%s/ping", newDomain))
```
lmars: We will need to add `dashboard.$CLUSTER_DOMAIN` to `/etc/hosts` in CI for this to work (i.e. [here](https://github.com/flynn/flynn/blob/v20151013.0/test/cluster/cluster.go#L557-L559)).

controller/worker/domain_migration/domain_migration.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	return err
+}
+
+func (c *context) createEvent(dm *ct.DomainMigration, err error) error {
+	e := ct.DomainMigrationEvent{DomainMigration: dm}
+	if err != nil {
+		e.Error = err.Error()
+	}
+	query := "INSERT INTO events (object_id, object_type, data) VALUES ($1, $2, $3)"
```
josephglanville: You can re-use the `event_insert` prepared query here.

jvatic:
@titanous @lmars comments addressed

jvatic:
It appears there's still one or two things to fix

controller/worker/domain_migration/domain_migration.go
```diff
@@ -0,0 +1,254 @@
...
+func dupRelease(release *ct.Release) *ct.Release {
+	return &ct.Release{
+		ArtifactID: release.ArtifactID,
+		Env:        release.Env,
+		Meta:       release.Meta,
+		Processes:  release.Processes,
+	}
+}
+
+func (c *context) maybeDeployController(oldCert, cert *tlscert.Cert, oldDomain, newDomain string) error {
```
jvatic: This should check for a deployment in progress and wait for it to complete first

controller/worker/domain_migration/domain_migration.go
```diff
@@ -0,0 +1,254 @@
...
+	release = dupRelease(release)
+	release.Env["DEFAULT_ROUTE_DOMAIN"] = newDomain
+	release.Env["CA_CERT"] = cert.CACert
+	if err := c.client.CreateRelease(release); err != nil {
+		return err
+	}
+	return c.client.DeployAppRelease(appName, release.ID)
+}
+
+func (c *context) maybeDeployRouter(oldCert, cert *tlscert.Cert, oldDomain, newDomain string) error {
```
jvatic: Ditto

controller/worker/domain_migration/domain_migration.go
```diff
@@ -0,0 +1,254 @@
...
+	release = dupRelease(release)
+	release.Env["TLSCERT"] = cert.Cert
+	release.Env["TLSKEY"] = cert.PrivateKey
+	if err := c.client.CreateRelease(release); err != nil {
+		return err
+	}
+	return c.client.DeployAppRelease(appName, release.ID)
+}
+
+func (c *context) maybeDeployDashboard(oldCert, cert *tlscert.Cert, oldDomain, newDomain string) error {
```
jvatic: Ditto

cli/cluster.go
```diff
@@ -195,3 +206,60 @@ func runClusterDefault(args *docopt.Args) error {
...
+			}
+			var e *ct.DomainMigrationEvent
+			if err := json.Unmarshal(event.Data, &e); err != nil {
+				return err
+			}
+			if e.Error != "" {
+				fmt.Println(e.Error)
+			}
+			if e.DomainMigration.FinishedAt != nil {
+				fmt.Printf(`Changed cluster domain from "%s" to "%s"`, dm.OldDomain, dm.Domain)
```
jvatic: Missing `\n`

controller/worker/domain_migration/domain_migration.go
jvatic: Done

controller/worker/domain_migration/domain_migration.go
jvatic: Done

test/test_domain_migration.go
jvatic: Done

cli/cluster.go
```diff
@@ -195,3 +206,60 @@ func runClusterDefault(args *docopt.Args) error {
...
+				return err
+			}
+			if e.Error != "" {
+				fmt.Println(e.Error)
+			}
+			if e.DomainMigration.FinishedAt != nil {
+				fmt.Printf("Changed cluster domain from \"%s\" to \"%s\"\n", dm.OldDomain, dm.Domain)
+				return nil
+			}
+		case <-time.After(2 * time.Minute):
```
lmars: Given this could take up to 2 mins without giving any output, we may want to warn the user so they don't `ctrl-c`, e.g.:

```
migrating domain (this can take up to 2 minutes)...
```

or we could print something every few seconds (e.g. `.`).

cli/cluster.go
```diff
@@ -195,3 +206,60 @@ func runClusterDefault(args *docopt.Args) error {
...
+			}
+			var e *ct.DomainMigrationEvent
+			if err := json.Unmarshal(event.Data, &e); err != nil {
+				return err
+			}
+			if e.Error != "" {
+				fmt.Println(e.Error)
+			}
+			if e.DomainMigration.FinishedAt != nil {
+				fmt.Printf("Changed cluster domain from \"%s\" to \"%s\"\n", dm.OldDomain, dm.Domain)
```
lmars: Use `%q` rather than `\"%s\"`

controller/worker/domain_migration/domain_migration.go
```diff
@@ -0,0 +1,306 @@
...
+	errChan := make(chan error)
+	createMissingRoutes := func(app *ct.App) {
+		errChan <- c.appCreateMissingRoutes(app.ID, oldCert, cert, oldDomain, newDomain)
+	}
+	for _, app := range apps {
+		go createMissingRoutes(app)
+	}
+	for range apps {
+		if err := <-errChan; err != nil {
+			return err
```
lmars: We should read all the errors from `errChan` to avoid leaking goroutines

jvatic: We can/should close `errChan` after this loop, but something like `for err := range errChan` will cause a deadlock.

titanous: You can buffer the channel to avoid leaking.

jvatic:
The integration test has a few different failures I've been having difficulty tracking down:

```
test_domain_migration.go:113:
    s.migrateDomain(t, dm)
test_domain_migration.go:60:
    t.Fatal("timed out waiting for final domain_migration event")
... Error: timed out waiting for final domain_migration event
```
https://ci.flynn.io/builds/20151015154750-712ae4cb

and

```
test_domain_migration.go:113:
    s.migrateDomain(t, dm)
test_domain_migration.go:84:
    t.Assert(routerRelease.Env["TLSKEY"], c.Not(c.Equals), "")
... obtained string = ""
... expected string = ""
```
https://ci.flynn.io/builds/20151015164125-155e362c

jvatic:
Oh, I think I know going on (for the second error at least): The private key for the tls cert is discarded when added to the db and so it's not present when deploying the router if the job has restarted.

jvatic:
Failing to ping dashboard (DNS should be getting set [here](https://github.com/flynn/flynn/blob/master/test/cluster/cluster.go#L557)):

```
test_domain_migration.go:119:
    s.migrateDomain(t, &ct.DomainMigration{
        OldDomain: dm.Domain,
        Domain:    dm.OldDomain,
    })
test_domain_migration.go:101:
    t.Assert(err, c.IsNil)
... value *url.Error = &url.Error{Op:"Get", URL:"http://dashboard.flynn-b14aee09666a2a53.local/ping", Err:(*net.OpError)(0xc208745d80)} ("Get http://dashboard.flynn-b14aee09666a2a53.local/ping: dial tcp: lookup dashboard.flynn-b14aee09666a2a53.local: no such host")
```
https://ci.flynn.io/builds/20151016184018-3ebbb65a
https://ci.flynn.io/builds/20151016182015-866d2d8b
https://ci.flynn.io/builds/20151016175354-92170bec

jvatic:
#1982 should have fixed the dashboard ping error, running the tests again

jvatic:
ping @titanous / @lmars for review

cli/cluster.go
```diff
@@ -195,3 +206,64 @@ func runClusterDefault(args *docopt.Args) error {
...
+	}
+	dm.OldDomain = release.Env["DEFAULT_ROUTE_DOMAIN"]
+
+	if !promptYesNo(fmt.Sprintf("Migrate cluster domain from %q to %q?", dm.OldDomain, dm.Domain)) {
+		fmt.Println("Aborted")
+		return nil
+	}
+
+	maxDuration := 2 * time.Minute
+	fmt.Printf("Migrating cluster domain (this can take up to %s)...\n", maxDuration)
```
lmars: This line should be in the example output above

controller/worker/domain_migration/domain_migration.go
```diff
@@ -0,0 +1,358 @@
...
+	createMissingRoutes := func(app *ct.App) {
+		errChan <- m.appCreateMissingRoutes(app.ID)
+	}
+	for _, app := range apps {
+		go createMissingRoutes(app)
+	}
+	defer close(errChan)
+	for range apps {
+		if err := <-errChan; err != nil {
+			return err
```
lmars: Returning an error early and thus closing the channel due to the defer could cause a panic when the remaining goroutines try to send their errors

titanous:
Looks like this needs a rebase.

jvatic:
@lmars @titanous comments addressed

controller/worker/domain_migration/domain_migration.go
```diff
@@ -0,0 +1,359 @@
...
+		return err
+	}
+	errChan := make(chan error, len(apps))
+	createMissingRoutes := func(app *ct.App) {
+		errChan <- m.appCreateMissingRoutes(app.ID)
+	}
+	for _, app := range apps {
+		go createMissingRoutes(app)
+	}
+	defer close(errChan)
```
titanous: This close isn't necessary.

titanous:
LGTM pending comment.

titanous:
@jvatic Please write up any remaining issues on this.

jvatic:
@titanous The TestDeployController test appears to fail with an unexpected old release still in the process of shutting down only on this branch (e.g. https://ci.flynn.io/builds/20151026194047-7640f6ae). I suspect a deployment is failing to wait for job down events.

jvatic:
If I'm right, fixing [this](https://github.com/flynn/flynn/blob/70f8acd618d65d85c6ef0bd3eddd61ece1ebd71d/controller/worker/deployment/job.go#L130-L131) would solve the issue.

lmars:
@jvatic #2035 should fix the TestDeployController issue

jvatic:
There are still a bunch of intermittent test failures to investigate:

https://ci.flynn.io/builds/20151028011622-4148614e
https://ci.flynn.io/builds/20151028013736-59a4a131
https://ci.flynn.io/builds/20151028015504-8041ff69
https://ci.flynn.io/builds/20151028030337-196e1afd

controller/worker/domain_migration/domain_migration.go
```diff
@@ -0,0 +1,358 @@
...
+		AppID:       d.AppID,
+		ObjectID:    d.ID,
+		ObjectTypes: []ct.EventType{ct.EventTypeDeployment},
+	}, events)
+	if err != nil {
+		return err
+	}
+	defer stream.Close()
+
+	log := m.logger.New("app", app, "deployment", d.ID)
```
jvatic: `app` should be something else as there is already a key with the same name. `deploy_app` maybe?

jvatic:
RE https://ci.flynn.io/builds/20151028011622-4148614e:

TestDomainMigration fails due to too many active connections to the router (https://gist.github.com/anonymous/75adba5833dafdee8951#file-default-router-app-72e20c98-bfbbed8d-c12b-4f6c-9ed7-3f4548077562-log-filtered-L134). Restricting the number of concurrent route updates should resolve this.

jvatic:
Looks like the timeout needs increasing in `TestDomainMigration`: https://ci.flynn.io/builds/20151028200555-cd38b1de (the migration completed about 10 seconds after it timed out.)

controller/worker/domain_migration/domain_migration.go
lmars: `app.name`?

lmars:
@jvatic 

> TestDomainMigration fails due to too many active connections to the router ... Restricting the number of concurrent route updates should resolve this.

Given the router [gets all routes and then filters them](https://github.com/flynn/flynn/blob/migrate-cluster-domain/router/api.go#L120-L142), maybe the worker should just get all the routes and do the filtering itself?

jvatic:
@lmars agreed, that will significantly reduce the number of requests. However the updates will still happen concurrently and need limiting.

jvatic:
@lmars, @titanous, I think this is ready to ship!

titanous:
LGTM

PR #1960, opened 2015-10-11
host: Add signal command

josephglanville:
Adds a `signal` command to `flynn-host` for signalling jobs.

Could potentially add a mapping from string => int signals to allow usage of any of SIGQUIT, QUIT or 3 as the SIGNAL argument.

### DIFF
--- a/host/cli/signal.go
+++ b/host/cli/signal.go
@@ -1,0 +1,41 @@
+package cli
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
+	"github.com/flynn/flynn/pkg/cluster"
+)
+
+func init() {
+	Register("signal", runSignal, `
+usage: flynn-host signal ID SIGNAL
+
+Signal a job`)
+}
+
+func runSignal(args *docopt.Args, client *cluster.Client) error {
+	id := args.String["ID"]
+	sig, err := strconv.Atoi(args.String["SIGNAL"])
+	if err != nil {
+		fmt.Println("invalid value for SIGNAL")
+		return err
+	}
+	hostID, err := cluster.ExtractHostID(id)
+	if err != nil {
+		fmt.Println("could not parse", id)
+		return err
+	}
+	hostClient, err := client.Host(hostID)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


host/cli/signal.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+	"strconv"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
+	"github.com/flynn/flynn/pkg/cluster"
+)
+
+func init() {
+	Register("signal", runSignal, `
+usage: flynn-host signal ID [SIGNAL]
```
lmars: The code requires a valid `SIGNAL` argument, so it shouldn't be wrapped in `[ ]`

host/cli/signal.go
```diff
@@ -0,0 +1,41 @@
...
+func runSignal(args *docopt.Args, client *cluster.Client) error {
+	id := args.String["ID"]
+	sig, err := strconv.Atoi(args.String["SIGNAL"])
+	if err != nil {
+		fmt.Println("invalid value for SIGNAL")
+		return err
+	}
+	hostID, err := cluster.ExtractHostID(id)
+	if err != nil {
+		fmt.Println("could not parse ", id)
```
lmars: `fmt.Println` adds spaces between the arguments, so you can drop the trailing space here.

josephglanville: Doh! Fixed.

host/cli/signal.go
josephglanville: Ahh yep, I was going to make it default to SIGINT but figured that probably doesn't make sense and didn't fully roll it back. Will fix it up.

lmars:
> Could potentially add a mapping from string => int signals to allow usage of any of SIGQUIT, QUIT or 3 as the SIGNAL argument.

I was thinking we could add a `flynn-host pid` command and then we could just run `kill -QUIT $(flynn-host pid ID)`, but that would only work for signalling jobs on the same host...

lmars:
LGTM

josephglanville:
I was thinking similar regarding `flynn-host pid`. Maybe we should add the pid to `flynn-host inspect`?

lmars:
> Maybe we should add the pid to flynn-host inspect

SGTM

PR #1962, opened 2015-10-11
Godeps: remove pq and go-sql

josephglanville:
Drops `pq` and `go-sql` as they are no longer used since migrating to `pgx`.


titanous:
LGTM

lmars:
LGTM

PR #1964, opened 2015-10-12
bootstrap: Report service status details in bootstrap

### LINKED ISSUE #1825: Show status failure details in bootstrap
If the status endpoint is not showing everything as healthy after bootstrap is complete, a message should be printed explaining what services are failing.

josephglanville:
This introduces another step into the bootstrap process that queries the status app to ascertain the health of the cluster before completing the bootstrap.

Example output with a single unhealthy service:
```
19:17:35.309024 wait status-wait
19:17:38.062046 status-check status-check
19:18:08.068476 status-check status-check unhealthy services detected!

The following services are reporting unhealthy, this likely indicates a problem with your deployment:
	dashboard

19:18:08.068492 log log-complete
19:18:08.068593 log log-complete

Flynn bootstrapping complete. Install the Flynn CLI (see https://flynn.io/docs/cli for instructions) and paste the line below into a terminal window:

flynn cluster add -p AnKjz+VOD1ndlsUKhGjc/JiiQ4QfsNVQu9Gjg4Cduf0= default dev.localflynn.com 339cb492858821690bf4d297540f2ce3

The built-in dashboard can be accessed at http://dashboard.dev.localflynn.com and your login token is 53303cfc65d00f8085ba08ec75e0b140
```

Closes  #1825

### DIFF
--- a/bootstrap/manifest_template.json
+++ b/bootstrap/manifest_template.json
@@ -510,8 +510,8 @@
     "url": "tcp://gitreceive.discoverd"
   },
   {
-    "id": "status-wait",
-    "action": "wait",
+    "id": "status-check",
+    "action": "status-check",
     "url": "http://status-web.discoverd"
   },
   {
--- a/bootstrap/status_check_action.go
+++ b/bootstrap/status_check_action.go
@@ -1,0 +1,81 @@
+package bootstrap
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+type StatusCheckAction struct {
+	ID     string `json:"id"`
+	URL    string `json:"url"`
+	Output string `json:"output"`
+}
+
+type StatusResponse struct {
+	Data StatusData `json:"data"`
+}
+
+type StatusData struct {
+	Status string                  `json:"status"`
+	Detail map[string]StatusDetail `json:"detail"`
+}
+
+type StatusDetail struct {
+	Status string `json:"status"`
+}
+
+func init() {
+	Register("status-check", &StatusCheckAction{})
+}
+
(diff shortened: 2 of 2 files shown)


bootstrap/manifest_template.json
```diff
@@ -515,6 +515,11 @@
     "url": "http://status-web.discoverd"
   },
```
titanous: I think we should drop this step. If the status endpoint is not green, it will never pass this point due to the 500. It's probably worth manually testing this out by breaking a service or two and confirming that the output is correct.

bootstrap/status_check_action.go
```diff
@@ -0,0 +1,83 @@
...
+	lookupDiscoverdURLHost(u, waitMax)
+
+	start := time.Now()
+	for {
+		req, err := http.NewRequest("GET", u.String(), nil)
+		if err != nil {
+			return err
+		}
+		header := make(http.Header)
+		header.Set("Content-Type", "application-json")
```
titanous: This header is wrong (should be `Accept` and `application/json`).

bootstrap/status_check_action.go
```diff
@@ -0,0 +1,83 @@
...
+		req.Header = header
```
titanous: Could assign directly to the allocated map:

```go
req.Header = make(http.Header)
req.Header.Set...
```

bootstrap/manifest_template.json
josephglanville: Yeah you are right. I tested the actual status check by adding a sleep to the start of the status-check step to give me time to go break something. But if it's broken at that point it does indeed not pass.

titanous:
LGTM

PR #1966, opened 2015-10-13
HA/recoverable log aggregation

titanous:
This PR is for review only, it is blocked on the #1672 due to the log handling changes.

The `logaggregator` component is now highly available, the `host/logmux` package has been rewritten, and the `host/logbuf` package deleted.

Logs now have sequence numbers associated with them, the tuple of `(host_id, timestamp, seq)` (used as a "cursor") makes a given log line unique and can be used for deduplication. Time is assumed to only move forward on each host (no exact time coordination between hosts is required), which allows us to only care about comparing sequence numbers if the timestamp is the same (we may need to tweak this a bit to handle leap seconds).

Logs are saved by `flynn-host` into rotated per-app local files in a slightly modified syslog format.

When a host starts up, as soon as discoverd is available, it subscribes to the `logaggregator` service. When a new aggregator instance comes up, the current cursor for the host is determined. Local logs are then replayed for each app from the given cursor.

The aggregator sorts incoming log messages by timestamp, which avoids problems when multiple hosts are replaying old logs in parallel. This does mean that logs may be in slightly unexpected order if hosts have large amounts of clock skew, but it seems like a reasonable tradeoff for simplicity.

If there is a leader when an aggregator instance is booting up, a log buffer snapshot will be fetched from it, reducing the number of log messages that need to be replayed from individual hosts.


host/logmux/logmux.go
```diff
@@ -2,140 +2,664 @@ package logmux
...
-		_, err := w.Write(rfc6587.Bytes(msg))
-		if err != nil {
-			g.Log(grohl.Data{"status": "error", "err": err.Error()})
+	host, _, _ := net.SplitHostPort(addr)
+	c, _ := client.New("http://" + host)
+	cursors, err := c.GetCursors()
+	if err != nil {
+		// TODO(titanous): retry
+		l.Error("failed to get cursors from aggregator", "error", err)
+		return
```
lmars: `conn` needs closing here

host/logmux/logmux.go
```diff
@@ -2,140 +2,664 @@ package logmux
...
+	if aggCursor != nil {
+		l.Info("got cursor", "cursor.timestamp", aggCursor.Time, "cursor.seq", aggCursor.Seq)
+	} else {
+		l.Info("no cursor for host")
+	}
+
+	appLogs, err := m.logFiles("")
+	if err != nil {
+		l.Error("failed to get local log files", "error", err)
+		return
```
lmars: `conn` needs closing here

logaggregator/server.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+type Server struct {
+	*Aggregator
+	Cursors *HostCursors
+
+	conf ServerConfig
+
+	ll, al net.Listener   // syslog and api listeners
+	lwg    sync.WaitGroup // syslog wait group
```
lmars: Could just give these fields descriptive names rather than commenting them, e.g. `syslogListener apiListener syslogwg`

logaggregator/server.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (s *Server) LoadSnapshotFile(path string) error {
+	f, err := os.Open(path)
+	if os.IsNotExist(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
```
lmars: `defer f.Close()`

logaggregator/main.go
```diff
@@ -28,320 +22,40 @@ func main() {
...
-func (a *Aggregator) run() {
-	for msg := range a.msgc {
-		a.getOrInitializeBuffer(string(msg.AppName)).Add(msg)
-		if afterMessage != nil {
-			afterMessage()
-		}
+	if err := srv.Start(); err != nil {
+		shutdown.Fatal(err)
 	}
+	<-make(chan struct{})
```
lmars: Could rename `Server.Start` to `Server.Serve` which blocks (rather than starting a goroutine and then blocking on this channel)

logaggregator/server.go
```diff
@@ -0,0 +1,221 @@
...
+	defer conn.Close()
+
+	connDone := make(chan struct{})
+	defer close(connDone)
+
+	go func() {
+		select {
+		case <-connDone:
+		case <-s.shutdown:
+			conn.Close()
```
lmars: This could double close with the defer above, is that an issue?

script/kill-flynn
```diff
@@ -62,7 +62,7 @@ kill_libvirt_lxc() {
     --exec "${flynn_host}"
 
   # force kill any remaining containers
-  virsh -c lxc:/// list --name | xargs -L 1 virsh -c lxc:/// destroy || true
+  sudo virsh -c lxc:/// list --name | xargs -L 1 sudo virsh -c lxc:/// destroy || true
```
lmars: `sudo` shouldn't be necessary since #1841, but I don't mind using it anyway.

lmars:
@titanous would be good to see a CI run, should we rebase #1672 on master? (I actually think we should merge that anyway)

titanous:
Yep, first step is rebasing.

lmars:
@titanous I have rebased #1672 on master.

logaggregator/main.go
titanous: I considered that, but the tests need a non-blocking version that returns errors and this was easier to manage.

titanous:
Comments addressed.

logaggregator/server.go
titanous: Done.

PR #1968, opened 2015-10-14
host: Don't display EndedAt if zero

### LINKED ISSUE #1958: flynn-host inspect returns the zero date as EndedAt for Running jobs
`flynn-host` should probably not print a timestamp for EndedAt for tasks that are still running.

josephglanville:
When a task is yet to end it's EndedAt time is zero.
This change prints the empty string instead of of the zero data in `flynn-host inspect`.

Closes #1958


titanous:
LGTM

PR #1970, opened 2015-10-14
bootstrap: Allow passing pre-generated TLS details in env vars

### LINKED ISSUE #1039: Allow changing clusters without breaking clients
In order to swap out a cluster without the dashboard CA cert changing or the CLI TLS pin changing, we should allow all of the following to be extracted and reused for bootstrap:

- CA certificate (the private key is discarded and unecessary)
- Cluster domain wildcard cert and keypair (used by controller, dashboard, router)

Bootstrap should be modified to allow specifying any/all of these parameters instead of generating them, and we should have documentation on how to extract them.

titanous:
This allows a cluster to be bootstrapped using a previous cluster's certificates, so a cluster can be seamlessly replaced without breaking clients.

Closes #1039 

/cc @temujin9

### DIFF
--- a/bootstrap/gen_tls_cert_action.go
+++ b/bootstrap/gen_tls_cert_action.go
@@ -1,6 +1,9 @@
 package bootstrap
 
 import (
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/pem"
 	"fmt"
 
 	"github.com/flynn/flynn/pkg/certgen"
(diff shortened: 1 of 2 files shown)


lmars:
LGTM (it would be nice to add a test for this, but I think we need the test binary to be able to easily boot new clusters first, possible since #1891)

kwando:
:heart:

PR #1980, opened 2015-10-16
controller,cli: Update route

### LINKED ISSUE #981: controller: Client has no method to update an existing route
The controller has `RouteList`, `GetRoute`, `CreateRoute`, and `DeleteRoute`, but no `UpdateRoute`. It should probably be added (but maybe after #959 is merged :)

jvatic:
Closes #981

### DIFF
--- a/cli/route.go
+++ b/cli/route.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
 	"github.com/flynn/flynn/controller/client"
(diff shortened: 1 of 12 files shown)


cli/route.go
```diff
@@ -162,6 +176,81 @@ func runRouteAddHTTP(args *docopt.Args, client *controller.Client) error {
...
+	appName := mustApp()
+
+	route, err := client.GetRoute(appName, id)
+	if err != nil {
+		return err
+	}
+
+	if service := args.String["--service"]; service != "" {
+		route.Service = service
+	}
```
titanous: This should probably exit early if the service is the same or empty.

cli/route.go
```diff
@@ -162,6 +176,81 @@ func runRouteAddHTTP(args *docopt.Args, client *controller.Client) error {
...
+
+	tlsCertPath := args.String["--tls-cert"]
+	tlsKeyPath := args.String["--tls-key"]
+	if tlsCertPath != "" && tlsKeyPath != "" {
```
titanous: This logic looks similar to add, can it be abstracted?

jvatic:
@titanous comments addressed

titanous:
Do we have any integration tests for configuring HTTPS routes with the CLI? If not, please add them since that functionality is important.

jvatic:
@titanous Integration test added

titanous:
LGTM

Issue #1119, opened 2015-03-01
Controller deploy can fail to set app release


lmars:
If the `web` process is the last one to restart, the deployer sometimes fails with "connection refused" when calling `SetAppRelease`.

See https://s3.amazonaws.com/flynn-ci-logs/20150301130013-72d02e32-build-8faaae3105ebb22e95c95d60d987bcdb21c2b7b7-2015-03-01-13-18-14.txt.

The deployer needs better retry logic when deploying the controller.

jvatic:
Closing on the basis that #1991 fixes the issue

PR #1981, opened 2015-10-16
controller/deployment: Retry SetAppRelease a few times when connection failure

jvatic:
#1119

### DIFF
--- a/controller/worker/deployment/context.go
+++ b/controller/worker/deployment/context.go
@@ -3,6 +3,7 @@ package deployment
 import (
 	"encoding/json"
 	"errors"
+	"strings"
 	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/que-go"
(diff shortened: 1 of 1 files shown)


controller/worker/deployment/context.go
```diff
@@ -114,9 +115,17 @@ func (c *context) HandleDeployment(job *que.Job) (e error) {
 		return err
 	}
 	log.Info("setting the app release")
-	if err := c.client.SetAppRelease(deployment.AppID, deployment.NewReleaseID); err != nil {
-		log.Error("error setting the app release", "err", err)
-		return err
+	for n := 2; n > 0; n-- {
+		if err := c.client.SetAppRelease(deployment.AppID, deployment.NewReleaseID); err != nil {
+			if strings.HasSuffix(err.Error(), "connection refused") && n > 0 {
+				log.Error("failed to connect to controller when setting app release, retrying", "err", err)
```
josephglanville: Should this be handling other types of retry-able network errors?

jvatic: Yeah, that sounds like a good idea. Also need to construct a test for this.

jvatic:
This appears to no longer be an issue (see #1991)

PR #1982, opened 2015-10-17
test: Fix setting dashboard domain

lmars:
Where it was being set previously was ineffective. I have added the domain in the correct place (just before running the test binary) and removed the ineffective `setLocalDNS`.

/cc @jvatic


titanous:
LGTM

jvatic:
:shipit:

PR #1983, opened 2015-10-17
boostrap: When checking status in bootstrap, retry if unhealthy

josephglanville:
The status check was racing with the dashboard starting up and reporting it unhealthy falsely.
By retrying until the max wait period is reached we avoid reporting false positives to the user.

### DIFF
--- a/bootstrap/status_check_action.go
+++ b/bootstrap/status_check_action.go
@@ -55,6 +55,11 @@ func (a *StatusCheckAction) Run(s *State) error {
 		}
 		if res.StatusCode == 200 {
 			s.StepData[a.ID] = &LogMessage{Msg: "all services healthy"}
+		} else if time.Now().Sub(start) < waitMax {
+			// if services are unhealthy wait out the wait period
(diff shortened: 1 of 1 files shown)


bootstrap/status_check_action.go
```diff
@@ -55,6 +55,10 @@ func (a *StatusCheckAction) Run(s *State) error {
 		}
 		if res.StatusCode == 200 {
 			s.StepData[a.ID] = &LogMessage{Msg: "all services healthy"}
+		} else if time.Now().Sub(start) < waitMax {
+			// if services are unhealthy wait out the wait period
+			// before reporting them as unhealthy to the user
+			continue
```
titanous: Should this `goto fail` for the `Sleep` call?

josephglanville: The `fail` label is for handling not being able  to communicate with the status service.
In this case we got a response but is was non-200, I had just forgotten to add the sleep so it was actually a busy loop, oops.

titanous:
LGTM

PR #1986, opened 2015-10-19
bootstrap: Check if ports are available prior to bootstrap

### LINKED ISSUE #1824: Add sanity checks to bootstrap
The bootstrap phase should have sanity checks in place that verify obvious failure modes like other services bound to necessary ports, firewalled ports, etc.

josephglanville:
This introduces a stage into bootstrap that runs ahead of starting any services to try ascertain if there are any resources in-use that might cause the bootstrap to fail. Currently this only checks that required ports are available.

Example output when `tcp:1111` is in use on a bootstrap target:
```
00:28:39.479409 check online-hosts
00:28:40.496625 require-env require-env
00:28:40.498279 resource-check resource-check
00:28:40.499529 resource-check resource-check error: conflicts detected!

The following hosts have conflicting services listening on ports Flynn is configured to use:
10.0.2.15: tcp:1111

After you correct the above errors re-run bootstrap to continue setting up Flynn.
```

Closes #1824

### DIFF
--- a/bootstrap/resource_check_action.go
+++ b/bootstrap/resource_check_action.go
@@ -1,0 +1,53 @@
+package bootstrap
+
+import (
+	"encoding/json"
+	"fmt"
+	"net"
+
+	"github.com/flynn/flynn/host/types"
+	"github.com/flynn/flynn/pkg/cluster"
+	"github.com/flynn/flynn/pkg/httphelper"
+)
+
+type ResourceCheckAction struct {
+	Ports []host.Port
+}
+
+func init() {
+	Register("resource-check", &ResourceCheckAction{})
+}
+
+func (a *ResourceCheckAction) Run(s *State) error {
+	conflicts := make(map[*cluster.Host][]host.Port)
+	for _, h := range s.Hosts {
+		req := host.ResourceCheck{Ports: a.Ports}
+		if err := h.ResourceCheck(req); err != nil {
+			if j, ok := err.(httphelper.JSONError); ok {
+				var resp host.ResourceCheck
+				if err := json.Unmarshal(j.Detail, &resp); err != nil {
+					return err
+				}
+				conflicts[h] = resp.Ports
+			} else {
+				return err
+			}
+		}
+	}
+	if len(conflicts) > 0 {
+		conflictMsg := "conflicts detected!\n\nThe following hosts have conflicting services listening on ports Flynn is configured to use:\n"
+		for host, ports := range conflicts {
+			hostIP, _, err := net.SplitHostPort(host.Addr())
+			if err != nil {
+				return err
+			}
+			conflictMsg += hostIP + ": "
+			for _, port := range ports {
+				conflictMsg += fmt.Sprintf("%s:%d ", port.Proto, port.Port)
+			}
+		}
+		conflictMsg += "\n\nAfter you correct the above errors re-run bootstrap to continue setting up Flynn."
+		return fmt.Errorf(conflictMsg)
+	}
+	return nil
+}
--- a/host/http.go
+++ b/host/http.go
@@ -1,7 +1,9 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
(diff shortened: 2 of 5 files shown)


josephglanville:
I'm not super happy with this this turned out but I was unable to come up with a reasonable strategy to extract the information generically from the manifest and provide a reasonable API for checking if said jobs would be runnable.

The issues I ran into were primarily that jobs that use host networking don't always specify the ports in the manifest they need and in the case of the router actually start listening immediately on a massive range of ports.

host/http.go
```diff
@@ -228,6 +230,45 @@ func (h *jobAPI) GetStatus(w http.ResponseWriter, r *http.Request, _ httprouter.
 	httphelper.JSON(w, 200, &h.host.status)
 }
 
+func checkPort(port host.Port) bool {
+	fmt.Println("Checking port", port.Port)
```
titanous: Remove this print.

host/http.go
```diff
@@ -228,6 +230,45 @@ func (h *jobAPI) GetStatus(w http.ResponseWriter, r *http.Request, _ httprouter.
...
+	}
+	if len(conflicts) > 0 {
+		msg, err := json.Marshal(conflicts)
+		if err != nil {
+			httphelper.Error(w, err)
+			return
+		}
+		httphelper.JSON(w, 409, httphelper.JSONError{
+			Code:    httphelper.ConflictErrorCode,
+			Message: string(msg),
```
titanous: Use the `Detail` field, a string would double-escape the JSON.

host/http.go
```diff
@@ -228,6 +230,45 @@ func (h *jobAPI) GetStatus(w http.ResponseWriter, r *http.Request, _ httprouter.
...
+	if err != nil {
+		return false
+	}
+	l.Close()
+	return true
+}
+
+func (h *jobAPI) ResourceCheck(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	var ports []host.Port
+	conflicts := make([]host.Port, 0)
```
titanous: No need to allocate:

```go
var conflicts []host.Port
```

host/http.go
```diff
@@ -228,6 +230,45 @@ func (h *jobAPI) GetStatus(w http.ResponseWriter, r *http.Request, _ httprouter.
...
+		msg, err := json.Marshal(conflicts)
+		if err != nil {
+			httphelper.Error(w, err)
+			return
+		}
+		httphelper.JSON(w, 409, httphelper.JSONError{
+			Code:    httphelper.ConflictErrorCode,
+			Message: string(msg),
+		})
+	}
```
titanous: This branch is missing a `return`.

host/http.go
```diff
@@ -228,6 +230,45 @@ func (h *jobAPI) GetStatus(w http.ResponseWriter, r *http.Request, _ httprouter.
...
+		return false
+	}
+	l.Close()
+	return true
+}
+
+func (h *jobAPI) ResourceCheck(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	var ports []host.Port
+	conflicts := make([]host.Port, 0)
+	if err := httphelper.DecodeJSON(r, &ports); err != nil {
```
titanous: If this is a "resource check" and not a "port check" it should use a top-level struct for extensibility.

titanous:
This is fine, simple is good.

josephglanville:
Comments addressed.

I think it's relevance as a resource-check will become more prominent when we make it easier for people to setup their own ZFS pools etc, sufficient space on both the ZFS pool and the Flynn working directories could be checked here before starting.

host/types/types.go
```diff
@@ -286,3 +286,7 @@ const (
 	JobEventStop   string = "stop"
 	JobEventError  string = "error"
 )
+
+type ResourceCheck struct {
+	Ports []Port `json"ports, omitempty"`
```
titanous: Remove the space after the comma.

bootstrap/resource_check_action.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (a *ResourceCheckAction) Run(s *State) error {
+	conflicts := make(map[*cluster.Host][]host.Port)
+	for _, h := range s.Hosts {
+		req := host.ResourceCheck{Ports: a.Ports}
+		if err := h.ResourceCheck(req); err != nil {
+			if j, ok := err.(httphelper.JSONError); ok {
+				var resp host.ResourceCheck
+				if err := json.Unmarshal(j.Detail, &resp); err != nil {
+					fmt.Printf(string(j.Detail))
```
titanous: Drop this print.

bootstrap/resource_check_action.go
```diff
@@ -0,0 +1,48 @@
...
+				conflicts[h] = resp.Ports
+			} else {
+				return err
+			}
+		}
+	}
+	if len(conflicts) > 0 {
+		conflictMsg := "conflicts detected!\n\nThe following hosts have conflicting services listening on ports Flynn is configured to use:\n"
+		for host, ports := range conflicts {
+			conflictMsg += host.Addr() + ": "
```
titanous: What do you think about stripping the port from the host IP? (`net.SplitHostPort`)

josephglanville: Yeah, I was going to ask how to do that. (and now I know! :D)

titanous:
LGTM

PR #1990, opened 2015-10-19
discoverd: add instance expiration

### LINKED ISSUE #1984: discoverd instance expiry is not implemented
Currently instance expiry is only used for a minor component of leader election, instances do not appear to expire on their own. This needs to be implemented.

@benbjohnson Do you have thoughts on the strategy for this?

benbjohnson:
## Overview

This commit adds the ability for the discoverd store to periodically check for instance expiration and then execute a consistent expiry command through the raft consensus.

Fixes #1984

### DIFF
--- a/discoverd/server/store.go
+++ b/discoverd/server/store.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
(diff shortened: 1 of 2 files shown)


titanous:
LGTM!

PR #2000, opened 2015-10-21
Discoverd improvements

josephglanville:
This PR is pulling out some of the work @lmars has done to support updating discoverd so that PR can be more focused on the actual deployment/update process when it's opened.


titanous:
LGTM

PR #2001, opened 2015-10-21
router, Godeps: Switch to github.com/vanillahsu/go_reuseport

josephglanville:
This is prep-work for the discoverd update process as UDP support is required for zero-downtime DNS.


titanous:
LGTM

PR #2002, opened 2015-10-21
Make more Procfile process types routable via HTTP

tombh:
Very new to Flynn, so correct me if there are other ways of doing this.

The problem I'm trying to solve is that I have multiple processes in my app that I want to accessible via the web. Specifically, I have an API, JS frontend and admin UI all running as individual processes in my Procfile, but I can only have one of them routed with `flynn route add http`.

How would it be to extend the "web" process type behaviour to also include process types ending in "-web"?

### DIFF
--- a/test/test_controller.go
+++ b/test/test_controller.go
@@ -329,10 +329,18 @@ func (s *ControllerSuite) TestAppDeleteCleanup(t *c.C) {
 	_, err := s.discoverdClient(t).Instances(service, 10*time.Second)
 	t.Assert(err, c.IsNil)
 
+	t.Assert(r.flynn("scale", "another-web=1"), Succeeds)
+	_, err = s.discoverdClient(t).Instances(app+"-another-web", 10*time.Second)
+	t.Assert(err, c.IsNil)
+
 	// create some routes
-	routes := []string{"foo.example.com", "bar.example.com"}
+	routes := []string{"foo.example.com", "bar.example.com", "another.example.com"}
 	for _, route := range routes {
-		t.Assert(r.flynn("route", "add", "http", route), Succeeds)
+		if route == "another.example.com" {
+			t.Assert(r.flynn("route", "add", "http", "-s", app+"-another-web", route), Succeeds)
+		} else {
+			t.Assert(r.flynn("route", "add", "http", route), Succeeds)
+		}
 	}
 	routeList, err := client.RouteList(app)
 	t.Assert(err, c.IsNil)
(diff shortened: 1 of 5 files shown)


titanous:
This approach is reasonable. We'd also need to specify a constructed service name on line 161 (I think `appname-proctype` is fine).

tombh:
So that a `web` proctype gets a service name of `appname-web` and `<procname>-web` proctypes get service names of `appname-<procname>-web`?

titanous:
Just `<appname>-<procname>` that will maintain the existing `appname-web` for `web` process types and other would be like `foo-api-web` if they have the`-web` suffix that triggers this.

tombh:
Just like this then? https://github.com/tombh/flynn/blob/make-more-process-types-web-routable/gitreceive/receiver/flynn-receive.go#L161

Can you point me in the right direction to getting a local dev env setup so I can try and write a test for this?

titanous:
https://flynn.io/docs/development

tombh:
Thanks, I've made a test now. Is there anything I can do about those unsigned commits causing the failing build?

titanous:
Yeah, you can sign them :)

https://flynn.io/docs/contributing#developer%E2%80%99s-certificate-of-origin

titanous:
And while you're at it, squash the commits together and add a `gitreceive/receiver: ` prefix to the commit message.

test/test_controller.go
```diff
@@ -329,10 +329,16 @@ func (s *ControllerSuite) TestAppDeleteCleanup(t *c.C) {
 	_, err := s.discoverdClient(t).Instances(service, 10*time.Second)
 	t.Assert(err, c.IsNil)
 
+	t.Assert(r.flynn("scale", "another-web=1"), Succeeds)
+
```
titanous: Add a test like this to ensure the service comes up before we test the route:

```go
_, err = s.discoverdClient(t).Instances(app+"-another-web", 10*time.Second)
t.Assert(err, c.IsNil)
```

tombh:
I can't tell why that build failed?

titanous:
Looks like an unrelated failure, I'm digging into it.

tombh:
Did you find out the problem?

titanous:
We've fixed a few unrelated test failures, please rebase on master and force-push.

tombh:
I don't think the `14:48:13.224 FAIL: test_release.go:101: ReleaseSuite.TestReleaseImages` failure is something I can fix here is it?

titanous:
LGTM, thanks!

tombh:
\o/

PR #2003, opened 2015-10-21
Update discoverd

josephglanville:
This patchset implements an in-place update of the discoverd service.

The process is as follows:

1. A discoverd deployment is triggered
2. The controller then creates a "deployment context" that describes the state of the discoverd deployment in the discoverd service metadata. This is implemented as a re-usable deployment strategy called "discoverd-meta"
3. The controller worker will then scale the formation up, as discoverd is an omnipresent job this means it will spawn new jobs on every node in the cluster.
4. These jobs will upon starting up check to see if there is already a discoverd process listening on the HTTP address they intend to bind to, if so they will connect to it.
5. They will then attempt to mark themselves as "performing" in the deployment context, ensuring only one node is marked performing at a time. This is safe because discoverd service metadata is linear.
6. Once it successfully marks itself as performing it will start it's DNS server in proxy mode (using SO_REUSEPORT) and issue a shutdown command to the process it's replacing, acquiring that processes current log position in the process.
7. It will block on the other node shutting down before starting it's Raft store and will wait until it's atleast caught up to the position that the process it replaced was before it starts the HTTP listener.
8. Once reaching this point the process will mark itself as "done" in the deployment context.
9. The controller worker watches the discoverd service meta event stream and when all nodes are marked "done" it will proceed to scaling down the formation and terminating the now dormant old jobs.
10. Deployment complete.

### DIFF
--- a/discoverd/client/client.go
+++ b/discoverd/client/client.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	dt "github.com/flynn/flynn/discoverd/types"
 	"github.com/flynn/flynn/pkg/dialer"
 	"github.com/flynn/flynn/pkg/httpclient"
 	hh "github.com/flynn/flynn/pkg/httphelper"
@@ -160,6 +161,10 @@ outer:
 			return nil, ErrTimedOut
 		}
 	}
+}
+
+func (c *Client) Shutdown() (res dt.ShutdownInfo, err error) {
+	return res, c.c.Post("/shutdown", nil, &res)
 }
 
 type service struct {
--- a/discoverd/deployment/deployment.go
+++ b/discoverd/deployment/deployment.go
@@ -1,0 +1,194 @@
+package deployment
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
+	"github.com/flynn/flynn/discoverd/client"
+	"github.com/flynn/flynn/pkg/attempt"
+	"github.com/flynn/flynn/pkg/stream"
+)
+
+type DeploymentState string
+
+const (
+	DeploymentStatePerforming DeploymentState = "performing"
+	DeploymentStateDone       DeploymentState = "done"
+)
+
+func NewDeployment(service string) (*Deployment, error) {
+	s := discoverd.NewService(service)
+	events := make(chan *discoverd.Event)
+	stream, err := s.Watch(events)
+	if err != nil {
+		return nil, err
+	}
+	return &Deployment{s, events, stream}, nil
+}
+
+// Deployment is a wrapper around service metadata for marking jobs as either
+// performing a deployment or done deploying
+type Deployment struct {
+	service discoverd.Service
+	events  chan *discoverd.Event
+	stream  stream.Stream
+}
+
+// MarkPerforming marks the given address as performing in the service metadata,
+// ensuring there is only one address marked as performing at any given time
+// (waiting for a ServiceMeta event and retrying if there is already an
+// address performing).
+func (d *Deployment) MarkPerforming(addr string, timeout int) error {
+outer:
+	for {
+		meta, err := d.service.GetMeta()
+		if err != nil {
+			return err
+		}
+
+		states, err := d.decode(meta)
+		if err != nil {
+			return err
+		}
+
+		performing := ""
+		for a, state := range states {
+			if a == addr {
+				// already marked as performing, nothing to do
+				return nil
+			}
+
+			if state == DeploymentStatePerforming {
+				performing = a
+				break
+			}
+		}
+
+		// if another address is performing, wait for a ServiceMeta
+		// event then try again
+		if performing != "" {
+			for {
+				select {
+				case event, ok := <-d.events:
+					if !ok {
+						return fmt.Errorf("service stream closed unexpectedly: %s", d.stream.Err())
+					}
+					if event.Kind == discoverd.EventKindServiceMeta {
+						continue outer
+					}
+				case <-time.After(time.Duration(timeout) * time.Second):
+					return fmt.Errorf("timed out waiting for %s to finish performing", performing)
+				}
+			}
+		}
+
+		// no performing addresses, attempt to mark addr
+		states[addr] = DeploymentStatePerforming
+
+		data, err := json.Marshal(states)
+		if err != nil {
+			return err
+		}
+		meta.Data = data
+
+		if err := d.service.SetMeta(meta); err == nil {
+			return nil
+		}
+	}
+	return nil
+}
+
+var attempts = attempt.Strategy{
+	Delay: 100 * time.Millisecond,
+	Total: 30 * time.Second,
+}
+
+// MarkDone marks the addr as done in the service metadata
+func (d *Deployment) MarkDone(addr string) error {
+	return attempts.Run(func() error {
+		meta, err := d.service.GetMeta()
+		if err != nil {
+			return err
+		}
+
+		states, err := d.decode(meta)
+		if err != nil {
+			return err
+		}
+
+		states[addr] = DeploymentStateDone
+
+		return d.set(meta, states)
+	})
+}
+
+// Wait waits for an expected number of "done" addresses in the service metadata
+func (d *Deployment) Wait(expected int, timeout int, log log15.Logger) error {
+	for {
+		actual := 0
+		select {
+		case event, ok := <-d.events:
+			if !ok {
+				return fmt.Errorf("service stream closed unexpectedly: %s", d.stream.Err())
+			}
+			if event.Kind == discoverd.EventKindServiceMeta {
+				states, err := d.decode(event.ServiceMeta)
+				if err != nil {
+					return err
+				}
+				log.Info("got service meta event", "states", states)
+				actual = 0
+				for _, state := range states {
+					if state == DeploymentStateDone {
+						actual++
+					}
+				}
+				if actual == expected {
+					return nil
+				}
+			}
+		case <-time.After(time.Duration(timeout) * time.Second):
+			return fmt.Errorf("timed out waiting for discoverd deployment (expected=%d actual=%d)", expected, actual)
+		}
+	}
+}
+
+// Reset sets the service metadata to null
+func (d *Deployment) Reset() error {
+	return attempts.Run(func() error {
+		if err := d.set(&discoverd.ServiceMeta{}, nil); err == nil {
+			return nil
+		}
+		meta, err := d.service.GetMeta()
+		if err != nil {
+			return err
+		}
+		return d.set(meta, nil)
+	})
+}
+
+func (d *Deployment) Close() error {
+	return d.stream.Close()
+}
+
+func (d *Deployment) decode(meta *discoverd.ServiceMeta) (map[string]DeploymentState, error) {
+	var states map[string]DeploymentState
+	if err := json.Unmarshal(meta.Data, &states); err != nil {
+		return nil, err
+	}
+	if states == nil {
+		states = make(map[string]DeploymentState)
+	}
+	return states, nil
+}
+
+func (d *Deployment) set(meta *discoverd.ServiceMeta, states map[string]DeploymentState) error {
+	data, err := json.Marshal(states)
+	if err != nil {
+		return err
+	}
+	meta.Data = data
+	return d.service.SetMeta(meta)
+}
--- a/discoverd/main.go
+++ b/discoverd/main.go
@@ -43,11 +43,12 @@ func main() {
 
 // Main represents the main program execution.
 type Main struct {
-	mu        sync.Mutex
-	status    host.DiscoverdConfig
-	store     *server.Store
-	dnsServer *server.DNSServer
-	ln        net.Listener
+	mu         sync.Mutex
+	status     host.DiscoverdConfig
+	store      *server.Store
+	dnsServer  *server.DNSServer
+	httpServer *http.Server
+	ln         net.Listener
 
 	logger *log.Logger
 
@@ -172,6 +173,10 @@ func (m *Main) Run(args ...string) error {
 
 // Close shuts down all open servers.
 func (m *Main) Close() error {
+	if m.httpServer != nil {
+		// Disable keep alives so that persistent connections will close
+		m.httpServer.SetKeepAlivesEnabled(false)
+	}
 	if m.store != nil {
 		m.store.Close()
 		m.store = nil
@@ -258,15 +263,16 @@ func (m *Main) openDNSServer(addr string, recursors, peers []string) error {
 func (m *Main) openHTTPServer(ln net.Listener, peers []string) error {
 	// If we have no store then simply start a proxy handler.
 	if m.store == nil {
-		go http.Serve(ln, &server.ProxyHandler{Peers: peers})
+		m.httpServer = &http.Server{Handler: &server.ProxyHandler{Peers: peers}}
+		go m.httpServer.Serve(ln)
 		return nil
 	}
 
 	// Otherwise initialize and start handler.
 	h := server.NewHandler()
 	h.Store = m.store
-	go http.Serve(ln, h)
-
+	m.httpServer = &http.Server{Handler: h}
+	go m.httpServer.Serve(ln)
 	return nil
 }
 
--- a/discoverd/main.go
+++ b/discoverd/main.go
@@ -16,7 +16,9 @@ import (
 	"time"
 
 	"github.com/flynn/flynn/discoverd/client"
+	dd "github.com/flynn/flynn/discoverd/deployment"
 	"github.com/flynn/flynn/discoverd/server"
+	dt "github.com/flynn/flynn/discoverd/types"
 	"github.com/flynn/flynn/host/types"
 	"github.com/flynn/flynn/pkg/cluster"
 	"github.com/flynn/flynn/pkg/shutdown"
@@ -77,6 +79,64 @@ func (m *Main) Run(args ...string) error {
 		return err
 	}
 
+	// Set up advertised address and default peer set.
+	advertiseAddr := MergeHostPort(opt.Host, opt.Addr)
+	if len(opt.Peers) == 0 {
+		opt.Peers = []string{advertiseAddr}
+	}
+
+	// Create a slice of peers with their HTTP address set instead.
+	httpPeers, err := SetPortSlice(opt.Peers, opt.Addr)
+	if err != nil {
+		return fmt.Errorf("set port slice: %s", err)
+	}
+
+	// if there is a discoverd process already running on this
+	// address perform a deployment by starting a proxy DNS server
+	// and shutting down the old discoverd job
+	var deploy *dd.Deployment
+	var shutdownInfo dt.ShutdownInfo
+
+	target := fmt.Sprintf("http://%s:1111", opt.Host)
+	m.logger.Println("checking for existing discoverd process at", target)
+	if err := discoverd.NewClientWithHTTP(target, &http.Client{}).Ping(); err == nil {
+		m.logger.Println("discoverd responding at", target, "taking over")
+
+		// update DISCOVERD environment variable so that the default
+		// discoverd client connects to the instance we intend to replace.
+		os.Setenv("DISCOVERD", target)
+		discoverd.DefaultClient = discoverd.NewClient()
+
+		deploy, err = dd.NewDeployment("discoverd")
+		if err != nil {
+			return err
+		}
+		m.logger.Println("Created deployment")
+		if err := deploy.MarkPerforming(advertiseAddr, 60); err != nil {
+			return err
+		}
+		m.logger.Println("marked", advertiseAddr, "as performing in deployent")
+		addr, resolvers := waitHostDNSConfig()
+		if opt.DNSAddr != "" {
+			addr = opt.DNSAddr
+		}
+		if len(opt.Recursors) > 0 {
+			resolvers = opt.Recursors
+		}
+		m.logger.Println("starting proxy DNS server")
+		if err := m.openDNSServer(addr, resolvers, httpPeers); err != nil {
+			return fmt.Errorf("Failed to start DNS server: %s", err)
+		}
+		m.logger.Printf("discoverd listening for DNS on %s", addr)
+
+		shutdownInfo, err = discoverd.NewClientWithURL(target).Shutdown()
+		if err != nil {
+			return err
+		}
+	} else {
+		m.logger.Println("failed to contact existing discoverd server, starting up without takeover")
+	}
+
 	// Open listener.
 	ln, err := net.Listen("tcp4", opt.Addr)
 	if err != nil {
@@ -86,12 +146,6 @@ func (m *Main) Run(args ...string) error {
 
 	// Multiplex listener to store and http api.
 	storeLn, httpLn := server.Mux(ln)
-
-	// Set up advertised address and default peer set.
-	advertiseAddr := MergeHostPort(opt.Host, opt.Addr)
-	if len(opt.Peers) == 0 {
-		opt.Peers = []string{advertiseAddr}
-	}
 
 	// Open store if we are not proxying.
 	if err := m.openStore(opt.DataDir, storeLn, advertiseAddr, opt.Peers); err != nil {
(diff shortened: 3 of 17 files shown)


josephglanville:
So I have had some time with this to get to know it a bit better.

What is working so far is spawning the new `discoverd` jobs, having them tell the old jobs to shutdown and then becoming active themselves.

While this works there are some kinks. Namely that the worker doing the deployment seems to stop receiving the service events. It's currently unclear what the exact cause is but I have a feeling it's because the old `discoverd` processes that have been issued shutdown commands possibly still have connections open. It is plausible it is an error in the client code but I looked at the retry logic and it looks sound.

There also seems to be the curious case of DNS resolution no longer working as expected. Namely the worker became unable to resolve the postgres discoverd address.

Because the above the actual deployment doesn't complete right now and the old `discoverd` processes are left running but the new cluster is active, sending broadcasts etc.

I did also see one case where the new instances had formed a working cluster but the discoverd service metadata was null, which shouldn't really be possible with how the code works so hopefully I don't see that again.

lmars:
@josephglanville 

> the worker doing the deployment seems to stop receiving the service events

Yes I saw this behaviour too. I expected [this change](https://github.com/flynn/flynn/commit/9a71e3c9bf50a13f70c4a62dbce215af4b8913f1#diff-e92f931b3419848b18a7209332fe268dR201) to cause the stream to close, thus leading to the worker's service watch [auto-reconnecting](https://github.com/flynn/flynn/blob/update-discoverd/discoverd/client/watch.go#L83-L86), but this doesn't seem to be happening. A next step to debug this could be to force the old process to close all open connections by just exec'ing a new binary (I assume `CLOEXEC` is set for all the TCP connections) and see if the watch re-connects to the new process.

josephglanville:
`discoverd` seems to be properly closing all connections on shutdown now.
However a new issue was discovered, in some cases the new store has tried to start listening before the former has been fully shutdown, possibly due to my change to move shutting down the store until after the listeners have been shutdown.
This appears in the log like this:
```
2015-10-29T13:25:54.606938Z app[app.host2-1133cf41-cb06-4d50-9776-06e6a95cb32d]: Failed to open store: opening tcp transport: listen tcp 198.51.100.3:1110: bind: address already in use
2015-10-29T13:25:56.030439Z app[app.host1-29947a72-8466-4506-b586-480194d86471]: 2015/10/29 13:25:56 [WARN] raft: Heartbeat timeout reached, starting election
2015-10-29T13:25:56.030538Z app[app.host1-29947a72-8466-4506-b586-480194d86471]: 2015/10/29 13:25:56 [INFO] raft: Node at 198.51.100.2:1110 [Candidate] entering Candidate state
2015-10-29T13:25:56.076641Z app[app.host0-c3e15607-76be-498d-80a3-a13f60217ad0]: 2015/10/29 13:25:56 [DEBUG] raft-net: 198.51.100.1:1110 accepted connection from: 198.51.100.1:37866
2015-10-29T13:25:56.076693Z app[app.host0-c3e15607-76be-498d-80a3-a13f60217ad0]: 2015/10/29 13:25:56 [WARN] raft: Rejecting vote from 198.51.100.2:1110 since we have a leader: 198.51.100.3:1110
2015-10-29T13:25:56.076645Z app[app.host1-29947a72-8466-4506-b586-480194d86471]: 2015/10/29 13:25:56 [ERR] raft: Failed to make RequestVote RPC to 198.51.100.3:1110: dial tcp 198.51.100.3:1110: connection refused
2015-10-29T13:25:56.199804Z app[app.host1-29947a72-8466-4506-b586-480194d86471]: 2015/10/29 13:25:56 [DEBUG] raft: Votes needed: 2
2015-10-29T13:25:56.199895Z app[app.host1-29947a72-8466-4506-b586-480194d86471]: 2015/10/29 13:25:56 [DEBUG] raft: Vote granted. Tally: 1
2015-10-29T13:25:56.772966Z app[app.host0-c3e15607-76be-498d-80a3-a13f60217ad0]: 2015/10/29 13:25:56 [WARN] raft: Heartbeat timeout reached, starting election
2015-10-29T13:25:56.773035Z app[app.host0-c3e15607-76be-498d-80a3-a13f60217ad0]: 2015/10/29 13:25:56 [INFO] raft: Node at 198.51.100.1:1110 [Candidate] entering Candidate state
2015-10-29T13:25:56.810807Z app[app.host0-c3e15607-76be-498d-80a3-a13f60217ad0]: 2015/10/29 13:25:56 [ERR] raft: Failed to make RequestVote RPC to 198.51.100.3:1110: dial tcp 198.51.100.3:1110: connection refused
2015-10-29T13:25:56.810908Z app[app.host1-29947a72-8466-4506-b586-480194d86471]: 2015/10/29 13:25:56 [DEBUG] raft-net: 198.51.100.2:1110 accepted connection from: 198.51.100.1:58185
2015-10-29T13:25:56.810976Z app[app.host1-29947a72-8466-4506-b586-480194d86471]: 2015/10/29 13:25:56 [INFO] raft: Duplicate RequestVote for same term: 3
2015-10-29T13:25:56.895375Z app[app.host0-c3e15607-76be-498d-80a3-a13f60217ad0]: 2015/10/29 13:25:56 [DEBUG] raft: Votes needed: 2
2015-10-29T13:25:56.895485Z app[app.host0-c3e15607-76be-498d-80a3-a13f60217ad0]: 2015/10/29 13:25:56 [DEBUG] raft: Vote granted. Tally: 1
2015-10-29T13:25:57.387091Z app[app.host2-ea0f8221-b41d-448e-8d93-e74d516c261f]: 2015/10/29 13:25:57 discoverd: shutting down
```
We can see the old job (`host2-ea0f8221-b41d-448e-8d93-e74d516c261f`) doesn't report shutdown until well after the new job (`host2-1133cf41-cb06-4d50-9776-06e6a95cb32d`) fails to start citing the port being in-use.

The other issues still remain, seems to be some flakiness in the client in properly disconnecting/reconnecting and after the new cluster is successfully setup there are some strange things going on, DNS in particular doesn't seem to be working right and we aren't receiving the down events we are looking for during formation scale down.

josephglanville:
Made some good progress here.

Found and fixed the DNS problem, it was two fold in that we weren't listening correctly on deploy and when I fixed that the TCP listener wasn't using SO_REUSEPORT. With that fixed DNS seems to continue to work throughout the deployment process.

There were some misc issues around races with connections closing vs listeners closing etc, most of those have been resolved bar a race with the Raft transport listener which I don't quite understand yet.

Still outstanding is some issues with the discoverd client, namely on reconnect we seem to miss service meta events and possibly related we never seem to get our job down events even when everything else seems to be working fine.

josephglanville:
This is now fully functional and the `discoverd` deployment deployment test is passing locally.
Going to get everything working in CI and backtrack and fix TODOs.

lmars:
@josephglanville great work so far! Let me know if you want to chat through any of the remaining TODOs.

josephglanville:
@lmars It would be good if you could share your thoughts on https://github.com/flynn/flynn/commit/48950268fdf65ded60d64176a9a5ebf365f35fea
It doesn't seem to cause problems but it does seem strange it wasn't like this before.
Specifically if you had a release with a single process type that used service events was it actually deployable without this change?

lmars:
@josephglanville yes I think it is just that previously we had no service-only deployments (i.e. there was always a process type that used job events), that change looks right.

josephglanville:
The integration suite is working locally but I can't work out why it's failing in CI.
I have a feeling it's related to the HostUpdate suite failing but I am not sure.
Might have to get some help to setup the CI style multi-vm setup locally (right now I am using the single machine multiple flynn-host process setup)

josephglanville:
Interesting, removing some of the debug code in the discoverd client has lead to some issues.Specifically re-enabling keepalives in the HTTP transport. I might need some help understanding how to fix that correctly without disabling that (assuming we want to keep it enabled which I can only assume we do).

josephglanville:
Found one of the problems.. it's actually a little nasty.

With this patchset we detect if we should be doing a deployment by detecting the presence of the DISCOVERD environment variable and checking if it's empty or "none", if so we consider ourselves the first discoverd on this node and procede to startup, else we attempt a takeover.

This works fine your cluster is bootstrapped and then remains up with the same number of nodes but falls down when either you lose a discoverd job and rectify is called or you add a new node to your cluster or you reboot an existing node. This is because the DISCOVERD environment variable will be populated by flynn-host and even though there is no other node to take over from it will attempt and then fail to do so.

I haven't worked out what the best course of action to resolve this is however.
One idea was to have it detect if there is a running instance rather than relying on the environment variable, that could be done by just connecting to the IP that discoverd intends to bind to and see if the other instance responds.

Thoughts @lmars?

lmars:
@josephglanville I think pinging the address that it intends to bind to makes sense.

josephglanville:
I don't know if it needs addressing here but when testing I uncovered a failure mode we might want to fix. Specifically if discoverd crashes on a node nothing can be done to restore it except use the cluster repair tool. This is because the local flynn-host daemon only communicates with it's local discoverd. If instead it had a cluster aware client it would stay registered as in instance and the scheduler would eventually correct the situation. Having a cluster-aware client would probably also just increase robustness in general.

titanous:
Yeah, the client should be cluster-aware. Please open a separate issue for that.

josephglanville:
Tracked the final issue in CI to postgres cluster having a blip due to discoverd outage on the node running the leader.

josephglanville:
@benbjohnson would you be able to shed some light on https://github.com/flynn/flynn/blob/master/discoverd/client/watch_test.go#L21?

I have been able to get reconnects working by disabling keepalives on the HTTP transport but I don't think this is the right solution.

josephglanville:
I think I might understand better now, that FIXME is specifically about fixing the state reconciliation that occurs on reconnect? The test when uncommented currently fails with:

```
waiting for leader: status=500
waiting for leader: status=500
waiting for leader: status=500
waiting for leader: status=500
discoverd leader established
waiting for leader: status=500
waiting for leader: status=500
waiting for leader: status=500
waiting for leader: status=500
discoverd leader established
discoverd process exited: signal: killed
waiting for leader: status=500
waiting for leader: status=500
waiting for leader: status=500
waiting for leader: status=500
discoverd leader established
watch_test.go:179:
    // check watchB emits missed events
    assertCurrent(eventsB, []*expectedEvent{
        {Addr: ":2222", Kind: discoverd.EventKindUpdate},
        {Addr: ":4444", Kind: discoverd.EventKindUp},
        {Addr: ":4444", Kind: discoverd.EventKindLeader},
        {Kind: discoverd.EventKindServiceMeta, ServiceMeta: serviceMeta},
        {Addr: ":3333", Kind: discoverd.EventKindDown},
    })
watch_test.go:158:
    c.Fatalf("expected %d events, got %d", len(expected), count)
... Error: expected 5 events, got 0
```

Which suggests that is the case.

josephglanville:
Ok, futher adventures lead me to why this remained a `FIXME`, it's a bit of a rabbit hole.

The premise of the test is rather simple.
Client A will do a bunch of writing while Client B is disconnected, when Client B reconnects we will test to see if we received all the missed events we expect.

However the current implementation is broken for a number of reasons.
The current test wants to do the following.
- 1. Spawn server A
- 2. Spawn server B
- 3. Create service entries with Client A
- 4. Kill server B
- 5. Create a bunch of other data with Client A
- 6. Restart server B
- 7. Wait for Client B to reconnect
- 8. Compare events to see if we got what we expected.

The core idea is sound minus a few caveats:
- Raft with 2 nodes is not good because it will (or atleast should...) become unwritable when Server B is terminated
- Current discoverd can't handle redirecting to other peers on ports that a different to it's own, so you can't even spawn 3 servers on the same IP to combat the above problem, would require loopback addresses. Which seems a bit nasty to put into what is meant to be a "unit" test.

Finding a way to execute the test against a server in standalone mode would probably be ideal but require some form of HTTP transport mock layer or something to disable requests.

I did hack in a cluster launcher and try get the test to work but ran into the redirect issue and gave up for now, I will have another hack at getting it to use multiple IPs tomorrow so I can atleast verify the behaviour of the client is what we expect so I can tick it off the list of possible causes of failures I have spotted.

josephglanville:
I tracked the last of the race conditions to the fact that the future returned from raft.Shutdown doesn't actually seem to wait for the listener to close. I have submitted a patch that closes it explicitly https://github.com/hashicorp/raft/pull/72 but I don't know if this is the right approach. Hopefully one of the Raft people with chime in.

benbjohnson:
@josephglanville The `FIXME` was because the `Store` wasn't replaying events while disconnected. I'm digging through the previous implementation (based on `etcd`) and I can't for the life of me figure out where it's different in this respect:

https://github.com/benbjohnson/flynn/tree/4f700317effeffffc9df1e9bb757ddc08bc7ed48/discoverd

As for the different port, that is an issue for sure. Currently we don't have a place to track the different ports for each instance, however, it's something that can be resolved by the TCP multiplexing mentioned in https://github.com/flynn/flynn/issues/2158. That'll give us just one port to manage.

josephglanville:
@titanous @lmars this is ready for review.

I will need to handle getting the Raft patch either merged upstream or whatever approach they prefer but the rest of the code is good to go. Do you want me to split it into individual commits or just squash it into a single commit?

discoverd/client/deployment.go
```diff
@@ -0,0 +1,195 @@
...
+)
+
+type DeploymentState string
+
+const (
+	DeploymentStatePerforming DeploymentState = "performing"
+	DeploymentStateDone       DeploymentState = "done"
+)
+
+func NewDeployment(service string) (*Deployment, error) {
```
titanous: This seems like it should be in a separate package, does it depend on any client internals?

discoverd/client/client.go
```diff
@@ -48,7 +48,10 @@ func NewClientWithURL(url string) *Client {
 		c: &httpclient.Client{
 			URL: url,
 			HTTP: &http.Client{
-				Transport:     &http.Transport{Dial: dialer.Retry.Dial},
+				Transport: &http.Transport{
+					Dial:              dialer.Retry.Dial,
+					DisableKeepAlives: true,
```
titanous: Was the keep-alive problem sorted out?

discoverd/client/client.go
```diff
@@ -162,6 +165,11 @@ outer:
 	}
 }
 
+func (c *Client) Shutdown() (uint64, error) {
+	var res uint64
+	return res, c.c.Get("/shutdown", &res)
```
titanous: This should be a `POST`

discoverd/client/deployment.go
```diff
@@ -0,0 +1,195 @@
...
+
+		data, err := json.Marshal(states)
+		if err != nil {
+			return err
+		}
+		meta.Data = data
+
+		if err := d.service.SetMeta(meta); err == nil {
+			return nil
+		}
```
titanous: This loop should probably have logging (especially the error here, but probably most of the code paths)

discoverd/client/deployment.go
```diff
@@ -0,0 +1,195 @@
...
+		case event, ok := <-d.events:
+			if !ok {
+				return fmt.Errorf("service stream closed unexpectedly: %s", d.stream.Err())
+			}
+			if event.Kind == EventKindServiceMeta {
+				states, err := d.decode(event.ServiceMeta)
+				if err != nil {
+					return err
+				}
+				log.Info("got service meta event", "states", states)
```
titanous: The creator of the deployment should probably pass a namespaced logger in for context.

discoverd/client/deployment.go
```diff
@@ -0,0 +1,195 @@
...
+				actual = 0
+				for _, state := range states {
+					if state == DeploymentStateDone {
+						actual++
+					}
+				}
+				if actual == expected {
```
titanous: Should this be `>=`?

discoverd/client/deployment.go
```diff
@@ -0,0 +1,195 @@
...
+	return d.stream.Close()
+}
+
+func (d *Deployment) decode(meta *ServiceMeta) (map[string]DeploymentState, error) {
+	var states map[string]DeploymentState
+	if err := json.Unmarshal(meta.Data, &states); err != nil {
+		return nil, err
+	}
+	if states == nil {
+		states = make(map[string]DeploymentState)
```
titanous: Could just do this at the top of the method, no need for a conditional.

discoverd/main.go
```diff
@@ -82,6 +83,58 @@ func (m *Main) Run(args ...string) error {
...
+	httpPeers, err := SetPortSlice(opt.Peers, opt.HTTPAddr)
+	if err != nil {
+		return fmt.Errorf("set port slice: %s", err)
+	}
+
+	// if there is a discoverd process already running on this
+	// address perform a deployment by starting a proxy DNS server
+	// and shutting down the old discoverd job
+	var deploy *discoverd.Deployment
+	var last_idx uint64
```
titanous: :camel:

discoverd/main.go
```diff
@@ -160,21 +215,43 @@ func (m *Main) Run(args ...string) error {
...
+	addr = net.JoinHostPort(ip.String(), "53")
+	return addr, status.Network.Resolvers
+}
+
 // Close shuts down all open servers.
-func (m *Main) Close() error {
-	if m.store != nil {
-		m.store.Close()
-		m.store = nil
+func (m *Main) Close() (last_idx uint64, err error) {
```
titanous: :camel:

discoverd/server/store.go
```diff
@@ -197,14 +199,31 @@ func (s *Store) Open() error {
...
+	defer s.mu.RUnlock()
+	if s.raft != nil {
+		return s.raft.LastIndex()
+	}
+	return 0
+}
+
 // Close shuts down the transport and store.
-func (s *Store) Close() error {
+func (s *Store) Close() (last_idx uint64, err error) {
```
titanous: :camel:

discoverd/server/handler.go
```diff
@@ -305,6 +310,15 @@ func (h *Handler) serveGetLeader(w http.ResponseWriter, r *http.Request, params
 // servePing returns a 200 OK.
 func (h *Handler) servePing(w http.ResponseWriter, r *http.Request, params httprouter.Params) {}
 
+func (h *Handler) serveShutdown(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
+	last_idx, err := h.Main.Close()
```
titanous: :camel:

discoverd/server/handler.go
```diff
@@ -50,12 +50,17 @@ func NewHandler() *Handler {
 
 	r.GET("/ping", h.servePing)
 
+	r.GET("/shutdown", h.serveShutdown)
```
titanous: `POST`

discoverd/server/handler.go
```diff
@@ -305,6 +310,15 @@ func (h *Handler) serveGetLeader(w http.ResponseWriter, r *http.Request, params
 // servePing returns a 200 OK.
 func (h *Handler) servePing(w http.ResponseWriter, r *http.Request, params httprouter.Params) {}
 
+func (h *Handler) serveShutdown(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
+	last_idx, err := h.Main.Close()
+	if err != nil {
+		hh.Error(w, err)
+		return
+	}
+	hh.JSON(w, 200, last_idx)
```
titanous: The index should be wrapped up in a struct so that we can provide more data fields in the future.

discoverd/server/dns.go
```diff
@@ -6,32 +6,49 @@ import (
...
 	TCPAddr   string
 	Domain    string
 	Recursors []string
 
-	Store interface {
-		Instances(service string) ([]*discoverd.Instance, error)
-		ServiceLeader(service string) (*discoverd.Instance, error)
-	}
+	store    DNSStore
+	storeMtx sync.Mutex
```
titanous: Could use a `sync.Atomic` instead (grep around for example usage).

discoverd/client/deployment.go
josephglanville: Agreed, it doesn't require access to any client internals.

discoverd/client/client.go
josephglanville: It was not. I have a PCAP if you want to take a swing at it though.

discoverd/client/client.go
josephglanville: Will do.

discoverd/client/deployment.go
josephglanville: Will fix.

discoverd/client/deployment.go
josephglanville: Agreed.

discoverd/client/deployment.go
josephglanville: `>=` would work but I can't think of a case were there would be more nodes marked done then the amount we are expecting?

discoverd/client/deployment.go
josephglanville: Will fix.

discoverd/main.go
josephglanville: Haha, the Ruby naming conventions slipped in. Will fix.

discoverd/server/handler.go
josephglanville: Will fix.

discoverd/server/handler.go
josephglanville: Agreed.

discoverd/server/dns.go
josephglanville: Yep I see how that could be done, will do.

josephglanville:
[capture.txt](https://github.com/flynn/flynn/files/41005/capture.txt)

Attached the PCAP, you will need to rename from TXT though as Github doesn't support uploads for "unsupported filetypes".

discoverd/server/dns.go
josephglanville: I looked into using `atomic.Value` but it seems that it can only store values of the same concrete type? We swap the store between the actual implementation and the proxy implementation which are different types. Is there a workaround that makes sense or given the limitations is the mutex just the more straightforward option?

titanous: Up to you. A pointer to an interface is a concrete type: https://play.golang.org/p/3LilYHYSir

discoverd/deployment/deployment.go
```diff
@@ -0,0 +1,196 @@
...
+
+// Deployment is a wrapper around service metadata for marking jobs as either
+// performing a deployment or done deploying
+type Deployment struct {
+	service discoverd.Service
+	events  chan *discoverd.Event
+	stream  stream.Stream
+}
+
+const deploymentTimeout = time.Minute
```
lmars: Should we be getting the timeout from the app now?

discoverd/main.go
```diff
@@ -82,6 +85,58 @@ func (m *Main) Run(args ...string) error {
...
+	}
+
+	// if there is a discoverd process already running on this
+	// address perform a deployment by starting a proxy DNS server
+	// and shutting down the old discoverd job
+	var deploy *dd.Deployment
+	var shutdownInfo dt.ShutdownInfo
+
+	target := fmt.Sprintf("http://%s:1111", opt.Host)
+	m.logger.Println("Trying to connect to:", target)
```
lmars: I would probably say something like the following to make it clearer in the logs what is happening:

```
checking for existing discoverd process listening on xxx:1111
```

discoverd/main.go
```diff
@@ -160,21 +217,45 @@ func (m *Main) Run(args ...string) error {
...
 	if m.dnsServer != nil {
 		m.dnsServer.Close()
 		m.dnsServer = nil
 	}
-	if m.httpListener != nil {
-		m.httpListener.Close()
-		m.httpListener = nil
+	var lastIdx uint64
+	if m.store != nil {
+		lastIdx, err = m.store.Close()
```
lmars: `info.LastIndex, err = m.store.Close()`

test/test_z_discoverd.go
```diff
@@ -0,0 +1,52 @@
...
+			}
+			if event.Status == "complete" {
+				debugf(t, "got deployment event: %s", event.Status)
+				break loop
+			}
+			if event.Status == "failed" {
+				t.Fatal("the deployment failed")
+			}
+			debugf(t, "got deployment event: %s %s", event.JobType, event.JobState)
+		case <-time.After(2 * time.Minute):
```
lmars: `app.DeployTimeout * time.Second`

lmars:
LGTM pending minor comments, great work :sunglasses:

discoverd/deployment/deployment.go
josephglanville: Yep, forgot to adjust for your DeployTimeout changes.

discoverd/main.go
josephglanville: Agreed.

discoverd/main.go
josephglanville: Yes, much nicer!

test/test_z_discoverd.go
josephglanville: Yep, ditto above.

josephglanville:
Ty for review, will update shortly.

discoverd/deployment/deployment.go
josephglanville: Actually I am not sure about this. The timeout should probably be configurable but it's not clear of `app.DeployTimeout` is directly comparable to the `deploymentTimeout` used here as it's used to both determine the amount of time we spend trying to mark ourselves performing and how long we wait for all `done` events to arrive. I am thinking if we do want to use app.DeploymentTimeout it should be given as an argument to `Wait()` as it's the only place this information is needed or even relevant.
The same could probably be said for the mark performing stage, it should be up to the application to choose how long is reasonable enough to wait for other nodes to finish.

josephglanville:
Looks like we have a rather nasty race somewhere still:
```
t=2015-11-26T09:10:12+0000 lvl=info msg="starting discoverd-meta deployment" app=worker fn=deployDiscoverdMeta
t=2015-11-26T09:10:12+0000 lvl=info msg="starting one-by-one deployment" app=worker fn=deployOneByOne
t=2015-11-26T09:10:12+0000 lvl=info msg="scaling new formation up by one" app=worker fn=deployOneByOne release_id=69608524-a8fb-416a-8d7e-d6f4a194e2d3 type=app
t=2015-11-26T09:10:12+0000 lvl=info msg="received deployment event" app=worker fn=HandleDeployment deployment_id=3e0bf8d2-c660-47c0-97b0-d6f3eda54196 app_id=82965c5d-bd0c-4cac-8f5a-5fe44d6c5795 strategy=discoverd-meta status= type=app state=starting
t=2015-11-26T09:10:12+0000 lvl=info msg="received deployment event" app=worker fn=HandleDeployment deployment_id=3e0bf8d2-c660-47c0-97b0-d6f3eda54196 app_id=82965c5d-bd0c-4cac-8f5a-5fe44d6c5795 strategy=discoverd-meta status= type=app state=starting
t=2015-11-26T09:10:12+0000 lvl=info msg="received deployment event" app=worker fn=HandleDeployment deployment_id=3e0bf8d2-c660-47c0-97b0-d6f3eda54196 app_id=82965c5d-bd0c-4cac-8f5a-5fe44d6c5795 strategy=discoverd-meta status= type=app state=starting
t=2015-11-26T09:10:12+0000 lvl=info msg="waiting for 3 job up event(s)" app=worker fn=deployOneByOne release_id=69608524-a8fb-416a-8d7e-d6f4a194e2d3 type=app
t=2015-11-26T09:10:12+0000 lvl=info msg="got service meta event" app=worker fn=deployOneByOne release_id=69608524-a8fb-416a-8d7e-d6f4a194e2d3 states="map[10.69.3.3:1110:done 10.69.3.4:1110:done 10.69.3.5:1110:performing]"
t=2015-11-26T09:10:12+0000 lvl=info msg="got service meta event" app=worker fn=deployOneByOne release_id=69608524-a8fb-416a-8d7e-d6f4a194e2d3 states=map[]
t=2015-11-26T09:10:12+0000 lvl=eror msg="error waiting for job up events" app=worker fn=deployOneByOne release_id=69608524-a8fb-416a-8d7e-d6f4a194e2d3 err="timed out waiting for discoverd deployment (expected=3 actual=0)"
```

Somehow we got a state `map[]` when we had already gotten valid data..  this probably means we are reading a stale state from somewhere.. which should be impossible as we wait for the store to catchup before re-openning HTTP API.

josephglanville:
That wasn't the cause of the failure though, turns out the `Deployment` member of `DeployJob` doesn't have it's `DeployTimeout` member properly populated.

@lmars should we copy `DeployTimeout` from the app to the deployment record so we can easily return it from the deployments endpoint? I could just write some fancy SQL but I think copying it is nicer in this case as the deployment is a historic record and it seems more "correct" to record the `DeployTimeout` at the time the deployment was created.

josephglanville:
Discussed with @lmars, will be copying `deploy_timeout` from apps to deployments on insertion, happening in: #2183

josephglanville:
The migration is safe and tested, however it does result in all previous deployments having a deploy_timeout of 30 even though they may have used another value. This is probably not worth doing anything about, just noting for posterity.

josephglanville:
@titanous fixed the need for the `DisableKeepAlives`.
This is now ready for final review changes before merge.
The changes are mostly around changes to support the new discoverd mux and the introduction of a shutdown guard in the HTTP handler.

titanous:
LGTM!!

lmars:
LGTM :smile:

PR #2008, opened 2015-10-21
HA/recoverable log aggregation

titanous:
Replaces #1966

The `logaggregator` component is now highly available, the `host/logmux` package has been rewritten, and the `host/logbuf` package deleted.

Logs now have sequence numbers associated with them, the tuple of `(host_id, timestamp, seq)` (used as a "cursor") makes a given log line unique and can be used for deduplication. Time is assumed to only move forward on each host (no exact time coordination between hosts is required), which allows us to only care about comparing sequence numbers if the timestamp is the same (we may need to tweak this a bit to handle leap seconds).

Logs are saved by `flynn-host` into rotated per-app local files in a slightly modified syslog format.

When a host starts up, as soon as discoverd is available, it subscribes to the `logaggregator` service. When a new aggregator instance comes up, the current cursor for the host is determined. Local logs are then replayed for each app from the given cursor.

The aggregator sorts incoming log messages by timestamp, which avoids problems when multiple hosts are replaying old logs in parallel. This does mean that logs may be in slightly unexpected order if hosts have large amounts of clock skew, but it seems like a reasonable tradeoff for simplicity.

If there is a leader when an aggregator instance is booting up, a log buffer snapshot will be fetched from it, reducing the number of log messages that need to be replayed from individual hosts.


titanous:
@lmars Please give this a final review.

lmars:
LGTM, although would be good to rebase and see a successful CI build

Issue #992, opened 2015-02-12
Remove martini


titanous:
We've mostly purged martini from the codebase, with a few small users left:

- [x] `dashboard/api`
- [x] `router`
- [x] `appliance/postgresql/api`
- [ ] `pkg/cors`

PR #2009, opened 2015-10-22
router: Replace martini with httprouter

jvatic:
Refs #992


josephglanville:
This also fixes all of the responses to use JSONError, which in turn means it's automatically afforded the additional robustness/retries etc it offers.

LGTM

PR #2016, opened 2015-10-23
discoverd/server: Bump raft apply timeout to 30 seconds

### LINKED ISSUE #1899: discoverd raft "timed out enqueuing operation"
I deleted and recreated the 'terminus' app, and an old container stuck around running, which was preventing a new container from spawning. Terminating the container let the new one start fine.

https://gist.github.com/e376d4894c149da5f3b5 has the debug-info, prior to terminating the errant container.

titanous:
The current timeout of five seconds is clearly not high enough, as users are hitting it in the wild in otherwise normal situations.

30 seconds is the timeout that Consul, the primary user of the raft package uses.

Closes #1899 

/cc @benbjohnson


lmars:
LGTM

benbjohnson:
:+1:

PR #2021, opened 2015-10-23
Godeps: Remove go-bindata symlink test data

titanous:
Multiple users reported build issues with the recursive symlinks.


jvatic:
LGTM

PR #2025, opened 2015-10-24
cli: Route export/import

### LINKED ISSUE #1956: Add option for including routes when importing / exporting apps.
Currently there is no way to export / import the different routes belonging to an application. Extra routes added is not remembered.

josephglanville:
Allows optional importing of routes.

Closes #1956

### DIFF
--- a/cli/export.go
+++ b/cli/export.go
@@ -19,8 +19,10 @@ import (
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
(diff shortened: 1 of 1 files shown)


cli/export.go
```diff
@@ -19,8 +19,10 @@ import (
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
+	h "github.com/flynn/flynn/pkg/httphelper"
```
titanous: We've been calling this `hh` elsewhere.

titanous:
LGTM pending comment.

cli/export.go
josephglanville: Ok, that does sound better.

PR #2026, opened 2015-10-24
bootstrap: Only start a single scheduler

### LINKED ISSUE #1906: bootstrap: Race when starting schedulers
During bootstrap, [the controller formation is scaled](https://github.com/flynn/flynn/blob/master/bootstrap/manifest_template.json#L231-L240) and then [schedulers are started](https://github.com/flynn/flynn/blob/master/bootstrap/manifest_template.json#L241-L248) ([one per host](https://github.com/flynn/flynn/blob/master/bootstrap/run_app_action.go#L106-L108) since the scheduler is an omni process type), but this leads to a race where the first scheduler syncs with the cluster + controller before bootstrap has finished starting all the schedulers, and it itself starts schedulers, leading to too many schedulers running.

This isn't particularly problematic as the scheduler will eventually rectify and stop the surplus schedulers, but we should try and avoid this if possible, perhaps by bootstrap only starting one scheduler and letting the scheduler start the rest.

josephglanville:
Starting more than one scheduler during bootstrap can lead to a race
where more schedulers than are required are started.

Closes #1906

### DIFF
--- a/bootstrap/run_app_action.go
+++ b/bootstrap/run_app_action.go
@@ -103,14 +103,20 @@ func (a *RunAppAction) Run(s *State) error {
 			count = 1
 		}
 		hosts := s.ShuffledHosts()
-		if a.ExpandedFormation.Release.Processes[typ].Omni {
+		proc := a.ExpandedFormation.Release.Processes[typ]
+
+		// Special case the scheduler, we only start one scheduler and allow
+		// it to start the rest of the jobs to avoid spawning too many jobs.
+		if proc.Service == "controller-scheduler" {
+			count = 1
+		} else if proc.Omni {
 			count = len(hosts)
 		}
 		for i := 0; i < count; i++ {
 			host := hosts[i%len(hosts)]
 			config := utils.JobConfig(a.ExpandedFormation, typ, host.ID())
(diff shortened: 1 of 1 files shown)


bootstrap/run_app_action.go
```diff
@@ -103,14 +103,20 @@ func (a *RunAppAction) Run(s *State) error {
 			count = 1
 		}
 		hosts := s.ShuffledHosts()
-		if a.ExpandedFormation.Release.Processes[typ].Omni {
+		proc := a.ExpandedFormation.Release.Processes[typ]
+
+		// Special case the scheduler, we only start one scheduler and allow
+		// it to start the rest of the jobs to avoid spawning too many jobs.
+		if proc.Service == "controller-scheduler" {
+			count = 1
```
titanous: Is there a sane way to do this in the bootstrap JSON instead of hard-coding?

josephglanville: Possibly.. The way to to do it would probably be to not have the release be omni in the run-app section but have it merged in at deploy-app.

josephglanville:
Closing this for now until I can work out a pure JSON way to do it.

PR #2032, opened 2015-10-27
cli/config: Fix multiple cluster conflict error

titanous:
Previously all configs with conflicts and multiple clusters resulted in the conflict error even if there were not multiple conflicts.


lmars:
LGTM

PR #2037, opened 2015-10-27
scheduler: Retry SyncJobs with a small delay if it fails

### LINKED ISSUE #2031: scheduler: Error getting jobs list after host update
After a host update, during a periodic sync, the scheduler encounters this error:

```text
t=2015-10-26T15:15:38+0000 lvl=eror msg="error getting jobs list" component=scheduler fn=SyncJobs host.id=3bff2486 err="Get http://10.69.9.8:1113/host/jobs: read tcp 10.69.9.8:1113: use of closed network connection"
```
Instead of retrying, it treats all of the jobs on that host as gone and starts new ones (on that host!) which then causes there to be too many jobs, which results in other jobs being killed.

This is the cause of most of the recent `TestReleaseImages` test failures.

The solution is probably to retry a few times before treating the host as dead. See also #2004.

https://ci.flynn.io/builds/20151026150943-c66764cd

lmars:
Fixes #2031

### DIFF
--- a/controller/scheduler/scheduler.go
+++ b/controller/scheduler/scheduler.go
@@ -328,17 +328,24 @@ func (s *Scheduler) Run() error {
 	return nil
 }
 
-func (s *Scheduler) SyncJobs() {
+func (s *Scheduler) SyncJobs() (err error) {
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

PR #2040, opened 2015-10-29
util/packer: Wait 60s before running provisioning scripts on EC2

josephglanville:
The idea here is to try workaround the race that is causing the apt failures.
If it doesn't help will probably revert it but if it does I will try track down the actual root cause properly (it's just proven elusive).

### DIFF
--- a/util/packer/ubuntu-14.04/template.json
+++ b/util/packer/ubuntu-14.04/template.json
@@ -73,7 +73,13 @@
       "ssh_username": "ubuntu"
     }
   ],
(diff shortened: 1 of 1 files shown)


util/packer/ubuntu-14.04/template.json
```diff
@@ -76,6 +76,11 @@
   "provisioners": [
     {
       "type": "shell",
+      "inline": ["sleep 60"],
+      "only": ["amazon-ebs"]
```
lmars: Probably worth a comment in case it solves the problem and we keep it here

josephglanville:
Merged this for now, will revert if it doesn't seem to help.

Issue #2004, opened 2015-10-21
scheduler nukes everything if host is unregistered


titanous:
A single-node cluster run by @temujin9 suffered an (as yet unexplained) hiccup where the `flynn-host` service instance was temporarily expired in discoverd.

The scheduler saw this and decided it'd be a great time to kill everything, including postgres:

```
t=2015-10-21T17:35:36+0000 lvl=info msg="handling host event" component=scheduler fn=HandleHostEvent event.type=down
t=2015-10-21T17:35:36+0000 lvl=info msg="host is down, stopping job event stream" component=scheduler fn=HandleHostEvent event.type=down host.id=ip1010046117
t=2015-10-21T17:35:36+0000 lvl=info msg="unfollowing host" component=scheduler fn=unfollowHost host.id=ip1010046117
t=2015-10-21T17:35:36+0000 lvl=info msg="removing job" component=scheduler fn=unfollowHost host.id=ip1010046117 job.id=ip1010046117-9b7af9a9-67cd-499b-8abd-e8c9427c07f2
t=2015-10-21T17:35:36+0000 lvl=info msg="triggering rectify" component=scheduler key="{AppID:5cfa0f99-e32f-4d08-987c-3b0572118cc1 ReleaseID:aef8635a-f956-42f0-aec2-3c42619f88e3}"
t=2015-10-21T17:35:36+0000 lvl=info msg="removing job" component=scheduler fn=unfollowHost host.id=ip1010046117 job.id=ip1010046117-6db3a55a-9e89-4eff-b4e8-3d36e880cce3
t=2015-10-21T17:35:36+0000 lvl=info msg="triggering rectify" component=scheduler key="{AppID:ff8d2703-448d-42cf-addc-b6e738e0f349 ReleaseID:de86f3ca-4da2-40a5-95d7-f994e25a6f05}"
...
```

https://gist.github.com/temujin9/bc52b9f648ca2e13dc77#file-gistfile1-txt-L548-L550

I haven't looked at why the scheduler did this yet, but it doesn't seem like the correct behavior.

Cluster debug info: https://gist.github.com/80a36e8c54d49b6123e1

titanous:
Looks like this is the culprit: https://github.com/flynn/flynn/blob/master/controller/scheduler/scheduler.go#L584-L590

lmars:
[...]

PR #2041, opened 2015-10-29
scheduler: Implement host sync and checking

### LINKED ISSUE #1921: scheduler: Reconnect host event streams intelligently
Currently when a formation, host, and leader event stream gets disconnected, the scheduler attempts to reconnect. We need to do the same with job streams from a host instead of immediately giving up on the host.

As part of this, also remove the automatic host unfollow that happens if we can't connect to a host client when scheduling a job, which will involve needing to pick a different host for the job.

Also figure out how to correctly handle stopping a job when we can't connect to the host running the job to be stopped.

lmars:
Closes #1921.

Fixes part of #2004.

### DIFF
--- a/controller/scheduler/scheduler.go
+++ b/controller/scheduler/scheduler.go
@@ -1103,8 +1103,7 @@ func (s *Scheduler) startHTTPServer(port string) {
 func (s *Scheduler) tickSyncJobs(d time.Duration) {
 	logger.Info("starting sync jobs ticker", "duration", d)
 	go func() {
-		ch := time.Tick(d)
-		for range ch {
+		for range time.Tick(d) {
 			s.triggerSyncJobs()
 		}
 	}()
@@ -1113,8 +1112,7 @@ func (s *Scheduler) tickSyncJobs(d time.Duration) {
 func (s *Scheduler) tickSyncFormations(d time.Duration) {
 	logger.Info("starting sync formations ticker", "duration", d)
 	go func() {
-		ch := time.Tick(d)
-		for range ch {
+		for range time.Tick(d) {
 			s.triggerSyncFormations()
 		}
 	}()
@@ -1123,8 +1121,7 @@ func (s *Scheduler) tickSyncFormations(d time.Duration) {
 func (s *Scheduler) tickSyncHosts(d time.Duration) {
 	logger.Info("starting sync hosts ticker", "duration", d)
 	go func() {
-		ch := time.Tick(d)
-		for range ch {
+		for range time.Tick(d) {
 			s.triggerSyncHosts()
 		}
 	}()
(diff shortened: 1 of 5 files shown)


controller/scheduler/scheduler.go
```diff
@@ -1075,6 +1120,16 @@ func (s *Scheduler) tickSyncFormations(d time.Duration) {
 	}()
 }
 
+func (s *Scheduler) tickSyncHosts(d time.Duration) {
+	logger.Info("starting sync hosts ticker", "duration", d)
+	go func() {
+		ch := time.Tick(d)
+		for range ch {
```
titanous: No need to assign to `ch`, could just put it inline.

titanous:
LGTM. What did you do to test this?

lmars:
@titanous a very crude test, but I added the following code to `flynn-host` to temporarily mark it as down:

```go
go func() {
	for range time.Tick(30 * time.Second) {
		d.mtx.Lock()
		d.hb.Close()
		time.Sleep(7 * time.Second)
		d.hb, _ = disc.AddServiceAndRegisterInstance("flynn-host", d.inst)
		d.mtx.Unlock()
	}
}()
```

I then observed the following in the scheduler log:

```
t=2015-10-29T21:07:24+0000 lvl=info msg="handling host event" component=scheduler fn=HandleHostEvent event.type=down
t=2015-10-29T21:07:24+0000 lvl=warn msg="host service is down, marking as unhealthy and triggering host checks" component=scheduler host.id=host
t=2015-10-29T21:07:24+0000 lvl=info msg="triggering host checks" component=scheduler
t=2015-10-29T21:07:24+0000 lvl=info msg="performing host checks" component=scheduler fn=PerformHostChecks
t=2015-10-29T21:07:24+0000 lvl=info msg="getting status of unhealthy host" component=scheduler fn=PerformHostChecks host.id=host
t=2015-10-29T21:07:24+0000 lvl=info msg="host is now healthy" component=scheduler fn=PerformHostChecks host.id=host
t=2015-10-29T21:07:31+0000 lvl=info msg="handling host event" component=scheduler fn=HandleHostEvent event.type=up
t=2015-10-29T21:07:31+0000 lvl=info msg="host is up, starting job event stream" component=scheduler fn=HandleHostEvent event.type=up host.id=host
t=2015-10-29T21:07:32+0000 lvl=info msg="triggering host sync" component=scheduler
t=2015-10-29T21:07:32+0000 lvl=info msg="syncing hosts" component=scheduler fn=SyncHosts
```

So the host check actually passed (as the host was still responsive) and nothing untoward happened.

I am investigating the CI failures, but I think they may be unrelated.

lmars:
The failures were unrelated, merging.

PR #2049, opened 2015-10-30
deployer: Wait for job events before deleting formation during rollback

### LINKED ISSUE #1160: deployer: Rolling back the controller breaks the cluster
If the controller fails to deploy, the deployer will do a rollback which calls `PutFormation` and `DeleteFormation` in quick succession (i.e. [here](https://github.com/flynn/flynn/blob/master/controller/deployer/main.go#L154-L164)) , which leads to the scheduler potentially scaling itself down before it has started new schedulers to replace itself, leaving the cluster in a broken state.

jvatic:
Closes #1160 

- [x] Write test

### DIFF
--- a/controller/events.go
+++ b/controller/events.go
@@ -143,6 +143,8 @@ func (c *controllerAPI) GetEvent(ctx context.Context, w http.ResponseWriter, req
 }
 
 func (c *controllerAPI) Events(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	l, _ := ctxhelper.LoggerFromContext(ctx)
+	log := l.New("fn", "Events")
 	var app *ct.App
 	if appID := req.FormValue("app_id"); appID != "" {
 		data, err := c.appRepo.Get(appID)
@@ -155,15 +157,18 @@ func (c *controllerAPI) Events(ctx context.Context, w http.ResponseWriter, req *
 
 	if req.Header.Get("Accept") == "application/json" {
 		if err := listEvents(ctx, w, req, app, c.eventRepo); err != nil {
+			log.Error("error listing events", "err", err)
 			respondWithError(w, err)
 		}
 		return
 	}
 
 	if err := c.maybeStartEventListener(); err != nil {
+		log.Error("error starting event listener", "err", err)
 		respondWithError(w, err)
 	}
 	if err := streamEvents(ctx, w, req, c.eventListener, app, c.eventRepo); err != nil {
+		log.Error("error streaming events", "err", err)
 		respondWithError(w, err)
 	}
 }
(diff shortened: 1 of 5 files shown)


controller/events.go
```diff
@@ -155,15 +157,18 @@ func (c *controllerAPI) Events(ctx context.Context, w http.ResponseWriter, req *
 
 	if req.Header.Get("Accept") == "application/json" {
 		if err := listEvents(ctx, w, req, app, c.eventRepo); err != nil {
+			log.Error("Error listing events", "err", err)
```
lmars: We tend to not capitalise error messages

test/test_scheduler.go
```diff
@@ -353,6 +353,97 @@ func (s *SchedulerSuite) TestTCPApp(t *c.C) {
...
+	t.Assert(err, c.IsNil)
+	defer watcher.Close()
+
+	// get the current controller formation
+	formation, err := client.GetFormation(app.ID, release.ID)
+	t.Assert(err, c.IsNil)
+
+	currentReleaseID := release.ID
+
+	// create a controller deployment what will fail
```
lmars: s/what/that

test/test_scheduler.go
```diff
@@ -353,6 +353,97 @@ func (s *SchedulerSuite) TestTCPApp(t *c.C) {
...
+	t.Assert(err, c.IsNil)
+
+	events := make(chan *ct.DeploymentEvent)
+	eventStream, err := client.StreamDeployment(deployment, events)
+	t.Assert(err, c.IsNil)
+	defer eventStream.Close()
+
+	// wait for the deploy to complete (this doesn't wait for specific events
+	// due to the fact that when the deployer deploys itself, some events will
+	// not get sent)
```
lmars: Can probably replace this comment with "wait for the deploy to fail"

test/test_scheduler.go
```diff
@@ -353,6 +353,97 @@ func (s *SchedulerSuite) TestTCPApp(t *c.C) {
...
+			}
+			debugf(t, "got deployment event: %s %s", e.JobType, e.JobState)
+			switch e.Status {
+			case "complete":
+				t.Fatal("the deployment succeeded when it should have failed")
+			case "failed":
+				break loop
+			}
+		case <-time.After(2 * time.Minute):
+			t.Fatal("timed out waiting for the deploy to complete")
```
lmars: s/complete/fail

test/test_scheduler.go
```diff
@@ -353,6 +353,97 @@ func (s *SchedulerSuite) TestTCPApp(t *c.C) {
...
+	}
+
+	// wait for jobs to come back up
+	hosts, err := s.clusterClient(t).Hosts()
+	expected := map[string]map[string]int{
+		currentReleaseID: {
+			"web":       formation.Processes["web"],
+			"worker":    formation.Processes["worker"],
+			"scheduler": len(hosts),
+		},
```
lmars: Since the deploy is now sorted, won't it always deploy the scheduler and then fail at web, so only resurrect the scheduler jobs (i.e. this should just be waiting for new scheduler jobs)?

jvatic: It will deploy the scheduler, web, then fail at worker.

jvatic:
@lmars comments addressed and rebased with master

test/test_scheduler.go
lmars: Ah ok, for some reason I read `worker` as `web` earlier :unamused: 

Ok so in that case do we expect any new worker jobs to start? Shouldn't the old ones still be intact?

jvatic: There was a typo: `worker` was being set to the web process. And yes.

lmars:
LGTM

PR #2051, opened 2015-10-31
test: Handle closed event stream in TestDomainMigration

lmars:
This was previously panicking if the events channel closed.

I've also added some debug messages and refactored a little bit.

/cc @jvatic


josephglanville:
LGTM

jvatic:
:shipit:

PR #2053, opened 2015-11-01
script: When using multi-host cluster use separate log directories

josephglanville:
The new logmux system needs isolated log directories per `flynn-host` as it no longer namespaces the jobs the same way.


titanous:
LGTM

Issue #2042, opened 2015-10-29
Speedup builds by using hosts squid-deb-proxy if available


rikur:
I was able to speed up my builds a lot by modifying the cedarish Dockerfile as follows:

```
FROM ubuntu-debootstrap:14.04

# Derived from https://github.com/heroku/stack-images/blob/master/Dockerfile

RUN echo 'deb http://archive.ubuntu.com/ubuntu trusty main' >/etc/apt/sources.list &&\
    echo 'deb http://archive.ubuntu.com/ubuntu trusty-security main' >>/etc/apt/sources.list &&\
    echo 'deb http://archive.ubuntu.com/ubuntu trusty-updates main' >>/etc/apt/sources.list &&\
    echo 'deb http://archive.ubuntu.com/ubuntu trusty universe' >>/etc/apt/sources.list &&\
    # try to use hosts deb cache
    echo 'Acquire { Retries "0"; HTTP { Proxy "http://172.17.42.1:8000"; }; };' > /etc/apt/apt.conf.d/10_docker_hostproxy &&\
...
```

I will try to find the time to see that it doesn't break anything if the flynn VM is not running squid-deb-proxy.

rikur:
```docker images -q | xargs docker rmi -f; make clean && make```
The cache will serve 165mb of data for apt, which for me saves at least 10 minutes of downloading time and ever more building time, since cedarish builds faster and remaining tasks are not blocked by it for as long.

lmars:
@rikur can we not rely on the Docker build cache?

rikur:
@lmars: this will cache the .deb and .pkg.gz files and speed up the initial build as well, mostly thanks to caching the sources for apt-get update (and that many Docker images are built in parallel). Docker cache wasn't being utilized due to Dockerfiles being very different, part of this PR would fix that.

Considering that most development is meant(?) to be done on the Vagrant image, another option would be to bake in the Docker build caches (and repos for buildpacks).

My change doesn't brake anything though, it simply tries to use a squid proxy - if available - on the Docker host. It doesn't break even if there is no cache available.

I will finish this up, but if you feel it's too hackish then don't apply it :)

lmars:
@rikur if it helps your use case then I have no issue with it, just wanted to understand the reasoning, happy to see it merged!

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

PR #2058, opened 2015-11-02
Install squid-deb-proxy during 'vagrant up' and try to use it in ever…

rikur:
…y Debian based Dockerfile. Silently falls back to not using the proxy if it's unreachable.

A proposal for #2042 

Things to consider:
 * Caching won't work if the gateway is not the default 172.17.42.1 for the containers
 * I only whitelisted the currently used apt sources, new ones need to be added when needed

### DIFF
--- a/appliance/postgresql/Dockerfile
+++ b/appliance/postgresql/Dockerfile
@@ -2,9 +2,12 @@ FROM ubuntu-debootstrap:14.04
 
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN apt-get update &&\
-    apt-get dist-upgrade -y &&\
-    apt-get -qy --fix-missing --force-yes install language-pack-en &&\
+# Try to use the Hosts squid-deb-proxy and fall back if it's unreachable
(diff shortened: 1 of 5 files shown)


appliance/postgresql/Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM ubuntu-debootstrap:14.04
 
 ENV DEBIAN_FRONTEND noninteractive
 
+# Try to use the Hosts squid-deb-proxy and fall back if it's unreachable
+RUN (echo 'Acquire { Retries "0"; HTTP { Proxy "http://172.17.42.1:8000"; }; };' > /etc/apt/apt.conf.d/10_docker_hostproxy && apt-get update -qq 2>&1 | grep -q 'Unable to connect to 172.17.42.1:8000' && rm /etc/apt/apt.conf.d/10_docker_hostproxy) || true
```
titanous: This should be combined with the `RUN` below to avoid creating another layer, and the config file should be removed at the end of the run so that it doesn't end up in the final image.

rikur:
Great idea, I will push an improvement ASAP

rikur:
I took a slightly different approach and unified the beginning of all the Dockerfiles to use Docker image cache as well.

titanous:
This looks good! Please squash all of the commits together, add a subsystem prefix of `all: ` to the commit message, and [sign the commit](https://flynn.io/docs/contributing#developer%E2%80%99s-certificate-of-origin). You can then rebase on master and force-push the result.

rikur:
Ok don't apply this, binds to same port as discoverd.

PR #2077, opened 2015-11-04
worker: Always scale old formations down after deployment

lmars:
The deployer previously only scaled the old formation down if there were running jobs for the old release, so if the jobs were crashed at the time, the formation would stay scaled up and the jobs would be restarted.


titanous:
LGTM

PR #2088, opened 2015-11-05
Fix backup restore singleton flag

titanous:
This ensures that a backup from a multi-node cluster restores correctly into a single node cluster.


josephglanville:
LGTM

lmars:
LGTM

PR #2090, opened 2015-11-06
script: Fix multi-host router IP to match new IPs

josephglanville:
When moving to the new IP addresses this address was missed.

### DIFF
--- a/script/run-integration-tests
+++ b/script/run-integration-tests
@@ -100,7 +100,7 @@ main() {
     args+=("--stream")
   fi
   if [[ "${size}" -gt "1" ]]; then
-    args+=("--router-ip=198.51.100.1")
+    args+=("--router-ip=192.0.2.200")
   fi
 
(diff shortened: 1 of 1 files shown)


josephglanville:
This doesn't seem to be the only thing wrong.. when I try run the multi-host integration suite on master + this fix on master I am getting about 70 failures. Something is very wrong but I don't know what precisely.

titanous:
I don't think we changed the IP on CI, just the `bootstrap-flynn` script which is only used for local development.

josephglanville:
Yeah this is when running `run-integration-tests` locally, when then calls `bootstrap-flynn`

josephglanville:
Ping @lmars 

I think this is required to work correctly with https://github.com/flynn/flynn/commit/8af0fdf5575dc200e0640d80520e569bd709bd3a

lmars:
LGTM

PR #2094, opened 2015-11-06
cli: support `flynn cluster add` when git is not present

### LINKED ISSUE #1835: Support cluster add without git present
Currently `flynn cluster add` requires git to be installed, we should have an option like `--no-git` that skips the git configuration and is suggested if the command fails due to a missing git binary.

imjorge:
Closes #1835


titanous:
LGTM, thanks!

PR #2103, opened 2015-11-07
status: Require key for status endpoint from public IP

### LINKED ISSUE #1787: Require key for status endpoint from public IP
When requesting the status endpoint from a public IP, it should require a key in the URL or Authorization header to avoid the (very minor) information disclosure risk.

imjorge:
Closes #1787

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/subtle"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


titanous:
We also need to generate and set the `AUTH_KEY` during bootstrap, like this:

https://github.com/flynn/flynn/blob/master/bootstrap/manifest_template.json#L119-L124
https://github.com/flynn/flynn/blob/master/bootstrap/manifest_template.json#L329

status/status.go
```diff
@@ -9,16 +10,92 @@ import (
...
+		n(net.ParseCIDR("172.16.0.0/12")),   // 172.16.0.0 - 172.31.255.255
+		n(net.ParseCIDR("192.168.0.0/16")),  // 192.168.0.0 - 192.168.255.255
+		n(net.ParseCIDR("127.0.0.0/8")),     // 127.0.0.0 - 127.255.255.255
+		n(net.ParseCIDR("fd00::/8")),        // fdxx:xxxx:xxxx...
+		n(net.ParseCIDR("::1/128")),         // ::1
+		n(net.ParseCIDR("192.0.2.0/24")),    // TEST-NET (for containers)
+		n(net.ParseCIDR("198.51.100.0/24")), // TEST-NET (for containers)
+		n(net.ParseCIDR("203.0.113.0/24")),  // TEST-NET (for containers)
+		n(net.ParseCIDR("100.64.0.0/10")),   // for containers
+	}
```
lmars: It is conventional to prefix functions which are expected to always work with `must` (e.g. like [regexp.MustCompile](https://golang.org/pkg/regexp/#MustCompile)), so I would probably make this:

```go
whitelist := []*net.IPNet{
	mustParseCIDR("10.0.0.0/8"),
	mustParseCIDR("172.16.0.0/12"),
	...
}
```

status/status.go
```diff
@@ -9,16 +10,92 @@ import (
...
+	return &statusHandler{h, k, whitelist}
+}
+
+func (s *statusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if s.k != "" {
+		var addr string
+		if fwdFor, ok := r.Header["X-Forwarded-For"]; ok && len(fwdFor) == 1 {
```
titanous: Do `r.Header.Get("X-Forwarded-For")` instead.

status/status.go
```diff
@@ -9,16 +10,92 @@ import (
...
+		}
+		if addr == "" {
+			if clientIP, _, err := net.SplitHostPort(r.RemoteAddr); err == nil {
+				addr = clientIP
+			}
+		}
+
+		var authed bool
+		if addr != "" {
+			if ip := net.ParseIP(addr); ip != nil {
```
titanous: Could probably drop the `addr != ""` check since `ip` will be `nil` in that case.

status/status.go
```diff
@@ -9,16 +10,92 @@ import (
...
+			}
+		}
+
+		if !authed {
+			_, password, _ := utils.ParseBasicAuth(r.Header)
+			if password == "" {
+				password = r.URL.Query().Get("key")
+			}
+			if len(password) == len(s.k) && subtle.ConstantTimeCompare([]byte(password), []byte(s.k)) == 1 {
+				authed = true
```
titanous: You can assign `authed` directly: `authed = len(password)...`

status/status.go
imjorge: Was there just for a case where modifications to the router make it not crunch multiple `X-Forwarded-For` headers into a single one. ignore that then?

titanous: Yeah, you can ignore that.

imjorge:
Implemented reviewer comments.

titanous:
LGTM

Issue #1100, opened 2015-02-27
Pin `flynn-host download` to specific components


lmars:
Currently `install-flynn` downloads a specific `flynn-host` binary, but then running `flynn-host download` downloads the latest components. It should instead download the components from the release it came from.

The following changes are necessary:

* Export `version.json` with the release version in the filename (e.g. `version-v20150226.0.json`) - this requires `script/export-components` to accept a version option and add it to the filename [here](https://github.com/flynn/flynn/blob/master/script/export-components#L40)
* Update `pinkerton.PullImages` to take an optional version string to construct the name of the file to download (i.e. [here](https://github.com/flynn/flynn/blob/master/pinkerton/context.go#L178)) and set that in [runDownload](https://github.com/flynn/flynn/blob/master/host/cli/download.go#L61-L69)
* Update [downloadGzippedFile](https://github.com/flynn/flynn/blob/master/host/cli/download.go#L97) to construct the filenames using the version

`flynn-host update` should continue to get the latest components.

PR #2105, opened 2015-11-08
script: Support installing explicit version

lmars:
I forgot to add this as part of closing #1100.

This means an explicit version can be installed like:

```bash
sudo FLYNN_VERSION="v20151104.1" bash < <(curl -fsSL https://dl.flynn.io/install-flynn)
```

/cc @temujin9


titanous:
LGTM

josephglanville:
LGTM

PR #2107, opened 2015-11-09
cli: add app info command

### LINKED ISSUE #88: App info command
A new app info command should show at least the git and web URL (first HTTP route) for the app.

imjorge:
Closes #88

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -64,6 +64,24 @@ Examples:
 	f488cfb478f54edea497bf6347c2eb80  postgres
 	9d5be7be873c41b9898032c08aa87597  controller
 `)
+
+	register("info", runInfo, `
+usage: flynn info
+
+Show information for an app.
+
+Examples:
+
+	$ flynn info
+	=== example
+	Git URL:  https://git.dev.localflynn.com/example.git
+	Web URL:  http://example.dev.localflynn.com
+
+	$ flynn -a example info
+	=== example
+	Git URL:  https://git.dev.localflynn.com/example.git
+	Web URL:  http://example.dev.localflynn.com
+`)
 }
 
 func runCreate(args *docopt.Args, client *controller.Client) error {
@@ -137,5 +155,38 @@ func runApps(args *docopt.Args, client *controller.Client) error {
 	for _, a := range apps {
 		listRec(w, a.ID, a.Name)
 	}
+	return nil
+}
+
+func runInfo(_ *docopt.Args, client *controller.Client) error {
+	appName := mustApp()
+
+	fmt.Println("===", appName)
+
+	w := tabWriter()
+	defer w.Flush()
+
+	if release, err := client.GetAppRelease(appName); err == nil || err == controller.ErrNotFound {
+		if err == controller.ErrNotFound || release.Env["SLUG_URL"] != "" {
+			listRec(w, "Git URL:", gitURL(clusterConf, appName))
+		}
+	} else {
+		return err
+	}
+
+	if routes, err := client.RouteList(appName); err == nil {
+		for _, k := range routes {
+			if k.Type == "http" {
+				route := k.HTTPRoute()
+				protocol := "https"
+				if route.TLSCert == "" {
+					protocol = "http"
+				}
+				listRec(w, "Web URL:", protocol+"://"+route.Domain)
+				break
+			}
+		}
+	}
+
 	return nil
 }
--- a/cli/main.go
+++ b/cli/main.go
@@ -45,6 +45,7 @@ Commands:
 	create      create an app
 	delete      delete an app
 	apps        list apps
+	info        show app information
 	ps          list jobs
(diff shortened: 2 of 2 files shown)


imjorge:
Perhaps we should merge existing "app" commands under `flynn app`:
- `flynn app create`
- `flynn app delete`
- `flynn app list`
- `flynn app` (could be used for listing info or, more verbose, `flynn app info`)

Or do that in another issue/PR.

cli/app.go
```diff
@@ -139,3 +157,37 @@ func runApps(args *docopt.Args, client *controller.Client) error {
 	}
 	return nil
 }
+
+func runInfo(_ *docopt.Args, client *controller.Client) error {
+	appName := mustApp()
+
+	log.Printf("=== %s\n", appName)
```
lmars: This could be `fmt.Println("===", appName)`

cli/app.go
```diff
@@ -139,3 +157,37 @@ func runApps(args *docopt.Args, client *controller.Client) error {
...
+
+	if routes, err := client.RouteList(appName); err == nil {
+		for _, k := range routes {
+			if k.Type == "http" {
+				route := k.HTTPRoute()
+				protocol := "https"
+				if route.TLSCert == "" {
+					protocol = "http"
+				}
+				listRec(w, "Web Url:", protocol+"://"+route.Domain)
```
lmars: s/Url/URL

lmars:
@imjorge please open a separate issue for namespacing under `flynn app`

imjorge:
@lmars implemented comments

imjorge:
created #2127 to address namespacing "app" commands under `flynn app`

cli/app.go
```diff
@@ -139,3 +157,37 @@ func runApps(args *docopt.Args, client *controller.Client) error {
...
+
+	fmt.Println("===", appName)
+
+	w := tabWriter()
+	defer w.Flush()
+
+	if remotes, err := gitRemotes(); err == nil {
+		for _, v := range remotes {
+			if v.Name == appName {
+				listRec(w, "Git Repo:", gitURL(v.Cluster, v.Name))
```
lmars: Since this is a URL, call it `Git URL` (a "git repo" to me is a directory containing `.git`)

lmars:
@imjorge after trying this out, I think it would be useful to always show the two fields but with something like `n/a` if they don't apply (and a reason why if possible), e.g.:

```
# when not in a git repository
$ flynn -a controller info
=== controller
Git URL:  n/a (current directory not a git repository)
Web URL:  https://controller.example.com

$ flynn -a blobstore info
=== blobstore
Git URL:  n/a (current directory not a git repository)
Web URL:  n/a (no routes configured)


# when in a git repository
$ flynn -a controller info
=== controller
Git URL:  n/a (app not in git remotes)
Web URL:  https://controller.example.com

$ flynn -a my-app info
=== my-app
Git URL:  https://git.example.com/my-app.git
Web URL:  https://controller.example.com
```

If the reasons prove too fiddly to add then I don't feel strongly that they are necessary, but I think at least always showing all the fields is useful, rather than the current:

```
$ flynn -a blobstore info
=== blobstore
```

cli/app.go
imjorge: > Since this is a URL, call it Git URL (a "git repo" to me is a directory containing .git)

used the [terminology from heroku toolbelt](https://devcenter.heroku.com/articles/using-the-cli#app-commands). Wanna me to change it?

imjorge:
@lmars i think that would unnecessarily complicate the code as more checks would have to be put in place.

/cc @titanous

imjorge:
furthermore, the output of this command should list capabilities. if an app has an HTTP route it should list its Web URL. Otherwise, it wont be shown. Always adding those lines, even for apps for which some of them do not make sense, is only noise.

cli/app.go
titanous: Yes, `Git URL` is correct.

cli/app.go
```diff
@@ -139,3 +157,37 @@ func runApps(args *docopt.Args, client *controller.Client) error {
...
+
+func runInfo(_ *docopt.Args, client *controller.Client) error {
+	appName := mustApp()
+
+	fmt.Println("===", appName)
+
+	w := tabWriter()
+	defer w.Flush()
+
+	if remotes, err := gitRemotes(); err == nil {
```
titanous: This should generate the git URL, not read it from the repo.

imjorge: Every app in Flynn has a Git URL?

titanous: Every app can be technically deployed via git, currently there isn't a very tight coupling. The only thing we could do is not show the git repo if slugrunner is not in use for the current release (if there are no releases we should show it).

imjorge: Done.

titanous:
LGTM

PR #2112, opened 2015-11-09
bootstrap: Set PG* env vars on postgres release

josephglanville:
Allows usage of the `flynn pg` tools with the default `postgres` database.


lmars:
LGTM

PR #2114, opened 2015-11-10
docs: Fix nav clutter on installation pages

jvatic:


jvatic:
Actually, it makes more sense for this to be the default toc level

PR #2115, opened 2015-11-11
controller: Add cluster backup endpoint

### LINKED ISSUE #2095: Move full cluster backup logic into the controller
Add an endpoint in the controller that acts the same as `flynn cluster backup`.

jvatic:
Closes #2095

### DIFF
--- a/cli/cluster.go
+++ b/cli/cluster.go
@@ -17,6 +17,7 @@ import (
 	cfg "github.com/flynn/flynn/cli/config"
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/pkg/backup"
 	"github.com/flynn/flynn/pkg/shutdown"
 )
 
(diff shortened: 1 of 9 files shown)


cli/cluster.go
```diff
@@ -337,55 +338,12 @@ func runClusterBackup(args *docopt.Args) error {
...
-			App:       app,
-			Release:   release,
-			Artifact:  artifact,
-			Processes: formation.Processes,
-		}
+	if bar != nil {
+		dest = io.MultiWriter(dest, bar)
 	}
-	if err := tw.WriteJSON("flynn.json", data); err != nil {
+	if err := clusterbackup.RunBackup(client, dest); err != nil {
```
lmars: I think the package should be `backup` and the function `Run`, so this becomes `backup.Run`, WDYT?

lmars:
Is it worth adding a test for this? Hit `/backup` and check we get a tar with the relevant files in it?

cli/cluster.go
jvatic: Yeah, that's simpler.

jvatic:
- [ ] Add integration test

jvatic:
@lmars comments addressed

jvatic:
- [ ] `Content-Disposition` header should be set
- [ ] should be able to use `key` param with this endpoint

lmars:
LGTM

PR #2123, opened 2015-11-11
logaggregator: do not send backlog when requested lines are zero

### LINKED ISSUE #2120: logaggregator: Full backlog is returned when lines=0
Full backlog is returned when `lines=0`. The reason being `backlog` is being set [here](https://github.com/flynn/flynn/blob/master/logaggregator/api.go#L69) without checking that `lines > 0`.

imjorge:
Closes #2120

### DIFF
--- a/logaggregator/api_test.go
+++ b/logaggregator/api_test.go
@@ -172,6 +172,70 @@ func (s *LogAggregatorTestSuite) TestAPIGetLogFollow(c *C) {
 	}
 }
 
+func (s *LogAggregatorTestSuite) TestAPIGetLogFollowWithoutBacklog(c *C) {
+	appID := "test-app"
+	msg1 := newMessageForApp(appID, "web.1", "log message 1")
+	msg2 := newMessageForApp(appID, "web.2", "log message 2")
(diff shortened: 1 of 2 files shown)


logaggregator/api_test.go
```diff
@@ -172,6 +172,70 @@ func (s *LogAggregatorTestSuite) TestAPIGetLogFollow(c *C) {
...
+		select {
+		case l := <-lines:
+			if l.err != nil {
+				return "", fmt.Errorf("could not read log output: %s", l.err)
+			}
+			return l.text, nil
+		case <-time.After(1 * time.Second):
+			return "", errors.New("timed out waiting for log output")
+		}
+	}
```
titanous: Looks like this re-uses all of the setup from `TestAPIGetLogFollow`, should we abstract into a setup method?

imjorge: We may. I will do it.

--jmf

imjorge:
Merged tests `TestAPIGetLogFollow` and `TestAPIGetLogFollowWithoutBacklog`.

titanous:
LGTM, thanks!

PR #2129, opened 2015-11-14
WIP: Discoverd client improvements

josephglanville:
Allows the discoverd client to connect to multiple peers and retry if the first peer is unavailable.

It will retry each server for up to a second every 100ms before moving to the next server.
It's possible to make this much better with a background goroutine to keep track of which servers are up so that requests always have low latency if there is a server available.

The main motivation for this is to allow `flynn-host` and `flynn-postgres` to not experience noticeable interuptions during `discoverd` deployments/updates but it should provide additional robustness in the face of network errors etc.

TODO
- [ ] Populate `DISCOVERD_PEERS` default environment variable with all of the peer URLs
- [ ] Sort the peer list so that the local server is prefered
- [x] Keep track of the current leader and direct writes there to avoid proxying

### DIFF
--- a/discoverd/client/client.go
+++ b/discoverd/client/client.go
@@ -1,261 +1,112 @@
 package discoverd
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"os"
 	"strings"
-	"time"
+	"sync"
 
-	"github.com/flynn/flynn/pkg/dialer"
 	"github.com/flynn/flynn/pkg/httpclient"
-	hh "github.com/flynn/flynn/pkg/httphelper"
 	"github.com/flynn/flynn/pkg/stream"
 )
 
-type Service interface {
-	Leader() (*Instance, error)
-	Instances() ([]*Instance, error)
-	Addrs() ([]string, error)
-	Leaders(chan *Instance) (stream.Stream, error)
-	Watch(events chan *Event) (stream.Stream, error)
-	GetMeta() (*ServiceMeta, error)
-	SetMeta(*ServiceMeta) error
-	SetLeader(string) error
-}
-
 var ErrTimedOut = errors.New("discoverd: timed out waiting for instances")
 
-type Client struct {
-	c *httpclient.Client
+type Config struct {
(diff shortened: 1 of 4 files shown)


josephglanville:
@lmars What would be a good way to get the DISCOVERD_PEERS variable to `flynn-host` so that it can  properly set it as a default env var?

titanous:
Writes should go to the raft leader (currently a redirect is in place for this, as non-leaders can't accept writes).

josephglanville:
It's directing writes to the leader now, I will change it to use a watch instead of polling to update the leader though.

josephglanville:
Changes to the discoverd update process have made this unnecessary for now but I am leaving it open as I work on a new approach that should be better than my first pass.

Specifically the new strategy will be as as follows:

1. The client will be initialised with an initial set of nodes
2. On performing the initial request the first server that responds successfully (even if it responds with a redirect because of it being a non-master and trying to service a write) then we will "pin" this server for future requests.
3. When doing further requests we will first attempt to use the pinned server, if that fails finding and then pinning a new server to service reads.
4. When doing a write if we get a redirect we will track the last redirect we received and mark this as the leader so we can avoid the redirects where possible.

The client will always be cluster aware but some requests (for now just Shutdown) will take an address of a specific node.

josephglanville:
Closing for now to prevent CI building stuff, will re-open when it's ready for people to look at.

titanous:
You've probably considered this, but I'd expect something along these lines:

- All reads start pointed towards the local instance, to prevent extra cross-network traffic.
- If the local instance is unavailable, reads go elsewhere and return to it when the local instance comes back.
- The leader is tracked through a stream of some type.

josephglanville:
Yep, just implemented falling back to preferred server after a TTL (default 60s).

As for leader tracking I ended up implementing it by updating the leader whenever we get a redirect.
This avoids the background goroutine and should only ever hit the redirect whenever the leader changes.

josephglanville:
Hmm unfortunately I can't reopen this because I force-pushed. :( I will open a new one.

PR #2130, opened 2015-11-14
Makefile: Fix git tag determination

lmars:
Using `git describe --exact-match` was just returning the first tag seen
for a commit. From `git help describe`:

    If an exact match is found, its name will be output and searching
    will stop.

So at the current master which is tagged with both v20151113.0 and
v20151114.0, GIT_TAG was being set to v20151113.0, whereas we want it to
be the latest value of v20151114.0:

    $ git tag --points-at master
    v20151113.0
    v20151114.0
    $ git describe --tags --exact-match --match "v*"
    v20151113.0

Using `git tag --sort | tail -n1` fixes this:

    $ git tag --list "v*" --sort "v:refname" --points-at HEAD | tail -n1
    v20151114.0


josephglanville:
LGTM

PR #2132, opened 2015-11-15
logaggregator: Use retry dialer in client

lmars:
This prevents errors like the on in CI build https://ci.flynn.io/builds/20151115010657-73c9f645 where the test got a connection refused as it was connecting to a logaggregator which just went down (but whose address was still being returned by discoverd).


titanous:
LGTM

PR #2135, opened 2015-11-15
controller: Add App.DeployTimeout

lmars:
Both the controller and postgres can take longer than the default 30s to deploy, and so we have some custom timeouts spread amongst the code, but the default timeout is used in cases like `flynn -a controller env set`, so is susceptible to timeouts.

I have added `App.DeployTimeout` which is the timeout in seconds, and added `Deployment.DeployTimeout` which is copied from the app so the client knows what the timeout is when creating a deployment without also looking up the app.

I have also made the schema change as a new migration now that we have a stable release :smile:

### DIFF
--- a/bootstrap/manifest_template.json
+++ b/bootstrap/manifest_template.json
@@ -213,7 +213,8 @@
     "app": {
       "name": "controller",
       "meta": {"flynn-system-app": "true"},
-      "strategy": "one-by-one"
+      "strategy": "one-by-one",
+      "deploy_timeout": 120
     }
   },
   {
@@ -223,7 +224,8 @@
     "app": {
       "name": "postgres",
       "meta": {"flynn-system-app": "true"},
-      "strategy": "postgres"
+      "strategy": "postgres",
+      "deploy_timeout": 120
     }
   },
   {
--- a/cli/release.go
+++ b/cli/release.go
@@ -18,7 +18,7 @@ func init() {
 	register("release", runRelease, `
 usage: flynn release [-q|--quiet]
        flynn release add [-t <type>] [-f <file>] <uri>
-       flynn release show [<id>]
+       flynn release show [--json] [<id>]
 
 Manage app releases.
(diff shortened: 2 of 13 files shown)


lmars:
So I just realised that, although I added a migration for the schema change, cluster updates won't pick up the 120s timeouts I have added in the bootstrap manifest. How should we be managing data migrations? Just put it in the same migration as the schema change?

josephglanville:
I think doing it in the migration is reasonable, find the postgres/controller apps, set it to 120 if null.

titanous:
Yeah, in the migration.

lmars:
Ok I added the data change to the migration, and I've also ensured `DeployAppRelease` never uses a zero timeout (which will happen for deploys during an update before the migration has run).

I added a `--json` flag to `flynn release show`, then verified everything works as expected by doing the following:

* bootstrap a cluster using master
* build a controller image using this branch
* create a release using the new image:

```
$ flynn -a controller release show --json | jq -r 'del(.id)' > controller.json
$ NEW_ID=$(jq -r '."flynn/controller"' version.json)
$ flynn -a controller release add -f controller.json "https://dl.flynn.io/tuf?name=flynn/controller&id=$NEW_ID"
Created release 4ce7eb40-e7ed-47ca-bd0a-484e9e2a68b9.
```

* Check the deploy timeouts:

```
$ flynn -a controller pg psql -- --command 'SELECT name, deploy_timeout FROM apps'                                                                   
     name      | deploy_timeout 
---------------+----------------
 flannel       |             30
 discoverd     |             30
 blobstore     |             30
 router        |             30
 gitreceive    |             30
 logaggregator |             30
 taffy         |             30
 dashboard     |             30
 status        |             30
 postgres      |            120
 controller    |            120
```

@titanous @josephglanville please review.

josephglanville:
LGTM.

`flynn release show --json` will be super useful too.

PR #2137, opened 2015-11-16
test/runner: Increase timeout for integration tests to 25m

josephglanville:
The margin for tests taking a little longer is getting pretty thin with some new tests that take longer (domain migration, discoverd deployment etc) the timeout needs to be increased a bit to cover this.
It may be possible to speed the tests up by increasing concurrency also.


lmars:
LGTM

PR #2142, opened 2015-11-17
discoverd: client keep headers on redirect

### LINKED ISSUE #1880: discoverd: Client does not maintain headers through a proxy redirect
Redirects are handled transparently by Go ([here](https://github.com/golang/go/blob/master/src/net/http/client.go#L351-L373)) but headers are not copied to the new request (see https://github.com/golang/go/issues/4800), meaning for example that stream requests (which set `Accept: text/event-stream`) which are redirected by a proxy server don't appear as stream requests to the destination server (causing issues like https://github.com/flynn/flynn/issues/1855).

imjorge:
Closes #1880

### DIFF
--- a/discoverd/client/client.go
+++ b/discoverd/client/client.go
@@ -48,7 +48,8 @@ func NewClientWithURL(url string) *Client {
 		c: &httpclient.Client{
 			URL: url,
 			HTTP: &http.Client{
-				Transport: &http.Transport{Dial: dialer.Retry.Dial},
+				Transport:     &http.Transport{Dial: dialer.Retry.Dial},
+				CheckRedirect: redirectPreserveHeaders,
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


imjorge:
Currently `Authorization` header is copied. Spec is not clear about this. curl does not pass that header unless explicitly set:

> When authentication is used, curl only sends its credentials to the initial host. If a redirect takes curl to a different host, it won't be able to intercept the user+password. See also --location-trusted on how to change this.
> 
>  --location-trusted
> (HTTP/HTTPS) Like -L, --location, but will allow sending the name + password to all hosts that the site may redirect to. This may or may not introduce a security breach if the site redirects you to a site to which you'll send your authentication info (which  is  plain-text in the case of HTTP Basic authentication).

titanous:
LGTM, thanks!

PR #2145, opened 2015-11-17
Add raft log cache

benbjohnson:
This commit wraps the log store in a raft.LogCache to improve I/O performance.

Benchmark tests with the log cache show about a 33% improvement in throughput on my local SSD, however, more importantly, the disk I/O seems to be much more consistent. With the log cache, it was consistently writing about 70MB/sec and without the log cache it would write anywhere from 8-50MB/sec The performance improvement should be more significant on spinning disk.

Signed-off-by: Ben Johnson <benbjohnson@yahoo.com>

### DIFF
--- a/discoverd/server/store.go
+++ b/discoverd/server/store.go
@@ -170,6 +170,13 @@ func (s *Store) Open() error {
 	}
 	s.stableStore = stableStore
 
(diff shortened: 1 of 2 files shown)


titanous:
LGTM

Issue #2030, opened 2015-10-26
non-leader schedulers attempt to handle "crashed" jobs


titanous:
[This code](https://github.com/flynn/flynn/blob/dc83867a0fccbf07f744734f5737bf4dedca1ad1/controller/scheduler/scheduler.go#L952-L960)  appears to be wrong. It is run no matter if the scheduler is a leader or not, so it doesn't know if the job is supposed to be stopping (for instance it might not have an updated formation yet that the leader is reacting to).

It's possible that I'm missing something about how this is supposed to work, but it looks wrong to me.

josephglanville:
This eventually calls `scheduleJobStart` to start the replacement job which has an `isLeader` check on it.
That said it does seem like non-leader instances do end up doing a lot of stuff that should probably be only done by the leader.

titanous:
Yeah, I saw that. I'm more worried about a job being marked crashed in the controller by a random non-leader scheduler when it was intentionally killed by the leader.

josephglanville:
Yeah, that was part of the stuff the "should probably be only done by the leader". :P

Maybe I miss-understand the design decisions but it seems to me anything that mutates state (either in the db or on hosts) should be carried out by the leader, basically the non-leader instances should basically sit idle waiting for election.

titanous:
The original intention was to prevent events from being lost. But I think that can be remedied by having the leader look at the jobs table occasionally and sync with reality.

lmars:
> This code appears to be wrong. It is run no matter if the scheduler is a leader or not, so it doesn't know if the job is supposed to be stopping (for instance it might not have an updated formation yet that the leader is reacting to).

As @josephglanville said, that code is responsible for scheduling a job restart, which will only actually schedule a new job if the scheduler is the leader, and also [marks the job as crashed in the internal state](https://github.com/flynn/flynn/blob/dc83867a0fccbf07f744734f5737bf4dedca1ad1/controller/scheduler/scheduler.go#L1025).

> I'm more worried about a job being marked crashed in the controller by a random non-leader scheduler when it was intentionally killed by the leader

[...]

PR #2153, opened 2015-11-17
controller: Convert from enums to tables

josephglanville:
Enums while great have proven to be somewhat troublesome when it comes to working with them in migrations.
Instead we will use single column tables which have several advantages, chief among them the ability to enumerate all possible values without the need to hardcode this value outside of the schema.

### DIFF
--- a/controller/schema.go
+++ b/controller/schema.go
@@ -186,5 +186,26 @@ $$ LANGUAGE plpgsql`,
 		`UPDATE apps SET deploy_timeout = 120 WHERE name = 'controller'`,
 		`UPDATE apps SET deploy_timeout = 120 WHERE name = 'postgres'`,
 	)
+	m.Add(4,
+		`CREATE TABLE deployment_strategies (name text PRIMARY KEY)`,
+		`INSERT INTO deployment_strategies (name) VALUES
+			('all-at-once'), ('one-by-one'), ('postgres')`,
+		`ALTER TABLE apps ALTER COLUMN strategy TYPE text`,
+		`ALTER TABLE apps ALTER COLUMN strategy SET DEFAULT 'all-at-once'`,
+		`ALTER TABLE apps ADD CONSTRAINT apps_strategy_fkey FOREIGN KEY (strategy) REFERENCES deployment_strategies (name)`,
+		`ALTER TABLE deployments ALTER COLUMN strategy TYPE text`,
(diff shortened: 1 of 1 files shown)


titanous:
Have you tested the migration against a cluster that has data in it?

josephglanville:
I tested the migration steps manually against a cluster with data in it.
I couldn't work out how to update to a development build though, that would be a better test.

The reason why this migration is safe is that the `ALTER COLUMN` effectively runs a cast from enum to text. The migration would fail and rollback if the foreign key constraint couldn't be validated because the `ADD CONSTRAINT` will run a full validation of the apps/deployments tables before succeeding.
It should also be safe to run on a running cluster without breaking anything because the entire migration runs within a single transaction.

The only non-clear part is probably that we reset the default value on apps.strategy, that is because otherwise the default value for the column would actually still be the enum but with a default cast to text.

The only thing I am not sure about is converting the `job_state` enum. Mainly because it's unlikely to be updated (famous last words lol) but also because there is the trigger/function that would need to be updated to handle checking valid state transitions. Converting from enum will remove the implicit order so I will probably have to add a sequence or int column to maintain the current behaviour.
A slightly overkill option would be to create the `job_states` table with 2 additional array columns that contain the valid from and to states. This would be the most flexible long term but kinda smells over over-engineering.

lmars:
> I tested the migration steps manually against a cluster with data in it.
> I couldn't work out how to update to a development build though, that would be a better test.

See the steps I took [here](https://github.com/flynn/flynn/pull/2135#issuecomment-156884881). We could also unit test this if we add a `migrateTo` function, then `migrateTo(3); <insert some data>; migrateTo(4)`.

Once the updater is finished (or even before), we should add integration tests which install the latest stable release, create some apps, run an update, then do some checks, which will give us more confidence for these migration changes.

> The only thing I am not sure about is converting the job_state enum. Mainly because it's unlikely to be updated (famous last words lol) but also because there is the trigger/function that would need to be updated to handle checking valid state transitions. Converting from enum will remove the implicit order so I will probably have to add a sequence or int column to maintain the current behaviour.

These values will most likely change soon. I'm going to add a `scheduled` state for jobs which are scheduled to start in the future (e.g. after a previous job has crashed, and the scheduler is backing off restarts).

As for the validation, this can be dropped once we stop all the schedulers watching events (see [this comment](https://github.com/flynn/flynn/issues/2030#issuecomment-151337726)) which I am going to be doing very soon (I pretty much have it changed locally).

titanous:
LGTM

PR #2154, opened 2015-11-17
host: Add configurable tags, broadcast via service discovery

lmars:
This is a prerequisite for implementing tag based constraints in the scheduler (i.e. run a job only on hosts tagged with `role=db`).

Tags can be set in a number of ways:

* Via the `--tags` flag when starting `flynn-host daemon`
* Via the host API (`POST /host/tags`)
* Via the CLI (`flynn-host tags set <hostid> key=val...`)

The tags are set in service discovery instance metadata so that they can be easily seen by services which already watch the `flynn-host` service (e.g. the scheduler), and the keys are prefixed with `tag:` so it is clear which metadata should be interpreted as tags.

### DIFF
--- a/host/discoverd.go
+++ b/host/discoverd.go
@@ -114,7 +114,13 @@ func (d *DiscoverdManager) UpdateTags(tags map[string]string) error {
 	d.mtx.Lock()
 	defer d.mtx.Unlock()
 	for k, v := range tags {
-		d.inst.Meta[host.TagPrefix+k] = v
+		name := host.TagPrefix + k
+		// treat empty tags as ones to delete
+		if v == "" {
+			delete(d.inst.Meta, name)
+			continue
+		}
+		d.inst.Meta[name] = v
 	}
 	if d.hb == nil {
 		return nil
--- a/host/discoverd.go
+++ b/host/discoverd.go
@@ -116,5 +116,8 @@ func (d *DiscoverdManager) UpdateTags(tags map[string]string) error {
 	for k, v := range tags {
 		d.inst.Meta[host.TagPrefix+k] = v
 	}
+	if d.hb == nil {
+		return nil
+	}
 	return d.hb.SetMeta(d.inst.Meta)
 }
--- a/host/discoverd.go
+++ b/host/discoverd.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/flynn/flynn/discoverd/client"
 	"github.com/flynn/flynn/host/logmux"
+	"github.com/flynn/flynn/host/types"
 	"github.com/flynn/flynn/pkg/shutdown"
 )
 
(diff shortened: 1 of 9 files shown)


lmars:
Realised these tags need to survive an update, will implement.

lmars:
Updated so that tags are maintained through an update: 4c08a55

Please review.

host/discoverd.go
```diff
@@ -105,3 +109,15 @@ func (d *DiscoverdManager) ConnectPeer(ips []string) error {
 	}
 	return err
 }
+
+func (d *DiscoverdManager) UpdateTags(tags map[string]string) error {
+	d.mtx.Lock()
+	defer d.mtx.Unlock()
+	for k, v := range tags {
+		d.inst.Meta[host.TagPrefix+k] = v
```
titanous: This doesn't provide a method for removing tags. Perhaps the entire tag set should be provided?

lmars: Perhaps we just remove any that are empty?

We could have `flynn-host tags del foo` which calls `UpdateTags(map[string]string{"foo": ""})`, and here we just `delete(d.inst.Meta, host.TagPrefix+k)` if `v` is empty?

titanous: I think having an empty value is something we might want? For tags that just "exist" or not.

imjorge: What about using a `-` prefix on a key to indicate that we want to drop it?

In this example:

`flynn-host tags set <hostid> key1=val1,key2=val2,key3=,-key4,-key5=,-key6=val6`

- key1 will be set to value1 (may be added or updated)
- key2 will be set to value2 (may be added or updated)
- key3 will be set to "" (empty) - tag that just "exist"
- key4, key5 and key6 are removed (if existing) - value is simply ignored

lmars: So at a low-level, we could just represent tags that just "exist" with the string `true`, leading to the following:

```
$ flynn-host tags set foo=true # foo gets set to "true"
$ flynn-host tags set foo      # foo gets set to "true"
$ flynn-host tags set foo=     # foo gets deleted
$ flynn-host tags del foo      # foo gets deleted
$ flynn-host tags set foo=bar  # foo gets set to "bar"
```

WDYT?

titanous: SGTM

lmars:
@titanous @imjorge comments addressed

host/discoverd.go
imjorge: > $ flynn-host tags set foo=     # foo gets deleted

This would surprise me. Not really clear for the user it would drop a tag. Even more when the single absence of `=` would cause it to get created instead:

> $ flynn-host tags set foo      # foo gets set to "true"

Aren't the `-` prefixing a tag more clearer?

lmars: So I believe this boils down to how we document things, we should definitely have `flynn-host tags del` to delete a tag, so the only issue is what happens when you set a tag to the empty string. My implementation deletes the tag (which we can document) and yours will actually set it to empty (with the extra syntax to delete), but I really don't think we want empty tags right now, so I'm going to go with what is here.

titanous:
LGTM

PR #2160, opened 2015-11-18
controller: Always connect job event stream during deployment

josephglanville:
Currently if you deploy a release that only contains processes that will use service events the job event stream will not be connected.
This breaks deployments of these releases because the deployer still expects to use job events to track the processes shutting down.


titanous:
LGTM

lmars:
LGTM

PR #2165, opened 2015-11-19
gitreceive/receiver: Set service on process type for zero-downtime deploy

titanous:
When the service is specified, it is used by the deployer to ensure that old jobs are not terminated until the new ones are registered in discoverd.


lmars:
LGTM

PR #2173, opened 2015-11-23
discoverd/client: Cluster awareness/failover

### LINKED ISSUE #2091: discoverd: Client should be cluster aware
The discoverd client should be aware of the full list of peers and be able to failover between them.

josephglanville:
This changeset makes the discoverd client cluster aware.

It can now be initialised with a connection string of comma delimited servers.
The first server is considered to be preferred and it will use it unless it's unavailable.
If it does become unavailable then the rest of the servers configured will be tried until a working one is found, this server will then be pinned for 60 seconds, after the TTL expires the client will attempt to use the preferred server once more.
Writes are directed to the leader by recording where redirects are sent, if at any point the leader becomes unavailable we will try issue the write to another server and have it redirect us to the new leader.

Closes #2091

### DIFF
--- a/discoverd/client/client.go
+++ b/discoverd/client/client.go
@@ -1,32 +1,23 @@
 package discoverd
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"net"
 	"net/http"
+	"net/url"
 	"os"
+	"strconv"
 	"strings"
-	"time"
+	"sync"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
 	dt "github.com/flynn/flynn/discoverd/types"
-	"github.com/flynn/flynn/pkg/dialer"
 	"github.com/flynn/flynn/pkg/httpclient"
 	hh "github.com/flynn/flynn/pkg/httphelper"
 	"github.com/flynn/flynn/pkg/stream"
 )
-
-type Service interface {
-	Leader() (*Instance, error)
-	Instances() ([]*Instance, error)
-	Addrs() ([]string, error)
-	Leaders(chan *Instance) (stream.Stream, error)
-	Watch(events chan *Event) (stream.Stream, error)
-	GetMeta() (*ServiceMeta, error)
-	SetMeta(*ServiceMeta) error
-	SetLeader(string) error
-}
 
 var ErrTimedOut = errors.New("discoverd: timed out waiting for instances")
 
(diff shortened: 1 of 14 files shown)


discoverd/client/client.go
```diff
@@ -1,261 +1,192 @@
...
+	leader     int
+	mu         sync.RWMutex
 }
 
-func NewClient() *Client {
-	url := os.Getenv("DISCOVERD")
-	if url == "" {
-		url = "http://127.0.0.1:1111"
+func NewClientWithConfig(config Config) *Client {
+	client := Client{}
```
lmars: We tend to write `var client Client`

lmars: Then again, looking at the next lines, this should probably be:

```go
client := &Client{
	servers: make([]*httpclient.Client, 0, len(config.Endpoints)),
}
```

discoverd/client/client.go
```diff
@@ -1,261 +1,192 @@
...
 
-func (c *Client) Ping() error {
-	return c.c.Get("/ping", nil)
+func formatURLs(urls []string) []string {
+	formatted := make([]string, len(urls))
+	for _, u := range urls {
+		if !strings.HasPrefix(u, "http") {
+			u = "http://" + u
+		}
+		formatted = append(formatted, u)
```
lmars: Given you have created a slice with a non-zero length, this either needs to index into that slice (i.e. `formatted[i] = u`), or just create the slice with a capacity but zero length.

discoverd/client/client.go
```diff
@@ -1,261 +1,192 @@
...
-type LeaderType string
-
-const (
-	LeaderTypeManual LeaderType = "manual"
-	LeaderTypeOldest LeaderType = "oldest"
-)
-
-type ServiceConfig struct {
-	LeaderType LeaderType `json:"leader_type"`
+func (c *Client) updateLeader(url string) {
```
lmars: This argument seems to be a `host`, rather than a `url`

discoverd/client/client.go
```diff
@@ -1,261 +1,192 @@
...
-		case EventKindCurrent:
-			break outer
-		case EventKindUp:
-			instances = append(instances, event.Instance)
+	for i := pinned; i < len(c.servers)+pinned; i++ {
+		k := i % len(c.servers)
+		hc := c.servers[k]
+		err := hc.Send(method, path, in, out)
+		if err != nil {
+			continue
```
lmars: We should save these errors so they can be returned if all the hosts fail

discoverd/client/client.go
josephglanville: Yep this makes more sense.

discoverd/client/client.go
josephglanville: I see.. will fix.

discoverd/client/client.go
josephglanville: Yep, rename of arg in order.

discoverd/client/client.go
josephglanville: Hmm, would the best approach here be to collect these into a slice and then return them as a big Error? I could probably create a bunch of custom errors like ErrNoHosts etc to handle the different ways we can blow up.

lmars: I think I'd like to see an error like:

```
error sending HTTP request: 10.0.0.1:1111 connection refused, 10.0.0.2:1111 something went wrong, 10.0.0.3:1111 connection reset
```

josephglanville: Yep, that is what I was thinking.

josephglanville:
So right now it will only failover to another server if the error is a `net.OpError`.
I think it should also failover if there is no current Raft leader because there could be a net split in play and if we keep trying peers we might find a peer on the majority side of the net split.

discoverd/client/client.go
```diff
@@ -1,266 +1,205 @@
...
-	return &Client{
-		c: &httpclient.Client{
-			URL: url,
-			HTTP: &http.Client{
-				Transport:     &http.Transport{Dial: dialer.Retry.Dial},
-				CheckRedirect: redirectPreserveHeaders,
-			},
-		},
-	}
+	return NewClientWithConfig(Config{Endpoints: formatURLs([]string{url})})
```
lmars: `formatURLs` should take `urls ...string` so you can call it here like `formatURLs(url)`

discoverd/client/client.go
```diff
@@ -1,266 +1,205 @@
...
-	return &service{
-		client: client,
-		name:   name,
+func isNetError(err error) bool {
+	switch err.(type) {
+	case *net.OpError:
+		return true
 	}
+	return false
 }
```
lmars:
```go
_, ok := err.(*net.OpError)
return ok
```

host/discoverd.go
```diff
@@ -78,6 +78,7 @@ func (d *DiscoverdManager) ConnectLocal(url string) error {
 	}
 	d.local.Store(true)
 
+	fmt.Println("Setting default env var DISCOVERD=", url)
```
lmars: Either drop this or use `log15`

josephglanville:
This is ready for a review pass.

Failing over on `ErrNoKnownLeader` will probably happen in a future PR as I don't want to touch the server much in this pass. (supporting it properly requires converting the server to use JSONError most likely)

discoverd/client/client.go
```diff
@@ -1,303 +1,251 @@
...
-		case <-time.After(timeout):
-			return nil, ErrTimedOut
+		if err != nil && isNetError(err) {
+			errors = append(errors, err.Error())
+			continue
+		} else if err != nil {
+			return nil, err
+		}
+		if time.Since(pinUpdated) > PinTTL {
+			k = 0 // TTL has timed out, use prefered server on next request
```
titanous: preferred

titanous:
This appears to rely on all available discoverd servers to be specified ahead of time. Will another pass implement getting an updated list from the server?

josephglanville:
Yes, I think when I work out what strategy will be used to update the list of peers for when we do reconfigurations etc I will also use it to have the client automatically keep the list of servers up to date.
In all likelihood it will probably be finding a good way to make something like `/raft/peers` stream changes. Potentially we could poll but that seems nasty given how good our tools for long-poll/streaming are.

josephglanville:
@titanous this is ready for some review. I implemented updates to the peer set by setting a HTTP header `DISCOVERD_PEERS` on each response that the client can use to keep in sync. This avoids the need for a goroutine.

discoverd/client/client.go
```diff
@@ -36,268 +26,256 @@ func init() {
...
-		case <-time.After(timeout):
-			return nil, ErrTimedOut
 		}
+		if err != nil && isNetError(err) {
+			errors = append(errors, err.Error())
+			continue
+		} else if err != nil {
+			return nil, err
+		}
+		// If the pinned server failed to fufil our request then update the pin
```
titanous: s/fufil/fulfill/

discoverd/client/client.go
```diff
@@ -36,268 +26,256 @@ func init() {
...
+		// We don't update the pin on leader requests
+		if k != pinned && !leaderReq {
+			c.updatePin(k)
+		}
+		if peers := rsp.Header.Get("DISCOVERD_PEERS"); peers != "" {
+			c.updateServers(strings.Split(peers, ","))
```
titanous: Should this include the raft index so that out of order request response processing doesn't nuke a new peer list?

josephglanville:
@titanous comments addressed. Added the index to prevent changes to peer requests from older indices. Also modified how the servers are maintained and updated to prevent creating new httpclient.Clients if the transport config is the same.

discoverd/client/client.go
```diff
@@ -36,268 +27,306 @@ func init() {
...
 		return nil
 	}
-	for attr, val := range via[0].Header {
-		if _, ok := req.Header[attr]; !ok {
-			req.Header[attr] = val
-		}
+	return &httpclient.Client{
+		URL: url,
+		HTTP: &http.Client{
+			CheckRedirect: checkRedirect,
```
titanous: Can this `http.Client` client be defined once globally? It doesn't look like it varies based on the URL.

josephglanville: I don't see why not, I wasn't aware you could share them across httpclient.Client instances but if so  may as well.

titanous: Yeah, I don't think there's any reason we can't do that.

discoverd/server/handler.go
```diff
@@ -108,15 +110,25 @@ func proxyWhitelisted(r *http.Request) bool {
...
+			return
+		}
+	} else {
+		// Send current peer list and index to the client so it can keep the list of
+		// peers in sync with the cluster.
+		peers, err := h.Store.GetPeers()
+		if err == nil {
+			w.Header().Set("Discoverd-Current-Peers", strings.Join(peers, ","))
+		}
+		w.Header().Set("Discoverd-Current-Index", strconv.FormatUint(h.Store.LastIndex(), 10))
```
titanous: Should we drop `-Current` from these so that it's just `Discoverd-Peers` and `Discoverd-Index`?

josephglanville: Yeah I guess current is redundant. It wouldn't be sending them to you if it didn't think they were current.

titanous:
LGTM

temujin9:
Thanks, guys. LMK when this drops, so I can start testing it out on our sandbox.

titanous:
@temujin9 This is in `v20160602.0`.

PR #2177, opened 2015-11-24
controller: avoid panic when streaming formation updates

### LINKED ISSUE #2175: controller: Panic subscribing to formation updates
In https://gist.github.com/anonymous/5e7cc17cc6c01953f907, job `controller-web-flynn1448137186-aa89fd57-4f8d-4733-b41e-5fe20ce65266` crashed with the following panic:

```
panic: send on closed channel

goroutine 199 [running]:
main.(*FormationRepo).sendUpdatedSince(0xc208163630, 0xc20811f8c0, 0xc20811f920, 0xecde2cf69, 0x0, 0xa30320, 0x0, 0x0)
	/tmp/tmp.Q2hgRVCoYK/src/github.com/flynn/flynn/controller/formation.go:284 +0x385
created by main.(*FormationRepo).Subscribe
	/tmp/tmp.Q2hgRVCoYK/src/github.com/flynn/flynn/controller/formation.go:265 +0x16c
```

imjorge:
Closes #2175

### DIFF
--- a/controller/formation.go
+++ b/controller/formation.go
@@ -21,13 +21,41 @@ type formationKey struct {
 	AppID, ReleaseID string
 }
 
+// we are wrapping the client specified channel to send formation updates in order to safely interrupt
+// `sendUpdateSince` goroutine if an unsubscribe happens before it is completed. otherwise we can get
+// a panic due to sending to a closed channel. See https://github.com/flynn/flynn/issues/2175 for more
+// details on this subject.
+type FormationSubscription struct {
+	mtx sync.RWMutex
+	ch  chan *ct.ExpandedFormation
+}
+
+func (f *FormationSubscription) notify(ef *ct.ExpandedFormation) bool {
+	f.mtx.RLock()
+	defer f.mtx.RUnlock()
+	if f.ch == nil {
+		return false
+	}
+	f.ch <- ef
+	return true
+}
+
+func (f *FormationSubscription) close() {
+	f.mtx.Lock()
+	defer f.mtx.Unlock()
+	if f.ch != nil {
+		close(f.ch)
+		f.ch = nil
+	}
+}
+
 type FormationRepo struct {
 	db        *postgres.DB
 	apps      *AppRepo
 	releases  *ReleaseRepo
 	artifacts *ArtifactRepo
 
-	subscriptions map[chan *ct.ExpandedFormation]struct{}
+	subscriptions map[*FormationSubscription]struct{}
 	stopListener  chan struct{}
 	subMtx        sync.RWMutex
 }
@@ -38,7 +66,7 @@ func NewFormationRepo(db *postgres.DB, appRepo *AppRepo, releaseRepo *ReleaseRep
 		apps:          appRepo,
 		releases:      releaseRepo,
 		artifacts:     artifactRepo,
-		subscriptions: make(map[chan *ct.ExpandedFormation]struct{}),
+		subscriptions: make(map[*FormationSubscription]struct{}),
 		stopListener:  make(chan struct{}),
 	}
 }
@@ -237,8 +265,8 @@ func (r *FormationRepo) publish(appID, releaseID string) {
 	r.subMtx.RLock()
 	defer r.subMtx.RUnlock()
 
-	for ch := range r.subscriptions {
-		ch <- f
+	for sub := range r.subscriptions {
+		sub.notify(f) // ignore failure for this subscriber (went away) and keep sending to remaining subscribers
 	}
 }
 
@@ -296,13 +324,12 @@ func (r *FormationRepo) unsubscribeAll() {
 	r.subMtx.Lock()
 	defer r.subMtx.Unlock()
 
-	for ch := range r.subscriptions {
-		r.unsubscribeLocked(ch)
-		close(ch)
+	for sub := range r.subscriptions {
+		r.unsubscribeLocked(sub)
 	}
 }
 
-func (r *FormationRepo) Subscribe(ch chan *ct.ExpandedFormation, stopCh <-chan struct{}, since time.Time) error {
+func (r *FormationRepo) Subscribe(ch chan *ct.ExpandedFormation, since time.Time, updated chan<- struct{}) (*FormationSubscription, error) {
 	// we need to keep the mutex locked whilst calling startListener
 	// to avoid a race where multiple subscribers can get added to
 	// r.subscriptions before a potentially failed listener start,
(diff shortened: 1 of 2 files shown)


imjorge:
The test uses a `time.Sleep` which is fragile.

controller/formation.go
```diff
@@ -303,7 +318,14 @@ func (r *FormationRepo) unsubscribeLocked(ch chan *ct.ExpandedFormation) {
 		for range ch {
 		}
 	}()
+	si, ok := r.subscriptions[ch]
 	delete(r.subscriptions, ch)
+	if ok {
+		si.uc <- struct{}{}
+		si.wg.Wait()
```
lmars: We want the `sendUpdatedSince` goroutine to stop if the channel is closed, which we can do by just wrapping the channel:

```go
type FormationSubscription struct {
	mtx sync.Mutex
	ch  chan *ct.ExpandedFormation
}

func (f *FormationSubscription) Notify(ef *ct.ExpandedFormation) bool {
	f.mtx.RLock()
	defer f.mtx.RUnlock()
	if f.ch == nil {
		return false
	}
	f.ch <- ef
	return true
}

func (f *FormationSubscription) Close() {
	f.mtx.Lock()
	defer f.mtx.Unlock()
	close(f.ch)
	f.ch = nil
}

// in sendUpdatedSince
for rows.Next() {
	// ...
	if ok := f.Notify(ef); !ok {
		return nil
	}
}
```

WDYT?

imjorge: But that way we are not ALSO selecting on stopCh inside sendUpdatedSince. Right?

--jmf

imjorge: But that way we are **not ALSO selecting** on `stopCh` inside `sendUpdatedSince`. Right?

lmars: Correct, but `stopCh` gets closed [at the same time as unsubscribing](https://github.com/flynn/flynn/blob/master/controller/formation.go#L400-L401), so this will have the same effect (stopping the `sendUpdatedSince` goroutine when the channel is unsubscribed).

imjorge: So drop `stopCh` altogether?

lmars: I reckon so yes

controller/formation.go
```diff
@@ -21,13 +21,35 @@ type formationKey struct {
...
+}
+
 type FormationRepo struct {
 	db        *postgres.DB
 	apps      *AppRepo
 	releases  *ReleaseRepo
 	artifacts *ArtifactRepo
 
-	subscriptions map[chan *ct.ExpandedFormation]struct{}
+	subscriptions map[chan *ct.ExpandedFormation]*formationSubscription
```
lmars: Make this `map[*formationSubscription]struct{}`

controller/formation.go
```diff
@@ -243,11 +265,10 @@ func (r *FormationRepo) unsubscribeAll() {
 
 	for ch := range r.subscriptions {
 		r.unsubscribeLocked(ch)
-		close(ch)
 	}
 }
 
-func (r *FormationRepo) Subscribe(ch chan *ct.ExpandedFormation, stopCh <-chan struct{}, since time.Time) error {
+func (r *FormationRepo) Subscribe(ch chan *ct.ExpandedFormation, since time.Time, sendUpdateSinceDone chan<- struct{}) error {
```
lmars: Just call this argument `updatesDone`, or maybe even just `updated`

controller/formation.go
```diff
@@ -260,13 +281,21 @@ func (r *FormationRepo) Subscribe(ch chan *ct.ExpandedFormation, stopCh <-chan s
...
+
+	go r.sendUpdatedSince(ch, f, since, sendUpdateSinceDone)
 	return nil
 }
 
-func (r *FormationRepo) sendUpdatedSince(ch chan *ct.ExpandedFormation, stopCh <-chan struct{}, since time.Time) error {
+func (r *FormationRepo) sendUpdatedSince(ch chan *ct.ExpandedFormation, f *formationSubscription, since time.Time, sendUpdatedSinceDone chan<- struct{}) error {
+	if sendUpdatedSinceDone != nil {
+		defer func() {
+			sendUpdatedSinceDone <- struct{}{}
```
lmars: Just close this channel

controller/formation.go
```diff
@@ -21,13 +21,35 @@ type formationKey struct {
 	AppID, ReleaseID string
 }
 
+type formationSubscription struct {
```
lmars: We should add a comment as to why we are wrapping the channel like this

controller/formation.go
imjorge: Done.

controller/formation.go
imjorge: Done.

controller/formation.go
imjorge: Done. Used `updated`.

controller/formation.go
imjorge: Done.

controller/formation.go
imjorge: Done.

controller/formation.go
```diff
@@ -182,8 +208,10 @@ func (r *FormationRepo) publish(appID, releaseID string) {
 	r.subMtx.RLock()
 	defer r.subMtx.RUnlock()
 
-	for ch := range r.subscriptions {
-		ch <- f
+	for sub := range r.subscriptions {
+		if !sub.notify(f) {
+			break
```
lmars: We shouldn't break if one of the subscribers has gone away (the other subscribers may still be open)

controller/formation.go
```diff
@@ -257,16 +284,24 @@ func (r *FormationRepo) Subscribe(ch chan *ct.ExpandedFormation, stopCh <-chan s
...
+	go r.sendUpdatedSince(sub, since, updated)
+	return sub, nil
 }
 
-func (r *FormationRepo) sendUpdatedSince(ch chan *ct.ExpandedFormation, stopCh <-chan struct{}, since time.Time) error {
+func (r *FormationRepo) sendUpdatedSince(sub *FormationSubscription, since time.Time, updated chan<- struct{}) error {
+	if updated != nil {
+		defer func() {
+			close(updated)
+		}()
```
lmars: `defer close(updated)`

controller/formation.go
```diff
@@ -281,29 +316,31 @@ func (r *FormationRepo) sendUpdatedSince(ch chan *ct.ExpandedFormation, stopCh <
...
-func (r *FormationRepo) unsubscribeLocked(ch chan *ct.ExpandedFormation) {
-	go func() {
+func (r *FormationRepo) unsubscribeLocked(sub *FormationSubscription) {
+	go func(ch chan *ct.ExpandedFormation) {
 		// drain to prevent deadlock while removing the listener
 		for range ch {
 		}
-	}()
-	delete(r.subscriptions, ch)
+	}(sub.ch)
```
lmars: There is no need to pass the channel to the goroutine, the variable binding will not change.

controller/formation.go
```diff
@@ -281,29 +316,31 @@ func (r *FormationRepo) sendUpdatedSince(ch chan *ct.ExpandedFormation, stopCh <
...
+	_, ok := r.subscriptions[sub]
+	delete(r.subscriptions, sub)
+	if ok {
+		sub.close()
```
lmars: No need to check if it is in `r.subscriptions`, just close it.

controller/formation.go
imjorge: Actually it matters since `sub.ch` will become `nil` when the subscription is closed.

[This deadlocks](https://play.golang.org/p/3MRVr6j___).

[This does not dealock](https://play.golang.org/p/GADxvT3LZ5).

controller/formation.go
imjorge: Done.

controller/formation.go
imjorge: Done.

controller/formation.go
imjorge: Done. Protected `sub.close()` to be idempotent (do not panic on already closed channel).

controller/formation.go
lmars: Ah yeh, ignore me :smile:

titanous:
@imjorge looks like this needs a rebase.

imjorge:
@titanous rebased.

lmars:
LGTM

josephglanville:
Appears to have introduced a hang (atleast in the controller unit tests)

lmars:
See #2208 re the controller hang.

PR #2183, opened 2015-11-26
controller: Copy deploy_timeout to deployment on insertion

josephglanville:
This will allow the controller endpoints `/apps/:app_id/deployments` and `/deployments/:id` to return the `deploy_timeout` value that was used at the time the deployment was created. Currently it always returns 0 despite whatever is set on the app record, which is somewhat surprising.


josephglanville:
The migration is safe and tested, however it does result in all previous deployments having a deploy_timeout of 30 even though they may have used another value. This is probably not worth doing anything about, just noting for posterity.

titanous:
LGTM

PR #2184, opened 2015-11-26
test: Trace commands in runner Upstart script

lmars:
This makes it easier to know what is going on when tailing the log file.


josephglanville:
LGTM

PR #2191, opened 2015-11-27
pkg/knownhosts: Fix parsing file with comments

jvatic:


pkg/knownhosts/knownhosts_test.go
```diff
@@ -219,6 +219,31 @@ func (S) TestMixedKeyTypes(c *C) {
...
+
+	// commented out host
+	host1Addr := "101.102.103.72"
+	input.WriteString("# ")
+	input.WriteString(host1Addr)
+	input.WriteString(sep)
+	input.Write(keyBytes)
+	input.WriteString("\n")
+
+	k, err := Unmarshal(bytes.NewReader(input.Bytes()))
```
titanous: You should be able to use `input` directly without wrapping in a `bytes.Reader`.

titanous:
LGTM

PR #2199, opened 2015-11-28
pkg/status: Return version in status

josephglanville:

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -3,6 +3,8 @@ package status
 import (
 	"encoding/json"
 	"net/http"
+
+	"github.com/flynn/flynn/pkg/version"
 )
 
 /*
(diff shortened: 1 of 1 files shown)


pkg/status/status.go
```diff
@@ -38,12 +41,14 @@ var (
...
-	Status Code             `json:"status"`
-	Detail *json.RawMessage `json:"detail,omitempty"`
+	Status  Code             `json:"status"`
+	Detail  *json.RawMessage `json:"detail,omitempty"`
+	Version string           `json:"version,omitempty"`
 }
 
 func New(healthy bool, detail interface{}) (Status, error) {
 	var s Status
+	s.Version = version.String()
```
titanous: Any reason to assign this here? The assignment in the handler should be sufficient.

titanous:
LGTM

PR #2201, opened 2015-11-28
test: Run domain migration tests after other tests

lmars:
Ideally this would be isolated from the rest, which is doable, but this is a quicker fix for now.


titanous:
LGTM

PR #2202, opened 2015-11-29
contorller: Don't call c.Assert in goroutine

lmars:
It results in a data race (see https://ci.flynn.io/builds/20151128100145-5210d2b0)


titanous:
LGTM

PR #2203, opened 2015-11-29
Tag based scheduling

lmars:
Still some changes I need to make + more testing, but opening for feedback.

Summary of changes:

* Add `Formation.Tags` which is a map of process types to tags for that process type
* Update `flynn scale` to handle adding tags (e.g. `flynn scale web=3,active=true db=2,active=true,disk=ssd`)
* Only place jobs on hosts whose tags are a superset of the process type's tags
* When either host or formation tags change, stop any jobs which no longer match the tags of the host they are on (e.g. if I have scaled to `web=3,active=true`, and I set `active=false` on one of my hosts, then I want the web job running on that host to be stopped and subsequently restarted on a different, still matching host)

Caveats:

* I originally planned to support more fine grained tags per process type (e.g. I want 3 web jobs on `type=big` and 1 on `type=small`) but the logic was hard to get right, I do think what I have done here though is a good start in getting to that
* I wanted to make `Formation.Processes` a `map[string]Process`, with `Process` something like `struct { Count int, Tags map[string]string }` but it would be difficult to manage that change during an update (the controller would need to handle both the old and new formats because the deployment will be posting the formations)

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -249,13 +249,13 @@ func checkOnlineHosts(expected int, state *State, urls []string, timeoutSecs int
 		if known >= expected {
 			state.Hosts = make([]*cluster.Host, 0, known)
 			for _, url := range urls {
-				h := cluster.NewHost("", url, nil)
+				h := cluster.NewHost("", url, nil, nil)
 				status, err := h.GetStatus()
 				if err != nil {
 					continue
 				}
 				online++
-				state.Hosts = append(state.Hosts, cluster.NewHost(status.ID, status.URL, nil))
+				state.Hosts = append(state.Hosts, cluster.NewHost(status.ID, status.URL, nil, nil))
 			}
 			if online >= expected {
 				break
--- a/cli/scale.go
+++ b/cli/scale.go
@@ -10,13 +10,22 @@ import (
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/controller/utils"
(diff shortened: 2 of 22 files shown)


josephglanville:
Would it be difficult to support anti-affinity? i.e
```
flynn scale web=3,active=true,disk=!ssd
```
To prevent jobs being scheduled on a host with a specific tag.

titanous:
For now that type of ant-affinity could be accomplished with `notssd=true` or whatever.

I wonder if we should set up the CLI syntax to support a "tag query language"? I don't think we should do more than exact matches with AND logic, but in the future we'll probably want to do more sophisticated stuff.

lmars:
>> Would it be difficult to support anti-affinity

> For now that type of anti-affinity could be accomplished with notssd=true or whatever

Yep I am currently expecting users to be a bit more explicit with their tags for this (e.g. `ssd=true` / `ssd=false`, or `disk=ssd` / `disk=mag`).

> I wonder if we should set up the CLI syntax to support a "tag query language"? I don't think we should do more than exact matches with AND logic, but in the future we'll probably want to do more sophisticated stuff.

Yes I think this is a good idea. So something like the following? 

```bash
flynn scale "web=3,active=true" "db=2,active=true AND disk=ssd"
```

It would be nice if it didn't require spaces (so needs to be quoted or spaces escaped), but I don't think it's a massive issue as long as it is documented.

WDYT?

titanous:
As long as we can distinguish the current comma-separated tags system from whatever new one we come up with on the CLI, I think we can stick with the current syntax.

lmars:
@josephglanville @titanous 

Updated to place jobs with tags which do not match any hosts in the "pending" state, and try to start them if host tags change (see 70ddcce).

Please review.

titanous:
LGTM. Great work!

PR #2207, opened 2015-11-30
Make updater aware of versions

josephglanville:
This introduces behaviour into the updater to manage skipping system applications that can't be updated if the running version is not recent enough.

### DIFF
--- a/updater/updater.go
+++ b/updater/updater.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
 	"time"
@@ -13,6 +14,8 @@ import (
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
 	"github.com/flynn/flynn/discoverd/client"
+	"github.com/flynn/flynn/pkg/status"
+	"github.com/flynn/flynn/pkg/version"
 	"github.com/flynn/flynn/updater/types"
 )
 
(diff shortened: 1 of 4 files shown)


updater/updater.go
```diff
@@ -41,6 +45,35 @@ func run() error {
 		return err
 	}
 
+	statusInstances, err := discoverd.GetInstances("status-web", 10*time.Second)
+	if err != nil {
+		log.Error("error looking up status service in service discovery", "err", err)
+		return err
+	}
+	req, err := http.NewRequest("GET", "http://"+statusInstances[0].Addr, nil)
```
lmars: Could just use `http://status-web.discoverd`

updater/updater.go
```diff
@@ -41,6 +45,35 @@ func run() error {
...
+	}
+	req.Header = make(http.Header)
+	req.Header.Set("Accept", "application/json")
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		log.Error("error getting cluster status", "err", err)
+	}
+	defer res.Body.Close()
+	if res.StatusCode != 200 {
+		log.Error("cluster status is unhealthy", "code", res.StatusCode)
```
lmars: Should this return here?

updater/updater.go
```diff
@@ -41,6 +45,35 @@ func run() error {
...
+	}
+	var statusWrapper struct {
+		Data struct {
+			Detail map[string]status.Status
+		}
+	}
+	if err := json.NewDecoder(res.Body).Decode(&statusWrapper); err != nil {
+		log.Error("error decoding cluster status JSON", "err", err)
```
lmars: Should this return here?

titanous:
Comments addressed.

titanous:
It looks like the discoverd updater is failing fairly often on CI. Haven't looked into the cause yet.

josephglanville:
discoverd deployment flakiness tracked in #2212

updater/updater.go
josephglanville: Fixed.

lmars:
LGTM

PR #2208, opened 2015-11-30
controller: Fix hang in controller tests

lmars:
The tests were hanging during `TestStreamDeployment`, and I noticed that pgx was blocking trying to acquire a connection. Looking at the code, we were creating transactions before reading from the database, and so I moved the transactions and it no longer hangs.

I don't know the specific reason this was an issue (I assume reading inside a transaction is an ok thing to do?).


titanous:
Maybe the connection pool was too small? In any event, there's no reason to start the transaction early.

titanous:
LGTM

PR #2212, opened 2015-12-01
discoverd: Close mux before closing store

josephglanville:
This works around a strange behavior during deployments where the mux
socket is unable to be bound to be the new process because it's still
in use. Despite having just been closed... by closing it before closing
the store the race is avoided.


titanous:
LGTM

Issue #2211, opened 2015-11-30
Segfault on 'flynn-host download'


temujin9:
```
===> 20:04:26.886 running "./flynn-host download --repository https://dl.flynn.io/tuf --tuf-db /etc/flynn/tuf.db --config-dir /etc/flynn --bin-dir /usr/local/bin"
t=2015-11-30T20:04:26+0000 lvl=info msg="initializing TUF client"
unexpected fault address 0x7efbfc81f029
fatal error: fault
[signal 0xb code=0x1 addr=0x7efbfc81f029 pc=0x4701c7]

goroutine 1 [running]:
runtime.gothrow(0xb570f0, 0x5)
        /usr/local/go/src/runtime/panic.go:503 +0x8e fp=0xc2080e1120 sp=0xc2080e1108
runtime.sigpanic()
        /usr/local/go/src/runtime/sigpanic_unix.go:29 +0x265 fp=0xc2080e1170 sp=0xc2080e1120
runtime.memmove(0xc20827c000, 0x7efbfc81f029, 0x5ce)
        /usr/local/go/src/runtime/memmove_amd64.s:216 +0x97 fp=0xc2080e1178 sp=0xc2080e1170
bytes.(*Reader).Read(0xc208285500, 0xc20827c000, 0x2000, 0x2000, 0x4f8217, 0x0, 0x0)
        /usr/local/go/src/bytes/reader.go:40 +0x12d fp=0xc2080e11d0 sp=0xc2080e1178
io.(*teeReader).Read(0xc2082a6720, 0xc20827c000, 0x2000, 0x2000, 0x7efbfc7e2340, 0x0, 0x0)
        /usr/local/go/src/io/io.go:495 +0x83 fp=0xc2080e1248 sp=0xc2080e11d0
io/ioutil.devNull.ReadFrom(0x0, 0x7efbfc7e4c70, 0xc2082a6720, 0x0, 0x0, 0x0)
        /usr/local/go/src/io/ioutil/ioutil.go:151 +0xb1 fp=0xc2080e12a0 sp=0xc2080e1248
io/ioutil.(*devNull).ReadFrom(0xc20802a300, 0x7efbfc7e4c70, 0xc2082a6720, 0x7efbfc7e49b0, 0x0, 0x0)
        <autogenerated>:9 +0xbb fp=0xc2080e12d8 sp=0xc2080e12a0
[...]

PR #2216, opened 2015-12-01
Fix TUF segfault

titanous:
Fixes segfault due to misuse of boltdb values (#2211).


josephglanville:
Revert is missing DCO but LGTM otherwise.

PR #2223, opened 2015-12-02
Added an hint to the PHP documentation

philiplb:
Added an hint to the PHP documentation about how to customize the webserver and PHP configuration.


titanous:
We shouldn't link off to the heroku docs, we'll need to write docs specifically for Flynn.

philiplb:
Fair enough. :)

PR #2233, opened 2015-12-03
Path based routing

### LINKED ISSUE #2069: Feature Request: Specify a path in the URL of a route
Hi,

maybe this is already possible? I'm very new to flynn. :)
Can I add a path to a route to specify an application?
We often use this scheme here to differentiate between dev, beta and live versions like this:
- dev: http://www.foo.bar/dev
- beta: http://www.foo.bar/staging
- live: http://www.foo.bar

### LINKED ISSUE #2102: Add path-based routing to the router.
I'd like to be able to use one subdomain and route my apps on the sub-folder level.
Like www.myapp.com/api
www.myapp.com/app
It'll be really helpful in cases where the end user doesn't want to deal with cors issues.

josephglanville:
This change set adds path based routing to the router.

Closes #2069 
Closes #2102

### DIFF
--- a/cli/route.go
+++ b/cli/route.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"strconv"
 	"strings"
@@ -43,6 +44,8 @@ Examples:
 
 	$ flynn route add http example.com
 
+	$ flynn route add http example.com/path/
+
 	$ flynn route add tcp
 `)
 }
@@ -79,8 +82,8 @@ func runRoute(args *docopt.Args, client *controller.Client) error {
 	w := tabWriter()
 	defer w.Flush()
 
-	var route, protocol, service, sticky string
-	listRec(w, "ROUTE", "SERVICE", "ID", "STICKY")
+	var route, protocol, service, sticky, path string
+	listRec(w, "ROUTE", "SERVICE", "ID", "STICKY", "PATH")
 	for _, k := range routes {
 		switch k.Type {
 		case "tcp":
@@ -96,8 +99,9 @@ func runRoute(args *docopt.Args, client *controller.Client) error {
 				protocol = "https"
 			}
 			sticky = fmt.Sprintf("%t", k.Sticky)
+			path = k.HTTPRoute().Path
 		}
-		listRec(w, protocol+":"+route, service, k.FormattedID(), sticky)
+		listRec(w, protocol+":"+route, service, k.FormattedID(), sticky, path)
 	}
 	return nil
 }
@@ -138,12 +142,18 @@ func runRouteAddHTTP(args *docopt.Args, client *controller.Client) error {
 		return err
 	}
 
+	u, err := url.Parse("http://" + args.String["<domain>"])
+	if err != nil {
+		return fmt.Errorf("Failed to parse %s as URL", args.String["<domain>"])
+	}
+
 	hr := &router.HTTPRoute{
 		Service: service,
-		Domain:  args.String["<domain>"],
+		Domain:  u.Host,
 		TLSCert: tlsCert,
 		TLSKey:  tlsKey,
 		Sticky:  args.Bool["--sticky"],
+		Path:    u.Path,
 	}
 	route := hr.ToRoute()
 	if err := client.CreateRoute(mustApp(), route); err != nil {
--- a/pkg/postgres/postgres.go
+++ b/pkg/postgres/postgres.go
@@ -18,6 +18,7 @@ const (
 	InvalidTextRepresentation = "22P02"
 	CheckViolation            = "23514"
 	UniqueViolation           = "23505"
+	RaiseException            = "P0001"
 )
 
 type Conf struct {
@@ -174,8 +175,8 @@ func IsUniquenessError(err error, constraint string) bool {
 	return false
 }
 
-func IsCheckViolation(err error) bool {
-	if e, ok := err.(pgx.PgError); ok && e.Code == CheckViolation {
+func IsPostgresCode(err error, code string) bool {
+	if e, ok := err.(pgx.PgError); ok && e.Code == code {
 		return true
 	}
 	return false
--- a/router/api.go
+++ b/router/api.go
@@ -55,22 +55,26 @@ func (api *API) CreateRoute(ctx context.Context, w http.ResponseWriter, req *htt
 
 	err := l.AddRoute(route)
 	if err != nil {
-		if err == ErrConflict {
-			rjson, err := json.Marshal(&route)
-			if err != nil {
-				log.Error(err.Error())
-				httphelper.Error(w, err)
-				return
-			}
-			httphelper.Error(w, httphelper.JSONError{
-				Code:    httphelper.ConflictErrorCode,
-				Message: "Duplicate route",
-				Detail:  rjson,
-			})
+		rjson, jerr := json.Marshal(&route)
+		if jerr != nil {
+			log.Error(jerr.Error())
+			httphelper.Error(w, jerr)
 			return
 		}
-		log.Error(err.Error())
-		httphelper.Error(w, err)
+		jsonError := httphelper.JSONError{Detail: rjson}
+		switch err {
+		case ErrConflict:
+			jsonError.Code = httphelper.ConflictErrorCode
+			jsonError.Message = "Duplicate route"
+		case ErrInvalid:
+			jsonError.Code = httphelper.ValidationErrorCode
+			jsonError.Message = "Invalid route"
+		default:
+			log.Error(err.Error())
+			httphelper.Error(w, err)
+			return
+		}
+		httphelper.Error(w, jsonError)
 		return
 	}
 	httphelper.JSON(w, 200, route)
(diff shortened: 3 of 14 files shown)


philiplb:
Awesome!

lmars:
Looks good so far...

josephglanville:
This is 99% done but I thought of a case we need to have a decent answer for.
If I have path based routes and I go to delete the default route, what should the behaviour be?
Should we prevent the user from deleting a route with dependent path based routes or should we allow them to and delete all the path based routes using a trigger?

josephglanville:
I went with making deletions of default routes with dependent routes fail. We can always the ability to delete all the routes for a domain in a single transaction later.

josephglanville:
@titanous @lmars this is ready for review.

cli/route.go
```diff
@@ -79,14 +79,15 @@ func runRoute(args *docopt.Args, client *controller.Client) error {
...
-	listRec(w, "ROUTE", "SERVICE", "ID", "STICKY")
+	var route, protocol, service, sticky, path string
+	listRec(w, "ROUTE", "SERVICE", "ID", "STICKY", "PATH")
 	for _, k := range routes {
 		switch k.Type {
 		case "tcp":
 			protocol = "tcp"
 			route = strconv.Itoa(k.TCPRoute().Port)
 			service = k.TCPRoute().Service
+			path = ""
```
titanous: This is unnecessary, `""` is the zero value of a string.

pkg/postgres/postgres.go
```diff
@@ -180,3 +181,10 @@ func IsCheckViolation(err error) bool {
 	}
 	return false
 }
+
+func IsRaiseException(err error) bool {
+	if e, ok := err.(pgx.PgError); ok && e.Code == RaiseException {
```
titanous: I wonder if this function should be generalized to take a Postgres error code constant instead? (could replace `IsCheckViolation` too)

router/api.go
```diff
@@ -55,22 +55,31 @@ func (api *API) CreateRoute(ctx context.Context, w http.ResponseWriter, req *htt
...
+		rjson, jerr := json.Marshal(&route)
+		if jerr != nil {
+			log.Error(jerr.Error())
+			httphelper.Error(w, jerr)
 			return
 		}
-		log.Error(err.Error())
-		httphelper.Error(w, err)
+		var message string
+		var code httphelper.ErrorCode
```
titanous: Instead of these variables, we could do one `JSONError` variable and assign into the struct fields directly.

router/http.go
```diff
@@ -238,7 +238,23 @@ func (h *httpSyncHandler) Set(data *router.Route) error {
...
+			log.Printf("router: default route exists, overwriting")
+			tree.backend = r
+		} else {
+			log.Printf("router: doesn't yet exist, inserting")
+			h.l.domains[strings.ToLower(r.Domain)] = NewTree(r)
+		}
+	} else {
+		log.Printf("router: adding route domain: %s, path: %s", r.Domain, r.Path)
+		// TODO(jpg): fix #1002 to make this safe. This can potentially
+		// insert nothing if the datastore it out of sync.
```
titanous: s/it out/is out

router/http.go
```diff
@@ -238,7 +238,23 @@ func (h *httpSyncHandler) Set(data *router.Route) error {
 	service.refs++
 	r.service = service
 	h.l.routes[data.ID] = r
-	h.l.domains[strings.ToLower(r.Domain)] = r
+	if data.Path == "/" {
+		log.Printf("router: adding default route for domain: %s", r.Domain)
```
lmars: Use `log15` (there is a global [logger variable](https://github.com/flynn/flynn/blob/master/router/server.go#L20))

router/http.go
```diff
@@ -238,7 +238,23 @@ func (h *httpSyncHandler) Set(data *router.Route) error {
...
+		if tree, ok := h.l.domains[strings.ToLower(r.Domain)]; ok {
+			log.Printf("router: default route exists, overwriting")
+			tree.backend = r
+		} else {
+			log.Printf("router: doesn't yet exist, inserting")
```
lmars: This will look odd in the logs, maybe change to `default route doesn't exist, inserting`

router/types/types.go
```diff
@@ -33,6 +33,10 @@ type Route struct {
 	// Sticky is whether or not to use sticky sessions for this route. It is only
 	// used for HTTP routes.
 	Sticky bool `json:"sticky,omitempty"`
+	// Path is the optional prefix to route to this service. It's exclusive with
+	// the TLS options and can only be set if a "default" route with the same domain
+	// and no Path already exists in the route table.
```
lmars: Probably worth mentioning it must have both a leading and trailing slash (is that necessary?).

josephglanville: It's not strictly necessary, the in-memory data-structure that the router engine uses for determining backends based on path can handle arbitrary input, however I wanted it to be consistent.

router/http.go
josephglanville: I'm thinking of removing those entirely now that I look at them, they were mostly useful when debugging.

router/types/types.go
lmars: Ok so personally I would expect to be able to omit at least the trailing slash, e.g:

```
flynn route add http --service assets foo.example.com /assets
```

And then looking at that, might be nice to not need the space?

```
flynn route add http --service assets foo.example.com/assets
```

router/schema.go
```diff
@@ -80,5 +80,57 @@ CREATE TRIGGER notify_http_route_update
 	AFTER INSERT OR UPDATE OR DELETE ON http_routes
 	FOR EACH ROW EXECUTE PROCEDURE notify_http_route_update()`,
 	)
+	m.Add(2,
+		`ALTER TABLE http_routes ADD COLUMN path varchar(255) NOT NULL`,
```
titanous: I there a reason to limit this to 255 characters?

router/schema.go
```diff
@@ -80,5 +80,57 @@ CREATE TRIGGER notify_http_route_update
...
+		RETURN NEW;
+	END IF;
+
+	-- If no path supplied then override it to '/', the default path
+	IF NEW.path = '' OR NULL THEN
+		NEW.path := '/';
+	END IF;
+
+	-- Validate the path
+	IF NEW.path !~* '^\/.*\/$|^\/$' THEN
```
titanous: I think this also might work:

```text
^\/(.*\/)?$
```

router/schema.go
josephglanville: I can't remember, but looking at it we have no reason to use length restricted field.. I will make it use `text` instead.

router/types/types.go
josephglanville: I changed it to work this way, it still wants a trailing slash otherwise it makes ensuring there isn't ambiguity a bit difficult.

test/test_cli.go
```diff
@@ -418,7 +418,29 @@ func (s *CLISuite) TestRoute(t *c.C) {
...
+
+	// flynn route add domain path duplicate
+	dupRoute = app.flynn("route", "add", "http", route+"/path/")
+	t.Assert(dupRoute, c.Not(Succeeds))
+	t.Assert(dupRoute.Output, c.Equals, "conflict: Duplicate route\n")
+
+	// flynn route add domain path invalid
+	invRoute := app.flynn("route", "add", "http", route+"/path")
+	t.Assert(invRoute, c.Not(Succeeds))
+	t.Assert(invRoute.Output, c.Equals, "validation_error: Invalid route\n")
```
titanous: Perhaps the error message could be nicer if it's just a missing trailling slash?

router/http_test.go
```diff
@@ -232,6 +232,57 @@ func (s *S) TestWildcardRouting(c *C) {
...
+		Service: "2",
+		Path:    "/2/",
+	}.ToRoute())
+
+	discoverdRegisterHTTPService(c, l, "1", srv1.Listener.Addr().String())
+	discoverdRegisterHTTPService(c, l, "2", srv2.Listener.Addr().String())
+
+	// Check that traffic received at the path is directed to correct backend
+	assertGet(c, "http://"+l.Addr, "foo.bar", "1")
+	assertGet(c, "http://"+l.Addr+"/2/", "foo.bar", "2")
```
titanous: What about without the trailing slash?

josephglanville: It will be directed even without the trailing slash, I couldn't work out which behaviour was preferable so settled on directing both as it's maybe a little less surprising for some.
I can add a test here to demonstrate.

titanous: Yes, that behavior makes sense, we should have a test to lock it in.

test/test_cli.go
josephglanville: We can, alternatively we could auto-add the trailing slash. What do you think is best?

titanous: I think adding the trailing slash is reasonable, we can add a new "wildcard" syntax later if we want to do raw substring prefix matching.

titanous:
LGTM!

temujin9:
Can we get an update to https://flynn.io/docs#routes to cover this?

titanous:
@jvatic please update the docs on the site.

jvatic:
Updated https://flynn.io/docs/cli#route, will update https://flynn.io/docs#routes shortly.

jvatic:
https://flynn.io/docs#routes is updated (#2312)

temujin9:
Thanks, Jesse.

Cheers,

--
Nathaniel Eliot
T9 Productions

PR #2234, opened 2015-12-03
util/release: Strip leading "v" from version in vagrant manifest

### LINKED ISSUE #2230: Vagrant manifest has broken versions
The Vagrant manifest does not support letters in the version string. Recently we started adding the `v` prefix to the version manifest, which completely breaks Vagrant:

```text
$ vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Checking if box 'flynn-base' is up to date...
A version of the box you're loading is formatted in a way that
Vagrant cannot parse: 'v20151201.3'. Please reformat the version
to be properly formatted. It should be in the format of
X.Y.Z.
```

lmars:
Fixes #2230.

### DIFF
--- a/util/release/types/manifest.go
+++ b/util/release/types/manifest.go
@@ -3,6 +3,7 @@ package release
 import (
 	"bytes"
 	"sort"
+	"strings"
 )
 
 const maxVersions = 5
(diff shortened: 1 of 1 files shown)


josephglanville:
LGTM

PR #2239, opened 2015-12-04
installer: Move prompts out of modal

jvatic:
This makes it possible to do other things (e.g. launch another cluster) before responding to the prompt.

screen shot 2015-12-03 at 21 07 52
screen shot 2015-12-03 at 21 07 55
screen shot 2015-12-03 at 21 07 15


lmars:
LGTM

PR #2243, opened 2015-12-05
controller: Copy app metadata to host jobs

josephglanville:


josephglanville:
@lmars thoughts? I was trying to implement only tailing logs for system jobs but `flynn-host` didn't have the app metadata which contains the `flynn-system-app` tag. This change copies that (and any other metadata) to the host when the jobs are launched via the scheduler.

There is also the problem of `flynn-system-app` not be populated on jobs started prior to the controller in bootstrap, separate problem though.

lmars:
LGTM (please open an issue for some system apps not having `flynn-system-app` set)

PR #2244, opened 2015-12-05
bootstrap: Populate flynn-system-app meta on jobs before controller start

josephglanville:
Populates the `flynn-system-app` meta value on jobs that are started before the controller.


titanous:
LGTM

lmars:
LGTM

PR #2245, opened 2015-12-05
host/cli: collect-debug-info only collect system app logs by default

josephglanville:
Right now this doesn't properly work when it has to fallback to on disk files.
Suggestions on what to do in that case?

### DIFF
--- a/host/cli/collect-debug-info.go
+++ b/host/cli/collect-debug-info.go
@@ -42,6 +42,7 @@ usage: flynn-host collect-debug-info [options]
 Options:
   --tarball          Create a tarball instead of uploading to a gist
   --include-env      Include sensitive environment variables
+  --include-apps     Include logs of non-system applications
   --job-lines=<num>  Maximum number of lines to fetch from each job [default: 1000]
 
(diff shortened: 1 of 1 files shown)


jvatic:
LGTM

lmars:
> Right now this doesn't properly work when it has to fallback to on disk files.

We could put the system logs in `/var/log/flynn/system`

josephglanville:
I'm thinking this is a good start. Redirecting system logs to a different directory turned out to be more complex than I had hoped and I don't think the benefit out-weighs that complexity.

Maybe we could prompt the user on fallback so they are aware that it's going to upload everything that is stored locally? (and have the chance to abort instead if they want)

titanous:
Yeah, a prompt on fallback is fine.

jvatic:
Note that the installer will need updating to handle the prompt

josephglanville:
@jvatic I think it would be reasonable for the installer to pass flag to skip prompt as there should be no non-system apps at installation time? Unless the installer can trigger debug logs outside of initial install/bootstrap?

jvatic:
@josephglanville Yeah, that's reasonable

PR #2251, opened 2015-12-07
installer: Prompt for credentials on auth failure

### LINKED ISSUE #2225: Cloud Installer: Azure - expired OAuth Token & Cluster delete
I wanted to create a new cluster but the registered OAuth token is expired. It's ok but after the expiration i cannot delete the cluster which is attached to an expired token and i cannot delete the expired token because is attached to a cluster. It's a catch-22 situation.

Maybe some token swapping method will be a solution?

I cloud not find out how to refresh those tokens at the Azure management portal.

jvatic:
Closes #2225


titanous:
LGTM

PR #2265, opened 2015-12-09
controller: Remove "crashed" and "failed" job states

lmars:
This consolidates "crashed" and "failed" job states into the "down" state, adding `job.ExitStatus` and `job.HostError` to distinguish between those states.

It also makes the states an explicit type.

### DIFF
--- a/controller/types/types.go
+++ b/controller/types/types.go
@@ -89,16 +89,33 @@ type Key struct {
 }
 
 type Job struct {
-	ID        string            `json:"id,omitempty"`
-	AppID     string            `json:"app,omitempty"`
-	ReleaseID string            `json:"release,omitempty"`
-	Type      string            `json:"type,omitempty"`
-	State     string            `json:"state,omitempty"`
-	Cmd       []string          `json:"cmd,omitempty"`
-	Meta      map[string]string `json:"meta,omitempty"`
-	CreatedAt *time.Time        `json:"created_at,omitempty"`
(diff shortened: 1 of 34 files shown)


jvatic:
The dashboard needs updating as it currently relies on `crashed` and `failed` states. I'll push the necessary changes to this branch later today.

controller/types/types.go
```diff
@@ -89,17 +89,34 @@ type Key struct {
...
-	CreatedAt *time.Time        `json:"created_at,omitempty"`
-	UpdatedAt *time.Time        `json:"updated_at,omitempty"`
+	ID         string            `json:"id,omitempty"`
+	AppID      string            `json:"app,omitempty"`
+	ReleaseID  string            `json:"release,omitempty"`
+	Type       string            `json:"type,omitempty"`
+	State      JobState          `json:"state,omitempty"`
+	Cmd        []string          `json:"cmd,omitempty"`
+	Meta       map[string]string `json:"meta,omitempty"`
+	ExitStatus int32             `json:"exit_status"`
```
jvatic: What do you think about making this a pointer so the JSON output of jobs in a non-down state don't include `"exit_status": 0`?

jvatic:
Pushed changes to dashboard.

controller/types/types.go
lmars: Yes that its better, changed in 9360c17

lmars:
This is ready for review.

titanous:
LGTM

PR #2270, opened 2015-12-09
host: Add jobs to state as early as possible

### LINKED ISSUE #2134: scheduler: Accepted jobs which do not start are not retried
If `host.AddJob` succeeds but the resulting job never starts (i.e. no job events are received for the job) the scheduler should try starting another job (and then stop one of the jobs if the original one ends up starting after a while).

Failure to handle this case can be seen in [this CI failure](https://ci.flynn.io/builds/20151115010657-73c9f645#default-jobs). The test scaled the `flies-patronize-timmins` app to two printer jobs in `SchedulerSuite.TestScale` but only got 1 up event because job `abb6a4b3-8b337462-88b2-45ca-9f57-445dfa9f99c1` (which the scheduler requested) did not start due to #1556.

lmars:
This means that if a job fails to start, event consumers will always be notified.

Fixes #2134.

### DIFF
--- a/host/http.go
+++ b/host/http.go
@@ -285,6 +285,8 @@ func (h *jobAPI) AddJob(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
 		return
 	}
 
+	h.host.state.AddJob(job)
+
 	go func() {
 		// TODO(titanous): ratelimit this goroutine?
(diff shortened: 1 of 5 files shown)


titanous:
LGTM

PR #2271, opened 2015-12-09
test: Don't deploy discoverd in single node cluster

lmars:
The deploy fails and leaves the cluster totally broken.


titanous:
LGTM

PR #2273, opened 2015-12-10
pkg/syslog/rfc6587: Fix parsing of lines that are the maximum length

titanous:
This fixes a bug that caused logaggregator to stop showing new log lines for some apps when they emitted lines that were 10,000 bytes or longer.

/cc @temujin9


josephglanville:
Good find, LGTM.

PR #2274, opened 2015-12-10
logaggregator: Add missing GET /snapshot endpoint

titanous:
This endpoint is always requested at startup, but the implementation was never committed.


josephglanville:
LGTM

lmars:
LGTM

PR #2280, opened 2015-12-12
cli: Remove legacy SSH git code

titanous:
We switched off of SSH git at the end of August, so it’s very unlikely that anyone is using it.


lmars:
LGTM

PR #2282, opened 2015-12-12
scheduler: Rectify omni job counts when a host goes down

### LINKED ISSUE #2277: scheduler: Omni jobs are restarted after a host goes down
When a host goes down, we [mark the jobs on that host as down](https://github.com/flynn/flynn/blob/master/controller/scheduler/scheduler.go#L706-L711) which [triggers a restart](https://github.com/flynn/flynn/blob/master/controller/scheduler/scheduler.go#L920-L922) because the formation has not changed (so the scheduler believes the job should still be running).

This is fine for normal jobs, but for omni jobs this means the job which was on the host that just went down gets started on a different host, and only when the next `SyncFormations` happens will the scheduler modify the omni job counts and realise it has too many jobs.

The scheduler should modify the omni job counts before marking jobs as stopped so that they are not restarted.

lmars:
Fixes #2277.

### DIFF
--- a/controller/scheduler/scheduler.go
+++ b/controller/scheduler/scheduler.go
@@ -546,9 +546,13 @@ func (s *Scheduler) HandlePlacementRequest(req *PlacementRequest) {
 	formation := req.Job.Formation
 	counts := s.jobs.GetHostJobCounts(formation.key(), req.Job.Type)
 	var minCount int = math.MaxInt32
-	for id, h := range s.hosts {
-		count, ok := counts[id]
-		if !ok || count < minCount {
+	for _, h := range s.SortedHosts() {
+		count, ok := counts[h.ID]
+		if !ok || count == 0 {
+			req.Host = h
+			break
+		}
+		if count < minCount {
 			minCount = count
 			req.Host = h
 		}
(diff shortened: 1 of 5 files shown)


titanous:
LGTM

lmars:
@titanous I realised what I had would continually multiply the counts when hosts change rather than multiplying the original count from the controller.

Have fixed by keeping a reference to the original processes, and made the test do more explicit tests on the jobs and their states when adding / removing hosts.

Please take another look.

controller/scheduler/scheduler.go
```diff
@@ -931,6 +953,11 @@ func (s *Scheduler) handleFormation(ef *ct.ExpandedFormation) (formation *Format
 	log := logger.New("fn", "handleFormation", "app.id", ef.App.ID, "release.id", ef.Release.ID)
 
 	defer func() {
+		// ensure the formation has the correct omni job counts
+		if changed := formation.RectifyOmni(len(s.hosts)); changed {
```
titanous: No need to assign to `changed` since the method returns a boolean.

titanous:
LGTM pending comment.

controller/scheduler/scheduler.go
lmars: Although not strictly needed, I added it as a kind of documentation that the return value indicates that the formation changed, but I think that is clear from the comment of the method, so I've removed it.

PR #2289, opened 2015-12-14
updater: Fix MinVersion logic

titanous:
- The app version should be before the MinVersion to be skipped, not the reverse.
- If an image is skipped it should actually be skipped.


lmars:
LGTM

PR #2291, opened 2015-12-14
router: Add default to HTTP route path column

titanous:
This allows updates and backup restores to work. Before this change they would fail with:

    ERROR: column "path" contains null values (SQLSTATE 23502)

### DIFF
--- a/router/schema.go
+++ b/router/schema.go
@@ -81,7 +81,7 @@ CREATE TRIGGER notify_http_route_update
 	FOR EACH ROW EXECUTE PROCEDURE notify_http_route_update()`,
 	)
 	m.Add(2,
-		`ALTER TABLE http_routes ADD COLUMN path text NOT NULL`,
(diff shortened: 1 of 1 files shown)


jvatic:
LGTM

router/schema.go
```diff
@@ -137,5 +137,11 @@ CREATE TRIGGER check_http_route_update
 	FOR EACH ROW
 	EXECUTE PROCEDURE check_http_route_update()`,
 	)
+	m.Add(3,
+		// Ensure the default is set on the path column. We set this above, but
+		// there was a release without it, so this will fix any old databases
+		// that have the broken release and have already run migration 2.
+		`ALTER TABLE http_routes ALTER COLUMN path SET DEFAULT '/'`,
```
lmars: Is it worth mentioning the specific broken release version?

lmars:
LGTM pending comment

PR #2295, opened 2015-12-15
controller: Add "pending" job state

lmars:
Pending jobs are ones which are scheduled to start in the future (either they have been created in-memory by the scheduler and a host is being picked to place the job on, or it is a scheduled restart of a crashed job).

Since pending jobs do not have a host ID, I have migrated the database so the primary key for jobs is the UUID part of the job ID, and the UUID is shown by the CLI whenever the job ID is not set (e.g. `flynn ps` showing pending jobs).

I have also changed the schedulers so that only the leader persists job state to the database, meaning we can drop the state transition validation and use a table rather than an enum.

### DIFF
--- a/cli/ps.go
+++ b/cli/ps.go
@@ -17,10 +17,11 @@ List flynn jobs.
 Example:
 
 	$ flynn ps
-	ID                                      TYPE  RELEASE
-	flynn-bb97c7dac2fa455dad73459056fabac2  web   b69d7fb5308a4684a09b160b82d267ec
-	flynn-c59e02b3e6ad49809424848809d4749a  web   b69d7fb5308a4684a09b160b82d267ec
-	flynn-46f0d715a9684e4c822e248e84a5a418  web   b69d7fb5308a4684a09b160b82d267ec
+	ID                                         TYPE  STATE    RELEASE
+	318810fb-4679-419b-aed4-b0838c71c0eb       web   pending  d2ab4264-a647-4dc2-ac8d-d5821a475962
+	5bc89fe1-d4b5-4021-a337-10dd2b391358       web   pending  d2ab4264-a647-4dc2-ac8d-d5821a475962
+	host-93612073-f06e-41d9-bdae-df3e45f8a11d  web   up       d2ab4264-a647-4dc2-ac8d-d5821a475962
+	host-53dba8f4-561b-460b-b75b-677e8b6660fb  web   up       d2ab4264-a647-4dc2-ac8d-d5821a475962
 `)
 }
 
@@ -34,15 +35,19 @@ func runPs(args *docopt.Args, client *controller.Client) error {
 	w := tabWriter()
 	defer w.Flush()
 
-	listRec(w, "ID", "TYPE", "RELEASE")
+	listRec(w, "ID", "TYPE", "STATE", "RELEASE")
 	for _, j := range jobs {
 		if j.Type == "" {
 			j.Type = "run"
 		}
-		if j.State != ct.JobStateUp {
+		if j.State != ct.JobStateUp && j.State != ct.JobStatePending {
 			continue
 		}
-		listRec(w, j.ID, j.Type, j.ReleaseID)
+		id := j.ID
+		if id == "" {
+			id = j.UUID
+		}
+		listRec(w, id, j.Type, j.State, j.ReleaseID)
 	}
 
 	return nil
--- a/cli/ps.go
+++ b/cli/ps.go
@@ -2,7 +2,9 @@ package main
 
 import (
 	"sort"
+	"time"
 
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/docker/docker/pkg/units"
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
 	"github.com/flynn/flynn/controller/client"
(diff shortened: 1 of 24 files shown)


josephglanville:
This migration will suffer from the same problem as the other enum conversions. I am working on a solution now but it will probably end up being that we setup a listen/notify on the migrations table and have every client of the database reset their connection pool.

titanous:
@josephglanville Please take a close look at the DB migration.

@lmars Have you tested the migration against a database that already has data in it?

lmars:
@titanous added a test of the migration in c1eb855, I will perform a manual test on a running cluster too.

controller/migrate_test.go
```diff
@@ -0,0 +1,55 @@
...
+	. "github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check"
+)
+
+type MigrateSuite struct{}
+
+var _ = Suite(&MigrateSuite{})
+
+// TestMigrations checks that migrating from one ID to the next works when
+// data exists in the database
+func (MigrateSuite) TestMigrations(c *C) {
```
titanous: This should be named for the specific migration.

titanous:
LGTM

josephglanville:
@lmars make sure you run it on a multi-node cluster. The problem only manifests when you have existing connections that have prepared queries that are effected by an `ALTER COLUMN TYPE`.
Specifically it will thrown an error that the type of the result doesn't match that of the cached query plan.

schema/controller/job.json
```diff
@@ -13,8 +13,16 @@
   "additionalProperties": false,
   "properties": {
     "id": {
+      "type": "string",
+      "description": "the job's cluster ID"
+    },
+    "uuid": {
```
jvatic: Why not make `id` the UUID and add a `cluster_id` property?

lmars: For backwards compatibility, the idea being that old clients will still be able to post `id` being the cluster ID, and newer clients use `uuid`. Not ideal I know, but I think necessary.

lmars:
@josephglanville can you write a failing test for the enum issue? Not really sure how to avoid it?

josephglanville:
@lmars this demonstrates the problem: https://github.com/flynn/flynn/commit/50852e49613bc9ea87118f37cc840cc66cc6183d

In this specific case the error is slightly different to what we have seen but the problem is the same.
We need a strategy to de-allocate all prepared queries and flush any session cached query plans whenever we run a migration. Listen/notify on the migrations table is my current best idea.

josephglanville:
The updated test at https://github.com/flynn/flynn/commit/46d5daf7126e4bd3aa32c3b1d6838dd77d47b8d9 shows that resetting the connection pool properly deals with this problem.

cli/ps.go
```diff
@@ -10,17 +12,32 @@ import (
...
+	ID                                         TYPE  STATE    CREATED             RELEASE
+	ab14754c-73b7-4212-a6d9-73b825587fd2       web   pending  2 seconds ago       1b1db8ef-ba4d-4314-85c1-d5895a44b27e
+	6ec25d6e-2985-4807-8e64-02dc23c348bc       web   pending  7 seconds ago       1b1db8ef-ba4d-4314-85c1-d5895a44b27e
+	host-f25797dc-c956-4337-89af-d49eff50f58e  web   up       14 seconds ago      1b1db8ef-ba4d-4314-85c1-d5895a44b27e
+	host-ef6249e3-b463-4fea-a7dd-b1302872f821  web   down     20 seconds ago      1b1db8ef-ba4d-4314-85c1-d5895a44b27e
+	host-cef95d8c-a632-4ae6-8a57-e13cbc24afa9  web   down     25 seconds ago      1b1db8ef-ba4d-4314-85c1-d5895a44b27e
+	host-95747b4f-fdcd-4c44-ab72-b3d9609668e4  run   down     54 seconds ago      cd698657-2955-4fa4-bc2f-8714b218a7a2
+	host-e8b4f9be-e422-481f-928c-4c82f0bb5e8b  web   down     About a minute ago  cd698657-2955-4fa4-bc2f-8714b218a7a2
+	host-feaff633-a37b-4565-9ade-24bae0cfae03  web   down     About a minute ago  cd698657-2955-4fa4-bc2f-8714b218a7a2
+	host-d84dc657-83b8-4a62-aab8-ad97bb994761  web   down     2 minutes ago       cd698657-2955-4fa4-bc2f-8714b218a7a2
```
titanous: Does it make sense to sort in this direction? If the list scrolls off the screen, you'd have to scroll up to see the most relevant data.

lmars: Hmm if it's a long list I'd probably pipe it to less, in which case I would see the most relevant data at the top? I don't feel strongly either way.

titanous: Let's do newest at the bottom.

lmars:
@titanous comments addressed

cli/ps.go
```diff
@@ -10,17 +12,32 @@ import (
 
 func init() {
 	register("ps", runPs, `
-usage: flynn ps
+usage: flynn ps [-a]
 
 List flynn jobs.
 
+Options:
+  -a, --all      Show all jobs (default is just running and pending)
```
titanous: s/just//

titanous:
LGTM pending comment.

PR #2296, opened 2015-12-15
router: Handle route deletions missed on reconnect

### LINKED ISSUE #1002: router: recover missed route deletions after sync failure

josephglanville:
This tracks which routes were removed while the sync is disconnected and removes them on reconnect.
Right now on reconnect we still fetch the entire route table from the database.

Closes #1002

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -45,6 +45,9 @@ type HTTPListener struct {
 	closed      bool
 	cookieKey   *[32]byte
 	keypair     tls.Certificate
+
+	preSync  func()
+	postSync func(<-chan struct{})
 }
 
 type DiscoverdClient interface {
@@ -127,7 +130,15 @@ func (s *HTTPListener) runSync(ctx context.Context, errc chan error) {
 
 		time.Sleep(2 * time.Second)
 
-		s.doSync(ctx, errc)
+		if s.preSync != nil {
+			s.preSync()
+		}
+
+		startc := s.doSync(ctx, errc)
+
+		if s.postSync != nil {
+			s.postSync(startc)
+		}
 
 		err = <-errc
 	}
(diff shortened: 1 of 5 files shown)


josephglanville:
There is a further change needed to make syncing safer for path based routes which involves sorting them before insertion into the routing trie.

I had a bigger overhaul in mind but it was a little ambitious so scaling back to fixing the correctness issues first.

josephglanville:
A bit of thought might actually suggest no changes are needed, because of the strict insertion/deletion order enforced by the constraint trigger. I will try modify the test to see if I can break it.

router/http_test.go
```diff
@@ -313,6 +316,123 @@ func (s *S) TestHTTPInitialSync(c *C) {
 	assertGet(c, "https://"+l.TLSAddr, "example.com", "1")
 }
 
+func (s *S) TestHTTPResync(c *C) {
+	connPids := make([]int32, 0)
```
titanous: Don't use `make` with `len = 0`:

```go
var connPids []int32
```

router/http_test.go
```diff
@@ -313,6 +316,123 @@ func (s *S) TestHTTPInitialSync(c *C) {
...
+		}
+		defer conn.Close()
+
+		// terminate all conns from the pool
+		for _, pid := range connPids {
+			if _, err := conn.Exec("select pg_terminate_backend($1)", pid); err != nil {
+				c.Fatalf("Unable to kill backend PostgreSQL process: %v", err)
+			}
+		}
+		connPids = make([]int32, 0)
```
titanous: Just truncate it:

```go
connPids = connPids[:0]
```

router/http.go
```diff
@@ -116,6 +116,10 @@ func (s *HTTPListener) startSync(ctx context.Context) error {
 	}
 }
 
+// Testing hooks
+var preResync func()
+var postResync func(startc <-chan struct{})
```
titanous: I'd move these to the listener struct so that multiple tests can be run concurrently.

titanous:
LGTM

PR #2303, opened 2015-12-16
controller: Make Job.HostError a pointer

lmars:
It can be NULL in the database, so fetching a NULL value into a string will result in a "Cannot decode NULL into type string" error.

This affects any users updating to the latest released version, so we need to release this soon.


titanous:
LGTM

PR #2306, opened 2015-12-16
controller: Migration listener

### LINKED ISSUE #2235: controller: "cached plan must not change result type"
```text
ERROR: cached plan must not change result type (SQLSTATE 0A000)
```

Sounds like a race between the DB migrations and preparing statements when creating connections.

https://gist.github.com/anonymous/2eedd040b6bf05bcd1da

josephglanville:
This change implements a listen/notify setup on the `schema_migrations` table in order to reset the connection pool after a migration takes place. This is to ensure any session level cached or prepared query plans can be refreshed.

Refs #2295 
Closes #2235

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -83,6 +83,9 @@ func main() {
 		}
 	}()
 
+	// Listen for database migration, reset connpool on new migration
+	go postgres.ResetOnMigration(db, logger, doneCh)
+
 	hb, err := discoverd.DefaultClient.AddServiceAndRegisterInstance("controller", &discoverd.Instance{
 		Addr:  addr,
 		Proto: "http",
--- a/pkg/postgres/migrate.go
+++ b/pkg/postgres/migrate.go
@@ -1,7 +1,11 @@
 package postgres
 
 import (
+	"strconv"
+	"time"
+
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/jackc/pgx"
+	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
 )
 
 type Migration struct {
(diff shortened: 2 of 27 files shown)


josephglanville:
A similar change will be needed for the router if/when it's changed to use prepared queries, this is why I implemented the core logic in `pkg/postgres` instead of in the controller itself.

Right now the datastore actually generates the queries semi on the fly which is probably terrible for performance with pgx. (which anonymously prepares queries on exec if they aren't prepared already..)
That said because it doesn't use prepared queries means it's probably immune from these sorts of errors right now.

controller/controller.go
```diff
@@ -83,6 +83,11 @@ func main() {
 		}
 	}()
 
+	// Listen for database migration, reset connpool on new migration
+	go func() {
+		postgres.ResetOnMigration(db, logger, doneCh)
```
titanous: No need to wrap in an anonymous function, you can just do `go postgres.ResetOnMigration...`

pkg/postgres/migrate.go
```diff
@@ -65,3 +73,30 @@ func (m Migrations) Migrate(db *DB) error {
...
+				}
+				log.Warn("new schema migration, resetting conn pool", "id", n.Payload)
+				db.Reset()
+			case _ = <-doneCh:
+				listener.Close()
+				return
+			}
+		}
+	}
+	return
```
titanous: I think this return is uncessary.

pkg/postgres/migrate.go
```diff
@@ -65,3 +73,30 @@ func (m Migrations) Migrate(db *DB) error {
...
+			case n, ok := <-listener.Notify:
+				if !ok {
+					log.Warn("migration listener disconnected, reconnecting in 5 seconds")
+					time.Sleep(5 * time.Second)
+					listener.Close()
+					break outer
+				}
+				log.Warn("new schema migration, resetting conn pool", "id", n.Payload)
+				db.Reset()
+			case _ = <-doneCh:
```
titanous: No need for `_ =`

josephglanville:
Comments addressed, cheers for review, still getting used to programming with channels.

pkg/postgres/migrate.go
titanous: Was the `return` required?

titanous:
LGTM

Issue #2308, opened 2015-12-16
Deployment of a Symfony 2.7 PHP application fails


philiplb:
Hi,

I'm trying to deploy an PHP application and suddenly, I get this:

> remote: [1G[K[1G-----> Compiled slug size is 109M[K
> remote: [1G[K-----> Creating release...[K
> remote: Error deploying app release: web process type failed to start, got down job event[K
> To https://git.........com/....git
> ! [remote rejected] master -> master (pre-receive hook declined)
> error: failed to push some refs to 'https://git.........com/....git'

Log:

> 2015-12-16T16:20:58.960826Z app[web.flynn0-db77e7bf-9304-4088-b08d-cf6003406723]: Using Apache2 VirtualHost-level configuration include 'apache_app.conf'
> 2015-12-16T16:20:59.043666Z app[web.flynn0-db77e7bf-9304-4088-b08d-cf6003406723]: 4 processes at 128MB memory limit.
> 2015-12-16T16:20:59.048682Z app[web.flynn0-db77e7bf-9304-4088-b08d-cf6003406723]: Starting php-fpm...
> 2015-12-16T16:21:01.050160Z app[web.flynn0-db77e7bf-9304-4088-b08d-cf6003406723]: Starting httpd...
> 2015-12-16T16:21:01.058308Z app[web.flynn0-db77e7bf-9304-4088-b08d-cf6003406723]: Application ready for connections on port 8080.
> 2015-12-16T16:21:01.242660Z app[web.flynn0-54f50d1f-41b6-4901-976e-86befc453cc6]: Going down, terminating child processes...
> 2015-12-16T16:23:19.570054Z app[web.flynn0-177dcf75-c18e-4d1e-b27b-fd6f4389c208]: tar: This does not look like a tar archive
> 2015-12-16T16:23:19.570420Z app[web.flynn0-177dcf75-c18e-4d1e-b27b-fd6f4389c208]: tar: Exiting with failure status due to previous errors
> 2015-12-16T16:23:19.938508Z app[web.flynn0-ec31f3cd-0597-4fbd-9498-76c9c047b068]: tar: This does not look like a tar archive
> 2015-12-16T16:23:19.938816Z app[web.flynn0-ec31f3cd-0597-4fbd-9498-76c9c047b068]: tar: Exiting with failure status due to previous errors

titanous:
That error typically indicates that the app slug was not successfully retrieved. Take a look at the blobstore log with `flynn -a blobstore log`

[...]

PR #2309, opened 2015-12-16
all: Use curl --fail whenever we pipe output with --silent

titanous:
This ensures that the shell pipeline terminates on error instead of continuing with useless output from curl.

Refs #2308


lmars:
LGTM

PR #2322, opened 2015-12-17
router: Only remove from routing tree if route matches

josephglanville:
This closes a small safety issue that can happen if a path based route is added, the sync connection is lost, that path route is then deleted and re-added (not updated) with a different backend service.
By checking the path backend matches we ensure when processing removals that we don't accidentally delete this new valid route from the routing table.


titanous:
LGTM

PR #2325, opened 2015-12-18
host: Perform explicit check for running containers

lmars:
I keep looking at CI build failures and finding #907 to be the cause, so this is an attempt at a workaround.

### DIFF
--- a/host/libvirt_lxc_backend.go
+++ b/host/libvirt_lxc_backend.go
@@ -778,41 +778,66 @@ func (c *libvirtContainer) watch(ready chan<- error, buffer host.LogBuffer) erro
 		}
 	}
 
-	g.Log(grohl.Data{"at": "watch_changes"})
-	for change := range c.Client.StreamState() {
-		g.Log(grohl.Data{"at": "change", "state": change.State.String()})
-		if change.Error != "" {
-			err := errors.New(change.Error)
-			g.Log(grohl.Data{"at": "change", "status": "error", "err": err})
-			c.Client.Resume()
-			c.l.state.SetStatusFailed(c.job.ID, err)
-			return err
-		}
-		switch change.State {
-		case containerinit.StateInitial:
-			g.Log(grohl.Data{"at": "wait_attach"})
-			c.l.state.WaitAttach(c.job.ID)
-			g.Log(grohl.Data{"at": "resume"})
-			c.Client.Resume()
-			g.Log(grohl.Data{"at": "resumed"})
-		case containerinit.StateRunning:
-			g.Log(grohl.Data{"at": "running"})
-			c.l.state.SetStatusRunning(c.job.ID)
+	// containerinit sometimes fails to send the StateRunning state change
+	// when the container is running, so we do an explicit check once a
+	// second to avoid the job being stuck in the "starting" state (see
+	// https://github.com/flynn/flynn/issues/907)
+	isRunning := false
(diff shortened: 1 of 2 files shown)


titanous:
Hmm, any ideas why the running event isn't arriving?

lmars:
> Hmm, any ideas why the running event isn't arriving?

No idea, I have looked quite a few times at it and got nowhere, so decided to just add this hack.

I haven't got as far as sysdig as it is hard to reproduce, but happens in approximately 1 in 5 failed CI jobs.

host/libvirt_lxc_backend.go
```diff
@@ -778,41 +778,64 @@ func (c *libvirtContainer) watch(ready chan<- error, buffer host.LogBuffer) erro
...
+				c.Client.Resume()
+				c.l.state.SetStatusFailed(c.job.ID, errors.New("container failed to start"))
+				return nil
+			}
+		case <-runningCheck:
+			if isRunning {
+				continue
+			}
+			g.Log(grohl.Data{"at": "running_check"})
+			if state, err := c.Client.GetState(); err == nil && state == containerinit.StateRunning {
```
titanous: This API call is cheap, why not do it once a second from the beginning?

lmars: Yep makes sense, updated in 841e809

lmars:
@titanous hmm it seems `c.Client.GetState` just blocks indefinitely, and I can't see it called anywhere else so maybe it has never worked? I am not overly familiar with the RPC code, any thoughts?

lmars:
@titanous found the cause of the hang, I was creating multiple state stream channels and not listening on them which blocks the rpcplus loop [here](https://github.com/flynn/flynn/blob/master/pkg/rpcplus/client.go#L181-L185).

Fixed in d9ae843, please review.

host/libvirt_lxc_backend.go
```diff
@@ -778,41 +778,66 @@ func (c *libvirtContainer) watch(ready chan<- error, buffer host.LogBuffer) erro
...
-			g.Log(grohl.Data{"at": "change", "status": "error", "err": err})
-			c.Client.Resume()
-			c.l.state.SetStatusFailed(c.job.ID, err)
-			return err
+	// containerinit sometimes fails to send the StateRunning state change
+	// when the container is running, so we do an explicit check once a
+	// second to avoid the job being stuck in the "starting" state (see
+	// https://github.com/flynn/flynn/issues/907)
+	isRunning := false
+	runningCheck := time.Tick(time.Second)
```
titanous: This should be a `time.Ticker` and get stopped when finished with it.

host/libvirt_lxc_backend.go
```diff
@@ -778,41 +778,66 @@ func (c *libvirtContainer) watch(ready chan<- error, buffer host.LogBuffer) erro
...
+				return nil
+			case containerinit.StateFailed:
+				g.Log(grohl.Data{"at": "failed"})
+				c.Client.Resume()
+				c.l.state.SetStatusFailed(c.job.ID, errors.New("container failed to start"))
+				return nil
+			}
+		case <-runningCheck:
+			if isRunning {
+				runningCheck = nil
```
titanous: Could just nil it out in `handleRunning`, then `isRunning` would not be required.

titanous:
LGTM

PR #2326, opened 2015-12-18
controller: Deploy web processes using service events

lmars:
When deploying the controller under heavy load, the web process can actually take a few seconds before it starts listening for HTTP requests, so we need to use service events so that there is always a functioning web process during the deployment.


titanous:
LGTM

PR #2329, opened 2015-12-19
controller: Close database listeners on shutdown

### LINKED ISSUE #2327: controller: Shutdown blocks closing DB pool
The controller currently does not terminate on SIGTERM as it gets blocked trying to close the DB connection pool, likely because we have listeners which are not explicitly closed on shutdown (so are blocking the pool from closing).

This means that every time we try and stop a web job, it takes 10s before it actually stops (due to the 10s SIGKILL timeout), so deploying the controller with 2 web process is spending 20s each time waiting. Given we deploy the controller 6 or 7 times in CI, this is adding 2-3 minutes to CI builds.

/cc @josephglanville

josephglanville:
I also patched the postgres listener to simply close connections on shutdown rather than trying to unlisten as I ran into the same RxMsg bug from the initial migration. I think I might have a better idea of the root cause of that bug though so I might try work with with pgx maintainer to fix it.

Closes #2327

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -37,6 +37,7 @@ import (
 var logger = log15.New("component", "controller")
 
 var ErrNotFound = errors.New("controller: resource not found")
+var ErrShutdown = errors.New("controller: shutting down")
 
 var schemaRoot = "/etc/flynn-controller/jsonschema"
 
@@ -231,6 +232,8 @@ func appHandler(c handlerConfig) http.Handler {
 		caCert:              []byte(os.Getenv("CA_CERT")),
 		config:              c,
 	}
+
+	shutdown.BeforeExit(api.Shutdown)
 
 	httpRouter := httprouter.New()
 
@@ -429,4 +432,10 @@ func createEvent(dbExec func(string, ...interface{}) error, e *ct.Event, data in
 func (c *controllerAPI) GetCACert(_ context.Context, w http.ResponseWriter, _ *http.Request) {
 	w.Header().Set("Content-Type", "application/x-x509-ca-cert")
 	w.Write(c.caCert)
+}
+
+func (c *controllerAPI) Shutdown() {
+	if c.eventListener != nil {
+		c.eventListener.CloseWithError(ErrShutdown)
+	}
 }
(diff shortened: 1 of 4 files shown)


pkg/postgres/migrate.go
```diff
@@ -84,6 +85,9 @@ func ResetOnMigration(db *DB, log log15.Logger, doneCh chan struct{}) {
 		for {
 			select {
 			case n, ok := <-listener.Notify:
+				if closed {
+					return
```
titanous: Is the goal here to drain the notify channel? Does it get closed at the end? Will more than one notification ever be sent after closing the connection?

pkg/postgres/migrate.go
```diff
@@ -93,6 +97,7 @@ func ResetOnMigration(db *DB, log log15.Logger, doneCh chan struct{}) {
 				log.Warn("new schema migration, resetting conn pool", "id", n.Payload)
 				db.Reset()
 			case <-doneCh:
+				closed = true
 				listener.Close()
```
lmars: Should this return?

josephglanville:
Simplified a bit, old behaviour was left over from a different attempt to fix the problem.

titanous:
LGTM

lmars:
@josephglanville this doesn't fully fix the issue, the event / formation listeners also need closing right?

controller/controller.go
```diff
@@ -232,6 +233,8 @@ func appHandler(c handlerConfig) http.Handler {
 		config:              c,
 	}
 
+	shutdown.BeforeExit(func() { api.Shutdown() })
```
titanous:
```go
shutdown.BeforeExit(api.Shutdown)
```

josephglanville:
@lmars the formation listeners currently look like they do the right thing. I wasn't able to reproduce it getting stuck on the event listener but I also couldn't work how it is meant to shutdown either so I changed it.

titanous:
LGTM

lmars:
@josephglanville 

> the formation listeners currently look like they do the right thing

I don't think they do. The listener is closed if `stopListener` receives (i.e. [here](https://github.com/flynn/flynn/blob/master/controller/formation.go#L314-L315)), but that only happens if [the number of subscribers hits zero](https://github.com/flynn/flynn/blob/master/controller/formation.go#L394-L399), and that won't happen because we are not unsubscribing clients on shutdown? I think we should just force stop the listener on shutdown.

PR #2330, opened 2015-12-19
test: Skip migrating cluster domain in TestExampleOutput

lmars:
The examples attempt to migrate the cluster domain to `127.0.0.1.xip.io` [here](https://github.com/flynn/flynn/blob/master/controller/examples/examples.go#L549-L552), which breaks the cluster and is unnecessary as the test [skips checking the output](https://github.com/flynn/flynn/blob/master/test/test_controller.go#L145) anyway.


titanous:
LGTM

PR #2333, opened 2015-12-19
scheduler: Fix restarting jobs which failed to start

lmars:
Jobs which failed have a zero `StartedAt` time and should be treated as though they have been running for a zero amount of time, so therefore should not have their restarts reset.


titanous:
LGTM

PR #2335, opened 2015-12-20
controller: Stop the formation listener on shutdown

### LINKED ISSUE #2327: controller: Shutdown blocks closing DB pool

lmars:
Fixes #2327.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -435,6 +435,8 @@ func (c *controllerAPI) GetCACert(_ context.Context, w http.ResponseWriter, _ *h
 }
 
 func (c *controllerAPI) Shutdown() {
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

josephglanville:
LGTM

PR #2339, opened 2015-12-28
host: Ensure new jobs are added to state when resurrecting

### LINKED ISSUE #2338: host: Panic when resurrecting jobs
I am not sure what's causing this error, but after a succesfull installation and run, setup none or more than 1 application. When I reboot my server the flynn-host fails to start. 

I mentioned this earlier in issue #2075 and @lmars would have fixed this in PR #2337 thought maybe to be related but the reboot survival, but I don't think this addresses the same problem. The reason I think this because I didn't see anyone else facing the same issue.

This time I included a full log:
https://gist.github.com/anonymous/9c8d0e6a90f208b8c028

Could it somehow have anything to do with the way IP are assigned to containers? I am really stuck here. Also wonder if other face(d) the same problem.

Thanks!

lmars:
New jobs from API requests are now added to the state independently of the backend (see #2270), so we need to do the same when resurrecting jobs.

Fixes #2338.

### DIFF
--- a/host/state.go
+++ b/host/state.go
@@ -135,6 +135,7 @@ func (s *State) Restore(backend Backend, buffers host.LogBuffers) (func(), error
 				newID := cluster.GenerateJobID(s.id, "")
 				log.Printf("resurrecting %s as %s", job.Job.ID, newID)
 				job.Job.ID = newID
+				s.AddJob(job.Job)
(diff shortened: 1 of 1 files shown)


josephglanville:
LGTM

PR #2340, opened 2015-12-28
script: Refactor multi-node bootstrap.

josephglanville:
* Always use multi-node logic - this deprecates `dev.localflynn.com` in favor of always using `<size>.localflynn.com` even for single node.
* Split out start-flynn-host into it's own script so it can be launched manually to add or respawn a node.

### DIFF
--- a/script/start-flynn-host
+++ b/script/start-flynn-host
@@ -1,0 +1,83 @@
+#!/bin/bash
+
+set -e
+
+ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+source "${ROOT}/script/lib/ui.sh"
+
+usage() {
+  cat <<USAGE >&2
+usage: $0 IP INDEX
+
+Start a flynn node with specific binary, ip and index
+USAGE
+}
+
+main() {
+  local destroy_vols=true
+
+  while true; do
+    case "$1" in
+      -z | --no-destroy-vols)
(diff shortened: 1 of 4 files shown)


josephglanville:
@lmars I'm using these changes to test cluster resurrection/monitor stuff. I think it's cleaner overall to have one set of IPs/domain names. WDYT?

lmars:
@josephglanville the only thing I'll say is that when booting a single node cluster it is nice if components listen on all addresses, so that for example the default discoverd address `http://127.0.0.1:1111` works (and subsequently commands like `flynn-host ps`).

We can of course get around this by setting `DISCOVERD=http://1.localflynn.com:1111`, so we should set that somewhere.

script/start-flynn-host
```diff
@@ -0,0 +1,83 @@
...
+  done
+
+  if [[ $# -ne 2 ]]; then
+    usage
+    exit 1
+  fi
+
+  local ip=$1
+  local index=$2
+
```
lmars: Looks like there is some trailing whitespace on both this line and line 41.

josephglanville:
@lmars added the `DISCOVERD` env var to the Vagrantfile so that unless you are setting up your own host it should be ready and good to go.

lmars:
LGTM

Issue #2346, opened 2015-12-31
Issue on using a elixir heroku compatible buildpack 


rachid:
This will be hard one because before writing this issue I was questioning myself where to do this, because a few projects are involved.

While deploying an elixir app and in this app I use dependencies which refer to a specific git repositories, the app always ends up in missing deps, while this is not the case.

To keep my story short I'll get to the point, Elixir uses Mix, which is a build tool that provides tasks for creating, compiling, testing Elixir projects, etc.

Now, in the file:
https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/scm/git.ex#L42-L45

mix uses the .git directory to check if dependencies fetch from git are up to date.
```
def checked_out?(opts) do
    # Are we inside a git repository?
    File.regular?(Path.join(opts[:dest], ".git/HEAD"))
end
```

Is this the place where the flynn is removing those dirs? 
https://github.com/flynn/flynn/blob/master/slugbuilder/builder/install-buildpack#L39

As an example I copy pasted the difference between a directory listing on a heroku app and a flynn app.
```
➜  oberm git:(master) heroku run ls deps/phoenix_haml/.git --app oberz
Running ls deps/phoenix_haml/.git on oberz... up, run.5258
HEAD  branches  config  description  hooks  index  info  logs  objects  packed-refs  refs
```

```
➜  oberm git:(master) flynn run ls deps/phoenix_haml/.git
ls: cannot access deps/phoenix_haml/.git: No such file or directory
```

Now my question is: is flynn deleting .git directories?
And secondly why does this happens and is there a way to get around this? Where does flynn stand when it comes to keep compatible with heroku's buildpack system?

titanous:
Yeah, the `.git` directory is omitted, but I wasn't aware that Heroku included it in app slugs. We do try to maintain Heroku compatibility, but I'm a bit hesitant here because including the full `.git` directory would mean that the slug size is always a superset of the repo size. This is problematic because many users I've talked to have large repos, and in the vast majority of situations the data will never be utilized.

Is the Heroku behavior documented anywhere?

rachid:
@titanous I am not sure about Heroku's precise implementation, but as far I as I can see, when running ls on the repo I am getting the following result on the app root:

Let's run ls .git on heroku again, but now on root:
```
➜  oberm git:(master) heroku run ls .git --app oberz
Running ls .git on oberz... up, run.1565
ls: cannot access .git: No such file or directory
```
[...]

PR #2349, opened 2016-01-05
slugbuilder: Only exclude '.git' at the root of the slug

josephglanville:
This avoids removing `.git` directories in subdirectories. We still want to exclude the top level `.git` directory because there is a large overhead to storing it when working with large repositories.

Refs #2346


titanous:
LGTM (add the issue reference to the commit or PR message)

PR #2356, opened 2016-01-09
router: Workaround currently broken test in Go 1.4.3

josephglanville:
Go 1.4.3 reduced some new request smuggling protection that seems to be running amok with non-GET/HEAD requests. For now we will disable checking for connection liveliness here and track it in #2358 to re-enable it when migrating to Go 1.5 and also notify/work with upstream to get it resolved on the 1.4.X branch.


titanous:
LGTM

PR #2360, opened 2016-01-10
slugbuilder: Make .slugignore Heroku compatible

### LINKED ISSUE #2359: .slugignore file is not used to ignore certain file(s)
I deployed latest Flynn to DO with CLI v20160107.0.

I deploy a [java app](https://github.com/paukiatwee/budgetapp) with `package.json` file at the root. I also have `.slugignore` file at the root to ignore `package.json` file so Heroku will not detect the app as NodeJS app.

However, Flynn failed to use `.slugignore` file to exclude `package.json` file before detect app type (and detected as NodeJS app).

Work around is to [specify custom buildpack](https://github.com/flynn/flynn/blob/master/docs/content/faq/how-do-i-use-custom-buildpacks.md)

However, I am not sure whether Flynn is going to support `.slugignore` file or not. If no, this no longer valid issue.

josephglanville:
This changes the behavior of the `.slugignore` file in `slugbuilder` to remove files from the build root before processing buildpacks rather than simply excluding them when archiving the slug.

It also attempts to implement line handling the same as Heroku, ignoring blank lines and allowing comments in the `.slugignore` file itself.

Closes #2359

### DIFF
--- a/slugbuilder/builder/build.sh
+++ b/slugbuilder/builder/build.sh
@@ -57,6 +57,38 @@ curl() {
   $(which curl) --fail --silent --retry 3 $@
 }
 
+# removes leading and trailing whitespace
+trim() {
+  local var="$*"
+  var="${var#"${var%%[![:space:]]*}"}"
+  var="${var%"${var##*[![:space:]]}"}"
+  echo -n "$var"
+}
+
+prune_slugignore() {
+  shopt -s nullglob
+  # read slugignore into array
+  local globs=()
+  local paths=()
+  readarray -t globs < "${build_root}/.slugignore"
+  # for line in slugignore
+  for glob in ${globs[@]}; do
+    # strip whitespace
+    glob=$(trim ${glob})
+    # ignore blank lines and comment lines
+    if [[ ${glob} == "" ]] || [[ ${glob:0:1} == "#" ]]; then
+      continue
+    fi
+    # remove leading slash(es)
+    glob="${glob#"${glob%%[!"/"]*}"}"
+    # append to build root and add to array of paths to remove
+    paths=("${paths[@]}" ${build_root}/${glob})
+  done
+  echo_title "Deleting ${#paths[@]} files matching .slugignore patterns."
+  rm -f ${paths[@]}
+  shopt -u nullglob
+}
+
 cd ${app_dir}
 
 ## Load source from STDIN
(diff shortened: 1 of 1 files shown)


titanous:
Ping @lmars for bash code review.

slugbuilder/builder/build.sh
```diff
@@ -57,6 +57,38 @@ curl() {
...
+    fi
+    # remove leading slash(es)
+    glob="${glob#"${glob%%[!"/"]*}"}"
+    # append to build root and add to array of paths to remove
+    paths=("${paths[@]}" ${build_root}/${glob})
+  done
+  echo_title "Deleting ${#paths[@]} files matching .slugignore patterns."
+  rm -f ${paths[@]}
+  shopt -u nullglob
+}
```
lmars: Would be nice to add a BATS based test for this (see [release.bats](https://github.com/flynn/flynn/blob/master/script/test/release.bats) for an example), or perhaps right it in Go and call it with the `.slugignore` file as input :smile:

slugbuilder/builder/build.sh
```diff
@@ -187,25 +222,13 @@ fi
...
-    . \
-    | cat
-fi
+tar \
+  --exclude='./.git' \
+  --use-compress-program=pigz \
+  -C ${build_root} \
+  -cf ${slug_file} \
+  . \
+  | cat
```
lmars: Why do we pipe this to `cat`?

lmars: I also think this should just be on one line:

```
tar --exclude='./.git' --use-compress-program=pigz -C "${build_root}" -cf "${slug_file}" .
```

slugbuilder/builder/build.sh
josephglanville: Considering how simple the entirety of `slugbuilder` is we should probably just write the whole thing in Go and do away with the bash here.

slugbuilder/builder/build.sh
josephglanville: Yeah I didn't look into that but I too am confused that it pipes to cat now that I look at it closer.

slugbuilder/builder/build.sh
lmars: Yep agreed, if you are confident this does the right thing (it looks good from looking at it) then we should merge and migrate to Go separately

lmars:
LGTM

PR #2361, opened 2016-01-10
discoverd: Add jitter to discoverd heartbeats, fixes #823.

### LINKED ISSUE #823: Add jitter to discoverd heartbeats
Heartbeats should include a random jitter to smooth out any potential for thundering herds.

philiplb:
Signed-off-by: Philip Lehmann-Boehm <philip@philiplb.de>

### DIFF
--- a/discoverd/client/heartbeat.go
+++ b/discoverd/client/heartbeat.go
@@ -3,6 +3,7 @@ package discoverd
 import (
 	"fmt"
 	"log"
+	"math/rand"
 	"os"
 	"strings"
 	"sync"
@@ -132,6 +133,7 @@ func (h *heartbeater) client() *Client {
 
 const (
 	heartbeatInterval        = 5 * time.Second
+	heartbeatJitter          = 2 * time.Second
 	heartbeatFailingInterval = 200 * time.Millisecond
 )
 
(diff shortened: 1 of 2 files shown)


discoverd/client/heartbeat.go
```diff
@@ -131,7 +133,9 @@ func (h *heartbeater) client() *Client {
 }
 
 const (
-	heartbeatInterval        = 5 * time.Second
+	// As a random amount of time of up to half of the heartbeatInterval is
+	// subtracted, it will average to an interval of 5s.
+	heartbeatInterval        = 6666 * time.Millisecond
```
lmars: Even with the comment, it took me a minute to understand why this was 6⅔ and not 5 (I now realise the heartbeat follows a uniform distribution `X = U(3⅓,  6⅔)`, so thus has an expected value of 5).

I think it would be clearer to make this value the mean value (e.g. `meanHeartbeatInterval = 5 * time.Second`) and add another constant like `maxHeartbeatJitter`, thus making heartbeat interval follow a uniform distribution between `meanHeartbeatInterval ± maxHeartbeatJitter`.

WDYT?

philiplb: Makes sense to me as readability is more important of course. Will adjust.

philiplb:
Updated the commit with a more readable version.

philiplb:
Can I improve this little patch more? :)

philiplb:
*sniff*

PR #2363, opened 2016-01-11
test/rootfs: Increase size of sparsefile to 32GB

josephglanville:
Running into out of space errors during builds, this should give it some more breathing room.


titanous:
LGTM

philiplb:
Does this mean, that this sparse file is always taking 32GB of hard drive?

josephglanville:
@philiplb this is only use on our CI server.
Also it's "sparse" so it only uses blocks which are written to, until they are written to they use no space on the underlying disk.

philiplb:
Got it, thanks. :)

PR #2368, opened 2016-01-13
test/runner: Bump max test runtime to 35 minutes

titanous:
We are regularly hitting ~26m test runs :(


josephglanville:
Did we work out why it was taking so long or did it go away with the reboot or whatever was done?

titanous:
I think this is happening when multiple parallel builds exhaust the tmpfs. The plan is to move to a new server with more memory in February.

PR #2374, opened 2016-01-18
controller: Avoid double close panic during shutdown

titanous:
During shutdown, `c.eventListener.CloseWithError(ErrShutdown)` is called this causes the pg listener channel to be closed which will race to be handled and could result in another call to `CloseWithError` which causes another close of `e.doneCh`.

To avoid this we check if the listener is closed in `CloseWithError` and skip the close logic if it has already been closed.


josephglanville:
LGTM

PR #2375, opened 2016-01-18
Cluster monitor/resurrection

josephglanville:
This change introduces a few differences to how resurrection works.
Mainly there is a new `flynn-host` subsystem I have dubbed the `cluster-monitor` that babysits the cluster and attempts to repair it if things go wrong.

The way resurrection works in `flynn-host` has fairly drastically changed also. There is no longer a resurrection flag for jobs and instead `flynn-host` has persistent slots for jobs that will be revived on restart. Only the latest write to these slots will stick so only the latest ran version of `discoverd`/`flannel` etc will be resurrected. Jobs like `postgres` and the `controller` jobs are no longer resurrected by `flynn-host` normally and are instead left to the `cluster-monitor`.

The `cluster-monitor` itself is currently a modified version of the cluster fixer. It's been through a few iterations but after much back and forth I still think it's a better idea to try extract functionality from the cluster fixer and have them rely on the same methods to repair damage to the cluster.
It's operation is quite simple, it starts off inert, waiting for a switch in it's service metadata to be flicked before it becomes active. This is done by the bootstrap process, which also writes in the number of hosts that were present during bootstrap.
The monitor once active then waits until either a fault is detected that can't be resolved by the controller (a deadline is reached and it decides the controller has taken too long to fix the problem) and that the same number of hosts that were present during bootstrap are online. It then effectively runs the same routine as the cluster fixer to attempt to fix the problem.


josephglanville:
Running into a few issues with the scheduler handling active jobs on hosts when they join the cluster.
Trying to track down the exact issue now but apart from that everything seems to work as expected.

josephglanville:
Turns out wasn't the scheduler. It's same strangeness in order of operations during resurrection.
So it's working "most" the time, the worst kind of working. :grin:

josephglanville:
When a host rejoins a healthy cluster (still has operations schedulers etc) then I am seeing something weird where my host resurrects it's flannel and discoverd jobs, it connects to the cluster via it's local discoverd and then the scheduler follows it and starts doing stuff.
For some reason the scheduler seems to think that it needs to run new discoverd/flannel jobs despite them already running on the host.

josephglanville:
So I have been unsucessful in working out why the scheduler can't seem to see the jobs I have running when it starts up.

Here is the log from the scheduler of a single node cluster using the new resurrection mechanism: https://gist.github.com/josephglanville/53918b3ab56022618383

It does initially work as the scheduler doesn't manage to terminate the existing discoverd process, but it does eventually succeed, breaking the cluster.

This is the only real issue remaining (aside from some code TODOs which I will get to). It does bring everything back to life it's just that the scheduler then proceeds to try it's hardest to break it.

titanous:
Have you investigated if the scheduler is getting access to the host API before resurrection has completed? It's possible that both the scheduler and the resurrection process start discoverd processes simultaneously because they both see no jobs running, then the count jumps to two and the scheduler kills the one it does know about (started by resurrection).

josephglanville:
I finally figured out the problem and now feel like a bit of an idiot. The way I changed the resurrection routine to start the new jobs was not correctly duping the jobs so it ended up modifying the ID of the old job in memory, which the host was then returning via HTTP API to the scheduler. Depending on the ordering of the jobs the scheduler would either see the old jobs first which is in the `stopped` state, then the new one in `started` and everything would be ok, or it would see them in the reverse order causing the scheduler to process a `started` -> `stopped` state transition for the job despite it still being running.
The fact it worked 50% of the time really threw me off heh.

josephglanville:
@titanous this is ready for a review pass.

I have tested it fairly extensively at this point and it works as far as I can tell 100% of the time if the required hosts return to the cluster.

There are some concerns that should probably be noted somewhere though. Namely, this system is only 100% tested and safe with clusters that haven't been grown beyond their original count at bootstrap time.
Secondly if the `postgres` formation has been edited to use more than 2 asyncs there is a potential for the postgres repair code to get confused.

Future thinking there is also a few things. First is the repair code should be enhanced to revive the sync if possible rather than replacing it, and if it does have to replace it it should clone the data volume of the master. Next is the matter of making persistent volumes safe.. for that to happen we have to first enhance the scheduler to be aware of the fact that certain volumes are of importance. We aren't able to code that into this layer, has to be done in the scheduler. The monitor will only contain special case code for the system postgres cluster as it's required for resurrection. So this work effectively becomes a dependency for persistent Redis clusters and the MySQL appliance.

josephglanville:
Also need to work out what it takes to deprecate and eventually remove the resurrection flag.
I didn't attempt to do that here I just stopped making use of it but that should also be done at some stage.

host/discoverd_wrapper.go
```diff
@@ -0,0 +1,89 @@
...
+	select {
+	case isLeader := <-d.leader:
+		return isLeader, nil
+	case <-time.After(30 * time.Second):
+		return false, errors.New("timed out waiting for current service leader")
+	}
+}
+
+func (d *discoverdWrapper) LeaderCh() chan bool {
+	return d.leader
```
titanous: This comment should note that only one receiver can consume the channel at a time.

host/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+func (m *Monitor) waitEnabled() {
+	for {
+		monitorMeta, err := m.monitorSvc.GetMeta()
+		if err != nil {
+			time.Sleep(5 * time.Second)
+			continue
+		}
+		var decodedMeta MonitorMetadata
+		if err := json.Unmarshal(monitorMeta.Data, &decodedMeta); err != nil {
+			continue
```
titanous: This will result in a tight loop fetching the metadata if the JSON is unparseable, log and add a delay?

host/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+	log := monitorLogger.New("fn", "checkCluster")
+	var faulted bool
+	hosts, err := m.c.Hosts()
+	if err != nil || len(hosts) < m.hostCount {
+		log.Info("waiting for hosts", "current", len(hosts), "want", m.hostCount)
+		return
+	}
+
+	controllerService := discoverd.NewService("controller")
+	controllerInstances, _ := controllerService.Instances()
```
titanous:
```go
discoverd.NewService("controller").Instances()
```

host/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+	if len(controllerInstances) == 0 {
+		log.Error("did not find any controller api instances")
+		faulted = true
```
titanous: Do we want this to trigger if discoverd is failing? Currently it does.

host/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	if _, err := discoverd.NewService("controller-scheduler").Leader(); err != nil && !discoverd.IsNotFound(err) {
+		log.Error("error getting scheduler leader, can't deterimine health")
```
titanous: s/deterimine/determine

titanous:
Awesome work! Just a few minor comments.

host/monitor.go
josephglanville: If discoverd is down/otherwise incapacitated won't the query to `Hosts()` result in an err and thus short circuiting the rest of the checkCluster function anyway?

titanous: Ah, yep.

titanous:
LGTM!

josephglanville:
Yay!

wolfskaempf:
I know this pull request has since been merged and this thread has become inactive, but are there any updates regarding the process of recovering from node-failure? I'm specifically referencing the lack of rebalancing of the database. The cluster seems to become functional again eventually — it just won't spread the database instances back out and I can't seem to find an elegant way of moving them manually (that doesn't involve killing individual containers). This issue is currently the main reason I'm holding off using Flynn in a production environment. Thank you!

PR #2384, opened 2016-01-21
test: Fix race in redis test

titanous:
The redis appliance is not always ready when the `resource add` command returns. Added a check that waits for the discoverd registration before continuing.

/cc @benbjohnson


benbjohnson:
LGTM

PR #2389, opened 2016-01-25
script: Add keep state option to start-flynn-host

josephglanville:
Helpful when writing scripts to test things like resurrection or host reboots.


lmars:
LGTM

Issue #2376, opened 2016-01-18
App deletion fails when postgres resources are still in use


josephglanville:
Postgres likes to refuse to delete databases that still have open connections.
There is 2 possible fixes for this issue, we either scale down all the process types for the application we are trying to delete first before deleting the resource (which should solve this class of problem).
Or we can solve it specifically for postgres by using the superuser to deny connections to the database and also sever existing connections.

titanous:
Killing all of the app processes first seems like a good idea (we should also nuke one-off processes).

NeilW:
Is this issue likely to be the cause of the app delete failure I came across during testing today?

```
neil@ubuntu-vm:~/Dev/go-flynn-example$ flynn delete
Are you sure you want to delete the app "goexample"? (yes/no): yes
unknown_error: Something went wrong
```

josephglanville:
@NeilW if it was using a postgres resource and the app was still scaled up. almost certainly yes.

To work around in the meantime you can simple scale your app down to zero before trying to remove the resource or delete the app. Once either of the 2 changes above land deletion will just work.

PR #2390, opened 2016-01-25
cli: Scale processes to 0 before deleting app

josephglanville:
By scaling down the processes before proceeding with deleting resources etc we can ensure they are no longer being used by the application.

This won't fix the case of trying to remove an in-use resource without deleting the app while the application still has processes running. This will need to be addressed on a per-provider/resource basis.

Refs #2376


titanous:
LGTM

Issue #1425, opened 2015-04-07
Router sticky sync failure


titanous:
The streaming router state sync does not appear to get the sticky flag when creating a new route.

**Reproduction**

1. Deploy an app.
1. Create a new route using this command: `flynn route add http --sticky sticky-test.dev.localflynn.com` (using the CLI from master, due to #1420)
1. Curl the route and note that there is no `Set-Cookie` header with a `_backend` cookie as expected.
1. Restart the router by killing it, curl again, note that the header is set.

benburkert:
I believe this is due to the router internally setting "sticky" on the proxy transport. There is one transport per backend service, so if a non-sticky route is added before a sticky one to the same service, both routes will be non-sticky: https://github.com/flynn/flynn/blob/master/router/http.go#L225

Unfortunately, this is not an easy fix b/c this area of the router very convoluted and complex. A while ago we started refactoring this stuff [here](https://github.com/benburkert/flynn/tree/router-route-table), but put it aside. I think we should fix this after landing that refactor.

josephglanville:
I just bumped into this when trying to add update support for leader routing, which has the same problem.  You currently can't have 2 routes to the same service with different proxy/transport level settings.

PR #2395, opened 2016-01-27
router: Leader routing

### LINKED ISSUE #2126: Please add a "--leader" switch to the "flynn route" command
Hi there!

I really like the way Flynn comes with data replication "from the moment you turn the key".

However, I'd like to be able to connect to the database remotely, so that I can set up an off-site postgresql instance as a streaming replica.

I asked in the #flynn IRC channel about my needs, and whilst I *can* add a route with "flynn -a postgres route add tcp -s postgres-postgres -p 5432", it looks like it **isn't** possible to configure the route to always send traffic to the leader.

Would it be possible for you to add a "--leader" switch to the "flynn route" command, to route traffic to the service leader only?

Thanks in advance for considering my feature request.

Kind Regards,

Hazel Smith

josephglanville:
This implements a `--leader` switch to allow routing only to the leader. If no leader is elected "no backends available" will be logged and a service unavailable message returned.

The main usecase of this feature is to expose data appliances like PostgreSQL to the outside world.
For instance to use this changeset to add a route for postgres one would use the following:
```
flynn -a postgres route add tcp --service postgres --leader
```

TODO
- [x] Update support
- [x] Unit tests

Closes #2126

### DIFF
--- a/cli/route.go
+++ b/cli/route.go
@@ -19,9 +19,9 @@ import (
 func init() {
 	register("route", runRoute, `
 usage: flynn route
-       flynn route add http [-s <service>] [-c <tls-cert> -k <tls-key>] [--sticky] <domain>
-       flynn route add tcp [-s <service>] [-p <port>]
-       flynn route update <id> [-s <service>] [-c <tls-cert> -k <tls-key>] [--sticky] [--no-sticky]
+       flynn route add http [-s <service>] [-c <tls-cert> -k <tls-key>] [--sticky] [--leader] [--no-leader] <domain>
+       flynn route add tcp [-s <service>] [-p <port>] [--leader]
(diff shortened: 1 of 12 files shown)


josephglanville:
Further investigation revealed it's not possible to have 2+ routes with the same service but different backend transport settings (sticky/leader, etc). See #1425

temujin9:
:+1: - this would help with seamless cluster migrations, at the data layer.

binarycrafts:
I think this makes sense to have for all appliances!

josephglanville:
@titanous @jvatic Unit tests implemented, code is now ready for review.

titanous:
LGTM pending comment

josephglanville:
@titanous did you forget to publish a comment here?

josephglanville:
@titanous I have added some additional unit and integration tests to verify defaults.

titanous:
LGTM

PR #2396, opened 2016-01-27
appliance/postgres: Forcefully close connections when deleting database

### LINKED ISSUE #2376: App deletion fails when postgres resources are still in use

josephglanville:
By first disallowing new connections and then terminating existing connections we can delete databases were `DROP` would otherwise fail.

Closes #2376

### DIFF
--- a/appliance/postgresql/api/server.go
+++ b/appliance/postgresql/api/server.go
@@ -15,6 +15,15 @@ import (
 	"github.com/flynn/flynn/pkg/shutdown"
 )
 
+const (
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

PR #2403, opened 2016-02-01
appliance/postgres: Extract state machine to pkg/sirenia

josephglanville:
This changeset extracts the implementation of the Manatee state machine from the `appliance/postgres` package along with it's simulator/test-harness.

I renamed it Sirenia as it's the greater family of sea cows that encompass both Manatees and Dugongs. :grinning: 

Of note is that it attempts a generalisation of the `xlog` interface. To implement this interface a database must have a history with a total ordering that can be serialised to a string format. (`xlog.Position`)

Apart from this requirement databases wishing to use the Sirenia state machine must also implement synchronous replication and be capable of blocking writes when it's downstream synchronous peer is unable to ack them.

### DIFF
--- a/appliance/postgresql/main.go
+++ b/appliance/postgresql/main.go
@@ -6,11 +6,16 @@ import (
 	"path/filepath"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
-	"github.com/flynn/flynn/appliance/postgresql/state"
 	"github.com/flynn/flynn/discoverd/client"
 	"github.com/flynn/flynn/pkg/httphelper"
 	"github.com/flynn/flynn/pkg/random"
 	"github.com/flynn/flynn/pkg/shutdown"
+	sd "github.com/flynn/flynn/pkg/sirenia/discoverd"
+	"github.com/flynn/flynn/pkg/sirenia/state"
+)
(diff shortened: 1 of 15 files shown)


appliance/postgresql/main.go
```diff
@@ -43,7 +44,7 @@ func main() {
 	}
 	inst := &discoverd.Instance{
 		Addr: ":5432",
-		Meta: map[string]string{"POSTGRES_ID": id},
+		Meta: map[string]string{"DB_ID": id},
```
titanous: Are there upgrade considerations for this change?

josephglanville: Yes I think so, though doing it live may prove to be somewhat difficult.
If we want the `postgres` deployment strategy to be made generic these variables would either need to be standardised or the prefix stored in controller/job or discoverd metadata.
Will think it over.

josephglanville:
Made update friendly, no longer modifies the metadata key used to identify postgres state machine members and instead makes it configurable so each user of the state machine can pick their own.

josephglanville:
@titanous this has been updated to be backwards compatible and dropped `Increment()` from the `XLog` interface as it's only used in the simulator.

josephglanville:
Going to close this and merge this work as part of the MariaDB integration.

PR #2406, opened 2016-02-02
appliance/mariadb: Add MariaDB appliance (WIP)

benbjohnson:
## Overview

This pull request adds support for a clustered MariaDB appliance with semisynchronous replication. This is currently a work-in-progress.


benbjohnson:
@josephglanville I pushed up what I have so far. I got the MariaDB processes to run on the same machine which I needed for the cluster testing. I also did some refactoring to try to match your pull request (#2403) although it still needs some work. I'll also rename the `appliance/postgresql/state` imports to `pkg/sirenia/state` once you merge yours in.

josephglanville:
@benbjohnson looks good so far. I will probably create an integration branch with all of the changes on it.

The big TODO seems to be transferring and applying the base backup. The issue here is we need a hot physical backup. The only reasonable way to get one is to use xtrabackup, which has a somewhat complicated process of restoring said backup.

The long story cut rather short is that xtrabackup copies data files directly so it also needs to watch the transaction log and apply it to the backup as well. This results in a 2 stage process in that after the backup has been created the transaction log needs to be replayed against the data directory before the database system can be started.

We will need to run xtrabackup on the master node so it will need to serve up an endpoint that can stream the backup. After we stream the backup over HTTP we can stage it a temp directory and run the "apply logs" stage, move the files into position then start the database system back up.

josephglanville:
I've done some integration work and pushed my changes to https://github.com/flynn/flynn/tree/mariadb-integration

josephglanville:
@benbjohnson I'm going to go ahead and write the xtrabackup integration. I will leave the HTTP server for now but I think we should just lift what we have in the postgres appliance and add an additional endpoint to stream the backup.

josephglanville:
First pass of the xtrabackup stuff is up. Stuff to consider. The posix user running the backups must have ability to read the data dir and all files in it, also we may want to create a special backup user as we will need to provide the password on the command line which means it will be in the process list in plain text.
It will need the RELOAD, LOCK TABLES, REPLICATION CLIENT grants.

josephglanville:
Going to close this and open a PR using the mariadb-integration branch.

PR #2414, opened 2016-02-03
test: Bump the expiry of the root manifest in the test TUF repo

lmars:
This was done by generating and then revoking a targets key:

```
export TUF_ROOT_PASSPHRASE="flynn-test"
export TUF_TARGETS_PASSPHRASE="flynn-test"
export TUF_SNAPSHOT_PASSPHRASE="flynn-test"
export TUF_TIMESTAMP_PASSPHRASE="flynn-test"

tuf gen-key --expires 10950 targets
tuf revoke-key --expires 10950 targets $ID

tuf add
tuf snapshot
tuf timestamp
tuf commit
```

The root expiry is now `2046-01-26T00:04:38Z`, so this shouldn't be an issue for a while :smile:


titanous:
LGTM

PR #2416, opened 2016-02-04
host: Start flynn-host at boot by default

titanous:
Disabled in EC2 and Vagrant images so that bootstrap scripts can do their thing.

### DIFF
--- a/host/upstart.conf
+++ b/host/upstart.conf
@@ -1,8 +1,8 @@
 description "Flynn layer 0"
 
-#start on (started libvirt-bin and started networking)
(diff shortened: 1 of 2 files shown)


josephglanville:
LGTM.

kwando:
This means a Flynn node can be rebooted now?

titanous:
Yes, I'm doing testing now but we believe we have fixed the reboot issues. A release will follow shortly if everything is working.

kwando:
Finally! :heart:

PR #2418, opened 2016-02-04
controller/scheduler: Wait for controller API to start

titanous:
Explicitly wait for the controller API to be registered in discoverd.

If we do not do this and it takes longer than expected to start, the scheduler may crash later when doing the initial job sync. This was observed during a restore from a backup by @temujin9.


lmars:
LGTM

PR #2419, opened 2016-02-06
installer: Ensure startScript only executes on first boot

josephglanville:
Only run the startScript on first boot.


titanous:
LGTM

PR #2420, opened 2016-02-06
Fix installer userdata

titanous:
Fixes several bugs in the userdata script, currently preventing EC2 installs with the latest release.

/cc @josephglanville

### DIFF
--- a/installer/cluster.go
+++ b/installer/cluster.go
@@ -579,6 +579,8 @@ iptables -A FORWARD -i eth0 -j DROP
 var startScript = template.Must(template.New("start.sh").Parse(`
 #!/bin/sh
 FIRST_BOOT="/var/lib/flynn/first-boot"
+mkdir -p /var/lib/flynn
+
 if [ ! -f "${FIRST_BOOT}" ]; then
   {{if .DataDisk}}
   zpool create -f flynn-default {{.DataDisk}}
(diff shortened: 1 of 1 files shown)


installer/cluster.go
```diff
@@ -578,8 +578,12 @@ iptables -A FORWARD -i eth0 -j DROP
 
 var startScript = template.Must(template.New("start.sh").Parse(`
 #!/bin/sh
+set -e -x
+
 FIRST_BOOT="/var/lib/flynn/first-boot"
-if [[ ! -f "${FIRST_BOOT}" ]]; then
+mkdir -p /var/lib/flynn
+
+if [ ! -f "${FIRST_BOOT}" ]; then
```
lmars: Any reason we aren't using `/bin/bash`? That is the change I would have made rather than this, but I don't think it *really* matters.

titanous: I don't think there is reason, no. That change could be made in a future patch, I don't think it really impacts anything.

installer/cluster.go
```diff
@@ -578,8 +578,12 @@ iptables -A FORWARD -i eth0 -j DROP
 
 var startScript = template.Must(template.New("start.sh").Parse(`
 #!/bin/sh
+set -e -x
+
 FIRST_BOOT="/var/lib/flynn/first-boot"
-if [[ ! -f "${FIRST_BOOT}" ]]; then
+mkdir -p /var/lib/flynn
```
lmars: We should probably do this in the install script too, no issue with it being here though.

lmars:
LGTM

PR #2421, opened 2016-02-07
Test fixes

### LINKED ISSUE #2182: scheduler: Host checks don't detect graceful shutdown
If a flynn-host daemon is gracefully shutting down, it will [stop heartbeating before stopping it's jobs](https://github.com/flynn/flynn/blob/master/host/host.go#L368-L375), and also before [closing the HTTP listener](https://github.com/flynn/flynn/blob/master/host/host.go#L299), which means when the scheduler receives the down event, it will successfully get the status from the host (because it is still listening for HTTP requests), so marks it as healthy again. Then, whilst all the jobs are stopping, the scheduler tries to restart them all because it think the host is healthy.

I think having the host should stop responding to status requests when it is shutting down, specifically before it stops heartbeating.

lmars:
These changes (hopefully) fix a bunch of errors I saw from looking at the past 20 or so CI failures.

Fixes #2182.

### DIFF
--- a/controller/scheduler/host.go
+++ b/controller/scheduler/host.go
@@ -20,6 +20,7 @@ type Host struct {
 	stop     chan struct{}
 	stopOnce sync.Once
 	done     chan struct{}
+	shutdown bool
 }
 
 func NewHost(h utils.HostClient) *Host {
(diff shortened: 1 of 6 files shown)


josephglanville:
LGTM

titanous:
LGTM

PR #2425, opened 2016-02-07
host: Fix displaying job ExitStatus in `flynn-host inspect`

lmars:
The field was changed to a pointer in commit 23babab.


titanous:
LGTM

PR #2426, opened 2016-02-07
host: Prevent multiple attachers to interactive jobs

### LINKED ISSUE #1697: host: Multiple job attach should be disabled when DisableLog is true
If `DisableLog` is set for a job, then clients get [attached to the job file descriptors](https://github.com/flynn/flynn/blob/master/host/libvirt_lxc_backend.go#L965-L989) (rather than the non-existent on disk logs), meaning if multiple clients attach to a job, the output will be distributed amongst the clients. This leads to unexpected behaviour (e.g. `flynn-host run` will miss some output if `flynn-host log` is run for the same job).

We should return an error if a client tries to attach to a job with `DisableLog` set which is already attached to another client.

lmars:
This means running `flynn-host log` for a job which has `DisableLog` set and is already attached now just returns as if there is no log.

I was lead to this by the fact that CI logs are currently around 100MB, with about 70MB of that being the output of the [controller backup test](https://github.com/flynn/flynn/blob/master/test/test_controller.go#L534-L537) because the postgres dump job is still running when the logs are dumped.

Fixes #1697.

### DIFF
--- a/host/attach.go
+++ b/host/attach.go
@@ -17,6 +17,20 @@ import (
 type attachHandler struct {
 	state   *State
 	backend Backend
+
+	// attached is a map of job IDs which are currently attached and is
+	// used to prevent multiple clients attaching to interactive jobs (i.e.
+	// ones which have DisableLog set)
+	attached    map[string]struct{}
+	attachedMtx sync.Mutex
+}
+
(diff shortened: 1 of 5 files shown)


titanous:
LGTM

PR #2427, opened 2016-02-08
worker: Stop DomainMigration jobs when worker is shutting down

lmars:
A `DomainMigration` job deploys the controller and so may result in the worker which is performing the migration killing itself, in which case the domain migration needs to stop waiting straight away so a new worker can continue with both the controller deploy and the domain migration.

I've also increased the stream reconnect timeout (see explanation in b9d3e51).


titanous:
LGTM

josephglanville:
LGTM

PR #2428, opened 2016-02-08
host: Mount volumes when restoring from statedb

josephglanville:
This changeset implements volume mounting when restoring the volume state db.

Previously it was assumed that some external process will ensure volumes would be mounted before they would be used by `flynn-host`. This behaviour turns out to be somewhat fragile and also non-portable as only Ubuntu ships with the modified version mountall designed to automount ZFS volumes.

### DIFF
--- a/host/volume/zfs/zfs.go
+++ b/host/volume/zfs/zfs.go
@@ -208,7 +208,7 @@ func (b *Provider) CreateSnapshot(vol volume.Volume) (volume.Volume, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := b.mountSnapshot(snap); err != nil {
+	if err := b.mountDataset(snap); err != nil {
 		return nil, err
 	}
 	b.volumes[id] = snap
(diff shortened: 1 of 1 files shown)


host/volume/zfs/zfs.go
```diff
@@ -452,11 +457,8 @@ func (b *Provider) RestoreVolumeState(volInfo *volume.Info, data json.RawMessage
 		dataset:   dataset,
 		basemount: record.Basemount,
 	}
-	// zfs should have already remounted filesystems; special remount case for snapshots
-	if v.IsSnapshot() {
-		if err := b.mountSnapshot(v); err != nil {
-			return nil, err
-		}
+	if err := b.mountDataset(v); err != nil {
```
titanous: What happens if the dataset is already mounted? (ie we didn't reboot)

josephglanville: It should be idempotent, atleast I am sure of the `zfs mount` command, I will test `mount -t zfs` to test that too but that code-path was already running. I think that probably wouldn't have mattered right now though as I don't think we are using the snapshotting features?

josephglanville: Just realised it doesn't matter if it's idempotent or not, `mountDataset()` does a check using `stat()` to check the device/fsid of the parent directory to determine if they 2 are different filesystems to determine if it's already mounted.

titanous:
LGTM

PR #2431, opened 2016-02-08
Fix bootstrap route race

titanous:
There is a race that can occur when the router discoverd registration has not propagated to the controller, resulting in a 500 from the controller. Retry a few times before giving up.


lmars:
LGTM

PR #2432, opened 2016-02-09
router: Use proxy instance per route rather than per service

josephglanville:
This fixes issues caused by sharing transport level options across routes.

Closes: #1425
Refs: #2395


titanous:
LGTM

PR #2433, opened 2016-02-09
Mariadb integration

josephglanville:
Integration branch for MariaDB appliance.


josephglanville:
@benbjohnson I am seeing `go test` hang with a defunct innobackupex process.
Have you seen this before?

josephglanville:
Source of the hang is `TestIntegration_BackupRestore`.
Looks like we might need better process handling, how about returning the `*exec.Cmd` or something from `Backup()` so that the caller can kill off the `innobackupex` process if needed?

benbjohnson:
@josephglanville Sorry, I had a temporary unit test committed that performed the commit and then waited indefinitely so I could manually verify that it worked. It was a temporary hack to test backups while I was getting the HTTP handler working:

https://github.com/flynn/flynn/blob/mariadb-integration/appliance/mariadb/process_test.go#L216

I'm guessing that's the cause of the hang. I removed the test in 5b75642 since it's being tested in the TwoNode test.

josephglanville:
All good, just wasn't sure if it was intention etc. :)

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1106 @@
...
+		}
+	} else {
+		if p.WaitUpstream {
+			if err := p.waitForUpstream(upstream); err != nil {
+				return err
+			}
+		}
+
+		if err := func() error {
+			// TODO(benbjohnson): Lock table during backup: "FLUSH TABLES WITH READ LOCK"
```
josephglanville: This isn't necessary. The master doesn't need to be locked when using `xtrabackup`, it will do a first pass copying physical files and keeping track of what happens in the log, then lock the tables and copy anything that changed since the backup started and also write the log and information about it into the backup stream.

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1106 @@
...
+				return err
+			}
+
+			logger.Info("restoring backup")
+			if err := p.Restore(resp.Body); err != nil {
+				logger.Error("error restoring backup", "err", err)
+				return err
+			}
+
+			// TODO(benbjohnson): Unlock tables after backup and log pos are retrieved.
```
josephglanville: Because the lock is unecessary the unlock is too. However you raise a good point about retrieving the log position of the backup, we can actually do that without starting the server after we unpack the backup with `xbstream`. `xtrabackup` creates a file called `xtrabackup_binlog_info`, this will contain the binlog position.

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1106 @@
...
+	dsn.Password = ""
+
+	db, err := sql.Open("mysql", dsn.String())
+	if err != nil {
+		logger.Error("error acquiring connection", "err", err)
+		return err
+	}
+	defer db.Close()
+
+	if _, err := db.Exec(fmt.Sprintf(`CREATE USER 'flynn'@'%%' IDENTIFIED BY '%s'`, p.Password)); err != nil {
```
josephglanville: Do we need to handle the user existing here or is `CREATE USER` idempotent in MySQL land?

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1106 @@
...
+	}
+	if _, err := db.Exec(`GRANT ALL PRIVILEGES ON *.* TO 'flynn'@'%'`); err != nil {
+		logger.Error("error granting privileges", "err", err)
+		return err
+	}
+	if _, err := db.Exec(`GRANT REPLICATION SLAVE ON *.* TO 'flynn'@'%'`); err != nil {
+		logger.Error("error granting replication privileges", "err", err)
+		return err
+	}
+	if _, err := db.Exec(`INSTALL PLUGIN rpl_semi_sync_master SONAME 'semisync_master.so'`); err != nil {
```
josephglanville: Ditto for plugin installation.

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1106 @@
...
+		logger.Error("error installing rpl_semi_sync_master", "err", err)
+		return err
+	}
+	if _, err := db.Exec(`FLUSH PRIVILEGES`); err != nil {
+		logger.Error("error flushing privileges", "err", err)
+		return err
+	}
+
+	// Enable semi-sync replication on the master.
+	if _, err := db.Exec(`SET GLOBAL rpl_semi_sync_master_enabled = ON`); err != nil {
```
josephglanville: We are also going to need to set a few other settings, may actually be better to set them in the configuration file though.

Specifically `rpl_semi_sync_master_wait_no_slave` is required to ensure the timeout will be observed even when the sync isn't connected. Otherwise the primary will automatically switch to async replication and can commit writes that are not replicated anywhere. The timeout itself too needs to be raised to a sufficiently high value that it will never be reached too, which is `rpl_semi_sync_master_timeout`.
We should also ensure that `rpl_semi_sync_master_wait_point` is set to `AFTER_SYNC` which is the safest/most-consistent.

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1106 @@
...
+			select {
+			case <-stopCh:
+				logger.Debug("canceled, stopping")
+				return
+			case <-time.After(checkInterval):
+			}
+		}
+
+		if enableWrites {
+			// TODO: enable writes, restart?
```
josephglanville: With MySQL we should be able to do this without a restart.
Because we don't give out `SUPER` willy nilly it should be safe enough just to set `read_only` to `1` and things should be great. To enable writes again we can just set it back to `0`.

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1106 @@
...
+		return
+	}
+
+	logger.Info("master status", "file", status.File, "position", status.Position)
+
+	return
+}
+
+// XLogPosition returns the current transaction log position.
+func (p *Process) XLogPosition() (xlog.Position, error) { panic("FIXME") }
```
josephglanville: I did a bit more reading. I think we should look into the MySQL GTID (Global Transaction ID). It seems to be a monotonically increasing counter that is persisted through the entire replication hierarchy. Seems like a much better solution than the file/offset setup which seems like it isn't comparable between layers of replication.

appliance/mariadb/process.go
benbjohnson: I was going off this section of the replication docs:

https://mariadb.com/kb/en/mariadb/setting-up-replication/#getting-the-masters-binary-log-co-ordinates

If `xtrabackup_binlog_info` provides the file/pos then I can ignore the lock. :+1:

appliance/mariadb/process.go
benbjohnson: I still need to handle the error. I was working on getting the happy path working first.

appliance/mariadb/process.go
```diff
@@ -0,0 +1,870 @@
...
+		db, err := p.connectLocal()
+		if err != nil {
+			logger.Error("error acquiring connection", "err", err)
+			return err
+		}
+		defer db.Close()
+
+		host, port, _ := net.SplitHostPort(upstream.Addr)
+		logger.Info("changing master", "host", host, "port", port)
+		if _, err := db.Exec(fmt.Sprintf(`SET GLOBAL gtid_slave_pos = "%s";`, slaveXLog)); err != nil {
```
josephglanville: I need to investigate if this is necessary. IMO this is something that innobackupex should be doing for us.

appliance/mariadb/process.go
```diff
@@ -0,0 +1,870 @@
...
+
+			elapsedTime := time.Since(startTime)
+			logger := logger.New(
+				"master_log_pos", masterXLog,
+				"slave_log_pos", slaveXLog,
+				"elapsed", elapsedTime,
+			)
+
+			// Update the slave status and reset start time if replication has progressed.
+			// FIXME(jpg): What is the intention here? To track slave xlog over each
```
josephglanville: I'm guessing the point of this is to loop and if the slave xlog position advances we should boot the start time back to give it more time to catch up?

titanous: Yeah, I think this is copied from the Postgres code which is in turn copied from the upstream Manatee code.

appliance/mariadb/process_test.go
```diff
@@ -0,0 +1,484 @@
...
+		Meta: map[string]string{
+			"MYSQL_ID":  id,
+			"HTTP_ADDR": fmt.Sprintf("127.0.0.1:%d", 54420+n),
+		},
+	}
+}
+
+func connect(c *C, n int, database string) *sql.DB {
+	dsn := DSN{
+		Host:     fmt.Sprintf("127.0.0.1:%d", 54320+uint16(n)),
```
titanous: I recently refactored the corresponding postgres tests to not reuse port numbers, as it is unreliable: https://github.com/flynn/flynn/commit/3ba9110c30b771a72a86f32bbea26c00c1e8c21c

benbjohnson: Thanks, I'll put that change in.

appliance/mariadb/handler.go
```diff
@@ -0,0 +1,105 @@
...
+
+// handleGetStatus handles request to GET /status.
+func (h *Handler) handleGetStatus(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	var status Status
+	status.Peer = h.Peer.Info()
+
+	info, err := h.Process.Info()
+	if err != nil {
+		// Log the error, but don't return a 500. We will always have some
+		// information to return, but postgres may not be online.
```
imjorge: s/postgres/mariadb/

appliance/mariadb/cmd/flynn-mariadb/main.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	err = discoverd.DefaultClient.AddService(serviceName, &discoverd.ServiceConfig{
+		LeaderType: discoverd.LeaderTypeManual,
+	})
+	if err != nil && !httphelper.IsObjectExistsError(err) {
+		shutdown.Fatal(err)
+	}
+	inst := &discoverd.Instance{
+		Addr: ":3306",
```
imjorge: use `DefaultPort` constant?

appliance/mariadb/process.go
josephglanville: Turns out this is necessary, else the slave starts trying to replicate from the start of the log.. which isn't so great for obvious reasons.

appliance/mariadb/process_test.go
benbjohnson: Fixed in cd4c3de

appliance/mariadb/handler.go
benbjohnson: Fixed in cd4c3de

appliance/mariadb/cmd/flynn-mariadb/main.go
benbjohnson: Fixed in cd4c3de

appliance/mariadb/process_test.go
```diff
@@ -0,0 +1,492 @@
...
+	waitReplSync(c, node1, 2)
+
+	time.Sleep(1 * time.Second)
+
+	// Read from the slave.
+	var value string
+	db2 := connect(c, node2, "foo")
+	defer db2.Close()
+	c.Assert(db2.QueryRow("SELECT value FROM bar").Scan(&value), IsNil)
+	c.Assert(value, Equals, "baz")
```
benbjohnson: @josephglanville This section here was previously querying `node1` instead of `node2` so it was always succeeding. I fixed that in cd4c3de but now I'm getting an error when I query it immediately (but it works if I add a `100ms` sleep) which makes me think it's not being synchronously replicated. I tried digging into it further but I'm not seeing why it's not running semi-sync.

benbjohnson:
@josephglanville I pushed up some minor fixes and clean up in cd4c3de. I also went through and verified that all the `Exec()` calls are idempotent (like `CREATE USER`) and fixed that where that wasn't the case.

appliance/mariadb/process_test.go
josephglanville: Ok, I will look into this.
We shouldn't be seeing commited data on `node1` before `node2`.

josephglanville:
@benbjohnson fixed the semi-sync issues. As a result I also had to skip setting up replication when running in singleton or the single node would wait forever for a slave to connect. On the plus side I did verify it works the way we intend. :smile:

josephglanville:
@benbjohnson changed the two node test to not attempt to do a consistent read as we don't do this for postgres and it's not a property we get with semi-sync replication (or pg synchronous replication for that matter).

Remaining TODOs are:
- [ ] zombie innobackupex children
- [ ] unit tests fail when all run at once but work when run individually (they fail on streaming the backup which is quite confusing..)
- [ ] provider databases API
- [ ] main binary plumbing
- [ ] deploy and add to controller during bootstrap

benbjohnson:
@josephglanville I'm looking into the unit tests failing when run all at once. It looks like something is preventing the processes from starting up in a timely manner and the backups and returning a `0` byte backup. I added a `waitForUpstream` and working status endpoint but I'm still tracking down exactly what is the cause for the stall. Just wanted to give you a heads up.

josephglanville:
@benbjohnson makes sense. We might need to make the backup http handler a little better. I was thinking of exposing the *exec.Cmd or wrapping it with something so that we can both call `Wait()` on the process itself but also get non-0 exit statuses and stderr and log them.

benbjohnson:
@josephglanville I added stderr logging for the `Cmd`. The instructions for `innobackupex` say that you need to check for `"completed OK!"` at the end of stderr so we'll have to make some kind of reader that wraps the stdout & stderr.

On the restore side we'll need to add some kind of confirmation at the end. Either magic numbers or a checksum. Otherwise we don't know how large the backup should be.

benbjohnson:
@josephglanville I finally tracked down the last concurrent testing bug in `appliance/mariadb`! Well, at least the last one I know about. It seems to work now. Fixed in 89249e4883ddfecc45de063cc1829ce0482f1a5f.

josephglanville:
@benbjohnson sorry about the unit tests the other night, was meant to push commit to fix it but forgot.

I am moving onto integrating up the stack using generalised primitives from the state machine. Namely I extracted and made the postgres client generic so it can be used by the deployment engine to automatically handle deployments for Sirenia compatible database appliances.

josephglanville:
Fixed the innobackupex zombies, calling `cmd.Wait()` does the trick.

josephglanville:
Pushed the provider API and a bunch of fixes to stuff.

Now that we are out of TODOs.. time for more..
- [ ] Integration tests - if possible try generalize the pg tests so they can be run on any Sirenia compatible DB
- [ ] Deployment bug, when deployment has replaced all peers the old peers don't seem to be terminating
- [ ] Cleanup - lots of the code (well the stuff I did atleast) got a bit hacked at times, cleanup everything and rebase into one big commit for the mariadb appliance and seperate commits for anything that was done for Sirenia state machine extraction.

appliance/mariadb/handler.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// Close the reader and confirm that it finished.
+	// Sets a trailer at the end of the HTTP response.
+	if err := r.Close(); err == nil {
+		// Temporary hack for writing trailers until Flynn uses Go 1.5.
+		w.(http.Flusher).Flush()
+		conn, buf, _ := w.(http.Hijacker).Hijack()
+		buf.WriteString("0\r\n") // eof
+		http.Header{backupTrailer: {"1"}}.Write(buf)
```
titanous: Would it make sense to put a SHA-512 here?

josephglanville: If Go has a streaming SHA-512 implementing I don't see why not. You can run it through both sides and then verifying it once the trailer is received.

benbjohnson: Sure, I can throw it in there.

benbjohnson: @titanous Added sha512 checksum in 2353d45.

josephglanville:
Status update. Fixed the deployment issue, we had forgotten to properly implement ReadWrite in the peer status endpoint.

I am now moving onto the integration tests for the deployment side of things.
@benbjohnson is working on implementing the CLI tools.

josephglanville:
So I have hit something concerning. When I originally hit it I thought it was just something strange in the dialer code but now that I attempted it with a custom dialer I am now certain it's a bug in either the driver or semi-sync replication or an interaction between the two.

I noted it in the code where I worked around it but will paste there for ease of reading:
```
        // XXX(jpg): For some reason the MySQL driver hangs if CREATE DATABASE
        // is sent as soon as initDb is called.
        // Unfortunately it could be:
        //   a) Bug in go-sql-driver/mysql
        //   b) Bug in MySQL semi-sync replication
        // Both of which are pretty bad, b) ofcourse being much worse.
        // For now just sleep but it does imply there are possible issues here
        // that could effect user applications when a new primary is promoted
        // or a new cluster is newly created and not yet had time to settle.
```

Now the workaround is just a small sleep but it does imply something darker going on as when I did a goroutine stacktrace dump we see the MySQL driver code hanging on trying to fill a buffer from the established connection, it seemingly hangs here forever:
```
goroutine 26 [IO wait, 13 minutes]:
net.(*pollDesc).Wait(0xc2080fced0, 0x72, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:84 +0x47
net.(*pollDesc).WaitRead(0xc2080fced0, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:89 +0x43
net.(*netFD).Read(0xc2080fce70, 0xc20812d000, 0x1000, 0x1000, 0x0, 0x7fdda3470dc8, 0xc2080dc1a8)
	/usr/local/go/src/net/fd_unix.go:242 +0x40f
net.(*conn).Read(0xc2080501a0, 0xc20812d000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/net.go:121 +0xdc
github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql.(*buffer).fill(0xc208043680, 0x4, 0x0, 0x0)
	/home/jpg/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql/buffer.go:68 +0x3bd
github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql.(*buffer).readNext(0xc208043680, 0x4, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/jpg/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql/buffer.go:97 +0x68
github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql.(*mysqlConn).readPacket(0xc208043680, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/jpg/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql/packets.go:30 +0xa1
github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql.(*mysqlConn).readResultSetHeaderPacket(0xc208043680, 0x100000003, 0x0, 0x0)
	/home/jpg/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql/packets.go:470 +0x40
github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql.(*mysqlConn).exec(0xc208043680, 0xc2081326e0, 0x1b, 0x0, 0x0)
	/home/jpg/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql/connection.go:293 +0x8a
github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql.(*mysqlConn).Exec(0xc208043680, 0xc2081326e0, 0x1b, 0xe66558, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/jpg/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql/connection.go:274 +0x2e1
database/sql.(*DB).exec(0xc208058640, 0xc2081326e0, 0x1b, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/database/sql/sql.go:898 +0x2e2
database/sql.(*DB).Exec(0xc208058640, 0xc2081326e0, 0x1b, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/database/sql/sql.go:875 +0xcd
main.func·001(0xc20809c3c0, 0xc2080b8080, 0xc208067620)
	/home/jpg/src/flynn/test/test_mariadb.go:56 +0x6fe
main.testSireniaDeploy(0xc2080503e0, 0xc2080503e8, 0xc20809c3c0, 0xc208067620)
	/home/jpg/src/flynn/test/sirenia.go:190 +0xd61
main.(*MariaDBSuite).TestDeploySingleAsync(0xc208053090, 0xc20809c3c0)
	/home/jpg/src/flynn/test/test_mariadb.go:88 +0x141
reflect.Value.call(0xab8780, 0xc208053090, 0x313, 0xae5650, 0x4, 0xc2080edf78, 0x1, 0x1, 0x0, 0x0, ...)
	/usr/local/go/src/reflect/value.go:419 +0x10e5
reflect.Value.Call(0xab8780, 0xc208053090, 0x313, 0xc2080edf78, 0x1, 0x1, 0x0, 0x0, 0x0)
	/usr/local/go/src/reflect/value.go:296 +0xbc
github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check.func·004(0xc20809c3c0)
	/home/jpg/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check/check.go:809 +0x401
github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check.func·002()
	/home/jpg/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check/check.go:703 +0x92
created by github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check.(*suiteRunner).forkCall
	/home/jpg/go/src/github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-check/check.go:704 +0x3fd
```

A more sophisticated workaround is probably to timeout the statement and redial but I think we should try get to the bottom of this one.

josephglanville:
@benbjohnson rebased on master because the base image updates result in super long builds if you switch between branches heh.

josephglanville:
In my efforts to understand hangs in the integration suite I was able to track the last issue to a deadlock between innobackupex and issuing DDL like "CREATE DATABASE" when the sync is not yet connected.

There is a reasonable workaround for this which is waiting until downstream replication has connected and caught up before attempting to issue DDL or other database wide locking operations.
This workaround is reasonable for a few reasons. Namely that this condition can only really occur during initial setup where we want to pull a backup from the primary to setup the Sync. Anytime after this there should always be one node with enough data to take over for the Sync without requiring a base backup.
Secondly is we control all operations that would happen during this window.

There is also a real fix. Which is to avoid taking full database locks when doing physical backups.
This can be achieved by changing the InnoDB settings to ensure the database is consistent on disk (we should probably do this anyway even if it costs performance) and working out a method to find the most recent event from an on disk database (or be able to work it out after we have started the slave ontop of it).

For now I am going to implement the workaround.

As we need to look at making volumes persistent to make this appliance production ready I will look into feasibility of doing dynamic mount/umount or dataset replacement then which would solve the issue for good and also probably decrease time required to initialise a new slave from an upstream.

josephglanville:
@benbjohnson I have rebased the commits and added your sign-off and co-authorship.

@titanous this is now rebased and ready for review.

benbjohnson:
@josephglanville :clap: :tada:

appliance/mariadb/cmd.go
```diff
@@ -0,0 +1,62 @@
...
+	"os"
+	"os/exec"
+	"sync/atomic"
+	"syscall"
+)
+
+// Cmd wraps exec.Cmd and provides helpers for checking for expected exits.
+type Cmd struct {
+	*exec.Cmd
+	stoppingValue atomic.Value
```
titanous: Add a comment indicating what type is stored in this value.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	fmt.Printf(`GRANT ALL ON %s.* TO '%s'@'%%'`, database, username)
+	if _, err := a.db.Exec(fmt.Sprintf(`GRANT ALL ON %s.* TO '%s'@'%%'`, database, username)); err != nil {
+		a.db.Exec(fmt.Sprintf(`DROP DATABASE "%s"`, database))
+		a.db.Exec(fmt.Sprintf(`DROP USER "%s"`, username))
+		httphelper.Error(w, err)
+		return
+	}
+
+	url := fmt.Sprintf("mysql://%s:%s@%s:5432/%s", username, password, serviceHost, database)
```
titanous: Is this correct? I thought the MySQL port was 3306?

kwando:
:heart:

josephglanville:
@titanous comments addressed.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -0,0 +1,129 @@
...
+
+	handler := httphelper.ContextInjector(serviceName+"-api", httphelper.NewRequestLogger(router))
+	shutdown.Fatal(http.ListenAndServe(addr, handler))
+}
+
+type API struct {
+	db *sql.DB
+}
+
+func (a *API) createDatabase(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
titanous: These handlers should take a context as the first argument. There are examples in a few other APIs including the router: https://github.com/flynn/flynn/blob/master/router/api.go

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -0,0 +1,129 @@
...
+		httphelper.Error(w, err)
+		return
+	}
+	fmt.Printf(`CREATE DATABASE "%s"`, database)
+	if _, err := a.db.Exec(fmt.Sprintf(`CREATE DATABASE %s`, database)); err != nil {
+		a.db.Exec(fmt.Sprintf(`DROP USER "%s"`, username))
+		httphelper.Error(w, err)
+		return
+	}
+	fmt.Printf(`GRANT ALL ON %s.* TO '%s'@'%%'`, database, username)
```
titanous: This and above should be sent through a logger, get it out of the context.

appliance/mariadb/handler.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	defer r.Close()
+
+	w.Header().Set("Trailer", backupChecksumTrailer)
+	w.WriteHeader(200)
+
+	hash := sha512.New()
+	if _, err := io.Copy(io.MultiWriter(w, hash), r); err != nil {
+		h.Logger.Error("error streaming backup", "err", err)
+	}
```
titanous: Missing return?

pkg/sirenia/state/state.go
```diff
@@ -147,15 +147,20 @@ func (x *PgConfig) IsNewDownstream(y *PgConfig) bool {
...
 
-type Postgres interface {
+type Database interface {
+	// TODO(jpg): need a more generic representation of database log position.
+	// the only important thing here is to ensure we have an ordering.
+	// the comparison of this should also be handed of to the underlying database wrapper
 	XLogPosition() (xlog.Position, error)
-	Reconfigure(*PgConfig) error
+	// Returns the implementation of the XLog interface for this database
+	XLog() xlog.XLog
```
titanous: @josephglanville At one point we discussed using an interface that supported comparison returned from `XLogPosition` instead of the singleton, as the only reason it is singleton is `Zero` which is only used be the simulator. Not sure if this is still the case, but wanted to point it out.

appliance/mariadb/dsn.go
```diff
@@ -0,0 +1,38 @@
...
+	User     string
+	Password string
+	Database string
+	Timeout  time.Duration
+}
+
+// String encodes dsn to a URL string format.
+func (dsn *DSN) String() string {
+	u := url.URL{
+		Host: fmt.Sprintf("tcp(%s)", dsn.Host),
```
titanous: This `tcp()` is weird, what is the basis for it?

appliance/mariadb/handler.go
```diff
@@ -0,0 +1,133 @@
...
+
+// ServeHTTP serves an HTTP request.
+func (h *Handler) ServeHTTP(w http.ResponseWriter, req *http.Request) { h.router.ServeHTTP(w, req) }
+
+// healthStatus returns the current status of the peer and process.
+func (h *Handler) healthStatus() status.Status {
+	info := h.Peer.Info()
+	if info.State == nil || info.RetryPending != nil ||
+		(info.Role != state.RolePrimary && info.Role != state.RoleSync && info.Role != state.RoleAsync) {
+		return status.Unhealthy
```
titanous: Is the exact same status handler logic used for postgres?

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1023 @@
...
+		"--port="+p.Port,
+		"--user=flynn",
+		"--password="+p.Password,
+		"--socket=",
+		"--stream=xbstream",
+		".",
+	)
+	cmd.Dir = p.DataDir
+	cmd.Stderr = os.Stderr
+	cmd.Stderr = &r.stderr
```
titanous: Double assignment to `cmd.Stderr`?

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1023 @@
...
+			// Close response and confirm backup from trailer.
+			if err := resp.Body.Close(); err != nil {
+				logger.Error("error closing backup body", "err", err)
+				return err
+			}
+
+			chk := hex.EncodeToString(hash.Sum(nil))
+			logger.Error("verifying backup checksum", "hash", chk)
+			if hdr := resp.Trailer.Get(backupChecksumTrailer); hdr != chk {
+				logger.Error("invalid backup checksum", "hash", chk)
```
titanous: This should log the actual and desired checksums.

appliance/mariadb/process.go
```diff
@@ -0,0 +1,1023 @@
...
+	}
+	defer db.Close()
+	var readOnly string
+	if err := db.QueryRow("SELECT @@read_only").Scan(&readOnly); err != nil {
+		return false, err
+	}
+	if readOnly == "0" {
+		return true, nil
+	}
+	return false, nil
```
titanous: This could be simplified to:

```go
return readOnly == "0", nil

File: appliance/mariadb/process.go [hunk 26]
```diff
@@ -0,0 +1,1023 @@
...
+	}
+
+	go func() {
+		defer close(doneCh)
+
+		startTime := time.Now().UTC()
+		logger := p.Logger.New(
+			"fn", "waitForSync",
+			"sync_name", downstream.Meta["MYSQL_ID"],
+			"start_time", log15.Lazy{func() time.Time { return startTime }},
```
titanous: Why is this lazy?

bootstrap/manifest_template.json
```diff
@@ -303,6 +304,58 @@
...
+    "processes": {
+      "mariadb": 3,
+      "web": 2
+    }
+  },
+  {
+    "id": "add-mysql-provider",
+    "action": "add-provider",
+    "name":"mysql",
+    "url":"http://mariadb-api.discoverd/databases"
```
titanous: Add spaces after the colon on this and the previous line.

bootstrap/manifest_template.json
```diff
@@ -303,6 +304,58 @@
...
+    },
+    "processes": {
+      "mariadb": 3,
+      "web": 2
+    }
+  },
+  {
+    "id": "add-mysql-provider",
+    "action": "add-provider",
+    "name":"mysql",
```
titanous: What do you think about calling the provider `mariadb`?

cli/mysql.go
```diff
@@ -0,0 +1,185 @@
...
+usage: flynn mysql console [--] [<argument>...]
+       flynn mysql dump [-q] [-f <file>]
+       flynn mysql restore [-q] [-f <file>]
+
+Options:
+	-f, --file=<file>  name of dump file
+	-q, --quiet        don't print progress
+
+Commands:
+	console  Open a console to a Flynn mysql database. Any valid arguments to mysql may be provided.
```
titanous: Should we use MariaDB consistently instead of MySQL?

titanous:
Some things that need to be done:

- [ ] Add to the status app
- [ ] Add data dump/restore to app backup/restore
- [ ] Add data dump/restore to cluster backup/restore

appliance/mariadb/dsn.go
josephglanville: This is used by the MySQL driver.

appliance/mariadb/handler.go
josephglanville: Yes. Is it possible to extract and re-use these routes and compose them with the extra MariaDB ones?

bootstrap/manifest_template.json
josephglanville: Don't feel strongly about it. Figured MySQL makes more sense as that is what most people will be looking for. MariaDB is more of an implementation detail.

cli/mysql.go
josephglanville: As above I think MariaDB is best kept is implementation detail.

bootstrap/manifest_template.json
benbjohnson: I can see changing the provider to `mariadb`. If we supported other variants in the future then there would need to be different options.

cli/mysql.go
benbjohnson: For the CLI, `mysql` makes sense to me since regardless of the implementation used it would likely go through the same CLI.

philiplb: May I throw this one in? MariaDB and MySQL are (in small details) not compatible to each other. With having MySQL as a name and using MariaDB in the background, you would block the route to having both integrated. I think it would make sense at some point, to offer both.

josephglanville: @philiplb it would not block the way as any multi-engine support would be server side. The CLI is strictly client side.

That said, to address the compatibility side for completeness:

While the 2 aren't strictly compatible they are from a client perspective unless you are doing something very very strange. If you are worried about this you should read the "Incompatibilities between MariaDB 10.0 and MySQL 5.6"  section of the MariaDB/MySQL compatibility page: https://mariadb.com/kb/en/mariadb/mariadb-vs-mysql-compatibility/

In real terms there is no practical difference if you were already limiting yourself to the features of MySQL (note MariaDB has a much large feature set). Most of the incompatibilities relate to replication and other HA features which Flynn already takes care of so you shouldn't need to touch them.

We avoid most of the nasty stuff by not using Galera replication as it's write set replication while not strictly incompatible has very surprising behaviour. If you have heard of client issues with regards to MariaDB or PXC vs MySQL this is probably the source of that information. In Flynn we just use standard master/slave replication to avoid these very problems to ensure the clustering is as minimally impacting on client applications as possible.

philiplb: All valid points, but I still don't see why not just call it "MariaDB" if it is a MariaDB. :) This would also hint the users to the additional features of Maria.
And the name "MySQL" wouldn't be taken on the server side? Assuming the Flynn team or someone from the outer side wants to implement support for the original MySQL DB.

bootstrap/manifest_template.json
josephglanville: I actually think it's more likely they would sit behind the `mysql` provider and simply have it direct the query appropriately. We will inevitably need this later to support dedicated DB clusters not just shared ones so it stands to reason that this functionality would be contained there. Once we support providing options to resources I think this will be more natural.

pkg/sirenia/state/state.go
josephglanville: I tried to eliminate the singleton but `Zero` is also used in the state machine to do various comparisons with the initial position. I think what we have now is actually as minimal as it can be barring Go tricks that I don't know yet.

appliance/mariadb/handler.go
josephglanville: I'm not sure what to do here. @benbjohnson can you chime in one how we should handle the failed Copy? I am assuming we should just close the reader we got back from Backup and just return? Or should we write something in the trailer?

benbjohnson: @titanous Yes, it's missing a return.

@josephglanville We should be able to simply close the reader and the let the client see that the trailer is missing.

josephglanville: Ok I will make that change.

josephglanville:
(most) Review comments addressed. I'm not sure about the lazy log message and for now I am going to leave the HTTP handler. When we move onto MongoDB I will see how much we can pull into the Sirenia package.

titanous:
Did restoration from cluster backup get implemented? (`host/cli/bootstrap.go`)

josephglanville:
@titanous Implemented the cluster restore side.

josephglanville:
@titanous this is ready for another look, I have implemented and tested restoring from a pre-MariaDB backup.

titanous:
@josephglanville Did you also test taking a backup from a cluster that doesn't have MariaDB running or released? (ie after restoring to it from a backup without MariaDB, or a cluster that predates the appliance)

josephglanville:
@titanous updated to address backing up when using the new code when mariadb doesn't exist.

There is a slight problem however that concerns both MariaDB and Redis, which is that the status app isn't aware that these features weren't added so it will report unhealthy. A result of this is restoring from a backup that doesn't have MariaDB or Redis in it with a version of the status app that expects it reports the cluster as unhealthy even though it is healthy.

A possible fix may be a feature matrix with toggles or something, not sure. Getting into the realm of stuff that actually needs to be thought about properly.

titanous:
One option is to split MariaDB and Redis out into an "optional" category that doesn't cause the overall status to go to unhealthy. I expect that some users will scale MariaDB and Redis down to save resources if they don't use them. Since they are not critical to the operation of anything I think this is an easy approach until we resolve more of the issues around generalizing data appliances.

titanous:
LGTM! Great work!

The status change can go in a separate PR, file an issue if you're not going to handle it immediately.

titanous:
:cake: :tada:

lmars:
:clap: :smile: :clap:

josephglanville:
:cake: :tada: :balloon: :+1: :clap:

tuananh:
🍰 👏

PR #2435, opened 2016-02-09
host: Don’t persist backend state for containers that aren’t running

### LINKED ISSUE #1461: flynn-host tries to reconnect to dead jobs
When starting, flynn-host tries to reconnect to containers that may still be running, but it doesn't filter out jobs that it knows have finished. This can cause it to take a long time to start if there are many finished jobs.

titanous:
This avoids trying to reconnect to definitely dead containers when restarting flynn-host.

Closes #1461


lmars:
LGTM

PR #2443, opened 2016-02-11
router: Specify service in release

lmars:
This means deployments will not succeed if the router job fails to register the service (e.g. if the `COOKIE_KEY` or TLS env vars are invalid).


titanous:
LGTM

josephglanville:
LGTM

Issue #2422, opened 2016-02-07
Improve artifact / image management


lmars:
Here are some proposed changes for improving image management and artifact creation:
- Switch to using OverlayFS rather than AUFS
  
  Given OverlayFS is now in the mainline kernel, other than having a minimum requirement on kernel version we would not need to install extra packages [as we do for AUFS](https://github.com/flynn/flynn/blob/master/script/install-flynn.tmpl#L102-L112). OverlayFS is also simpler than AUFS and is reported to be faster.
- Support creating artifacts from a job's filesystem changes
  
  `flynn-host` creates a layer for storing the filesystem changes made by a job, but this is temporary and removed when the job stops. We should support keeping this layer around so that it can be fetched, stored in the blobstore and an artifact created for it. We would likely want to support include / exclude rules so we could run a build job but only save the result of the build and not any intermediary files.
- Support multiple artifacts per release
  
  Rather than jobs having a single image pulled from the release's only artifact, we should support releases having an ordered list of artifacts which are overlayed under the job's layer. We could then migrate slugs from being tarball URLs stored in environment variables to artifacts which are overlayed over the slugrunner artifact (meaning slugs wouldn't be fetched / extracted every time a job starts, they would be cached in the host's image storage). This would also give us the flexibility to have arbitrary build pipelines which create artifacts and make them available to jobs (e.g. have a Ruby artifact which can be added to a release if a job needs Ruby), moving us closer to supporting #1865.

Thoughts very much welcome on all of the above, but my particular concerns:
- storing layers in the blobstore means they are stored in postgres, and so are replicated across the cluster, but then when `flynn-host` pulls the layer, it will also have it stored, so we are in effect storing the layer twice on some hosts. It would be better if `flynn-host` itself was the service from which layers were pulled (so an artifact URL would be something like `http://leader.flynn-host.discoverd/image/<id>`) but then we need to think about HA / replication at that level, is that something we want?
- overlaying artifacts means higher layers can "overwrite" files from lower layers, which could cause incompatibilities (e.g. if the base layer has `libfoo` symlinked to `libfoo.2.0` and a later layer changes that symlink to `libfoo.3.0`, library loading may break). I guess we need to make this clear (and maybe convince people to use [Nix](http://nixos.org/nix/) :wink:)

josephglanville:
Few thoughts.

* If we are going to move to OverlayFS we also need to look at security. I don't know how well it plays with user namespaces for example which is really important for being able to give containers root.

[...]

PR #2448, opened 2016-02-15
Add file artifacts

lmars:
Summary:

* Add a `tar` artifact type along with an `Attributes` field containing type specific attributes (compression and target path for `tar` artifacts)
* Rename `release.Artifact` to `release.ImageArtifact` (this is always required for a release to be runnable)
* Add `release.TarArtifacts` which is an optional list of `tar` artifacts
* Add `ExpandedFormation.TarArtifacts` and ensure it's always populated
* Add `Job.TarArtifacts` and extract them into the job's layer before starting the job
* Create tar artifacts for slugs

This is part of the changes proposed in #2422.

### DIFF
--- a/controller/schema.go
+++ b/controller/schema.go
@@ -317,6 +317,24 @@ $$ LANGUAGE plpgsql`,
 			AFTER INSERT ON release_artifacts
 			FOR EACH ROW EXECUTE PROCEDURE check_release_artifacts()`,
 		`INSERT INTO release_artifacts (release_id, artifact_id) (SELECT release_id, artifact_id FROM releases WHERE artifact_id IS NOT NULL)`,
+
+		// set "git=true" for releases with SLUG_URL set
+		`UPDATE releases SET meta = jsonb_merge(CASE WHEN meta = 'null' THEN '{}' ELSE meta END, '{"git":"true"}') WHERE env ? 'SLUG_URL'`,
+
+		// create file artifacts for any releases with SLUG_URL set
+		`DO $$
+		DECLARE
+			release RECORD;
+		BEGIN
+			FOR release IN SELECT * FROM releases WHERE env ? 'SLUG_URL' LOOP
+				WITH artifact AS (
+					INSERT INTO artifacts (type, uri, meta)
+					VALUES ('file', release.env->>'SLUG_URL', '{"blobstore":"true"}')
+					RETURNING *
+				)
+				INSERT INTO release_artifacts (release_id, artifact_id) (SELECT release.release_id, artifact_id FROM artifact);
+			END LOOP;
+		END $$`,
 		`ALTER TABLE releases DROP COLUMN artifact_id`,
 	)
 }
--- a/controller/schema.go
+++ b/controller/schema.go
@@ -288,6 +288,37 @@ $$ LANGUAGE plpgsql`,
 		)`,
 		`INSERT INTO event_types (name) VALUES ('cluster_backup')`,
 	)
+	migrations.Add(15,
+		`ALTER TABLE artifacts ADD COLUMN meta jsonb`,
+		`CREATE TABLE release_artifacts (
+			release_id uuid NOT NULL REFERENCES releases (release_id),
+			artifact_id uuid NOT NULL REFERENCES artifacts (artifact_id),
+			created_at timestamptz NOT NULL DEFAULT now(),
+			deleted_at timestamptz,
+			PRIMARY KEY (release_id, artifact_id))`,
+
+		// add a check to ensure releases only have a single "docker"
+		// artifact, and that artifact is added first
+		`CREATE FUNCTION check_release_artifacts() RETURNS OPAQUE AS $$
+			BEGIN
+			    IF (
+			      SELECT COUNT(*)
+			      FROM release_artifacts r
+			      INNER JOIN artifacts a ON r.artifact_id = a.artifact_id
+			      WHERE r.release_id = NEW.release_id AND a.type = 'docker'
+			    ) != 1 THEN
+			      RAISE EXCEPTION 'must have exactly one artifact of type "docker"' USING ERRCODE = 'check_violation';
+			    END IF;
+
+			    RETURN NULL;
+			END;
+			$$ LANGUAGE plpgsql`,
+		`CREATE TRIGGER release_artifacts_trigger
+			AFTER INSERT ON release_artifacts
+			FOR EACH ROW EXECUTE PROCEDURE check_release_artifacts()`,
+		`INSERT INTO release_artifacts (release_id, artifact_id) (SELECT release_id, artifact_id FROM releases WHERE artifact_id IS NOT NULL)`,
+		`ALTER TABLE releases DROP COLUMN artifact_id`,
+	)
 }
 
 func migrateDB(db *postgres.DB) error {
--- a/controller/types/types.go
+++ b/controller/types/types.go
@@ -80,6 +80,10 @@ func (r *Release) FileArtifactIDs() []string {
 	return r.ArtifactIDs[1:len(r.ArtifactIDs)]
 }
 
+func (r *Release) IsGitDeploy() bool {
+	return r.Meta["git"] == "true"
+}
+
 type ProcessType struct {
 	Cmd         []string           `json:"cmd,omitempty"`
 	Entrypoint  []string           `json:"entrypoint,omitempty"`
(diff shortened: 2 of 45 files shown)


controller/types/types.go
```diff
@@ -15,12 +15,13 @@ import (
...
-	App       *App                         `json:"app,omitempty"`
-	Release   *Release                     `json:"release,omitempty"`
-	Artifact  *Artifact                    `json:"artifact,omitempty"`
-	Processes map[string]int               `json:"processes,omitempty"`
-	Tags      map[string]map[string]string `json:"tags,omitempty"`
-	UpdatedAt time.Time                    `json:"updated_at,omitempty"`
+	App           *App                         `json:"app,omitempty"`
+	Release       *Release                     `json:"release,omitempty"`
+	ImageArtifact *Artifact                    `json:"artifact,omitempty"`
+	TarArtifacts  []*Artifact                  `json:"tar_artifacts,omitempty"`
```
titanous: What do you think about just a list of artifacts: `Artifacts []*Artifact` instead of splitting these into two fields?

lmars: The reason I went with two fields is because most of the time code is concerned with assigning / changing the "image" artifact, it always needs to be present for the release to be run, and there should only be one of them, which I feel is clearer when it has an explicit field rather than appearing in a slice.

WDYT?

jvatic:
This looks like it will require some minor dashboard changes. I'll take a closer look when I'm back.

controller/types/types.go
```diff
@@ -15,12 +15,13 @@ import (
...
 type ExpandedFormation struct {
-	App       *App                         `json:"app,omitempty"`
-	Release   *Release                     `json:"release,omitempty"`
-	Artifact  *Artifact                    `json:"artifact,omitempty"`
-	Processes map[string]int               `json:"processes,omitempty"`
-	Tags      map[string]map[string]string `json:"tags,omitempty"`
-	UpdatedAt time.Time                    `json:"updated_at,omitempty"`
+	App           *App                         `json:"app,omitempty"`
+	Release       *Release                     `json:"release,omitempty"`
+	ImageArtifact *Artifact                    `json:"artifact,omitempty"`
```
jvatic: The dashboard shouldn't be effected if the JSON field name stays the same (otherwise there's a [single reference](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/views/app-history.js.jsx#L73) that would need updating using the new name).

host/libvirt_lxc_backend.go
```diff
@@ -464,6 +469,14 @@ func (l *LibvirtLXCBackend) Run(job *host.Job, runConfig *RunConfig) (err error)
 		}
 	}
 
+	for _, artifact := range job.TarArtifacts {
+		g.Log(grohl.Data{"at": "extract_tar", "uri": artifact.URI, "target_path": artifact.Attributes.TarTargetPath})
+		if err := fetchAndExtractTar(rootPath, artifact); err != nil {
```
titanous: Can we do this in containerinit instead?

controller/schema.go
```diff
@@ -241,6 +241,45 @@ $$ LANGUAGE plpgsql`,
 	migrations.Add(10,
 		`ALTER TABLE formations ADD COLUMN tags jsonb`,
 	)
+	migrations.Add(11,
+		`ALTER TABLE artifacts ADD COLUMN attributes jsonb`,
+		`CREATE TABLE release_tar_artifacts (
```
titanous: Should this just be `release_artifacts`? Is there a reason to specialize it for tarballs? We could enforce the "base image" requirement in the database and move the artifact field out of the release and into this join table.

titanous:
@josephglanville Please take a look at the SQL.

host/libvirt_lxc_backend.go
lmars: This was my initial thought but I put it in the host as then we can add artifact caching later. We could of course do the fetch in the host and extract in containerinit, but without caching we would then be fetching to disk unnecessarily.

controller/schema.go
lmars: Ok I wasn't super happy with this either, I'll play around see what I come up with.

host/libvirt_lxc_backend.go
titanous: Let's start with it in containerinit, we can move it out if necessary for caching.

lmars:
@titanous moved the image artifact into the join table with a constraint to make sure the first artifact is an image, and it is the only one: 8b52687

I have kept the separation of image / tar artifacts on `ExpandedFormation` and `host.Job` as I think it's useful.

I have also added `release.LegacyArtifactID` for backwards compatibility with clients who only expect a single artifact, but we should migrate things like the dashboard to use `release.ArtifactIDs` (/cc @jvatic).

jvatic:
@lmars it looks like this should be the only change the dashboard requires: https://gist.github.com/jvatic/a5bd30e29b65b5fba425

jvatic:
@lmars added a commit with the dashboard change

lmars:
@titanous I had a change of heart and have replaced the notion of "tar" artifacts with simpler "file" artifacts and just downloaded them into `/artifacts` for the job (so in the case of slugs, leaving the slugrunner [responsible for extracting the slug](https://github.com/flynn/flynn/blob/29ee939e4fde6bfc313259b074f2a13fa62d6456/slugrunner/runner/init#L11-L12)).

I think this gives us a bit more flexibility for using artifacts in the future without having to change the controller / containerinit.

I'll probably squash all the commits together before merging.

lmars:
Updated the migration to add file artifacts for releases with `SLUG_URL` set: d5b073c

lmars:
Just to re-summarise the changes here:

* Add a `file` artifact type which is expected to have a URI pointing at a single file (e.g. a tarball)
* Update releases to have many artifacts, with the first artifact always having type `docker` and serving as the image to run the release with, then an optional list of `file` artifacts after that
* Add `ExpandedFormation.FileArtifacts` and ensure it's always populated from the release
* Add `Job.FileArtifacts` and fetch them into `/artifacts` in the job's filesystem layer before starting the job
* Create file artifacts for slugs (so slugs are available to the slugrunner at `/artifacts/slug.tgz`)

@titanous @jvatic @josephglanville please review.

jvatic:
- [ ] run `script/generate-examples controller`
- [ ] [update dashboard to handle `release.artifacts`](https://gist.github.com/jvatic/a5bd30e29b65b5fba425)

lmars:
@jvatic ah looks like I dropped your commit along the way, re-added: aa9e872

controller/types/types.go
```diff
@@ -40,12 +41,42 @@ func (a *App) System() bool {
...
+}
+
+func (r *Release) FileArtifactIDs() []string {
+	if len(r.ArtifactIDs) < 1 {
+		return nil
+	}
+	return r.ArtifactIDs[1:len(r.ArtifactIDs)]
+}
+
+func (r *Release) GitDeploy() bool {
```
jvatic: Do you think `IsGitDeploy` would be a better name for this?

host/containerinit/init.go
```diff
@@ -400,6 +402,15 @@ func setupCommon(c *Config, log log15.Logger) error {
 		return err
 	}
 
+	// fetch file artifacts now that the network is configured
+	for _, artifact := range c.FileArtifacts {
+		log.Info("fetching artifact", "uri", artifact.URI)
+		if err := fetchFileArtifact(artifact); err != nil {
```
jvatic: Should these be fetched concurrently?

lmars:
@jvatic comments addressed. I ran `script/generate-examples controller` but the diff seems to have some unexpected changes, can you take a look?

jvatic:
@lmars hmm, looks like the example generator needs updating... lets move this change into another PR so it doesn't hold this one up.

lmars:
@jvatic ok, I have dropped the examples update commit

jvatic:
@lmars cool, I'll open another PR for the examples/generator update

lmars:
@jvatic @titanous I have squashed the commits and manually tested app & cluster export / imports, ready for a final review.

host/containerinit/init.go
```diff
@@ -586,6 +608,27 @@ func babySit(process *os.Process) int {
...
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	res, err := http.Get(artifact.URI)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+	_, err = io.Copy(file, res.Body)
```
titanous: This should check for a 200.

titanous:
Sorry for the massive delay in reviewing. LGTM pending comment.

PR #2456, opened 2016-02-16
router: Use a retry dialer in the client

lmars:
This prevents errors in the controller when trying to create routes during a router deployment as seen in some domain migration jobs.


josephglanville:
LGTM

PR #2457, opened 2016-02-16
host: Always use logfmt output when logging

lmars:
When run as an Upstart job, stdout is connected to a pty which ultimately gets logged to a file, which means color codes and escape characters end up being logged to the file (e.g. see [this host log](https://gist.githubusercontent.com/anonymous/5da86fa3300212c642e1/raw/0dceef9f4483588819b893ff1e82929bf5f92327/upstart-flynn-host.log)).


titanous:
LGTM

josephglanville:
LGTM

Issue #2459, opened 2016-02-16
pinkerton: Incompatible with newer versions of Docker


josephglanville:
Pinkerton is unable to interact with the 1.10.X and greater versions of Docker, most likely due to a backwards incompatible update to the on-disk representation of the image repository.

titanous:
We no longer use the Docker image representation on-disk.

PR #2460, opened 2016-02-16
util/packer: Pin docker version

josephglanville:
Pinkerton isn't currently compatible with more recent Docker
versions so we must pin it for now.

Refs: #2459


titanous:
LGTM

lmars:
LGTM

lmars:
I've added a commit to also pin in CI.

PR #2470, opened 2016-02-18
host/containerinit: Print goroutine trace to log fd on SIGUSR2

titanous:
This will help debug an issue we’re seeing where containerinit processes stop health checking and responding to RPCs.


josephglanville:
LGTM

lmars:
LGTM

PR #2471, opened 2016-02-18
all: Update base images

lmars:
`flynn/busybox:trusty-20160217` was released after merging https://github.com/flynn/docker-busybox/pull/1.

I have not changed the slugrunner / slugbuilder base image references because they are [built after cedarish](https://github.com/flynn/flynn/blob/master/slugbuilder/Tupfile#L2), so will always use the latest cedarish as part of the build.


titanous:
LGTM

PR #2488, opened 2016-02-23
controller/client: Convert time to UTC when streaming formations

josephglanville:
This ensures the controller client will work on systems not configured for UTC.


titanous:
LGTM

Issue #2486, opened 2016-02-22
Scheduler nuked a cluster after discoverd blip


lmars:
A user in IRC had a cluster where nothing other than `flynn-host` was running on all 3 nodes in the cluster. Here is the debug-info from each:

`10.0.0.4`: https://gist.github.com/anonymous/217f1e5c6791be897429
`10.0.0.5`: https://gist.github.com/anonymous/14e8e235844b2fd1113e
`10.0.0.6`: https://gist.github.com/anonymous/7d43556b81879befda16

The discoverd logs from `.4` show errors related to connections to the other nodes at `2016/02/20 02:17:10`:

```
2016/02/20 02:17:10 [INFO] raft: aborting pipeline replication to peer 10.0.0.6:1111
2016/02/20 02:17:10 [INFO] raft: aborting pipeline replication to peer 10.0.0.5:1111
2016/02/20 02:17:10 [INFO] raft: pipelining replication to peer 10.0.0.6:1111
2016/02/20 02:17:10 [INFO] raft: pipelining replication to peer 10.0.0.5:1111
```

It then marked a bunch of services as expired, thus broadcasting many down / leader events:

```
t=2016-02-20T02:17:13+0000 lvl=info msg="marking instance for expiry" component=discoverd fn=EnforceExpiry service=cluster-monitor instance.id=876debd453fc6f909386f8cc61e3c810 instance.addr=10.0.0.6:1113
t=2016-02-20T02:17:13+0000 lvl=info msg="marking instance for expiry" component=discoverd fn=EnforceExpiry service=discoverd instance.id=84c5e3cb37240d31091b6aea0a74936c instance.addr=10.0.0.6:1111
t=2016-02-20T02:17:13+0000 lvl=info msg="broadcasting down event" component=discoverd fn=broadcast event=down service=cluster-monitor instance.id=876debd453fc6f909386f8cc61e3c810 instance.addr=10.0.0.6:1113
t=2016-02-20T02:17:13+0000 lvl=info msg="broadcasting down event" component=discoverd fn=broadcast event=down service=discoverd instance.id=84c5e3cb37240d31091b6aea0a74936c instance.addr=10.0.0.6:1111
t=2016-02-20T02:17:13+0000 lvl=info msg="broadcasting leader event" component=discoverd fn=broadcast event=leader service=cluster-monitor instance.id=719bfebcef0ee4ab467a2d2ef1edb8d5 instance.addr=10.0.0.5:1113
[...]

PR #2489, opened 2016-02-23
controller: Check rows.Err when listing/streaming formations

josephglanville:
When reading back formations from the database it's possible some errors won't be returned on query and instead will be contained in `rows.Err()`.
For instance if the connection pool returns a dead connection.

It's likely this was the root cause of #2486


titanous:
LGTM, great find!

Issue #1084, opened 2015-02-25
Replace Procfile with Flynnfile


lmars:
We currently use a `Procfile` to declare process types, but this only supports declaring how to start the job, whereas process types are [more complex than that](https://github.com/flynn/flynn/blob/master/controller/types/types.go#L42-L51).

We should replace this with a `Flynnfile` which can be parsed by the receiver and supports declaring all aspects of a process type (e.g. mounted volumes, ports, services, healthchecks etc.).

The `Flynnfile` could also declare pre / post deploy hooks once supported.

lmars:
We should ensure that we keep backwards compatibility.

thomasfedb:
I would be careful about removing the `Procfile` altogether, unless you can get foreman to support your format.

lmars:
@thomasfedb 

> I would be careful about removing the Procfile altogether, unless you can get foreman to support your format.

Flynn doesn't use Foreman, it parses the `Procfile` directly and extracts the process types (see [here](https://github.com/flynn/flynn/blob/master/slugbuilder/builder/build.sh#L149-L155)), so by "Replace Procfile" I mean "Replace the file which Flynn uses to define processes". Users can still have a `Procfile` in their project for use locally (which will be a fallback), but by default Flynn will look for a `Flynnfile`.

I don't think trying to redefine the `Procfile` format to support Flynn is going to be useful to a lot of people that use Procfiles (I'm pretty sure the current format is simply `TYPE: COMMAND`).

thomasfedb:
Right, so it was more hijacking to beging with anyway.

lmars:
The `Flynnfile` could have a `packages` section which uses [heroku-buildpack-apt](https://github.com/ddollar/heroku-buildpack-apt) to install packages using `apt-get`.

geovanisouza92:
@lmars I'd recent problems with heroku-buildpack-apt on flynn buildpack caching, and I think that one session exclusively for this can be nice, although I feel that this special case is not so special.

Anyway, creating another file format can be nice to fit with flynn ecosystem, but could detour flynn of nice existing tools like Foreman/Honcho or even Docker Compose, that I particularly think that is not so awesome substitute for Fig ('cause is just a rebrand and rewrite of the tool), without any additions.

So I think it may be nice to have a new file format to configure everything in one place, but I worry that this new format can accumulate so many others like Procfile, [fig | docker-compose].yml, terraform.yml , etc, etc, etc.

lmars:
[...]

PR #2491, opened 2016-02-23
controller: Bump default deploy timeout to two minutes

titanous:
This will allow deploying apps with larger slugs.


josephglanville:
LGTM.

Another thing to add to the long list of Flynnfile features in #1084

Issue #2494, opened 2016-02-24
Sometimes, a flynn node doesn't survive a reboot


philiplb:
Hi,

playing around with reboots and sometimes, a node doesn't come back. Resulting in
```
smf@flynn1:~$ status flynn-host
flynn-host start/running, process 1517
smf@flynn1:~$ flynn-host ps
14:52:59.793319 host.go:140: could not list hosts: Get http://127.0.0.1:1111/services/flynn-host/instances: dial tcp 127.0.0.1:1111: connection refused
```

And on another node:
```
smf@flynn0:~$ sudo flynn-host fix --min-hosts 3
[sudo] password for smf:
14:53:18.495260 host.go:140: expected at least 3 hosts, but 2 found
```

I rebooted the failing node again and all went back to normal. Wheras the actuall application I'm running took something like 15 minutes to come back. It was running on the rebooted node.

josephglanville:
@philiplb can you gist the entirety of `/var/log/upstart/flynn-host.log`

philiplb:
@josephglanville Sure: https://gist.github.com/philiplb/87e535f6bfdff9c004a1

jontonsoup:
+1 on this. I get shaky reboot stability.

titanous:
This should be resolved by #2495. Please open a new issue if you encounter it again with that patch.

philiplb:
Alright, thanks. :)

PR #2495, opened 2016-02-24
host: Always resurrect jobs

josephglanville:
Being able to connect to discoverd nodes on other peers doesn't necessarily mean the scheduler is healthy. We should resurrect our own jobs regardless so that if the cluster is damaged then the cluster-monitor will be able to fix it.


titanous:
Is this related to #2494?

titanous:
LGTM

josephglanville:
Yes, I suspect this is the cause of #2494.

PR #2503, opened 2016-02-29
Backup restore fixes

titanous:
First fix for some issues we've seen during backup restores. This will avoid overloading flynn-host when starting everything up.


lmars:
LGTM

Issue #2461, opened 2016-02-16
flynn-host fd exhaustion caused by failed stop


titanous:
I just ran into a cluster where `flynn-host` ran out of fds for the following reasons:
1. The scheduler wanted to terminate a job running on the host and sent a `DELETE` request to the `flynn-host` daemon and there were no available connection to reuse so it established a new TCP connection for the HTTP request to flynn-host.
2. `flynn-host` sent a `Stop` RPC to the `containerinit` process.
3. For some reason `containerinit` never responded to the RPC, so a HTTP response was never returned to the scheduler.
4. GOTO 1

This loop resulted in fd exhaustion in the `flynn-host` daemon. We should have safeguards that prevent this scenario.

titanous:
Here's an example stuck RPC goroutine:

```text
goroutine 264225 [chan receive, 5009 minutes]:
github.com/flynn/flynn/pkg/rpcplus.(*Client).Call(0xc2082eb0e0, 0xbab210, 0x14, 0x96e2e0, 0xc2096cc4a8, 0x94df20, 0x10ef0a0, 0x0, 0x0)
	/tmp/tmp.Z0ELp3CIwu/src/github.com/flynn/flynn/pkg/rpcplus/client.go:412 +0xd6
github.com/flynn/flynn/host/containerinit.(*Client).Signal(0xc208035538, 0xf, 0x0, 0x0)
	/tmp/tmp.Z0ELp3CIwu/src/github.com/flynn/flynn/host/containerinit/init.go:156 +0xe3
main.(*libvirtContainer).Stop(0xc208372050, 0x0, 0x0)
	/tmp/tmp.Z0ELp3CIwu/src/github.com/flynn/flynn/host/libvirt_lxc_backend.go:940 +0x47
main.(*LibvirtLXCBackend).Stop(0xc208064a00, 0xc209653e12, 0x31, 0x0, 0x0)
	/tmp/tmp.Z0ELp3CIwu/src/github.com/flynn/flynn/host/libvirt_lxc_backend.go:954 +0x85
main.(*Host).StopJob(0xc208057e00, 0xc209653e12, 0x31, 0x0, 0x0)
	/tmp/tmp.Z0ELp3CIwu/src/github.com/flynn/flynn/host/http.go:78 +0x6ff
main.(*jobAPI).StopJob(0xc2080342b8, 0x7f314658b638, 0xc209484990, 0xc209c53e10, 0xc209403220, 0x1, 0x1)
[...]

PR #2505, opened 2016-02-29
Fix containerinit hang

titanous:
Before this patch, the discoverd heartbeater would attempt to log errors to `os.Stderr`, which for reasons that have not been fully investigated would hang permanently. The process would never restart heartbeating when discoverd came back and all RPC operations would also become wedged. This prevented the process from being killed and resulted in the fd exhaustion reported in #2461.

Logging to the fd we use for all other log messages from containerinit solves this problem.

### DIFF
--- a/discoverd/client/client.go
+++ b/discoverd/client/client.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


lmars:
LGTM, good find!

PR #2512, opened 2016-03-02
discoverd: Bump leader wait timeout at start

titanous:
If the discoverd cluster doesn’t come up within this timeout, we can prematurely break the bootstrap process, bump it to a nice high ten minutes.


josephglanville:
LGTM

PR #2515, opened 2016-03-02
router: Fix hard-coded high port in tests

titanous:
Use our low port allocator. High ports can conflict with ephemeral kernel usage.


josephglanville:
LGTM

PR #2529, opened 2016-03-05
host: Make max job concurrency configurable

josephglanville:
Allow configuring the maximum number of jobs the API will allow to be started concurrently from the command line.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -56,6 +56,7 @@ options:
   --peer-ips=IPLIST      join existing cluster using IPs
   --bridge-name=NAME     network bridge name [default: flynnbr0]
   --no-resurrect         disable cluster resurrection
+  --max-job-concurrency  maximum number of jobs to start concurrently
 	`)
 }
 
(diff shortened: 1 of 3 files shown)


josephglanville:
@lmars where is it best to raise this for CI?

host/host.go
```diff
@@ -56,6 +56,7 @@ options:
   --peer-ips=IPLIST      join existing cluster using IPs
   --bridge-name=NAME     network bridge name [default: flynnbr0]
   --no-resurrect         disable cluster resurrection
+  --max-job-concurrency  maxiumum number of jobs to start concurrently
```
titanous: `maximum`

host/http.go
```diff
@@ -47,6 +47,8 @@ type Host struct {
 
 	listener net.Listener
 
+	maxJobConcurrency uint64
```
titanous: Doesn't this need to be an `int`?

josephglanville: I thought so too but uint64 works too and is slightly nicer in that it it alleviates the need to further validate for positive input.

titanous:
@josephglanville https://github.com/flynn/flynn/blob/master/test/cluster/cluster.go#L487

host/host.go
```diff
@@ -168,6 +169,12 @@ func runDaemon(args *docopt.Args) {
 		hostID = strings.Replace(hostname, "-", "", -1)
 	}
 
+	var maxJobConcurrency uint64
+	maxJobConcurrency = 4
```
titanous:
```go
var maxJobConcurrency uint64 = 4
```

titanous:
LGTM

lmars:
LGTM

PR #2537, opened 2016-03-07
worker: Watch job events per release when deploying

### LINKED ISSUE #2536: worker: Job events for multiple releases handled incorrectly during deploy
When the worker initiates a scaling event during a deployment and then waits for the appropriate job events, it waits for events for the release being scaled and discards events for any other releases (see [here](https://github.com/flynn/flynn/blob/master/controller/worker/deployment/job.go#L227-L229) and [here](https://github.com/flynn/flynn/blob/master/controller/worker/deployment/job.go#L250-L252)).

This is particularly problematic when deploying the controller as the worker stops itself, and then if the new worker resumes the deploy before the old worker dies, the first thing the new worker does is start another new worker, so may then receive the event for the old worker whilst it is waiting for the new one to come online, leading to an eventual timeout waiting for the down event (as it has already been seen and dropped).

A potential fix is to have a buffered channel for both the old and new release, and then modify [`waitForJobEvents`](https://github.com/flynn/flynn/blob/master/controller/worker/deployment/job.go#L190) to only receive on the channel for the release it cares about.

/cc @titanous

lmars:
Fixes #2536.

### DIFF
--- a/controller/worker/deployment/context.go
+++ b/controller/worker/deployment/context.go
@@ -9,7 +9,6 @@ import (
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
 	"github.com/flynn/flynn/controller/worker/types"
-	"github.com/flynn/flynn/discoverd/client"
 	"github.com/flynn/flynn/pkg/attempt"
 	"github.com/flynn/flynn/pkg/postgres"
 )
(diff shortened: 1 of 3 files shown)


josephglanville:
LGTM, though I do wonder if it conflicts with the changes I have on the `mariadb-integration` branch.

lmars:
@josephglanville I'm happy to help with the rebase if necessary, let me know.

PR #2542, opened 2016-03-08
scheduler: Ensure omni jobs do not become unbalanced

### LINKED ISSUE #2511: scheduler: Don't allow omni jobs to become unbalanced
The scheduler should check that omni jobs are correctly distributed across nodes in the cluster, rather than just checking that the total number of jobs running is correct (which is [what it does for non-omni jobs](https://github.com/flynn/flynn/blob/master/controller/scheduler/scheduler.go#L598-L613)).

### LINKED ISSUE #2521: Unbalanced routers after bootstrap
This is the second time we've seen two routers on one node in a three node cluster, and zero on another.

https://gist.github.com/titanous/81ed9409d35e53f2d8ff
https://gist.github.com/460bf195cc5ab5ff097f
https://gist.github.com/a3d30cc5cce5dc21b320
https://gist.github.com/90450bc2d70a373b1ab2

lmars:
Fixes #2511
Fixes #2521

### DIFF
--- a/controller/scheduler/scheduler.go
+++ b/controller/scheduler/scheduler.go
@@ -95,6 +95,10 @@ type Scheduler struct {
 	// their tags not matching any hosts, and is used to try and place the
 	// jobs when host tags change
 	pendingTagJobs map[string]*Job
+
+	// pause and resume are used by tests to control the main loop
+	pause  chan struct{}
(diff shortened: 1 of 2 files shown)


titanous:
LGTM

PR #2551, opened 2016-03-10
scheduler: Don't try to start jobs which have already started

### LINKED ISSUE #2545: Scheduler tries to re-create failed job with same job ID
- Scheduler calls `AddJob` against host.
- `AddJob` takes longer than the scheduler timeout, and the job ends up being created
- The scheduler, thinking the job create failed, tries to place the exact same job with the same ID on the same host. This fails, but the scheduler keeps on trying.

When a `AddJob` call fails, a new job ID should be generated and we should probably re-run the placement logic. Also, we should confirm that flynn-host does not allow creation of a job with an existing ID.

lmars:
This updates the scheduler to stop trying to add a job when it knows it has already started (regardless of whether AddJob continues to fail).

The host has also been updated to reject duplicate jobs, meaning the scheduler can continue to try starting a job with the same ID until it actually starts without leading to duplicate jobs.

I've also thrown in an update to the receiver to add `type=slugbuilder` to slugbuilder jobs as the scheduler now handles this correctly.

Closes #2545.

### DIFF
--- a/controller/scheduler/scheduler.go
+++ b/controller/scheduler/scheduler.go
@@ -35,7 +35,7 @@ const (
 var (
 	ErrNotLeader        = errors.New("scheduler is not the leader")
 	ErrNoHosts          = errors.New("no hosts found")
-	ErrJobStopped       = errors.New("job has been marked as stopped")
+	ErrJobNotPending    = errors.New("job is no longer pending")
 	ErrNoHostsMatchTags = errors.New("no hosts found matching job tags")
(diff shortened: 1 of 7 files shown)


titanous:
LGTM

lmars:
P.S. although I haven't added explicit tests, I tested this manually by adding random sleeps to the host's HTTP API and verifying the scheduler exhibits the correct behaviour.

PR #2552, opened 2016-03-10
pkg/httpclient: Don't use a timeout in ResumingStream

### LINKED ISSUE #2546: flynn-host scheduler stream timeout?
While looking over some logs, I noticed that the scheduler appears to be reconnecting to the job stream from `flynn-host` every 5s. This is probably due to a request timeout.

lmars:
Fixes #2546.


titanous:
LGTM

PR #2557, opened 2016-03-11
Fix containerinit registeration on error

titanous:
This fixes a bug in the discoverd registrar where the first registration fails and then there is no subsequent tries because a struct field (`h.hb`) that is supposed to be nil is not nil.


josephglanville:
LGTM, great find(s).

PR #2560, opened 2016-03-12
appliance/mariadb: Use more restrictive quoting

josephglanville:
MySQL can do some strange things if you don't use their non-standard backtick quoting for database and field names.


titanous:
LGTM

PR #2561, opened 2016-03-12
MongoDB Appliance

josephglanville:
Opening PR to facilitate easier collaboration.

### DIFF
--- a/appliance/mongodb/cmd.go
+++ b/appliance/mongodb/cmd.go
@@ -1,0 +1,62 @@
+package mongodb
+
+import (
+	"os"
+	"os/exec"
+	"sync/atomic"
+	"syscall"
+)
+
+// Cmd wraps exec.Cmd and provides helpers for checking for expected exits.
+type Cmd struct {
+	*exec.Cmd
+	stoppingValue atomic.Value // bool
+	stopped       chan struct{}
+	err           error
+}
+
+// NewCmd returns a new instance of Cmd that wraps cmd.
+func NewCmd(cmd *exec.Cmd) *Cmd {
+	c := &Cmd{
+		Cmd:     cmd,
+		stopped: make(chan struct{}, 1),
+	}
+	c.stoppingValue.Store(false)
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+	return c
+}
+
+// Start executes the command.
+func (cmd *Cmd) Start() error {
+	if err := cmd.Cmd.Start(); err != nil {
+		return err
+	}
+	go cmd.monitor()
+	return nil
+}
+
+// Stop marks the command as expecting an exit and stops the underlying command.
+func (cmd *Cmd) Stop() error {
+	cmd.stoppingValue.Store(true)
+	if err := cmd.Process.Signal(syscall.SIGKILL); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Stopped returns a channel that is closed when the process exits
+func (cmd *Cmd) Stopped() <-chan struct{} { return cmd.stopped }
+
+// Err returns an error if cmd stopped unexpectedly.
+// Must wait for the Stopped channel to return first.
+func (cmd *Cmd) Err() error { return cmd.err }
+
+// monitor checks for process exit and returns
+func (cmd *Cmd) monitor() {
+	err := cmd.Wait()
+	if !cmd.stoppingValue.Load().(bool) {
+		cmd.err = err
+	}
+	close(cmd.stopped)
+}
--- a/appliance/mongodb/cmd/flynn-mongodb-api/main.go
+++ b/appliance/mongodb/cmd/flynn-mongodb-api/main.go
@@ -1,0 +1,148 @@
+package main
+
+import (
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"strings"
+
+	"github.com/flynn/flynn/discoverd/client"
+	"github.com/flynn/flynn/pkg/httphelper"
+	"github.com/flynn/flynn/pkg/random"
+	"github.com/flynn/flynn/pkg/resource"
+	"github.com/flynn/flynn/pkg/shutdown"
+	"github.com/julienschmidt/httprouter"
+	"gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+)
+
+var serviceName = os.Getenv("FLYNN_MONGO")
+var serviceHost string
+
+func init() {
+	if serviceName == "" {
+		serviceName = "mongodb"
+	}
+	serviceHost = fmt.Sprintf("leader.%s.discoverd", serviceName)
+}
+
+func main() {
+	defer shutdown.Exit()
+
+	api := &API{}
+
+	router := httprouter.New()
+	router.POST("/databases", api.createDatabase)
+	router.DELETE("/databases", api.dropDatabase)
+	router.GET("/ping", api.ping)
+
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = "3000"
+	}
+	addr := ":" + port
+
+	hb, err := discoverd.AddServiceAndRegister(serviceName+"-api", addr)
+	if err != nil {
+		shutdown.Fatal(err)
+	}
+	shutdown.BeforeExit(func() { hb.Close() })
+
+	handler := httphelper.ContextInjector(serviceName+"-api", httphelper.NewRequestLogger(router))
+	shutdown.Fatal(http.ListenAndServe(addr, handler))
+}
+
+type API struct{}
+
+func (a *API) createDatabase(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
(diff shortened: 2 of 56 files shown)


josephglanville:
@benbjohnson I have the replica set configuration stuff mostly worked out now but having issues with auth etc. Once I have that resolved things should be looking better.

josephglanville:
Persisting the full peer toplogy through Reconfigure has caused some breakage in the Sirenia unit tests too, will sort that out tomorrow.

josephglanville:
@benbjohnson most stuff is working, adding/removing nodes. However promotions are failing because the sync is blocking trying to get it's own oplog position. Probably another mgo mode issue. Probably need to special case doing the local oplog with Direct and the appropriate mgo mode, should be able to work that out from the role (p.config().Role == state.RolePrimary etc).

I'm not super happy with the hacks that needed to go into the state machine, I am going to explore other options once we get it working. Possibly copying the Sirenia state machine and gutting it and trying to work out what can be shared.

josephglanville:
@benbjohnson pushed a bunch of fixes to the tests though we still need to sort out the 4 node and removal tests. If you have spare bandwith after the other stuff if you could take a look that would be great otherwise I will get to them in a bit.

benbjohnson:
Sure, I can take a look at those after the backup/restore.

josephglanville:
@benbjohnson unit tests are fixed minus the assertDownstream stuff for now. syncFrom is proving to be somewhat finicky. I think MongoDB only treats it as advice so I am probably going to switch it to just checking that the appropriate replica set configuration can be recieved from each node.

benbjohnson:
@josephglanville The `mongodump` & `mongorestore` tools are also finicky. They provide an `--archive` flag for dumping out as a single file which is great but it is impossible to dump from a database "X" and restore to a database "Y" with `--archive`. If you dump from "X" then you always have to restore to "X".

I reimplemented archive using `tar` and some bash scripts. Seems to be working.

josephglanville:
Heh, it wouldn't be mongo if it wasn't finicky.

I'm trying to work out why deployments in the Sirenia integration tests are failing now.

appliance/mongodb/dump.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
josephglanville: Use -e

appliance/mongodb/restore.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
josephglanville: Ditto.

josephglanville:
@benbjohnson got deploys working and the integration tests pass 99% of the time. Though I did see them fail once or twice with the test spinning on waiting for read/write. When I checked the response didn't contain read_write meaning Info() bombed out before reporting back from isReadWrite().
Added some debug logging so if we see it again we can crush it.

josephglanville:
There is a data race somewhere in the MongoDB process wrapper somewhere. Trying to track it down now.

josephglanville:
@titanous need a hand with this data race, been staring at it for quite a while. I think it's some nuanced Go thing that I don't know yet as I have double and triple checked shared values that I thought would require locking.

The data race apparently occurs in `Process#waitForSyncInner` when it does a read of a value written by `Process#reconfigure` try as I might I can't figure out which one.

appliance/mongodb/cmd.go
```diff
@@ -0,0 +1,62 @@
...
+func (cmd *Cmd) Stop() error {
+	cmd.stoppingValue.Store(true)
+	if err := cmd.Process.Signal(syscall.SIGKILL); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Stopped returns a channel that returns an error if stopped unsuccessfully.
+func (cmd *Cmd) Stopped() <-chan struct{} { return cmd.stopped }
```
titanous: The comment doesn't appear to match the method.

appliance/mongodb/cmd/flynn-mongodb-api/main.go
```diff
@@ -0,0 +1,150 @@
...
+		{"pwd", password},
+		{"roles", []bson.M{
+			{"role": "dbOwner", "db": database},
+		}},
+	}, &userResponse); err != nil {
+		httphelper.Error(w, err)
+		return
+	}
+
+	// Assign any required roles
```
titanous: This comment doesn't appear to match the code.

appliance/mongodb/cmd/flynn-mongodb/main.go
```diff
@@ -0,0 +1,101 @@
...
+	handler := mongodb.NewHandler()
+	handler.Process = process
+	handler.Peer = peer
+	handler.Heartbeater = hb
+	handler.Logger = log.New("component", "http")
+
+	shutdown.Fatal(http.ListenAndServe(":"+httpPort, handler))
+}
+
+func ip2id(ip net.IP) uint32 {
```
titanous: `2` should be spelled out as `To`

appliance/mongodb/dump.sh
```diff
@@ -0,0 +1,12 @@
...
+# Passthrough arguments and dump to temporary path.
+TMPFILE=`mktemp -d /tmp/dump.XXXXXXXXXX`
+/usr/bin/mongodump $@ -o $TMPFILE
+
+# Archive dump directory, gzip, & write to STDOUT.
+cd $TMPFILE
+tar -cf - *
+
+# Remove dump directory.
+rm -rf $TMPFILE
```
titanous: This should be in a trap so that it runs even if one of the other steps fails.

appliance/mongodb/cmd/flynn-mongodb/main.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	id := string(idBytes)
+	if len(id) == 0 {
+		id = random.UUID()
+		if err := ioutil.WriteFile(idFile, []byte(id), 0644); err != nil {
+			shutdown.Fatalf("error writing instance ID: %s", err)
+		}
+	}
+
+	keyFile := filepath.Join(dataDir, "Keyfile")
```
titanous: Maybe just call this file `password`?

josephglanville: The `Keyfile` thing is MongoDB convention. It may be possible to change the name with a flag but I figured it was easier to leave as is.

titanous: Ah, that's fine.

appliance/mongodb/process.go
```diff
@@ -0,0 +1,952 @@
...
+	if err != nil {
+		return err
+	}
+	defer session.Close()
+
+	if session.Run(bson.D{{"replSetReconfig", config}, {"force", true}}, nil); err != nil {
+		return err
+	}
+	// XXX(jpg): Prevent mongodb implosion if a reconfigure comes too soon after this one
+	time.Sleep(5 * time.Second)
```
titanous: It'd be great to have an upstream JIRA ticket to reference here.

appliance/mongodb/process_test.go
```diff
@@ -0,0 +1,535 @@
...
+	topology := &state.State{Singleton: true, Primary: instance(p)}
+	err = p.Reconfigure(&state.Config{Role: state.RolePrimary, State: topology})
+	c.Assert(err, IsNil)
+
+	err = p.Start()
+	c.Assert(err, IsNil)
+	defer p.Stop()
+
+	session := connect(c, p)
+	/*_, err = session.Exec("CREATE DATABASE test")*/
```
titanous: Should this be deleted?

appliance/mongodb/process_test.go
```diff
@@ -0,0 +1,535 @@
...
+	keyFile = filepath.Join(p.DataDir, "Keyfile")
+	err = ioutil.WriteFile(keyFile, []byte("password"), 0600)
+	c.Assert(err, IsNil)
+	err = p.Reconfigure(&state.Config{Role: state.RolePrimary, State: topology})
+	c.Assert(err, IsNil)
+	c.Assert(p.Start(), IsNil)
+	defer p.Stop()
+
+	session = connect(c, p)
+	/*_, err = session.Exec("CREATE DATABASE foo")*/
```
titanous: Ditto.

pkg/sirenia/state/state.go
```diff
@@ -56,6 +56,27 @@ func (s *State) Clone() *State {
...
 
+func (x *State) Equal(y *State) bool {
+	if x == nil || y == nil {
+		return x == y
+	}
+	if len(x.Async) != len(y.Async) || len(x.Deposed) != len(y.Deposed) {
+		return false
+	}
+	// TODO(jpg): Probably nicer way to do this in Go.
+	for i := 0; i < len(x.Async); i++ {
```
titanous:
```go
for i := range x.Async {
```

titanous:
LGTM pending minor comments! :tada:

PR #2562, opened 2016-03-12
Pre-Go1.6 fixes

titanous:
These are all of the patches for races and other brokenness found when running tests on Go 1.5/1.6. They should all be backwards compatible and fix real issues, so this PR is to get the fixes into master without being blocked by any straggling issues.


josephglanville:
LGTM

PR #2566, opened 2016-03-13
dashboard: Fix scaling app processes when no formation

### LINKED ISSUE #2410: Deploying with no web process makes dashboard break
When I don't have a web process in my procfile, I can't scale my worker process.

jvatic:
Closes #2410


titanous:
LGTM

PR #2569, opened 2016-03-13
host/cli: Send stderr from log stream to stdout by default

### LINKED ISSUE #2539: flynn-host log emits log lines to stderr
We should only emit actual errors to stderr by default unless a flag is specified (similar to the flag in `flynn log`).

titanous:
Closes #2539


josephglanville:
LGTM

lmars:
LGTM

PR #2588, opened 2016-03-16
Installer restore backup

### LINKED ISSUE #2098: installer: Restore from backup
Allow restoring from backup when launching a cluster.

There should be a file input for the user to pass in the backup file. The installer will upload it to the server and run bootstrap with it (`flynn-host bootstrap --from-backup`).

jvatic:
- [x] Installer needs to find controller key, ca cert, tls pin, and login token
- [x] Extract old domain from flynn.json in backup
- [x] Use old domain when bootstrapping
- [x] Run domain migration after restore is complete
- [x] A progress bar should be displayed for the upload

Closes #2098


jvatic:
This is ready for review

installer/cluster.go
```diff
@@ -509,10 +724,112 @@ func (c *BaseCluster) bootstrapTarget(t *TargetServer) error {
...
+	if err != nil {
+		return nil, err
+	}
+	stdout, err := sess.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+	sess.Stderr = os.Stderr
+
+	if err := sess.Start(fmt.Sprintf("flynn-host inspect $(flynn-host ps | grep %s | head -1 | sed -e 's/ .*//' | tr -d '\r\n')", appName)); err != nil {
```
jvatic: Should be able to use curl here instead and avoid parsing non-json output

jvatic:
ping @titanous for review

installer/app/src/views/backup-selector.js.jsx
```diff
@@ -0,0 +1,46 @@
...
+	getInitialState: function () {
+		return {
+			errorMsg: null
+		};
+	},
+
+	__handleFileChange: function (file) {
+		if ( !file.type.match(/\btar\b/) ) {
+			this.setState({
+				errorMsg: 'Must be a tar archive'
```
titanous: We don't enforce a file extension when exporting the backup.

installer/cluster.go
```diff
@@ -46,6 +53,70 @@ func (c *BaseCluster) HandleAuthenticationFailure(cluster Cluster, err error) bo
...
+		if percent > prevPercent {
+			prevPercent = percent
+			c.SendProgress(&ProgressEvent{
+				ID:          "upload-backup",
+				Description: fmt.Sprintf("Uploading... [%d%%]", percent),
+				Percent:     percent,
+			})
+		}
+	}}
+	tee := io.TeeReader(backup, tracker)
```
titanous: Could use a `MultiWriter` (tracker+pipe) here to avoid two `TeeWriter`s.

installer/cluster.go
```diff
@@ -46,6 +53,70 @@ func (c *BaseCluster) HandleAuthenticationFailure(cluster Cluster, err error) bo
...
 
+type receiveBackupWriter struct {
+	Callback func(read int)
+	read     int
+}
+
+func (w *receiveBackupWriter) Write(p []byte) (int, error) {
+	n := len(p)
+	w.read = w.read + n
+	w.Callback(w.read)
```
titanous: I think this is slightly less work to use as a `Reader` instead of a `Writer`. Proxy the reads and do the tracking there.

installer/cluster.go
```diff
@@ -46,6 +53,70 @@ func (c *BaseCluster) HandleAuthenticationFailure(cluster Cluster, err error) bo
...
+				continue
+			}
+			if err := json.NewDecoder(tr).Decode(&data); err != nil {
+				c.SendLog(fmt.Sprintf("Error decoding backup data: %s", err))
+				return
+			}
+			c.oldDomain = data.Controller.Release.Env["DEFAULT_ROUTE_DOMAIN"]
+			c.ControllerKey = data.Controller.Release.Env["AUTH_KEY"]
+		}
+	}()
```
titanous: Should this log an error if the necessary data isn't provided?

installer/cluster.go
```diff
@@ -370,6 +441,113 @@ func (c *BaseCluster) sshConfig() (*ssh.ClientConfig, error) {
...
+	}
+	stdin, err := sess.StdinPipe()
+	if err != nil {
+		return err
+	}
+	sess.Stdout = os.Stdout
+	sess.Stderr = os.Stderr
+
+	c.backupPath = "/tmp/flynn-backup.tar"
+	if err := sess.Start(fmt.Sprintf("tee > %s", c.backupPath)); err != nil {
```
titanous: Why `tee` instead of `cat`? Won't that end up downloading it back down from the server?

installer/cluster.go
```diff
@@ -509,10 +717,87 @@ func (c *BaseCluster) bootstrapTarget(t *TargetServer) error {
...
+	sess, err := t.SSHClient.NewSession()
+	if err != nil {
+		return nil, err
+	}
+	stdout, err := sess.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := sess.Start(fmt.Sprintf("curl -u ':%s' -H 'Host: controller.%s' localhost/apps/%s/release", c.ControllerKey, c.oldDomain, appName)); err != nil {
```
titanous: Instead, you could the SSH client as a [`net.Dialer`](https://godoc.org/golang.org/x/crypto/ssh#Client.Dial) with the controller client.

titanous: Or can you use the approach below used for the domain migration?

installer/cluster.go
jvatic: Not sure why, ends up doing the same thing as `cat`, will change to use `cat`.

installer/cluster.go
jvatic: Interesting. I think the reason I didn't use the below approach is due to not having the pin at this point.

installer/cluster.go
```diff
@@ -509,10 +717,87 @@ func (c *BaseCluster) bootstrapTarget(t *TargetServer) error {
...
+	if err != nil {
+		return nil, err
+	}
+	return release.Env, nil
+}
+
+func (c *BaseCluster) migrateDomain(dm *ct.DomainMigration, t *TargetServer) (*ct.DomainMigration, error) {
+	c.SendLog(fmt.Sprintf("Migrating domain (%s to %s)", dm.OldDomain, dm.Domain))
+
+	client, err := cc.NewClientWithConfig(fmt.Sprintf("http://%s", t.IP), c.ControllerKey, cc.Config{
```
jvatic: Should probably be using https here.

installer/cluster.go
titanous: You can create the client without the pin.

installer/installer.go
```diff
@@ -248,6 +248,13 @@ func (i *Installer) LaunchCluster(c Cluster) error {
 		Cluster:   base,
 		ClusterID: base.ID,
 	})
+	if base.IsRestoringBackup() {
+		base.SendProgress(&ProgressEvent{
+			ID:          "upload-backup",
+			Description: "Upload waiting...",
```
titanous: `Upload pending cluster start...`

jvatic:
@titanous ready for another review pass

installer/cluster.go
```diff
@@ -487,14 +631,49 @@ func (c *BaseCluster) bootstrapTarget(t *TargetServer) error {
...
+		if err != nil {
+			return err
+		}
+
+		b, _ := pem.Decode([]byte(routerEnv["TLSCERT"]))
+		if b == nil {
+			return fmt.Errorf("Error parsing TLSCERT")
+		}
+		sha := sha256.Sum256(b.Bytes)
+		c.ControllerPin = base64.StdEncoding.EncodeToString(sha[:])
```
jvatic: This isn't being used before the domain migration anymore, so no need to fetch the router env or calculate the pin.

installer/cluster.go
```diff
@@ -370,6 +383,144 @@ func (c *BaseCluster) sshConfig() (*ssh.ClientConfig, error) {
...
+
+	c.backupPath = "/tmp/flynn-backup.tar"
+	if err := sess.Start(fmt.Sprintf("cat > %s", c.backupPath)); err != nil {
+		return err
+	}
+
+	var uploadErr error
+	go func() {
+		if _, err := io.Copy(stdin, c.backup); err != nil {
+			stdin.Close()
```
titanous: This close is redundant.

installer/cluster.go
```diff
@@ -509,12 +678,70 @@ func (c *BaseCluster) bootstrapTarget(t *TargetServer) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("Error opening domain migration event stream: %s", err)
+	}
+	defer stream.Close()
+
+	if err := client.PutDomain(dm); err != nil {
+		return nil, fmt.Errorf("Error starting domain migration: %s", err)
+	}
+
+	timeout := time.After(2 * time.Minute)
```
titanous: Bump this up to five minutes.

titanous:
Make sure you run a restore from backup with the race detector on, I see a few spots that might be tricky but I didn't fully evaluate the dataflow.

installer/cluster_backup_receiver.go
```diff
@@ -0,0 +1,132 @@
...
+	for _, ch := range cbr.subs {
+		ch <- err
+	}
+	cbr.subs = make([]chan error, 0)
+	cbr.subsMux.Unlock()
+}
+
+func (cbr *ClusterBackupReceiver) Wait() error {
+	if cbr.err != nil {
+		return cbr.err
```
titanous: This needs to be wrapped in a the mutex.

installer/cluster_backup_receiver.go
```diff
@@ -0,0 +1,132 @@
...
+			break
+		}
+		io.Copy(ioutil.Discard, r)
+	}()
+	return cbr
+}
+
+func (cbr *ClusterBackupReceiver) Read(p []byte) (int, error) {
+	if cbr.err != nil {
+		return 0, cbr.err
```
titanous: This needs to be wrapped in the mutex (it is not sufficient to just protect the write, all reads must be protected with the same mutex)

jvatic:
@titanous Comments addressed, though there's still a few race conditions not related to this PR (on `BaseCluster.State` for example, which is proving a bit tricky to resolve without creating effective deadlock cases and I haven't seen any problems created by it in practice).

titanous:
Please file issues for any other races found.

titanous:
Can you try updating ql with the latest version of godep? I think there are some files that might not get pulled in.

jvatic:
Updated ql via godep

titanous:
LGTM

jvatic:
Squashed commits

PR #2589, opened 2016-03-16
Fix host-wait timeout

titanous:
This was previously hard-coded. Also, bump the default to a less aggressive two minutes.


josephglanville:
LGTM

PR #2590, opened 2016-03-17
Add CPU limits and resource partitions

titanous:
### CPU limits

CPU limits are implemented as a resource denominated in “milliCPUs”, a concept from Google. When doing placement, a request for 1000 milliCPUs will be equivalent to a single CPU core.

A limit of 1000 milliCPUs results in the Linux default of 1024 CPU shares, and the request portion of the resource spec can be used for scheduling in the future.

CPU shares are relative, the more shares a process has, the higher priority it is. When a host is under load, a job with 2000 milliCPUs will get twice the CPU time as a job with the default of 1000.

Limits can be specified on a per-process basis with:

```text
flynn limit set web cpu=1500
```

### Resource Partitions

Resource partitions complement CPU limits in order to isolate system apps from user apps. A top-level set of cgroups is added that all jobs are nested under. Currently these cgroups just provide CPU share allocations, but in the future they can be used for any other cgroup control. The partitions are only throttled when all partitions are under equal load, full bursting is allowed, unused shares are given to the other partitions on a proportional basis. With that in mind, the default allocations are:

- `system`, all Flynn-managed system apps, with 25% of available CPU time
- `user`, all user app processes, with 50% of available CPU time
- `background`, slugbuilder builds, with 25% of available CPU time

Allocations can be modified by passing the `--partitions` flag to `flynn-host daemon` like this:

```text
flynn-host daemon --partitions "system=cpu_shares:4096 background=cpu_shares:4096 user=cpu_shares:8192"
```

Refs #1496

### DIFF
--- a/cli/limit.go
+++ b/cli/limit.go
@@ -31,22 +31,22 @@ Commands:
 Examples:
 
 	$ flynn limit
-	web:     max_fd=10000  memory=1GB
-	worker:  max_fd=10000  memory=1GB
+	web:     cpu=1000  max_fd=10000  memory=1GB
+	worker:  cpu=1000  max_fd=10000  memory=1GB
 
-	$ flynn limit set web memory=512MB max_fd=12000
+	$ flynn limit set web memory=512MB max_fd=12000 cpu=500
 	Created release 5058ae7964f74c399a240bdd6e7d1bcb
 
 	$ flynn limit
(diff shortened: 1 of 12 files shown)


josephglanville:
LGTM

lmars:
Late to the party, but LGTM :smile:

PR #2592, opened 2016-03-18
host/cli: Use current MariaDB artifact URI when restoring from backup

titanous:
If this is not done, restores to a newer version of Flynn will fail with:

    run-app mariadb error: bootstrap: host error while launching job: "registry: unexpected status 403"'

/cc @josephglanville


lmars:
LGTM

PR #2596, opened 2016-03-19
host: Add 64MB /dev/shm to each container

### LINKED ISSUE #2570: Missing /dev/shm on base image
I'm encountering errors when starting up celery worker processed on flynn. It seems that celery's default worker pool makes use of shared memory, which I imagine Python expects to find in `/dev/shm`.

While a number of fixes [seem to exist](https://www.google.com/search?q=missing+%2Fdev%2Fshm), namely:

```bash
sudo ln -s /run/shm /dev/shm  # untested
```

However, I cannot gain permission to run this as part of the deployment process.

Can/should the base image be updated to include `/dev/shm`?

titanous:
Closes #2570

/cc @josephglanville

### DIFF
--- a/host/libvirt/types.go
+++ b/host/libvirt/types.go
@@ -80,8 +80,9 @@ type FSDriver struct {
 }
 
 type FSRef struct {
-	Dir  string `xml:"dir,attr,omitempty"`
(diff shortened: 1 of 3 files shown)


josephglanville:
LGTM

Issue #2504, opened 2016-02-29
Corrupt cluster backup


titanous:
We've received reports that some backups taken with `flynn cluster backup` were corrupt. The command exited successfully, but a restore failed with:

```text
ERROR:  relation "artifacts" does not exist
LINE 1: UPDATE artifacts SET uri = '\''https://dl.flynn.io/tuf?name=fly...
```

We should carefully review the entire backup code path for error checks to ensure that if a command fails we don't return success.

/cc @temujin9

titanous:
@temujin9 I have been unable to replicate this or find any code that explains it. Would you please look at one or two of the corrupt backups to confirm that there is no proprietary/sensitive data and then email me a link?

temujin9:
I can go digging, but the proprietary data thing might take a little while.

I can say that the cause of this was almost certainly the disk /tmp was on filling up during the postgres dump portion of the backup. Moving the file for that onto a larger partition has prevented recurrences.

titanous:
Okay, I will try to replicate again with a full tmpfs. Last time I tried I got the correct error and exit code:

```text
error dumping database: error running export: write /home/vagrant/go/src/github.com/flynn/flynn/foo/postgres.sql.gz699451114: no space left on device
```

titanous:
Just knowing the file size of the postgres dump file in the backup would be a start, I don't need to see the actual file.

temujin9:
So, on one of our clusters, the dump size went from 1.9GB (probably good) to 12.5KB (probably broken) from one hour's backup to the next.

titanous:
If you run `tar xvf backup.dump` against a broken backup, is there just a `flynn.json` file?

temujin9:
`tar` doesn't recognize my most recent instance of this issue as a valid tar file. `file` similarly classifies it as 'binary data' instead of anything more useful.

PR #2598, opened 2016-03-19
Fix cluster backup bugs

titanous:
This fixes an issue with corrupt backups (#2504) and ensures that we don't fail if the user intentionally disabled MariaDB.


josephglanville:
LGTM

PR #2602, opened 2016-03-19
updater: Fix updating slug environment variables

### LINKED ISSUE #2577: `flynn-host update` replaces slugrunner image with slugbuilder
@adamcharnock has a cluster that has been updated with `flynn-host update` that is using the slugbuilder image where the slugrunner image should be. We should audit the updater to figure out how this can happen. [This code](https://github.com/flynn/flynn/blob/eeb8d80c17ff255329d68210493a67734eec69f1/updater/updater.go#L224-L235) looks suspicious.

lmars:
Fixes #2577.


titanous:
LGTM

PR #2607, opened 2016-03-20
test: Update the CI build log page

### LINKED ISSUE #2601: CI log viewer should only show build log
We have so many logs that the browser gets overloaded. Instead we should provide a command to call that will download all of the host/job logs for the CI run.

lmars:
Summary of changes:

* Only serve the main build log when visiting `/builds/:id`, and it is served either as a HTML / SSE stream, or a plain text stream (if using `curl` for example). The log comes from either S3 or on disk if still running
* Add a download link to download the full log as a plain text file, the content being a valid multipart message (so can be parsed and individual files extracted with something like `munpack(1)`, but I just use `less(1)` :smile:)
* Make the build log look more like a terminal screen (this included vendoring [this JS lib](https://github.com/drudru/ansi_up) which doesn't have a `LICENSE` file but does include a License section in the vendored Readme.md, I assume that is ok?)

Here is how the build log page looks now:

---

screen shot 2016-03-20 at 22 28 49

---

Closes #2601.


titanous:
Awesome! LGTM.

PR #2614, opened 2016-03-21
controller/worker/deployment: Use configured deploy timeout when waiting for events

### LINKED ISSUE #2604: Timeouts for environment variables and releases
With the version v20160314.0 on a freshly setup cluster I still get timeouts. For example, (re-)deployment of env vars:
> timed out waiting for job events: map[web:map[up:1]]

titanous:
This hard-coded timeout is too tight for many larger apps.

Closes #2604.

@lmars I think it makes more sense to use this timeout as the amount of time to wait for a single process to start. We should probably just completely drop some of the other timeouts related to watching the deployment stream.


josephglanville:
LGTM

Issue #2610, opened 2016-03-21
catastrophic cluster failure


temujin9:
Not certain the original cause, but initial complaint from monitoring was loss of a scheduler. https://gist.github.com/dbf8464b1c8661edc735 https://gist.github.com/b2a3888d77da2d1b1a44 https://gist.github.com/a82293750eefefb1b537

titanous:
Discoverd remediation is #2612.

@lmars will follow-up with analysis of the scheduler bug that was triggered by discoverd registrations flapping due to very slow disk IO.

lmars:
During the discoverd flaps, the scheduler decided that the discoverd formation (along with some other formations) was in the wrong state:


From https://gist.github.com/dbf8464b1c8661edc735:

```
t=2016-03-21T02:42:32+0000 lvl=info msg="expected different from actual" component=scheduler fn=formationDiff app.id=275d0efd-6b77-4d76-95e7-bc59b814466f release.id=69bab3d7-636c-4352-b85f-c8a6e539fb6b expected=map[app:0] actual=map[app:3] diff=map[app:-3]
t=2016-03-21T02:42:32+0000 lvl=info msg="formation in incorrect state" component=scheduler fn=handleFormationDiff app.id=275d0efd-6b77-4d76-95e7-bc59b814466f release.id=69bab3d7-636c-4352-b85f-c8a6e539fb6b diff=map[app:-3]
t=2016-03-21T02:42:32+0000 lvl=info msg="stopping 3 app jobs" component=scheduler fn=handleFormationDiff app.id=275d0efd-6b77-4d76-95e7-bc59b814466f release.id=69bab3d7-636c-4352-b85f-c8a6e539fb6b
```

I haven't got to the bottom of why this was the case, but I am thinking the scheduler should do an extra check to make sure there is another active formation when scaling down a formation for a system app.

@titanous WDYT?

titanous:
Yeah, that makes sense but we'll need to mark critical and non-critical apps for that purpose.

titanous:
The discoverd flaps were fixed in #2618.

temujin9:
Okay, I think this is now addressed. Closing it; I'll reopen if I see a recurrence.

PR #2615, opened 2016-03-21
scheduler: Prevent completely scaling down critical apps

### LINKED ISSUE #2613: scheduler: Don't completely scale down critical system apps
In #2610, the scheduler decided the discoverd formation should be scaled down, but there was no other active discoverd formation, so this left the cluster in an unrecoverable state.

We should mark certain system apps as critical (at least discoverd + scheduler), and only scale those down if there is another active formation (i.e. if they are being deployed).

lmars:
This adds a sanity check to make sure we do not scale down "critical" apps such as discoverd, because if we do then clusters break hard.

We should not actually ever be getting these critical apps being scaled down like this, but it happened in #2610, and I suspect it was the controller not handling an error and returning empty formations rather than some agent actively scaling things down.

I have also updated the scheduler tests to use log messages for synchronisation (rather than the make shift events we had previously) which feels a little cleaner.

Fixes #2613.

### DIFF
--- a/controller/scheduler/formation.go
+++ b/controller/scheduler/formation.go
@@ -21,18 +21,28 @@ func (fs Formations) Add(f *Formation) *Formation {
 
 type Processes map[string]int
 
-func (p Processes) Equals(other Processes) bool {
+func (p Processes) Diff(other Processes) Processes {
+	diff := make(map[string]int)
 	for typ, count := range p {
-		if other[typ] != count {
-			return false
+		if typ == "" {
+			continue
 		}
+		diff[typ] = count - other[typ]
 	}
+
 	for typ, count := range other {
-		if p[typ] != count {
-			return false
+		if typ == "" {
+			continue
+		}
+		if _, ok := p[typ]; !ok {
+			diff[typ] = -count
 		}
 	}
-	return true
+	return diff
+}
+
+func (p Processes) Equals(other Processes) bool {
+	return p.Diff(other).IsEmpty()
 }
 
 func (p Processes) IsEmpty() bool {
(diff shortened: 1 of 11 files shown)


titanous:
We need a strategy for adding the new metadata to bootstraps from backups.

lmars:
@titanous added a migration to set the metadata: 7ade8a4

controller/scheduler/formation.go
```diff
@@ -44,6 +54,15 @@ func (p Processes) IsEmpty() bool {
 	return true
 }
 
+func (p Processes) IsScaleDown() bool {
+	for _, count := range p {
+		if count < 0 {
```
titanous: Should this check if the process is being scaled all the way to zero? There are reasons to scale down say the controller from 2 to 1 without a new release.

lmars: Yes makes sense, I have modified it to only prevent scaling a critical process down to zero: f79faa1

controller/schema.go
```diff
@@ -249,6 +249,28 @@ $$ LANGUAGE plpgsql`,
...
+			    SELECT * FROM jsonb_each(merge_data)
+			)
+			SELECT *
+			FROM jsonb_each(data)
+			WHERE key NOT IN (SELECT key FROM to_merge)
+			UNION ALL
+			SELECT * FROM to_merge
+		    ) t;
+		$$`,
+		`UPDATE apps SET meta = jsonb_merge(meta, '{"flynn-system-critical":"true"}') WHERE name IN ('discoverd', 'postgres', 'controller')`,
```
titanous: Should flannel be included here? The controller and postgres need it to communicate.

titanous:
LGTM pending comment.

PR #2618, opened 2016-03-21
discoverd/server: Fix regression sending heartbeats through raft

titanous:
b727a82 introduced a regression that sent all heartbeats through raft, instead of just new instances. Fix that regression while maintaining the correct mutex usage.


josephglanville:
LGTM, good find!

PR #2623, opened 2016-03-22
Godeps: Update digitalocean client

titanous:
The latest version of Godeps rewrote a few things.

Also, the deprecated code.google.com reference to oauth2 is replaced with the new one.


josephglanville:
LGTM

PR #2630, opened 2016-03-24
test/rootfs: Use upstart stop directly

josephglanville:


josephglanville:
@lmars fixes the rootfs build issue, still not sure why the service wrapper doesn't work but whatever

titanous:
LGTM

PR #2632, opened 2016-03-25
appliance/redis: Use automatic leader election for now

titanous:
We can change to manual later when the redis appliance is HA.


lmars:
@titanous looks like the Redis tests are failing?

josephglanville:
The Redis tests were explicitly checking for manual leader election, fixed that and rebased on master.

PR #2634, opened 2016-03-25
controller: Don't require auth for GET /ca-cert

jvatic:


titanous:
Maybe we should do this: If an auth header is provided for the endpoint, check it so that it still errors out when the auth token is invalid?

PR #2490, opened 2016-02-23
dashboard: Add UI for provisioning resources

jvatic:
screen shot 2016-03-12 at 06 36 54 am
screen shot 2016-03-12 at 06 38 24 am
screen shot 2016-03-12 at 06 38 29 am
screen shot 2016-03-12 at 06 38 40 am
screen shot 2016-03-12 at 06 38 47 am
screen shot 2016-03-12 at 06 39 53 am
screen shot 2016-03-12 at 06 40 55 am
screen shot 2016-03-12 at 06 42 53 am
screen shot 2016-03-12 at 06 42 59 am
screen shot 2016-03-12 at 06 43 07 am
screen shot 2016-03-12 at 06 46 32 am

Closes #2468 

screen shot 2016-03-12 at 06 33 01 am
screen shot 2016-03-12 at 06 33 56 am
screen shot 2016-03-12 at 06 36 40 am

Closes #2469 

*Updated Mar 12, 2016, 6:47 AM EST*

PR #2635, opened 2016-03-26
controller: Don't require auth for GET /ca-cert

jvatic:
Accidentally closed #2634 via #2490, reopening here with comment addressed.

\\cc @titanous


titanous:
LGTM

PR #2636, opened 2016-03-26
appliance/mariadb: do not startup by default

### LINKED ISSUE #2595: MariaDB should not start by default
We should start the cluster with it scaled down and start it when the first mysql resource request comes in.

benbjohnson:
## Overview

This pull request sets the number of `mariadb` processes to zero on start up and instead checks the formation when a mysql resource is added. The `mariadb-api` handles scaling the cluster on the first resource added.

Fixes #2595

Signed-off-by: Ben Johnson <benbjohnson@yahoo.com>

### DIFF
--- a/appliance/mariadb/cmd/flynn-mariadb-api/main.go
+++ b/appliance/mariadb/cmd/flynn-mariadb-api/main.go
@@ -2,20 +2,26 @@ package main
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
 	"strings"
+	"sync"
+	"time"
 
 	_ "github.com/flynn/flynn/Godeps/_workspace/src/github.com/go-sql-driver/mysql"
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/julienschmidt/httprouter"
 	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/net/context"
+	"github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
 	"github.com/flynn/flynn/appliance/mariadb"
+	"github.com/flynn/flynn/controller/client"
 	"github.com/flynn/flynn/discoverd/client"
 	"github.com/flynn/flynn/pkg/httphelper"
 	"github.com/flynn/flynn/pkg/random"
 	"github.com/flynn/flynn/pkg/resource"
 	"github.com/flynn/flynn/pkg/shutdown"
+	sirenia "github.com/flynn/flynn/pkg/sirenia/client"
 )
 
 var serviceName = os.Getenv("FLYNN_MYSQL")
@@ -31,14 +37,7 @@ func init() {
 func main() {
 	defer shutdown.Exit()
 
-	dsn := &mariadb.DSN{
-		Host:     serviceHost + ":3306",
-		User:     "flynn",
-		Password: os.Getenv("MYSQL_PWD"),
-		Database: "mysql",
-	}
-	db, err := sql.Open("mysql", dsn.String())
-	api := &API{db}
+	api := &API{}
 
 	router := httprouter.New()
 	router.POST("/databases", httphelper.WrapHandler(api.createDatabase))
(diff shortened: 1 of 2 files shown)


titanous:
I think whether or not `scaleUp` has run successfully can be cached for the life of the process. Also, it should be wrapped in a mutex so that concurrent requests don't trigger it multiple times and wait for the first call to finish.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -119,9 +134,148 @@ func (a *API) dropDatabase(ctx context.Context, w http.ResponseWriter, req *http
...
+	// Copy processes and increase database processes to 3.
+	processes := make(map[string]int, len(formation.Processes))
+	for k, v := range formation.Processes {
+		processes[k] = v
+	}
+	processes["mariadb"] = 3
+
+	// Otherwise begin scaling the database processes up.
+	logger.Info("watching job events", "app", app, "release_id", release.ID)
+	expected := client.ExpectedScalingEvents(formation.Processes, processes, release.Processes, 1)
```
titanous: Can we just wait for the discoverd event and then poll the status endpoint for read/write?

benbjohnson:
@titanous Do we need to be concerned about someone scaling down a mysql after the check has been performed once? Good point about the mutex though.

titanous:
@benbjohnson No, because that would have been explicitly done by an administrator, we don't want to override that.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -119,9 +134,148 @@ func (a *API) dropDatabase(ctx context.Context, w http.ResponseWriter, req *http
...
+		logger.Info("database is running, scaling not necessary")
+		return nil
+	}
+
+	// Copy processes and increase database processes to 3.
+	processes := make(map[string]int, len(formation.Processes))
+	for k, v := range formation.Processes {
+		processes[k] = v
+	}
+	processes["mariadb"] = 3
```
titanous: This should check the `SINGLETON` env var.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
benbjohnson: I was trying to mirror the `flynn scale` CLI functionality. What's the difference from waiting for discoverd events?

benbjohnson:
@titanous Good point.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
titanous: These events are based on when the job is started, they don't mean that the job is up and running. In this case we can be less generic and use discoverd directly, which is a more accurate indicator of when the process-level API is available.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -119,9 +137,173 @@ func (a *API) dropDatabase(ctx context.Context, w http.ResponseWriter, req *http
...
+func (a *API) scaleUp() error {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
+
+	// Ignore if already scaled up.
+	if a.scaledUp {
+		return nil
+	}
+
+	app := "mariadb"
```
titanous: This should read `FLYNN_APP_ID` from the environment.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -119,9 +137,173 @@ func (a *API) dropDatabase(ctx context.Context, w http.ResponseWriter, req *http
...
+	if os.Getenv("SINGLETON") == "true" {
+		processes["mariadb"] = 1
+	} else {
+		processes["mariadb"] = 3
+	}
+
+	// Otherwise begin scaling the database processes up.
+	logger.Info("watching discoverd events", "app", app, "release_id", release.ID)
+	events := make(chan *discoverd.Event)
+	if _, err := discoverd.DefaultClient.Service("mariadb").Watch(events); err != nil {
```
titanous: The stream should be assigned and the close of it deferred.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -119,9 +137,173 @@ func (a *API) dropDatabase(ctx context.Context, w http.ResponseWriter, req *http
...
+
+		// Attempt to connect to database.
+		if err := func() error {
+			db, err := a.connect()
+			if err != nil {
+				return err
+			}
+			defer db.Close()
+
+			if _, err := db.Exec("SELECT 1"); err != nil {
```
titanous: I think the `WaitForReadWrite` method in the sirenia client is more appropriate.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -119,9 +137,173 @@ func (a *API) dropDatabase(ctx context.Context, w http.ResponseWriter, req *http
...
+	logger.Info("updating formation", "app", app, "release_id", release.ID)
+	formation.Processes = processes
+	if err := client.PutFormation(formation); err != nil {
+		logger.Error("put formation error", "app", app, "release_id", release.ID, "err", err)
+		return err
+	}
+
+	// Wait for discoverd event from service.
+	logger.Info("waiting for scale up", "app", app, "release_id", release.ID)
+	if err := a.waitForUpEvent(events); err != nil {
```
titanous: I think we can just use `discoverd.GetInstances` here.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
benbjohnson: Fixed in b6e2e16.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
benbjohnson: Fixed in b6e2e16.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
benbjohnson: Fixed in b6e2e16.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
benbjohnson: I tried to get `GetInstances()` to work but it kept returning a timeout. I removed the discoverd event wait in b6e2e16 and just used the sirenia client to check for the cluster.

appliance/mariadb/cmd/flynn-mariadb-api/main.go
```diff
@@ -119,9 +138,127 @@ func (a *API) dropDatabase(ctx context.Context, w http.ResponseWriter, req *http
...
+	defer timer.Stop()
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-timer.C:
+			return errors.New("timed out while starting mariadb cluster")
+		case <-ticker.C:
+		}
```
titanous: Is the for loop redundant with the `WaitForReadWrite` call? I think it does this internally.

benbjohnson: It was failing immediately if it tried to reach the `leader` DNS before it was available.

titanous: @josephglanville What do you think about changing that method to retry on errors?

josephglanville:
Superceded by #2781

PR #2637, opened 2016-03-26
installer: Print more readable startup errors

### LINKED ISSUE #2573: Error while multiple flynn install are running
The command : ``flynn install`` must not be started multiple times.
Otherwise, we get this error : https://gist.github.com/Sabmit/c268ca8ead2f035ae513

This panic error should be catched and a human readable error should be displayed.

jvatic:
Closes #2573

### DIFF
--- a/installer/installer.go
+++ b/installer/installer.go
@@ -12,6 +12,7 @@ import (
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/digitalocean/godo"
 	log "github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
 	"github.com/flynn/flynn/pkg/httphelper"
+	"github.com/flynn/flynn/pkg/shutdown"
 )
 
 var ClusterNotFoundError = errors.New("Cluster not found")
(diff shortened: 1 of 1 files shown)


installer/installer.go
```diff
@@ -36,10 +37,10 @@ func NewInstaller(l log.Logger) *Installer {
 		logger:        l,
 	}
 	if err := installer.openDB(); err != nil {
-		panic(err)
+		shutdown.Fatalf("Error opening database: %s", err)
```
titanous: This should explicitly check for the "another instance is running" error and phrase it as such.

installer/installer.go
```diff
@@ -36,10 +37,13 @@ func NewInstaller(l log.Logger) *Installer {
 		logger:        l,
 	}
 	if err := installer.openDB(); err != nil {
-		panic(err)
+		if err.Error() == "resource temporarily unavailable" {
+			shutdown.Fatal("Error: Another instance is already running.")
```
titanous:
```text
Error: Another `flynn install` process is already running.
```

titanous:
LGTM

PR #2648, opened 2016-03-31
Update pgx

josephglanville:
Update pgx to bring in various improvements to connection pooling and a fix for a possible deadlock when resetting the connection pool.

Also change the way the `flynn` user is created in the PG appliance.
It looks like there was always a bug here but the new version of pgx shook it out.


titanous:
LGTM

Issue #2398, opened 2016-01-28
Consensus Cluster Reconfiguration Tool


dashkb:
Flynn operators need a tool which allows them to add and remove nodes from the Raft cluster.

kwando:
+1

titanous:
This was implemented in #2682.

PR #2652, opened 2016-04-01
Godeps: Update hashicorp/raft

josephglanville:
Prerequisite work for consensus reconfiguration and other discoverd improvements, ie. #2398.
The main highlights here are leader step down is properly implemented and there are a bunch of correctness fixes.


titanous:
LGTM

PR #2655, opened 2016-04-01
Monotonic timeouts

titanous:
If there is any time skew, subtracting absolute times is not reliable. The timers in Go use monotonic time, which does not have this problem.

I'm not certain that this is necessary, but I have seen a few weird things on CI with no good explanation, and this change is correct.


josephglanville:
LGTM

Issue #2666, opened 2016-04-06
Installer: invalid memory address or nil pointer dereference


terracatta:
Currently evaluating several PaaS solutions on AWS, when trying to use Flynn's installer after entering valid AWS credentials on the site (and click "Launch") I get the following error:

```
$ flynn install          
Open http://localhost:55211 in your browser to continue.
INFO[04-06|16:01:37] Event: new_cluster:  
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x20 pc=0x1cae47]

goroutine 48 [running]:
github.com/flynn/flynn/installer.func·028(0x20980a3f0, 0x28, 0x0, 0x0, 0x220974a330, 0x209904c30, 0x0, 0x0)
	/tmp/tmp.AaXXtTCrZT/src/github.com/flynn/flynn/installer/persistence.go:84 +0x57
path/filepath.Walk(0x20980a3f0, 0x28, 0x20986abd8, 0x0, 0x0)
	/tmp/tmp.AaXXtTCrZT/src/github.com/flynn/flynn/util/_toolchain/go/src/path/filepath/path.go:392 +0xb6
github.com/flynn/flynn/installer.listSSHKeyNames(0x0, 0x0, 0x0)
	/tmp/tmp.AaXXtTCrZT/src/github.com/flynn/flynn/installer/persistence.go:91 +0x9c
github.com/flynn/flynn/installer.(*AWSCluster).createKeyPair(0x209805520, 0x0, 0x0)
	/tmp/tmp.AaXXtTCrZT/src/github.com/flynn/flynn/installer/aws_cluster.go:248 +0x42
github.com/flynn/flynn/installer.*AWSCluster.(github.com/flynn/flynn/installer.createKeyPair)·fm(0x0, 0x0)
	/tmp/tmp.AaXXtTCrZT/src/github.com/flynn/flynn/installer/aws_cluster.go:155 +0x39
github.com/flynn/flynn/installer.func·002()
	/tmp/tmp.AaXXtTCrZT/src/github.com/flynn/flynn/installer/aws_cluster.go:166 +0x449
created by github.com/flynn/flynn/installer.(*AWSCluster).Run
	/tmp/tmp.AaXXtTCrZT/src/github.com/flynn/flynn/installer/aws_cluster.go:180 +0x8c

```

Repeating this process results in the same error.

jvatic:
[...]

PR #2667, opened 2016-04-06
installer: Add missing error check

jvatic:
Refs #2666 (this should fix the issue, or at least return a meaningful error).


titanous:
LGTM

Issue #2597, opened 2016-03-19
Cluster backup endpoint should indicate failure


titanous:
The workflow should probably look like:
1. Create backup resource.
2. Start streaming backup from separate endpoint.
3. Check backup info endpoint for expected SHA and any error logs.

/cc @jvatic

jvatic:
Here's my thoughts:

- Add a backups table (id, created_at, completed_at, sha, size, data)
- `POST /backups` saves a db record (id, created_at), starts job to generate tar and save into db (completed_at, sha, size, data)
- `GET /backups/:backups_id` returns backup record with `Accept: application/json`, otherwise streams backup (waiting for the job to complete if it hasn't)
- `GET /backup` does the same as the above endpoint, except assumes the latest backup and creates one if it doesn't exist for backwards compatibility (this endpoint can eventually be removed once API versioning is in place)
- `DELETE /backups/:backups_id` removes the backup from the db (I think in this case deleting the record instead of soft delete makes the most sense as backups can be rather large)
- Backup script should ensure records in backups table are not part of the postgres dump, though `pg_dumpall` doesn't appear to support the `--exclude-table-data` option so this could be an issue (maybe just store the tar data someplace else?)

titanous:
The data should not be saved in the database. For now lets have the GET endpoint trigger the jobs and stream inline directly to the client (we can do something more sophisticated later). The database should store the status of the backup as well as the error log.

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

PR #2668, opened 2016-04-06
controller: Backup refactor

jvatic:
Refs #2597

### DIFF
--- a/controller/backup.go
+++ b/controller/backup.go
@@ -1,26 +1,172 @@
 package main
 
 import (
+	"crypto/sha512"
+	"encoding/hex"
 	"fmt"
+	"io"
 	"net/http"
+	"strings"
 	"time"
 
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/jackc/pgx"
 	"github.com/flynn/flynn/Godeps/_workspace/src/golang.org/x/net/context"
 	"github.com/flynn/flynn/controller/client"
+	ct "github.com/flynn/flynn/controller/types"
 	"github.com/flynn/flynn/pkg/backup"
+	"github.com/flynn/flynn/pkg/ctxhelper"
+	"github.com/flynn/flynn/pkg/httphelper"
+	"github.com/flynn/flynn/pkg/postgres"
 )
 
+type BackupRepo struct {
+	db *postgres.DB
+}
+
+func NewBackupRepo(db *postgres.DB) *BackupRepo {
+	return &BackupRepo{db: db}
+}
+
+func (r *BackupRepo) GetLatest() (*ct.ClusterBackup, error) {
+	b := &ct.ClusterBackup{}
+	if err := r.db.QueryRow("backup_select_latest").Scan(&b.ID, &b.Status, &b.SHA512, &b.Size, &b.Error, &b.CreatedAt, &b.UpdatedAt, &b.CompletedAt); err != nil {
+		if err == pgx.ErrNoRows {
+			err = ErrNotFound
+		}
+		return nil, err
+	}
+	return b, nil
+}
+
+func (r *BackupRepo) Add(b *ct.ClusterBackup) error {
+	tx, err := r.db.Begin()
+	if err != nil {
+		return err
+	}
+	if err := tx.QueryRow("backup_insert", b.Status, b.SHA512, b.Size, b.Error, b.CompletedAt).Scan(&b.ID, &b.CreatedAt, &b.UpdatedAt); err != nil {
+		tx.Rollback()
+		return err
+	}
+	if err := createEvent(tx.Exec, &ct.Event{
+		ObjectID:   b.ID,
+		ObjectType: ct.EventTypeClusterBackup,
+	}, b); err != nil {
+		tx.Rollback()
+		return err
+	}
+	return tx.Commit()
+}
+
+func (r *BackupRepo) Update(b *ct.ClusterBackup) error {
+	tx, err := r.db.Begin()
+	if err != nil {
+		return err
+	}
+	if err := tx.QueryRow("backup_update", b.ID, b.Status, b.SHA512, b.Size, b.Error, b.CompletedAt).Scan(&b.UpdatedAt); err != nil {
+		tx.Rollback()
+		return err
+	}
+	if err := createEvent(tx.Exec, &ct.Event{
+		ObjectID:   b.ID,
+		ObjectType: ct.EventTypeClusterBackup,
+	}, b); err != nil {
+		tx.Rollback()
+		return err
+	}
+	return tx.Commit()
+}
+
 func (c *controllerAPI) GetBackup(ctx context.Context, w http.ResponseWriter, req *http.Request) {
-	w.Header().Set("Content-Type", "application/tar")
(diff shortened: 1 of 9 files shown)


jvatic:
ping @titanous for review

controller/schema.go
```diff
@@ -271,6 +271,21 @@ $$ LANGUAGE plpgsql`,
 		$$`,
 		`UPDATE apps SET meta = jsonb_merge(meta, '{"flynn-system-critical":"true"}') WHERE name IN ('discoverd', 'flannel', 'postgres', 'controller')`,
 	)
+	migrations.Add(14,
+		`CREATE TYPE backup_status AS ENUM ('running', 'complete', 'error')`,
```
titanous: I think we're using join tables instead of enums. Take a look at the migrations above for examples.

controller/schema.go
```diff
@@ -271,6 +271,21 @@ $$ LANGUAGE plpgsql`,
...
+		`CREATE TABLE backups (
+			backup_id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+			status backup_status NOT NULL,
+			hash text,
```
titanous: This column should be named for the hash algorithm and can be a `bytea`.

controller/backup.go
```diff
@@ -1,26 +1,143 @@
...
+
+	b := &ct.ClusterBackup{
+		Status: ct.ClusterBackupStatusRunning,
+	}
+	if err := c.backupRepo.Add(b); err != nil {
+		handleError(err)
+		return
+	}
+
+	h := sha1.New()
```
titanous: Use SHA-512.

controller/backup.go
```diff
@@ -1,26 +1,143 @@
...
+}
+
+func NewBackupRepo(db *postgres.DB) *BackupRepo {
+	return &BackupRepo{db: db}
+}
+
+func (r *BackupRepo) GetLatest() (*ct.ClusterBackup, error) {
+	b := &ct.ClusterBackup{}
+	if err := r.db.QueryRow("backup_select_latest").Scan(&b.ID, &b.Status, &b.Hash, &b.Size, &b.Error, &b.CreatedAt, &b.UpdatedAt, &b.CompletedAt); err != nil {
+		if err.Error() == "no rows in result set" {
```
titanous: I think there is a type to match against? Check the code for other objects.

titanous:
This doesn't close the issue, as it looks like a first patch towards the final multi-step API. I'm fine with merging as-is, but it looks like we need another iteration.

jvatic:
Updated description not to close referenced issue

jvatic:
@titanous comments addressed

schema/controller/cluster_backup.json
```diff
@@ -1,11 +1,52 @@
...
+    "id": {
+      "$ref": "/schema/controller/common#/definitions/id"
+    },
+    "status": {
+      "$ref": "#/definitions/backup_status"
+    },
+    "sha512": {
+      "description": "base64 encoded sha512 of backup",
+      "type": "string",
+      "format": "base64"
```
titanous: I think we should hex-encode the hash as all standard tools work with hex hashes, not base64.

controller/schema.go
```diff
@@ -271,6 +271,24 @@ $$ LANGUAGE plpgsql`,
...
+			status text NOT NULL,
+			sha512 bytea,
+			size integer,
+			error text,
+			created_at timestamptz NOT NULL DEFAULT now(),
+			updated_at timestamptz NOT NULL DEFAULT now(),
+			completed_at timestamptz,
+			deleted_at timestamptz
+		)`,
+		`ALTER TABLE backups ADD CONSTRAINT backups_status_fkey FOREIGN KEY (status) REFERENCES backup_statuses (name)`,
```
titanous: This can be dropped and you should be able to put `REFERENCES backup_statuses (status)` after `status text NOT NULL` instead.

schema/controller/cluster_backup.json
jvatic: This is poorly worded, it's a base64 of the hex encoded hash (it doesn't appear possible to JSON encode a `[]byte` without it being a base64 encoded string, so probably best just to use a string for that field).

jvatic:
@titanous comments addressed

controller/types/types.go
```diff
@@ -349,3 +350,20 @@ type DomainMigrationEvent struct {
...
+	ClusterBackupStatusRunning  string = "running"
+	ClusterBackupStatusComplete string = "complete"
+	ClusterBackupStatusError    string = "error"
+)
+
+type ClusterBackup struct {
+	ID          string     `json:"id,omitempty"`
+	Status      string     `json:"status"`
+	SHA512      string     `json:"sha512,omitempty"`
+	Size        int32      `json:"size,omitempty"`
```
titanous: Any reason this is an `int32` instead of an `int`?

controller/schema.go
```diff
@@ -271,6 +271,23 @@ $$ LANGUAGE plpgsql`,
...
 	)
+	migrations.Add(14,
+		`CREATE TABLE backup_statuses (name text PRIMARY KEY)`,
+		`INSERT INTO backup_statuses (name) VALUES
+			('running'), ('complete'), ('error')`,
+		`CREATE TABLE backups (
+			backup_id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+			status text NOT NULL REFERENCES backup_statuses (name),
+			sha512 text,
+			size integer,
```
titanous: Use `bigint`, no reason not to.

controller/client/client.go
```diff
@@ -813,6 +813,14 @@ func (c *Client) Backup() (io.ReadCloser, error) {
 	return res.Body, err
 }
 
+// GetBackupMeta returns metadata for latest backup
+func (c *Client) GetBackupMeta() (*ct.ClusterBackup, error) {
+	h := http.Header{"Accept": []string{"application/json"}}
+	b := &ct.ClusterBackup{}
+	_, err := c.RawReq("GET", "/backup", h, nil, b)
```
titanous: Is this `RawReq` necessary? Doesn't `Get` set the `Accept` header?

pkg/sizewriter/sizewriter.go
```diff
@@ -0,0 +1,32 @@
+package sizewriter
+
+import "io"
+
+type SizeWriter struct {
```
titanous: Since this is only used in one place you can just put it inline.

controller/types/types.go
jvatic: Something in the db stack was complaining about it being an int during deserialization.

controller/client/client.go
jvatic: Nope, it's only set for event streams.

titanous: Hmm, add a commit that sets it and see what happens!

jvatic: Looks like this doesn't break anything :) (https://ci.flynn.io/builds/20160409135114-7ba1d9c3)

jvatic:
@titanous comments addressed, and added detail to controller commit message.

titanous:
LGTM

PR #2074, opened 2015-11-04
Docs: Using postgres and ZFS device migration

nobert:
Addresses #1189

Issue #1286, opened 2015-03-16
Add docs on how to operate flynn


alex:
Once I've installed flynn, how do I:

* Add capacity
* Remove capacity
* Upgrade flynn
* Monitor/observe system health
* Find logs to debug production issues
* etc :-)

Cheers.

PS: I'm very excited about flynn

philiplb:
+1! :-)

aaronshaf:
+1

gmaclennan:
+1 how to add capacity

titanous:
Documentation covering how to operate Flynn in production is available at https://flynn.io/docs/production

Please open new issues for anything you notice that is missing.

Issue #68, opened 2014-08-11
PROXY protocol support in router


titanous:
We should support the [PROXY protocol](http://haproxy.1wt.eu/download/1.5/doc/proxy-protocol.txt) for both sides of TCP connections.

blurrcat:
Hi, do we have a timeline for this?

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

PR #2673, opened 2016-04-07
docs: Add Operating Flynn page

### LINKED ISSUE #478: Documentation for Adding Nodes to Running Cluster
I apologize if this was already covered, but I wasn't able to find anything related to adding nodes to an existing (running) cluster.

Or, if this is not current a supported feature (i.e. you must always bootstrap a cluster with the number of nodes you wish to have), it may be nice to have a note/documentation page somewhere about this.

### LINKED ISSUE #1189: Docs for migrating from zfs backing file to real devs
Currently we provision a zpool on a 100GB sparse file so that the user doesn't have to set up backing disks. We should document migrating the datasets on this zpool to a zpool backed by real disks. We should also document the caveats of using the flat file and the considerations that go into building a production zpool.

### LINKED ISSUE #2070: Documentation Request: System Requirements
Hi,

it would be great if the documentation could state the system requirements for flynn based on your experience. Today, I guessed how much cores, memory and hard drive space would make sense for a single node. I went for 1 core, 1GB RAM and 32GB hdd. But that's just a guess.
Some kind of base memory needed. The application specific part can't be specified of course. :)

### LINKED ISSUE #2659: Update instruction need to be a little clearer
I think to help avoid the issue i have had where i was just doing `flynn-host bootstrap --from-backup backup.tar` rather than 

```
$ sudo \
    CLUSTER_DOMAIN=demo.localflynn.com \
    flynn-host bootstrap \
    --min-hosts 3 \
    --discovery https://discovery.flynn.io/clusters/53e8402e-030f-4861-95ba-d5b5a91b5902 \
    --from-backup backup.tar
```
as the what I did does restore from the backup but you only get the 1 node not the complete cluster. I appreciate it was probably me being a bit stupid but it will help elimate any doubt.

titanous:
First pass.

@jvatic Please implement redirects from the added JSON file in the server.

Closes #2074 
Closes #1189
Closes #2659
Closes #2070
Closes #478
Refs #1286


jvatic:
@titanous Can you add `redirects_path` to [docs/lib/flynn-docs.rb](https://github.com/flynn/flynn/blob/98a54080452c6cd8e83a096dfc687d4933142563/docs/lib/flynn-docs.rb#L16-L18)?

titanous:
@jvatic Done.

jvatic:
LGTM

docs/content/operating.html.md
```diff
@@ -0,0 +1,233 @@
...
+
+Each host should have a minimum of 1GB of memory, and inter-host network packets
+should have a latency of less than 2ms. Deploying a single Flynn cluster across
+higher latency WAN links is not recommended, as it can have a negative impact on
+cluster consensus.
+
+## Storage
+
+Flynn uses ZFS to store data. By default, a ZFS pool is created in a sparse file
+on the existing filesystem. This is not recommended for production for
```
lmars: I have seen some users have `/var/lib/flynn` on a dedicated disk, so is it worth clarifying what "existing filesystem" means? For example:

```
By default, a ZFS pool is created in a sparse file on the
same filesystem that `/var/lib/flynn/volumes` exists on.
```

docs/content/operating.html.md
```diff
@@ -0,0 +1,233 @@
...
+
+Flynn has a built-in router that handles all incoming HTTP, HTTPS and TCP
+traffic. It is lightweight and runs on every Flynn host, this allows easy
+deployment of Flynn without requiring another load balancer in front of Flynn.
+
+We recommend running Flynn with low-TTL round-robin DNS A records pointing at
+each host. For automatic failover, health checks should be configured that
+automatically remove a host's A record if it is unhealthy.
+
+Alternatively, a TCP load balancer for ports 80 and 443 can be configured in
```
lmars: Saying "TCP load balancer" may cause people to use say an ELB TCP listener, when an ELB HTTP listener would be more appropriate (`X-Forwarded-For`, connection pooling etc.)

titanous: Yeah, that's what we want. We don't want ELB mucking with the HTTP requests and we want to handle TLS termination.

titanous: #68 will solve the source IP issue.

docs/content/operating.html.md
```diff
@@ -0,0 +1,233 @@
...
+
+A firewall preventing external access must always be configured on or in front
+of Flynn hosts, only these ports should be allowed:
+
+* 80 (HTTP)
+* 443 (HTTPS)
+* 3000 to 3500 (user-defined TCP services, optional)
+
+Internal cross-host cluster communication happens on a variety of ports and
+should not be restricted.
```
lmars: I think it is worth explicitly mentioning that both TCP & UDP should be unrestricted (I have seen users just allowing TCP).

docs/content/operating.html.md
```diff
@@ -0,0 +1,233 @@
...
+unavailable for a few seconds while they are updated.
+
+To perform an in-place update of the entire cluster, run `flynn-host update`.
+
+## Adding Hosts
+
+Hosts may be added to an existing cluster by running `flynn-host init` with the
+discovery token or list of host IPs that was used to start the cluster. Care
+should be taken to ensure that the same version of Flynn is installed on all
+hosts.
```
lmars: We should add a note on how to determine the version of an existing cluster, and then how to install that same version on a new host (setting `FLYNN_VERSION=xxx` when running the install script).

titanous:
Comments addressed.

philiplb:
> We recommend running Flynn with low-TTL round-robin DNS A records pointing at
> each host. For automatic failover, health checks should be configured that
> automatically remove a host's A record if it is unhealthy.

What does that mean, how would I implement the A records? After adding them in the DNS server, where would I configure them within Flynn, if so?

lmars:
@philiplb your DNS records are not managed from within Flynn, but by whatever service your domain's nameserver records point at.

You can use a service like [AWS Route53](https://aws.amazon.com/route53/) to manage your DNS records with health checks as proposed in the document.

lmars:
LGTM

philiplb:
Of course, this is what I meant with DNS server. Ok, this simply says to have a failover machine for each of the (initial) nodes.

PR #2674, opened 2016-04-07
host/cli: Only start one scheduler instance when restoring from backup

### LINKED ISSUE #2540: ratelimit hit during bootstrap with backup restore
```text
17:21:51.657699 run-app controller
17:21:55.617274 run-app controller error: ratelimited: maximum concurrent AddJob calls running, try again later'
```

We should retry in this case.

titanous:
Subsequent job requests can race with the first scheduler and cause a ratelimit error during bootstrap. We only need one, as the rest will be started by the scheduler.

Closes #2540


josephglanville:
LGTM

PR #2682, opened 2016-04-11
Discoverd reconfiguration

josephglanville:
Allows reconfiguration of the discoverd consensus by promoting proxy peers to members of the Raft peer set and demoting existing peers to proxies. It can also forcibly remove peers that are no longer available.


discoverd/client/client.go
```diff
@@ -180,10 +180,26 @@ func (c *Client) Shutdown() (res dt.ShutdownInfo, err error) {
 	return res, c.c.Post("/shutdown", nil, &res)
 }
 
+func (c *Client) Promote() error {
+	return c.c.Post("/promote", nil, nil)
+}
+
+func (c *Client) Demote() error {
+	return c.c.Post("/demote", nil, nil)
```
titanous: What do you think about putting these endpoints under the `/raft` prefix?

discoverd/main.go
```diff
@@ -144,16 +158,29 @@ func (m *Main) Run(args ...string) error {
...
+	go m.mux.Serve()
+
+	m.dataDir = opt.DataDir
+
+	// if the advertise addr is not in the peer list we are proxying
+	proxying := true
+	for _, addr := range m.peers {
+		if addr == m.advertiseAddr {
+			proxying = false
+		}
```
titanous: This code should break out of the outer loop after setting the flag.

discoverd/main.go
```diff
@@ -144,16 +158,29 @@ func (m *Main) Run(args ...string) error {
...
-		return fmt.Errorf("Failed to open store: %s", err)
+	if !proxying {
+		if err := m.openStore(true); err != nil {
+			return fmt.Errorf("Failed to open store: %s", err)
+		}
 	}
 
 	// Notify user that we're proxying if the store wasn't initialized.
-	if m.store == nil {
+	if proxying {
```
titanous: This could be an `else` with the statement above.

discoverd/main.go
```diff
@@ -144,16 +158,29 @@ func (m *Main) Run(args ...string) error {
...
 		fmt.Fprintln(m.Stderr, "advertised address not in peer set, joining as proxy")
```
titanous: This should print to the logger.

discoverd/main.go
```diff
@@ -246,10 +283,124 @@ func waitHostDNSConfig() (addr string, resolvers []string) {
...
+	// error out if the store is loaded
+	if m.store != nil {
+		return fmt.Errorf("can't join, already a raft peer")
+	}
+	// error out of the handler isn't running in proxy mode
+	if !m.handler.Proxy.Load().(bool) {
+		return fmt.Errorf("can't promote, handler not in proxy mode")
+	}
+	// TODO(jpg): Replace this with better way of getting leader
+	randAddr := m.peers[rand.Intn(len(m.peers))]
```
titanous: Use `random.Math`, it is pre-seeded.

discoverd/main.go
```diff
@@ -246,10 +283,124 @@ func waitHostDNSConfig() (addr string, resolvers []string) {
...
+		return err
+	}
+	// TODO(jpg): Need updated peer list
+	//if !inPeerList(m.peers, m.advertiseAddr) {
+	//	m.peers = append(m.peers, m.advertiseAddr)
+	//}
+	//m.logger.Println("setting peers")
+	//m.store.SetPeers(m.peers)
+
+	// wait for our local store to catch up
```
titanous: Is this aa TODO?

discoverd/main.go
```diff
@@ -246,10 +283,124 @@ func waitHostDNSConfig() (addr string, resolvers []string) {
...
+	if leaderAddr == m.advertiseAddr {
+		m.store.RemovePeer(m.advertiseAddr)
+	} else {
+
+		// ask the leader to remove us from the peer set
+		m.logger.Println("asking the leader to remove us from the raft set")
+		discoverd.NewClientWithURL(leaderAddr).RaftRemovePeer(m.advertiseAddr)
+		//m.store.SetPeers(nil)
+	}
+	time.Sleep(3 * time.Second)
```
titanous: Sleep?

discoverd/client/client.go
josephglanville: Yep makes sense.

discoverd/main.go
josephglanville: This can probably be removed now, it shouldn't be necessary.

discoverd/main.go
josephglanville: Should have been a TODO comment here to look into waiting until our local peer store has applied the removePeer command from the leader. I haven't quite worked that out yet. It's maybe not super required because as long as it's been applied by the leader there should be no adverse effects of us simply shutting down the store.

temujin9:
Okay, I hate to be "that guy", but I have to: this is now the last thing sitting in the critical path for my internal general availability release of Flynn at Bazaarvoice. Can I get a conservative estimate of how much longer until this lands?

temujin9:
Okay, followed up with @titanous and got a fair estimate. I'll write a runbook for our immediate needs, so we can go GA without waiting breathlessly on potentially finicky code.

josephglanville:
Just an update on this. During development I ran into some rather difficult to solve issues which delayed progress on this. However with those resolved this is getting fairly close.

ping @titanous for another review pass.

temujin9:
@josephglanville I'm glad to hear the difficulty has been surmounted. Thanks for the update

discoverd/main.go
```diff
@@ -260,6 +278,106 @@ func waitHostDNSConfig() (addr string, resolvers []string) {
...
+
+	// Update the dns server to use proxy store
+	m.dnsServer.SetStore(&server.ProxyStore{Peers: m.peers})
+	m.handler.Proxy.Store(true)
+
+	// If we are the leader remove ourselves directly,
+	// otherwise request the leader remove us from the peer set.
+	leaderAddr := m.store.Leader()
+	if leaderAddr == "" {
+		return server.ErrNoKnownLeader
```
titanous: In this case and other error cases below the handler and DNS server will not get reset back to the original, is that okay?

discoverd/server/handler.go
```diff
@@ -82,14 +91,33 @@ type Handler struct {
...
 
 func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if h.Shutdown.Load().(bool) {
 		hh.Error(w, ErrShutdown)
 		return
 	}
+	// If running in proxy mode then redirect requests to a random peer
+	if h.Proxy.Load().(bool) && !proxyWhitelisted(r) {
+		// TODO(jpg): Should configuring the node in proxy mode with no peers be impossible?
+		host := h.Peers[rand.Intn(len(h.Peers))]
```
titanous: Use `random.Math`, it is seeded properly.

josephglanville:
@titanous This is ready for another review pass.

titanous:
LGTM!

temujin9:
Thank you, guys! I look forward to seeing this in a release.

PR #2683, opened 2016-04-12
controller: Fix domain migration route creation

### LINKED ISSUE #2616: Domain migration fails to create routes for some apps
After restoring from a backup and running a domain migration, the status app and the user app from the backup (there was only one in the backup) didn't have routes for the new domain, only the old one. The other system apps correctly had a route for both domains.

jvatic:
Create new routes for apps where old route doesn't have a TLSCert.

Closes #2616 

- [x] Add test


titanous:
LGTM

titanous:
LGTM

PR #2690, opened 2016-04-13
Update operating.html.md

tjboring:
Document how to get the cluster auth_key to send in the cluster health URL.

### DIFF
--- a/docs/content/operating.html.md
+++ b/docs/content/operating.html.md
@@ -148,14 +148,14 @@ route, add the `--routes` flag.
 ## Monitoring
 
 Flynn provides a status endpoint over HTTP that exposes the health of system
-components at `http://status.$CLUSTER_DOMAIN?key=$auth_key` (for example,
-`http://status.1.localflynn.com?key=$auth_key`). The status endpoint returns a status code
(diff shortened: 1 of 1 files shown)


docs/content/operating.html.md
```diff
@@ -148,11 +148,16 @@ route, add the `--routes` flag.
 ## Monitoring
 
 Flynn provides a status endpoint over HTTP that exposes the health of system
-components at `http://status.$CLUSTER_DOMAIN` (for example,
-`http://status.1.localflynn.com`). The status endpoint returns a status code
+components at `http://status.$CLUSTER_DOMAIN?key=$auth_key` (for example,
+`http://status.1.localflynn.com?key=$auth_key`). The status endpoint returns a status code
```
titanous: This is only necessary if the request doesn't come from a private IP, which should be noted here. Also, the convention we are using is all caps variables, so `$AUTH_KEY`.

titanous: Also, there should be a forward slash before the question mark.

tjboring: done and done

PR #2701, opened 2016-04-16
discoverd: Use AppliedIndex instead of LastIndex

josephglanville:
Newer versions of the Raft library update LastIndex differently
as such it's not as good for our usecase. However they also
added a new function AppliedIndex which returns the index of the
last log successfully applied to the FSM.
This is even better than what we had previously as the other did
have a possibility of returning an index of a log entry that was
yet to be applied on the leader.


lmars:
LGTM

PR #2718, opened 2016-04-20
cli: Respect -y flag in `remote add`

### LINKED ISSUE #2716: "flynn remote add flynn -y" skips the "-y" flag
Hi,

currently, I'm setting up our Jenkins to deploy an application to the Flynn cluster.
As he doesn't know the git remote (yet), I need to add it in order to push. To do so, I use this:
```bash
flynn -a myapp remote add flynn -y
```

As the remote already exists on the flynn cluster, it prompts me with a confirmation which isn't skipped by the "-y" flag.
Current workaround:
```bash
echo "yes" | flynn -a myapp remote add flynn
```

I think this is a bug?
And would this be the recommended way of getting the remote into a freshly cloned repository?

titanous:
Closes #2716

### DIFF
--- a/cli/remote.go
+++ b/cli/remote.go
@@ -44,12 +44,14 @@ func runRemote(args *docopt.Args, client *controller.Client) error {
 		return nil
 	}
 
-	update, err := promptReplaceRemote(remote)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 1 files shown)


josephglanville:
LGTM

Issue #2719, opened 2016-04-21
flynn mysql command available but not documented


holtgrewe:
`flynn --help` does not show mysql command, `flynn mysql --help` works, though.

PR #2721, opened 2016-04-21
Adding redis and mysql to `flynn --help`

holtgrewe:
Addresses #2719

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -49,6 +49,7 @@ Commands:
 	ps          list jobs
 	kill        kill a job
 	log         get app log
+	mysql       manage mysql/mariadb database
 	scale       change formation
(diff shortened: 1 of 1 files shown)


jvatic:
@holtgrewe The commit message needs a `cli: ` prefix and a [DCO](https://flynn.io/docs/contributing#developer%E2%80%99s-certificate-of-origin). Also use `Closes` rather than `Addresses` so the issue will be closed automatically upon merge.

holtgrewe:
Thanks for the info. I created this patch using the Github web interface.

I don't think that this contribution reaches any sufficient level of creation. Is it enough that I state here: I place this contribution in the public domain and you simply create this patch?

PR #2733, opened 2016-04-24
scheduler: Formation stream fixes

lmars:
In [this CI failure](https://ci.flynn.io/builds/20160423202834-61474f73), the cluster fell over during a controller deployment because new schedulers were started, old ones were killed but the new schedulers all crashed with:

```
20:47:41.968254 scheduler.go:159: timed out waiting for current formation list
```

Whilst I am unsure why the scheduler was not getting the current event from the formation stream, I've made two patches which will improve the situation:

* Only register in service discovery once the current formation event has been received. This means that if there is the above error, the old schedulers will still be around and the deployment will just fail (rather than the old schedulers already being killed, and then the new schedulers also crashing, leaving no running schedulers)
* Return an error to the client if `sendUpdatedSince` encounters an error (and therefore does not send the current event)

I've also added a slightly unrelated patch so that the scheduler explicitly identifies jobs for which it needs to look up a formation, because I was seeing lots of errors looking up formations for one-off jobs.

### DIFF
--- a/controller/utils/utils.go
+++ b/controller/utils/utils.go
@@ -39,6 +39,7 @@ func JobConfig(f *ct.ExpandedFormation, name, hostID string, uuid string) *host.
 	metadata["flynn-controller.app"] = f.App.ID
 	metadata["flynn-controller.app_name"] = f.App.Name
 	metadata["flynn-controller.release"] = f.Release.ID
+	metadata["flynn-controller.formation"] = "true"
 	metadata["flynn-controller.type"] = name
 	job := &host.Job{
 		ID:       id,
--- a/controller/formation.go
+++ b/controller/formation.go
@@ -28,9 +28,10 @@ type formationKey struct {
 type FormationSubscription struct {
 	mtx sync.RWMutex
 	ch  chan *ct.ExpandedFormation
+	err error
 }
 
-func (f *FormationSubscription) notify(ef *ct.ExpandedFormation) bool {
+func (f *FormationSubscription) Notify(ef *ct.ExpandedFormation) bool {
 	f.mtx.RLock()
 	defer f.mtx.RUnlock()
 	if f.ch == nil {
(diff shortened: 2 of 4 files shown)


controller/utils/utils.go
```diff
@@ -39,6 +39,7 @@ func JobConfig(f *ct.ExpandedFormation, name, hostID string, uuid string) *host.
 	metadata["flynn-controller.app"] = f.App.ID
 	metadata["flynn-controller.app_name"] = f.App.Name
 	metadata["flynn-controller.release"] = f.Release.ID
+	metadata["flynn-controller.formation"] = f.App.ID + ":" + f.Release.ID
```
titanous: I'm not sure I follow this. This appears to be redundant information, since the AppID and ReleaseID are already in the meta. Do we never use `JobConfig` for one-off jobs? Can we mark the jobs more explicitly instead of repeating existing information?

lmars: `JobConfig` takes a formation argument, so it's always used when starting a job for a formation (there are currently three uses: [scheduler](https://github.com/flynn/flynn/blob/v20160423.0/controller/scheduler/scheduler.go#L1335-L1337), [bootstrapper](https://github.com/flynn/flynn/blob/v20160423.0/bootstrap/run_app_action.go#L109-L120) and [host/fixer](https://github.com/flynn/flynn/blob/v20160423.0/host/fixer/fixer.go#L335-L343)), but the controller `RunJob` method [also sets the AppID and ReleaseID](https://github.com/flynn/flynn/blob/master/controller/jobs.go#L286-L288) so that they are associated with the app and release, but these are not formation jobs.

So whilst the actual formation ID of `AppID:ReleaseID` is redundant, the extra information serves to differentiate one-off jobs from formation jobs. We could just make this `flynn-controller.formation = true` if you prefer?

titanous: Yeah, lets do the boolean flag.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #2741, opened 2016-04-25
appliance/postgresql: Don’t log transaction errors

### LINKED ISSUE #2739: Reduce postgres log verbosity
We should stop logging failed queries, the failures should be duplicated in the calling app's logs and they drown out the important logs around failovers.

titanous:
The default log level is `ERROR` which logs transaction errors to the pg log. These errors are also provided to the client, which can log them if they are relevant. Log messages that admins care about will have level of `LOG` or higher.

Closes #2739


josephglanville:
LGTM

lmars:
LGTM

PR #2756, opened 2016-04-28
discoverd: Deployment stability fixes

josephglanville:
This fixes 2 things.

One is to close the heartbeater before shutting down, this makes shutdowns an order of magnitude faster and thus considerably less disruptive.

The other is to introduce a wait between shutting down the target and starting up on deployment equal to 2x the election timeout. This is to work around an issue with the Raft library. I am working on a more permanent solution to this in my reconfigure and smarter client branch but this is a stop gap to get everything working again as the chances of the issue manifesting with this sleep in place are very low.

### DIFF
--- a/discoverd/main.go
+++ b/discoverd/main.go
@@ -135,6 +135,10 @@ func (m *Main) Run(args ...string) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

PR #2761, opened 2016-04-28
controller: Avoid creating duplicate artifacts in migration

### LINKED ISSUE #2758: "discoverd: timed out waiting for instances" failure on bootstrap
Failure on bootstrap:
```
17:45:48.092460 wait-controller wait-controller
17:46:18.094218 host.go:142: error getting controller instance: discoverd: timed out waiting for instances'
```

https://gist.github.com/5c8fbcd0103f3efe97a537780ba5377a is the debug-info from the boostrapping node.

lmars:
Although in general we shouldn't change migrations that have already been released, this migration only made it into a release a few hours ago, and does not affect clusters which successfully migrated (i.e. those that did not have multiple releases with the same `SLUG_URL` set).

Fixes #2758.


josephglanville:
LGTM

Tricky stuff migrations are.

PR #2762, opened 2016-04-28
scheduler: Include restarted jobs when counting host jobs

### LINKED ISSUE #2757: Router flapping
https://bazaarvoice.box.com/s/s4i6gy9phfvo9n2i93lh15rjj17runmb

The router on 10.100.45.56 is flapping, and apparently causing both developer-facing and user-facing issues. Doesn't look like #2638, but maybe I'm missing something.

lmars:
`job.restartTimer` is non-nil for any job that was a restart of another job, even when that job is running. This means that checking for a nil `job.restartTimer` causes `GetHostJobCounts` to ignore jobs which are restarts of others, regardless of whether those jobs are running or not, and thus can lead to too many jobs running on a single host.

This caused errors in a user's cluster because omni jobs were being placed unevenly, leading to the surplusOmniJobs routine stopping them, but then anti-entropy then placing the job unevenly again, leading to the omni job flapping indefinitely.

Fixes #2757.

### DIFF
--- a/controller/scheduler/job.go
+++ b/controller/scheduler/job.go
@@ -202,7 +202,7 @@ func (s sortJobs) Sort()              { sort.Sort(s) }
 func (j Jobs) GetHostJobCounts(key utils.FormationKey, typ string) map[string]int {
 	counts := make(map[string]int)
 	for _, job := range j {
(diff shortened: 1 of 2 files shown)


titanous:
LGTM

Issue #2750, opened 2016-04-27
Garbage Collection


lmars:
We need a garbage collector to automatically remove the following potentially stale items:
- Releases + Slugs (these can currently be removed manually with `flynn release delete <id>`)
- Volumes
- Jobs (these remain in `flynn-host` indefinitely and are retrieved every 30s by the scheduler in SyncJobs)

One approach for releases is to only keep the most recent N versions, but as @philiplb mentioned in #1569:

> Keeping the N last versions might be troublesome in situations were you deploy N times in a row (shit happens). Now you can't rollback to the version from yesterday.
> 
> Proposal: Keep the last N versions and delete the other versions only if they reach a certain, configurable minimum age.

Once we merge #2744, we should have a cron-like component that makes periodic job requests to do things like this.

lmars:
FYI not cleaning up old jobs lead to a scheduler bug: #3110.

The bug is fixed, but it does seem like wasted effort to be getting the status of thousands of stopped jobs every 30s, especially if they are already marked as stopped in the controller.

lmars:
A large number of stopped jobs (e.g. 10,000) leads to quite high average bandwidth between hosts as each scheduler fetches ~10MB of jobs from each host every 30s, leading to about 6mbps average.

We should at the very least have a `flynn-host` command to just delete old jobs from the local state database.

mhlz:
Is there any progress on the command to delete old jobs? We have an app that failed to start over a day or so in our development environment which makes the online log view impossible to open (browser hangs and sometimes outright crashes). It's not a big deal since we can still see the logs with `flynn log` but it'd be nice to get this fixed at some point.

lmars:
@mhlz we updated the scheduler to only sync active jobs in #3767 (which prevents a large number of stopped jobs slowing the scheduler down) so removing old jobs is now less of a priority (it is beneficial to keep jobs in the host to help with fixing broken clusters by resurrecting old jobs).

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

PR #2769, opened 2016-04-30
controller,worker: Add garbage collection to delete old slug releases

lmars:
This adds app garbage collection to remove old slug releases based on the value of `app.Meta["gc.max_inactive_slug_releases"]`:

* if not set or set to `false` then do nothing
* if set to an integer, ensure the app only has that many releases with distinct slugs

Garbage collection cycles are scheduled after every deployment.

For now I have left it as opt in so we can test it out in the real world and then potentially make it the default in the future.

To opt in, run:

```
$ flynn meta set gc.max_inactive_slug_releases=N
```

where `N` is a positive integer.

Ref #2750.

### DIFF
--- a/cli/release.go
+++ b/cli/release.go
@@ -318,10 +318,14 @@ func runReleaseDelete(args *docopt.Args, client controller.Client) error {
 			return nil
 		}
 	}
-	res, err := client.DeleteRelease(releaseID)
+	res, err := client.DeleteRelease(mustApp(), releaseID)
 	if err != nil {
 		return err
 	}
-	log.Printf("Deleted release %s (deleted %d files)", releaseID, len(res.DeletedFiles))
+	if len(res.RemainingApps) > 0 {
+		log.Printf("Release scaled down for app but not fully deleted (still associated with %d other apps)", len(res.RemainingApps))
+	} else {
+		log.Printf("Deleted release %s (deleted %d files)", releaseID, len(res.DeletedFiles))
+	}
 	return nil
 }
--- a/controller/app.go
+++ b/controller/app.go
@@ -303,6 +303,23 @@ func (c *controllerAPI) DeleteApp(ctx context.Context, w http.ResponseWriter, re
 	}
 }
 
+func (c *controllerAPI) ScheduleAppGarbageCollection(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	gc := &ct.AppGarbageCollection{AppID: c.getApp(ctx).ID}
+	args, err := json.Marshal(gc)
+	if err != nil {
+		respondWithError(w, err)
+		return
+	}
+
+	job := &que.Job{Type: "app_garbage_collection", Args: args}
+	if err := c.que.Enqueue(job); err != nil {
+		respondWithError(w, err)
+		return
+	}
+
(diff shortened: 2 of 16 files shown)


titanous:
Does this take into account the fact that multiple apps can reference the same release? We should only garbage collect if the maximum GC parameter across all referencing releases allows it.

lmars:
@titanous releases cannot be deleted if they are currently referenced as an app release (enforced by [this trigger](https://github.com/flynn/flynn/blob/master/controller/schema.go#L351-L361) in the database) but yes there is a chance that a given release could be within the last N of another app, but there is no reliable way to determine the last N releases for a given app, we are currently just relying on the fact that slug releases are most likely created in order for a single app.

I do wonder whether we need the flexibility of releases belonging to multiple apps, the situation would be simpler if releases belong to a single app and we just copy them if we want them to relate to another app. WDYT?

titanous:
Yeah, copying probably makes sense. There should be few actual instances of this, as it's only available via the API.

titanous:
It looks like this is currently just called generic "garbage collection" throughout the type names, but is scoped to releases/artifacts of a single app. Should we call it "App Garbage Collection" or "App Release GC" or similar? We may have GC for other things in the controller in the future.

lmars:
@titanous I renamed things to "App Garbage Collection" and I also changed release deletion so that it is namespaced by app and only deletes the release if there are no formations for other apps.

I've made a breaking change to the controller API (moving `DELETE /releases/:release_id` to `DELETE /apps/:app_id/releases/:release_id`), but given we only added the route in the last week or so and it is not in a stable release, I don't think we need to support backwards compatibility.

titanous:
Does this ignore releases with formations currently deployed? Should the parameter be called `gc.max_inactive_slug_releases`?

lmars:
@titanous 

> Does this ignore releases with formations currently deployed? Should the parameter be called gc.max_inactive_slug_releases?

It did not, so I have updated the worker to only delete scaled down releases and renamed the metadata to `gc.max_inactive_slug_releases` as suggested.

I have also updated release deletion so that artifacts are only deleted once all associated releases have been deleted.

The test has been updated to handle the above two cases.

titanous:
LGTM

dottodot:
Can you clarify if this will help with the ever increasing disk use issue. Also I'm still unsure how to tell when something has been added to the stable branch or not.

lmars:
@dottodot 

> Can you clarify if this will help with the ever increasing disk use issue

It should do yes.

> I'm still unsure how to tell when something has been added to the stable branch or not

Yes this is currently not very well exposed, I will open an issue to discuss approaches for making this information more available.

This particular change is not yet in stable (currently `v20160512.0`) but it did appear in nightly from version `v20160515.0`.

dottodot:
Ok might have to give the nightly a go then as my only issue with flynn at the moment is I have to build a new cluster every week as it keep running out of space when pushing lots of updates.

lmars:
@dottodot with the latest stable version, you can actually delete old releases, it is just a manual process using `flynn release delete <id>` (rather than being done for you by the changes merged here).

dottodot:
Ah brillant, hadn't realised that.

PR #2770, opened 2016-04-30
Remove grohl

josephglanville:
Removes the grohl logging framework and converts users to log15.

Refs #34


host/attach.go
```diff
@@ -52,9 +55,8 @@ func (h *attachHandler) ServeHTTP(w http.ResponseWriter, req *http.Request, _ ht
 
 func (h *attachHandler) attach(req *host.AttachReq, conn io.ReadWriteCloser) {
 	defer conn.Close()
-
-	g := grohl.NewContext(grohl.Data{"fn": "attach", "job.id": req.JobID})
-	g.Log(grohl.Data{"at": "start"})
+	log := h.logger.New("fn", "attach")
+	log.Info("starting", "job.id", req.JobID)
```
lmars: It would be useful for the job ID to be in all log messages (so stick it in the call to `logger.New`)

lmars:
LGTM pending comment

PR #2779, opened 2016-05-04
Cedarish updates

titanous:


lmars:
We don't currently have a strategy, but we need to update the dependent images with a tag so that they use the updated cedarish image.

The easiest thing would be to build it locally and push it to Docker Hub with a tag like `flynn/cedarish:20160504` then use that as a reference.

titanous:
In the past I've just cleared the Docker cache on the build machine, which should result in the right references, right?

lmars:
@titanous I guess that works too, but I think using an explicit tag gives us a bit more confidence that it has been dealt with when reviewing a PR.

lmars:
LGTM

josephglanville:
LGTM

lmars:
FYI I built the cedarish image manually on the build machine, tagged it as both `flynn/cedarish:latest` and `flynn/cedarish:20160510` and pushed to Docker Hub.

PR #2781, opened 2016-05-05
MariaDB: Scale up on first resource request

### LINKED ISSUE #2595: MariaDB should not start by default

josephglanville:
This changes the MariaDB appliance to scale up the database formation only after the first resource is requested. This saves cluster resources in the case MariaDB goes unutilized.

Fixes #2595


titanous:
LGTM

Issue #2760, opened 2016-04-28
Unrecoverable cluster after restart


quentez:
In order to perform system updates on the different nodes of a cluster, I restarted them one by one. I didn't wait for confirmation that the cluster was in a healthy state in-between restarts, assuming that it would recover eventually (I wasn't worried about a few minutes of downtime, since this wasn't a production system).

After all the nodes were restarted, the cluster never fully came back online. It was stuck trying to start the controller app on one of the nodes (after a while there was 10+ instances of the controller running on that node). I ended up having to boot a whole new cluster.

This is the main error that was logged again and again as the cluster was trying to recover: https://gist.github.com/quentez/2445a9fbbd854f16151637e8488961f5 It never moved past that state.

You can find the full logs for all the 4 nodes here: https://db.tt/m6BQnzWF

lmars:
@quentez I've had a good look through the logs and I did find a couple of bugs (fixed in #2790) but these did not necessarily lead to the cluster not recovering.

The underlying issue was that, even though postgres jobs were running, the postgres cluster never recovered, so the controller was just hanging on startup [waiting for postgres](https://github.com/flynn/flynn/blob/v20160506.0/controller/controller.go#L61) (hence the continual `did not find any controller api instances` errors).

The initial reboot happened between `2016-04-21T04:13:55+0000` and `2016-04-21T04:15:35+0000`, and here are some postgres logs from each of the 4 hosts in the few minutes after that time:

### `Flynn0001`

```
 t=2016-04-21T04:15:59+0000 lvl=info msg="got discoverd state" app=postgres component=peer fn=handleDiscoverdState index=483 state=true
 t=2016-04-21T04:15:59+0000 lvl=info msg="starting state evaluation" app=postgres component=peer fn=evalClusterState
 t=2016-04-21T04:15:59+0000 lvl=info msg="assuming sync role" app=postgres component=peer role=sync fn=assumeSync
 t=2016-04-21T04:15:59+0000 lvl=info msg="starting state evaluation" app=postgres component=peer fn=evalClusterState
 t=2016-04-21T04:15:59+0000 lvl=info msg="reconfiguring database" app=postgres component=peer fn=applyConfig
 t=2016-04-21T04:15:59+0000 lvl=dbug msg="starting database" app=postgres component=peer fn=applyConfig
 t=2016-04-21T04:15:59+0000 lvl=info msg="starting up as standby" app=postgres component=postgres fn=assumeStandby upstream=100.100.65.2:5432
[...]

PR #2790, opened 2016-05-06
Scheduler fixes for shut down hosts

lmars:
When a host is shutting down, we need to make sure we don't restart omni jobs on other hosts, and also prepare for jobs on that host stopping.

Evidence of races were found in the logs of #2760.


titanous:
LGTM

PR #2794, opened 2016-05-07
test: Trigger a more reliable error in TestAddFailingJob

lmars:
We were previously relying on the artifact pull from example.com to return a 404, thus generating a "repo not found" error, but example.com recently starting returning a 500 instead. We probably shouldn't be relying on the reliability of example.com :wink:


titanous:
LGTM

Issue #2797, opened 2016-05-07
test: ControllerSuite.TestAppDeleteCleanup failed


lmars:
```
20:51:13.074 FAIL: test_controller.go:330: ControllerSuite.TestAppDeleteCleanup

++ 20:50:58.318 /bin/cp -r apps/http /tmp/check-605394647632969758/0/repo

++ 20:50:58.320 /usr/bin/git init
Initialized empty Git repository in /tmp/check-605394647632969758/0/repo/.git/

++ 20:50:58.323 /usr/bin/git add .

++ 20:50:58.325 /usr/bin/git commit -am init
[master (root-commit) 4bdaedb] init
 3 files changed, 22 insertions(+)
 create mode 100644 Godeps/Godeps.json
 create mode 100644 Procfile
 create mode 100644 main.go

++ 20:50:58.328 /home/ubuntu/go/src/github.com/flynn/flynn/test/../cli/bin/flynn create app-delete-cleanup-22d38b8f
Created app-delete-cleanup-22d38b8f

++ 20:50:58.384 /usr/bin/git push flynn master
remote: -----> Building app-delete-cleanup-22d38b8f...        
remote: -----> Go app detected        
remote: -----> Checking Godeps/Godeps.json file.        
remote:  !             
remote:  !     Deprecated version of go (go1.4.2)        
remote:  !     See https://devcenter.heroku.com/articles/go-support#go-versions for supported version information.        
remote:  !             
remote: -----> Installing go1.4.2... done        
remote: -----> Running: godep go install -tags heroku .        
remote: -----> Discovering process types        
remote:        Procfile declares types -> web, another-web        
remote: -----> Compiled slug size is 1.7M        
remote: -----> Creating release...        
remote: =====> Application deployed        
remote: =====> Waiting for web job to start...        
remote: Failed to scale web process type        
To https://git.flynn-a1bde3de3ce1a59f.local/app-delete-cleanup-22d38b8f.git
 ! [remote rejected] master -> master (pre-receive hook declined)
error: failed to push some refs to 'https://git.flynn-a1bde3de3ce1a59f.local/app-delete-cleanup-22d38b8f.git'

test_controller.go:337:
    t.Assert(r.git("push", "flynn", "master"), Succeeds)
[...]

PR #2798, opened 2016-05-07
host/cli: Include artifact URIs in `flynn-host inspect` output

lmars:
This will help diagnose #2797 (i.e. is the image artifact set as the file artifact and vice versa, or does the job just have an image artifact and no file artifacts).


titanous:
LGTM

Issue #2763, opened 2016-04-29
Docker receive app


lmars:
This is a proposal for adding a `docker-receive` app which would receive Docker images from clients and deploy them as a Flynn release.

The app would have two process types:

* `registry` - this would be the [Docker registry](https://github.com/docker/distribution/blob/master/docs/index.md) with a blobstore [storage driver](https://github.com/docker/distribution/blob/master/docs/storagedrivers.md)
* `api` - this would receive [notifications from the registry](https://github.com/docker/distribution/blob/master/docs/notifications.md) when images are pushed and create image artifacts in the controller

We would add a Flynn CLI command to wrap a call to `docker push`, subscribe to artifact events and deploy a release. We would need to come up with a way of defining the process types for the release (could be part of the CLI command, read from a local file or extracted from the Docker image).

/cc @titanous @jvatic @josephglanville thoughts?

titanous:
This probably depends on v2 registry support, right? (#2196)

lmars:
> This probably depends on v2 registry support, right? (#2196)

Ah yes forgot to mention, yes that is a prerequisite.

lmars:
FYI I have a branch with an initial pass at implementing this: [docker-receive](https://github.com/flynn/flynn/compare/docker-receive).

So far I have made the following changes:

* Added a `docker-receive` app which uses the Docker registry with a [custom blobstore storage driver](https://github.com/flynn/flynn/blob/docker-receive/docker-receive/blobstore/blobstore.go) + repository middleware to [create artifacts when manifests are pushed](https://github.com/flynn/flynn/blob/docker-receive/docker-receive/main.go#L90-L106) 
* Added directory support to the blobstore (needed to implement [the storage driver `List` method](https://github.com/docker/distribution/blob/master/registry/storage/driver/storagedriver.go#L69-L71))
* Migrated pinkerton to pull from a V2 registry

There are quite a few TODOs in there, but from a high level it does what we need (see [the test](https://github.com/flynn/flynn/blob/docker-receive/test/test_docker_receive.go) :smile: ).

I'll probably clean up and make some incremental PRs soon.

lmars:
[...]

PR #2800, opened 2016-05-09
Blobstore enhancements

lmars:
A number of enhancements to the blobstore to aid in supporting a blobstore backend for a Docker registry (see #2763).

The directory indexes (which are GET requests to `/` with a `dir` param) are simpler versions of listing objects with a prefix from S3 (see [here](http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html#v2-RESTBucketGET-responses-examples-ex3)), and copying files (using the `X-Blobstore-Copy-From` header) is similar to [copying an object in S3](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html), which means we should still be able to make the blobstore an S3 gateway.

### DIFF
--- a/blobstore/README.md
+++ b/blobstore/README.md
@@ -7,12 +7,51 @@ files. It is like a simpler S3. All operations fall under these three HTTP
 verbs:
 
  * PUT: write a file: `curl -X PUT -T /path/to/local/file
-   http://blobstorehost/path/to/remote/file`
- * GET: read a file: `curl http://blobstorehost/path/to/remote/file`
+   http://blobstore.discoverd/path/to/remote/file`
+ * GET: read a file: `curl http://blobstore.discoverd/path/to/remote/file`
  * DELETE: delete a file: `curl -X DELETE
-   http://blobstorehost/path/to/remote/file`
+   http://blobstore.discoverd/path/to/remote/file`
 
-There are no directory indexes. Parent directories are automatically created.
+Files can be copied by setting the `Blobstore-Copy-From` header in a PUT
+request:
+
+```shell
+# create /file1.txt
+echo data | curl -X PUT --data-binary @- http://blobstore.discoverd/file1.txt
+
+# copy it to /file2.txt
+curl -X PUT --header "Blobstore-Copy-From: /file1.txt" http://blobstore.discoverd/file2.txt
+
+# read /file2.txt
+curl http://blobstore.discoverd/file2.txt
+data
+
+```
+
+
+Parent directories are automatically created, and directory indexes can be
+accessed with a GET request to the root path with the `dir` query param:
+
+```shell
+# create 4 files in different directories
+curl -X PUT --data-binary "data" http://blobstore.discoverd/foo.txt
+curl -X PUT --data-binary "data" http://blobstore.discoverd/dir1/foo.txt
+curl -X PUT --data-binary "data" http://blobstore.discoverd/dir2/foo.txt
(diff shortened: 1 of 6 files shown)


blobstore/README.md
```diff
@@ -7,12 +7,51 @@ files. It is like a simpler S3. All operations fall under these three HTTP
...
  * PUT: write a file: `curl -X PUT -T /path/to/local/file
-   http://blobstorehost/path/to/remote/file`
- * GET: read a file: `curl http://blobstorehost/path/to/remote/file`
+   http://blobstore.discoverd/path/to/remote/file`
+ * GET: read a file: `curl http://blobstore.discoverd/path/to/remote/file`
  * DELETE: delete a file: `curl -X DELETE
-   http://blobstorehost/path/to/remote/file`
+   http://blobstore.discoverd/path/to/remote/file`
+
+Files can be copied by setting the `X-Blobstore-Copy-From` header in a PUT
```
titanous: The `X-` prefix is deprecated.

blobstore/README.md
```diff
@@ -7,12 +7,51 @@ files. It is like a simpler S3. All operations fall under these three HTTP
...
+# list all top-level files and directories
+curl http://blobstore.discoverd/
+["/dir1/","/dir2/","/dir3/","/foo.txt"]
+
+# list files in /dir1
+curl http://blobstore.discoverd/?dir=/dir1
+["/dir1/foo.txt"]
+
+# list files in /dir2
+curl http://blobstore.discoverd/?dir=/dir2
```
titanous: Why a separate URL param and not just `GET /dir2`?

lmars: So my first pass at this did have `GET /dir` return a directory listing but this lead to the need to handle the following:

* don't allow both `/dir` and `/dir/foo.txt` to be regular files (since then `GET /dir` would be ambiguous)
* update postgres to know whether `GET /dir` should lookup a file or do a prefix search

The solution I came up with was to model directories explicitly in postgres by adding a `dir` column and adding a `MkdirAll` method (see [this commit](https://github.com/flynn/flynn/commit/d7a76e331aff67b7e2446bcc6d02c996b57f67b8)), but then I realised that S3 doesn't model directories like this, so if we want the blobstore to become a gateway to S3 like services, we should probably stick closer to their API, hence why I went with `GET /` for directory listings and `GET /path` to get files.

It is also much simpler (although if you think there is a simpler way to accomplish your approach without the complexity in my commit, let me know).

titanous: Okay, that's fine. I've never like REST anyway :)

lmars: :wink:

lmars:
@titanous comments addressed

blobstore/postgres_filesystem.go
```diff
@@ -105,9 +125,17 @@ create:
...
+		return err
+	}
+	defer src.Close()
+	return p.Put(dstPath, src, "")
+}
+
 func (p *PostgresFilesystem) Delete(name string) error {
-	err := p.db.Exec("DELETE FROM files WHERE name = $1", name)
-	return err
+	return p.db.Exec("DELETE FROM files WHERE name LIKE $1 || '%'", name)
```
titanous: Hmm, not sure about this. Won't `DELETE /foo` end up deleting `/foo.txt`?

lmars:
@titanous PostgreSQL Delete method updated to use a regular expression and test added.

titanous:
LGTM

PR #2816, opened 2016-05-11
appliance/mariadb: Skip controller checks if database is Read-Write

titanous:
This avoids a hard dependency on the controller being up. The controller is not up during cluster restores, so before this patch restores that included a MySQL dataset could not proceed.

/cc @josephglanville


josephglanville:
LGTM

PR #2817, opened 2016-05-11
worker: Don't retry deploys when ErrSkipRollback is returned

### LINKED ISSUE #2807: Deployment rollback ends up scaling app down
I'm not sure exactly what caused this, but a failed deploy on @tjboring's cluster ended up rolling back to a null formation instead of the correct previous scale.

Here are the logs from the worker and the recent events for the app: https://gist.github.com/titanous/330f393f20d66b8e508601d7964e4e67

I have the rest of the cluster logs, let me know if you need anything else.

lmars:
Fixes #2807.


titanous:
LGTM

PR #2824, opened 2016-05-13
pinkerton: Support pulling from V2 Docker registry

### LINKED ISSUE #2196: pinkerton: Support Docker registry V2
We currently expect to pull Docker images from a Docker V1 registry (see [here](https://github.com/flynn/flynn/blob/master/pinkerton/registry/docker.go)) but V1 has been deprecated.

We should support pulling from a V2 registry, perhaps vendoring [some of the Docker code](https://github.com/docker/docker/blob/master/registry/session.go).

lmars:
I have removed a lot of our code for pulling from a Docker registry and switched to using [TagStore.Pull from Docker 1.9.1](https://github.com/docker/docker/blob/v1.9.1/graph/pull.go#L63-L65). Our code for pulling from a TUF repository remains.

We need to stick with Docker 1.9.1 for now because later versions have a different on-disk image format, so requires more refactoring of pulling TUF images and checking out images (see #2459).

With the new Docker code, fixing #509 is probably straight forward, but I will leave that for a subsequent PR (tags still work as they did previously).

I have added a test which starts a Docker registry with some test files, pulls an image, checks it out and verifies the check out has the correct files.

Further testing will be added as part of adding the `docker-receive` app (#2763), but I have done some manual testing. Both `pinkerton pull` and `flynn release add` work with URIs like:

```
https://docker.io?name=ubuntu&tag=wily
https://docker.io?name=ubuntu&id=sha256:4094004b00ef9e67ffd5b537673295da2f25fb9a73ea015437c3bd1d288fb1e8
https://lmars:xxx@docker.io?name=lmars/private&id=sha256:xxx
```

Closes #2196.


titanous:
LGTM

PR #2829, opened 2016-05-13
appliance/postgres: Tune WAL checkpoint settings

josephglanville:
Improves performance when database is doing a large amount of writes by allowing postgres more time to complete snapshots and to perform them less frequently.


titanous:
LGTM

PR #2835, opened 2016-05-16
controller: Don't include deleted formations in formation stream

lmars:
Schedulers in a cluster restored from a backup were failing to stream formation updates with the following error:

```
t=2016-05-16T15:51:50+0000 lvl=info msg="connecting formation event stream" component=scheduler fn=streamFormationEvents
t=2016-05-16T15:51:50+0000 lvl=warn msg="formation event stream disconnected" component=scheduler fn=streamFormationEvents err="Server error: controller: resource not found"
```

The stream was being closed because the controller was trying to expand deleted formations, hitting `ErrNotFound` when looking up the associated release and returning that to the stream.

I have updated the `formation_list_since` query (along with some others) to check for `deleted_at IS NULL`.

### DIFF
--- a/controller/formation_test.go
+++ b/controller/formation_test.go
@@ -71,6 +71,55 @@ func (s *S) TestFormationStreaming(c *C) {
 	c.Assert(out.Processes, IsNil)
 }
 
+func (s *S) TestFormationStreamDeleted(c *C) {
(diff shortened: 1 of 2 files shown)


josephglanville:
Good catch, LGTM.

PR #2843, opened 2016-05-19
MariaDB recovery

josephglanville:
This changset introduces attempted recovery of MariaDB via both the CLI fix tool and the cluster monitor.


lmars:
LGTM

PR #2845, opened 2016-05-19
blobstore: Support non-zero file offset for PUT requests

lmars:
This is necessary to support [this interface](https://github.com/docker/distribution/blob/6ba799b69ed19ebc0c8ef8325ccaf5d105facfc5/registry/storage/driver/storagedriver.go#L58-L62) when using the blobstore as a backend for a Docker registry.


josephglanville:
LGTM

PR #2846, opened 2016-05-19
controller: Fix TestStreamAppLifeCycleEvents

lmars:
This test intermittently fails in CI as it gets events generated by other tests, so I have updated it to ignore those other events.


lmars:
Hmm the test is still failing, I'll look into it some more...

lmars:
Ok updated, please review.

josephglanville:
LGTM

PR #2847, opened 2016-05-19
script: Add script to configure Docker with a cluster's TLS CA certificate

lmars:
Part of the docker-receive changes (#2763) but opening as a separate PR so we can get the tests to pass in CI.


titanous:
LGTM

PR #2857, opened 2016-05-22
Implement Docker Receive App

### LINKED ISSUE #2763: Docker receive app

lmars:
This adds a `docker-receive` app which receives Docker images from `docker push` and creates corresponding artifacts, and `flynn docker push` which also deploys them to the cluster.

Running `flynn cluster add ...` will attempt to configure the local Docker client using `docker login`, but this can fail if the Docker daemon does not trust the Flynn cluster's CA certificate, in which case a warning is printed like:

```
WARN: docker configuration failed with a TLS error.
WARN:
WARN: Copy the TLS CA certificate /home/ubuntu/.flynn/ca-certs/default.pem
WARN: to /etc/docker/certs.d/docker.1.localflynn.com/ca.crt
WARN: on the docker daemon's host and restart docker.
```

Following the instructions then re-running the `flynn cluster add ...` command (or `flynn docker login`) should work.

Assuming things are configured, here is an example of deploying elasticsearch:

```
$ docker pull elasticsearch:2.3.3

$ flynn create --remote "" elasticsearch
Created elasticsearch

$ flynn -a elasticsearch docker push elasticsearch:2.3.3
flynn: getting image config with "docker inspect -f {{ json .Config }} elasticsearch:2.3.3"
flynn: tagging Docker image with "docker tag --force elasticsearch:2.3.3 docker.1.localflynn.com/elasticsearch:latest"
flynn: pushing Docker image with "docker push docker.1.localflynn.com/elasticsearch:latest"
The push refers to a repository [docker.1.localflynn.com/elasticsearch] (len: 1)
a8eb754d1a89: Pushed 
6162aa7ca453: Pushed 
...
4e59e6df754e: Pushed 
3059b4820522: Pushed 
latest: digest: sha256:1752ca12bbedb99734ca1ba3ec35720768a95ad83b7b6c371fc37a28b98ea351 size: 61216
flynn: image pushed, waiting for artifact creation
flynn: deploying release using artifact URI http://docker-receive.discoverd?name=elasticsearch&id=sha256:1752ca12bbedb99734ca1ba3ec35720768a95ad83b7b6c371fc37a28b98ea351
flynn: image deployed, scale it with 'flynn scale app=N'

$ flynn -a elasticsearch scale app=1
scaling app: 0=>1

18:05:38.308 ==> app fd8266b1-453d-4733-9417-2ad9ac33a418 pending
18:05:38.308 ==> app host0-fd8266b1-453d-4733-9417-2ad9ac33a418 starting
18:05:41.018 ==> app host0-fd8266b1-453d-4733-9417-2ad9ac33a418 up

scale completed in 2.728163927s

$ flynn -a elasticsearch run curl http://elasticsearch.discoverd:9200
{
  "name" : "Hate-Monger",
  "cluster_name" : "elasticsearch",
  "version" : {
    "number" : "2.3.3",
    "build_hash" : "218bdf10790eef486ff2c41a3df5cfa32dadcfde",
    "build_timestamp" : "2016-05-17T15:40:04Z",
    "build_snapshot" : false,
    "lucene_version" : "5.5.0"
  },
  "tagline" : "You Know, for Search"
}
```

Closes #2763.

### DIFF
--- a/cli/cluster.go
+++ b/cli/cluster.go
@@ -24,7 +24,7 @@ import (
 func init() {
 	register("cluster", runCluster, `
 usage: flynn cluster
-       flynn cluster add [-f] [-d] [--git-url <giturl>] [--no-git] [--docker-url <dockerurl>] [--docker] [-p <tlspin>] <cluster-name> <domain> <key>
+       flynn cluster add [-f] [-d] [--git-url <giturl>] [--no-git] [--docker-push-url <url>] [--docker] [-p <tlspin>] <cluster-name> <domain> <key>
        flynn cluster remove <cluster-name>
        flynn cluster default [<cluster-name>]
        flynn cluster migrate-domain <domain>
@@ -44,7 +44,7 @@ Commands:
             -d, --default             set as default cluster
             --git-url=<giturl>        git URL
             --no-git                  skip git configuration
-            --docker-url=<dockerurl>  Docker URL
+            --docker-push-url=<url>   Docker push URL
             --docker                  configure Docker to push to the cluster
             -p, --tls-pin=<tlspin>    SHA256 of the cluster's TLS cert
 
@@ -108,7 +108,7 @@ func runCluster(args *docopt.Args) error {
 		if gitURL == "" {
 			gitURL = "(none)"
 		}
-		dockerURL := s.DockerURL
+		dockerURL := s.DockerPushURL
 		if dockerURL == "" {
 			dockerURL = "(none)"
 		}
@@ -123,11 +123,11 @@ func runCluster(args *docopt.Args) error {
 
 func runClusterAdd(args *docopt.Args) error {
 	s := &cfg.Cluster{
-		Name:      args.String["<cluster-name>"],
-		Key:       args.String["<key>"],
-		GitURL:    args.String["--git-url"],
-		DockerURL: args.String["--docker-url"],
-		TLSPin:    args.String["--tls-pin"],
+		Name:          args.String["<cluster-name>"],
+		Key:           args.String["<key>"],
+		GitURL:        args.String["--git-url"],
+		DockerPushURL: args.String["--docker-push-url"],
+		TLSPin:        args.String["--tls-pin"],
 	}
 	domain := args.String["<domain>"]
 	if strings.HasPrefix(domain, "https://") {
@@ -138,8 +138,8 @@ func runClusterAdd(args *docopt.Args) error {
 	if s.GitURL == "" && !args.Bool["--no-git"] {
 		s.GitURL = "https://git." + domain
 	}
-	if s.DockerURL == "" && args.Bool["--docker"] {
-		s.DockerURL = "https://docker." + domain
+	if s.DockerPushURL == "" && args.Bool["--docker"] {
+		s.DockerPushURL = "https://docker." + domain
 	}
 
 	if err := config.Add(s, args.Bool["--force"]); err != nil {
@@ -153,7 +153,7 @@ func runClusterAdd(args *docopt.Args) error {
 	}
 
 	var caPath string
-	if s.GitURL != "" || s.DockerURL != "" {
+	if s.GitURL != "" || s.DockerPushURL != "" {
 		client, err := s.Client()
 		if err != nil {
 			return err
@@ -176,8 +176,8 @@ func runClusterAdd(args *docopt.Args) error {
 		}
 	}
 
-	if s.DockerURL != "" {
-		host, err := s.DockerHost()
+	if s.DockerPushURL != "" {
+		host, err := s.DockerPushHost()
 		if err != nil {
 			return err
 		}
@@ -236,7 +236,7 @@ func runClusterRemove(args *docopt.Args) error {
 
 		cfg.RemoveGlobalGitConfig(c.GitURL)
 
-		if host, err := c.DockerHost(); err == nil {
+		if host, err := c.DockerPushHost(); err == nil {
 			dockerLogout(host)
 		}
 
@@ -336,7 +336,7 @@ func runClusterMigrateDomain(args *docopt.Args) error {
 				cluster.TLSPin = dm.TLSCert.Pin
 				cluster.ControllerURL = fmt.Sprintf("https://controller.%s", dm.Domain)
 				cluster.GitURL = fmt.Sprintf("https://git.%s", dm.Domain)
-				cluster.DockerURL = fmt.Sprintf("https://docker.%s", dm.Domain)
+				cluster.DockerPushURL = fmt.Sprintf("https://docker.%s", dm.Domain)
 				if err := config.SaveTo(configPath()); err != nil {
 					return fmt.Errorf("Error saving config: %s", err)
 				}
@@ -357,7 +357,7 @@ func runClusterMigrateDomain(args *docopt.Args) error {
 
 				// try to run "docker login" for the new domain, but just print a warning
 				// if it fails so the user can fix it later
-				if host, err := cluster.DockerHost(); err == nil {
+				if host, err := cluster.DockerPushHost(); err == nil {
 					if err := dockerLogin(host, cluster.Key); err == ErrDockerTLSError {
 						printDockerTLSWarning(host, caFile.Name())
 					}
--- a/cli/cluster.go
+++ b/cli/cluster.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"log"
+	"net/url"
 	"os"
 	"os/exec"
 	"strings"
@@ -24,7 +26,7 @@ import (
 func init() {
 	register("cluster", runCluster, `
 usage: flynn cluster
-       flynn cluster add [-f] [-d] [--git-url <giturl>] [--no-git] [-p <tlspin>] <cluster-name> <domain> <key>
+       flynn cluster add [-f] [-d] [--git-url <giturl>] [--no-git] [--docker-url <dockerurl>] [--docker] [-p <tlspin>] <cluster-name> <domain> <key>
        flynn cluster remove <cluster-name>
        flynn cluster default [<cluster-name>]
        flynn cluster migrate-domain <domain>
@@ -44,6 +46,8 @@ Commands:
             -d, --default             set as default cluster
             --git-url=<giturl>        git URL
             --no-git                  skip git configuration
+            --docker-url=<dockerurl>  Docker URL
+            --docker                  configure Docker to push to the cluster
             -p, --tls-pin=<tlspin>    SHA256 of the cluster's TLS cert
 
     remove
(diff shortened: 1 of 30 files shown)


lmars:
FYI I have moved request authorization into `httphelper.IsAuthorized` and in doing so replaced [this explicit check for `/ca-cert`](https://github.com/flynn/flynn/blob/v20160520.0/controller/controller.go#L322) with [this one](https://github.com/flynn/flynn/blob/docker-receive/controller/controller.go#L320), but this lead to a failure of [this test](https://github.com/flynn/flynn/blob/v20160520.0/controller/controller_test.go#L558-L565).

I am unsure why we allow users to get the CA cert without authorization, but then return a 401 if they request it with an invalid key (my change just doesn't do an auth check for the CA cert path), so I have just removed the test.

@jvatic you introduced the logic in 4549ee4, do we need to return a 401 if the key is invalid, but serve it if there is no key provided?

jvatic:
We return a 401 so that users don't get a false-positive when using an invalid key even though `/ca-cert` is an unprotected endpoint. The `key` URL param was also limited to event streams and the `/backup` endpoint as basic auth can reasonably be used everywhere for all other requests.

lmars:
> We return a 401 so that users don't get a false-positive when using an invalid key even though /ca-cert is an unprotected endpoint.

Hmm I don't really see why we need to do that, if `/ca-cert` is unprotected, why would the user send an `Authorization` header? I guess we also set `WWW-Authenticate` in the response, but the user doesn't need to authenticate, so why not just give them the cert?

> The key URL param was also limited to event streams and the /backup endpoint as basic auth can reasonably be used everywhere for all other requests.

Is there any harm in accepting either the header or the query param for all protected paths rather than special casing?

lmars:
P.S. this discussion is not really related to this PR, let me know if you would rather I separated that change out into a different PR.

jvatic:
Hmm... I can't think of any strong reason to keep it the way it was.

cli/cluster.go
```diff
@@ -138,22 +144,42 @@ func runClusterAdd(args *docopt.Args) error {
 		return errors.New(fmt.Sprintf("Cluster %q does not exist and cannot be set as default.", s.Name))
 	}
 
+	var caPath string
+	if !args.Bool["--no-git"] && !args.Bool["--no-docker"] {
```
titanous: I think this should be an *or* condition, not an *and*?

cli/cluster.go
```diff
@@ -138,22 +144,42 @@ func runClusterAdd(args *docopt.Args) error {
...
+	if !args.Bool["--no-docker"] {
+		host, err := s.DockerHost()
 		if err != nil {
-			return fmt.Errorf("Error writing CA certificate: %s", err)
+			return err
 		}
-		if err := cfg.WriteGlobalGitConfig(s.GitURL, caPath); err != nil {
+		if err := dockerLogin(host, s.Key); err != nil {
+			if e, ok := err.(*exec.Error); ok && e.Err == exec.ErrNotFound {
+				err = errors.New("Executable 'docker' was not found. Use --no-docker to skip docker configuration")
```
titanous: Will this default to erroring out if Docker is not installed? We should definitely not do that.

lmars: I decided to treat this the same as configuring git: unless you specify `--no-docker`, an error will be returned if Docker configuration fails.

Do you think we should default to configuring Docker but ignore configuration errors (just print a warning and move on), or perhaps switch it so that Docker configuration is only done if `--docker-url` (or `--docker`) is specified?

titanous: I think we should only do it if the flag is specified for now.

cli/cluster.go
```diff
@@ -44,6 +44,8 @@ Commands:
             -d, --default             set as default cluster
             --git-url=<giturl>        git URL
             --no-git                  skip git configuration
+            --docker-url=<dockerurl>  Docker URL
```
titanous: Hmm, I think that this may be confused for a Docker daemon URL, any ideas for disambiguation?

lmars: How about `--docker-push-url`?

titanous: SGTM.

lmars:
@titanous comments addressed

cli/cluster.go
```diff
@@ -113,20 +123,24 @@ func runCluster(args *docopt.Args) error {
...
 		s.ControllerURL = domain
 	} else {
 		s.ControllerURL = "https://controller." + domain
 	}
-	if s.GitURL == "" {
+	if s.GitURL == "" && !args.Bool["--no-git"] {
 		s.GitURL = "https://git." + domain
 	}
+	if s.DockerPushURL == "" && args.Bool["--docker"] {
+		s.DockerPushURL = "https://docker." + domain
```
titanous: If we only set this when the `--docker` flag is set, we should probably have another command that configures Docker so that you can set it up later? (and prompt if not configured and using a Docker command)

pkg/httphelper/httphelper.go
```diff
@@ -254,3 +256,58 @@ func DecodeJSON(req *http.Request, i interface{}) error {
...
+
+func mustParseCIDR(s string) *net.IPNet {
+	_, n, err := net.ParseCIDR(s)
+	if err != nil {
+		panic(fmt.Sprintf("ParseCIDR(%q): %v", s, err))
+	}
+	return n
+}
+
+func IsInternal(req *http.Request) bool {
```
titanous: This is named incorrectly, it should be `IsFromPrivateIP` or similar.

docker-receive/auth.go
```diff
@@ -0,0 +1,47 @@
...
+// Authorized implements the auth.AccessController interface and authorizes a
+// request if it is either considered to be internal (e.g. image pulls using
+// docker-receive.discoverd) or includes the correct basic auth key
+func (a *Auth) Authorized(ctx context.Context, accessRecords ...auth.Access) (context.Context, error) {
+	req, err := context.GetRequest(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if !httphelper.IsInternal(req) && !httphelper.IsAuthorized(req, []string{a.key}) {
```
titanous: Using `IsInternal` is not safe here, this would allow access from the rest of EC2 if security groups are misconfigured (or the Digital Ocean private network, and so on). Should we just put the auth details in the artifact URL for now?

host/libvirt_lxc_backend.go
```diff
@@ -668,6 +675,36 @@ func (l *LibvirtLXCBackend) Run(job *host.Job, runConfig *RunConfig) (err error)
 	return nil
 }
 
+// maybeResolveDiscoverdURI resolves a discoverd host in the given URI to an
+// address using the configured discoverd URL as the host is likely not using
+// discoverd to resolve DNS queries
+func (l *LibvirtLXCBackend) maybeResolveDiscoverdURI(uri string) (string, error) {
```
titanous: `maybe` should only be used in a function name when it panics if there is an error.

test/test_docker_receive.go
```diff
@@ -0,0 +1,72 @@
...
+	stream, err := client.StreamEvents(ct.StreamEventsOptions{
+		ObjectTypes: []ct.EventType{ct.EventTypeArtifact},
+	}, events)
+	t.Assert(err, c.IsNil)
+	defer stream.Close()
+
+	// push the Docker image to docker-receive
+	//
+	// TODO: use cluster.DockerPushURL once CI configures it properly
+	// u, err := url.Parse(s.clusterConf(t).DockerPushURL)
```
titanous: Ditto.

docker-receive/blobstore/blobstore.go
```diff
@@ -0,0 +1,223 @@
...
+	n int64
+}
+
+func (r *readCounter) Read(p []byte) (int, error) {
+	n, err := r.r.Read(p)
+	atomic.AddInt64(&r.n, int64(n))
+	return n, err
+}
+
+func (d *driver) WriteStream(ctx context.Context, path string, offset int64, reader io.Reader) (int64, error) {
```
titanous: Is there a reason we're not using the new `FileWriter` interface? https://github.com/docker/distribution/commit/ff03381d4906d9d90dd92b1ad1050a63147464c7

brettneese:
Any ETA on when this will get merged? It's something we're really eager to see in our Flynn cluster soon. ;-)

cli/cluster.go
lmars: I've added a `flynn docker set-push-url` command: 83caf95

test/test_cli.go
```diff
@@ -1254,3 +1254,52 @@ func (s *CLISuite) TestSlugReleaseGarbageCollection(t *c.C) {
...
+		`ENV FOO=BAR`,
+		`CMD ["/bin/pingserv"]`,
+	)
+
+	// create app
+	client := s.controllerClient(t)
+	app := &ct.App{Name: "cli-test-docker-push"}
+	t.Assert(client.CreateApp(app), c.IsNil)
+
+	// TODO: remove this once CI configures DockerPushURL correctly
```
lmars: Well, this PR changes CI so that it sets `DockerPushURL`, but it won't be in use until we merge and restart CI, so I have added a hack to get the tests to pass with a plan to drop the commit just before merge.

Not ideal but after a couple of tries, I couldn't get CI to support both master and this branch.

Also note I have moved the two hacks into a single place: d537450

docker-receive/blobstore/blobstore.go
lmars: We are pinned to docker/docker 1.9.1 due to #2459. which means docker/distribution is pinned at 2.1.0, which does not include the `FileWriter` interface.

We could get around this by vendoring docker/distribution at a different version inside docker-receive/Godeps, but I don't think there is much benefit.

lmars:
@titanous comments addressed (I removed the refactor of controller / status auth, and changed docker-receive to always require auth and put the credentials in the artifact URI: 818412d)

lmars:
@brettneese 

> Any ETA on when this will get merged?

Hopefully in the next few days :smile:

cli/docker.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
+	cfg "github.com/flynn/flynn/cli/config"
+	"github.com/flynn/flynn/controller/client"
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/host/types"
+)
+
+func init() {
+	register("docker", runDocker, `
+usage: flynn docker set-push-url <url>
```
titanous: This should have a default URL (`flynn docker set-push-url` should work with no arguments)

cli/docker.go
```diff
@@ -0,0 +1,304 @@
...
+
+func runDockerLogin() error {
+	cluster, err := getCluster()
+	if err != nil {
+		return err
+	}
+	host, err := cluster.DockerPushHost()
+	if err != nil {
+		if err == cfg.ErrNoDockerPushURL {
+			err = errors.New("ERROR: Docker push URL not configured, set it with 'flynn docker set-push-url URL'")
```
titanous: drop ` URL` after adding default option (ditto below)

cli/docker.go
lmars: Do you think the default should be `s/controller/docker` on `ControllerURL`?

lmars: Or a route lookup on the app?

titanous: Yeah, the substitution is fine for now.

lmars:
@titanous I have modified `set-push-url` to use a default (83f09a7) and also updated `flynn export` to export pushed images (ee9d2e3).

imjorge:
@lmars does backup/restore contemplates the docker images?

lmars:
@imjorge since backup/restore backs up all the PostgreSQL databases, and these Docker images are being stored in PostgreSQL via the blobstore, yes a cluster backup/restore will manage these images.

titanous:
LGTM!

brettneese:
👍 👍 👍 👍 

Thanks gents.

PR #2871, opened 2016-05-25
host: Remove noisy “getting job” log lines

titanous:
These were added to track down a deadlock that is long since fixed.


lmars:
LGTM

PR #2890, opened 2016-06-03
router: Manage TLS certificates independent of routes

### LINKED ISSUE #2442: router: Move TLS certificate details to a different object
We should move the TLS certificates/keys out of the Route data structure/table and into a separate table/datatype. This sets us up to do a bunch of stuff:

- Use the same certificate for multiple routes
- Clean fallback certificates without env vars
- Allows changing system app certificates in one place
- Wildcard certificates (#267)
- OCSP stapling (#1049)
- CT enrollment and SCTs in handshakes (#1052)
- Chain verification and bundling (#1251)
- Let's Encrypt integration (#1995)

jvatic:
Closes #2442

### DIFF
--- a/router/api.go
+++ b/router/api.go
@@ -30,6 +30,11 @@ func apiHandler(rtr *Router) http.Handler {
 	r.GET("/routes", httphelper.WrapHandler(api.GetRoutes))
 	r.GET("/routes/:route_type/:id", httphelper.WrapHandler(api.GetRoute))
 	r.DELETE("/routes/:route_type/:id", httphelper.WrapHandler(api.DeleteRoute))
+	r.POST("/certificates", httphelper.WrapHandler(api.CreateCert))
+	r.GET("/certificates/:id", httphelper.WrapHandler(api.GetCert))
+	r.GET("/certificates/:id/routes", httphelper.WrapHandler(api.GetCertRoutes))
+	r.DELETE("/certificates/:id", httphelper.WrapHandler(api.DeleteCert))
+	r.GET("/certificates", httphelper.WrapHandler(api.GetCerts))
 	r.GET("/events", httphelper.WrapHandler(api.StreamEvents))
 
 	r.HandlerFunc("GET", "/debug/*path", pprof.Handler.ServeHTTP)
@@ -197,6 +202,89 @@ func (api *API) DeleteRoute(ctx context.Context, w http.ResponseWriter, req *htt
 			return
 		default:
 			log.Error(err.Error())
+			httphelper.Error(w, err)
+			return
+		}
+	}
+	w.WriteHeader(200)
+}
+
+func (api *API) CreateCert(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	var cert *router.Certificate
+	if err := json.NewDecoder(req.Body).Decode(&cert); err != nil {
+		httphelper.Error(w, err)
+		return
+	}
+
+	l := api.router.HTTP.(*HTTPListener)
+	err := l.AddCert(cert)
+	if err != nil {
+		jsonError := httphelper.JSONError{}
+		switch err {
+		case ErrConflict:
+			jsonError.Code = httphelper.ConflictErrorCode
+			jsonError.Message = "Duplicate cert"
+		case ErrInvalid:
+			jsonError.Code = httphelper.ValidationErrorCode
+			jsonError.Message = "Invalid cert"
+		default:
+			httphelper.Error(w, err)
+			return
+		}
+	}
+	httphelper.JSON(w, 200, cert)
+}
+
+func (api *API) GetCert(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	params, _ := ctxhelper.ParamsFromContext(ctx)
+
+	l := api.router.HTTP.(*HTTPListener)
+	cert, err := l.GetCert(params.ByName("id"))
+	if err == ErrNotFound {
+		httphelper.ObjectNotFoundError(w, "certificate not found")
+		return
+	}
+	if err != nil {
+		httphelper.Error(w, err)
+		return
+	}
+
+	httphelper.JSON(w, 200, cert)
+}
+
+func (api *API) GetCertRoutes(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	params, _ := ctxhelper.ParamsFromContext(ctx)
+
+	l := api.router.HTTP.(*HTTPListener)
+	routes, err := l.GetCertRoutes(params.ByName("id"))
+	if err != nil {
+		httphelper.Error(w, err)
+		return
+	}
+	httphelper.JSON(w, 200, routes)
+}
+
+func (api *API) GetCerts(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	l := api.router.HTTP.(*HTTPListener)
+	certs, err := l.GetCerts()
+	if err != nil {
+		httphelper.Error(w, err)
+		return
+	}
+	httphelper.JSON(w, 200, certs)
+}
+
+func (api *API) DeleteCert(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	params, _ := ctxhelper.ParamsFromContext(ctx)
+
+	l := api.router.HTTP.(*HTTPListener)
+	err := l.RemoveCert(params.ByName("id"))
+	if err != nil {
+		switch err {
+		case ErrNotFound:
+			httphelper.ObjectNotFoundError(w, "certificate not found")
+			return
+		default:
 			httphelper.Error(w, err)
 			return
 		}
--- a/router/data_store.go
+++ b/router/data_store.go
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/jackc/pgx"
@@ -17,10 +20,15 @@ var ErrInvalid = errors.New("router: invalid route")
 
 type DataStore interface {
 	Add(route *router.Route) error
+	AddCert(cert *router.Certificate) error
 	Update(route *router.Route) error
 	Get(id string) (*router.Route, error)
+	GetCert(id string) (*router.Certificate, error)
 	List() ([]*router.Route, error)
+	ListCerts() ([]*router.Certificate, error)
+	ListCertRoutes(id string) ([]*router.Route, error)
 	Remove(id string) error
+	RemoveCert(id string) error
 	Sync(ctx context.Context, h SyncHandler, startc chan<- struct{}) error
 	Ping() error
 }
@@ -45,10 +53,12 @@ type pgDataStore struct {
 }
 
 const (
-	routeTypeHTTP = "http"
-	routeTypeTCP  = "tcp"
-	tableNameHTTP = "http_routes"
-	tableNameTCP  = "tcp_routes"
+	routeTypeHTTP              = "http"
+	routeTypeTCP               = "tcp"
+	tableNameHTTP              = "http_routes"
+	tableNameTCP               = "tcp_routes"
+	tableNameCertificates      = "certificates"
+	tableNameRoutesCertificate = "route_certificates"
 )
 
 // NewPostgresDataStore returns a DataStore that stores route information in a
@@ -77,8 +87,8 @@ func (d *pgDataStore) Ping() error {
 }
 
 const sqlAddRouteHTTP = `
-INSERT INTO ` + tableNameHTTP + ` (parent_ref, service, leader, domain, tls_cert, tls_key, sticky, path)
-	VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
+INSERT INTO ` + tableNameHTTP + ` (parent_ref, service, leader, domain, sticky, path)
+	VALUES ($1, $2, $3, $4, $5, $6)
 	RETURNING id, created_at, updated_at`
 
 const sqlAddRouteTCP = `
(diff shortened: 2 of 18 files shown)


titanous:
@josephglanville Please review at least the SQL/schemas/migrations.

router/data_store.go
```diff
@@ -89,38 +98,224 @@ INSERT INTO ` + tableNameTCP + ` (parent_ref, service, leader, port)
...
+		TLSKey:    cert.TLSKey,
+		CreatedAt: cert.CreatedAt,
+		UpdatedAt: cert.UpdatedAt,
+	}
+	return nil
+}
+
+const sqlGetCert = `
+SELECT ` + selectColumnsHTTPCert + `, ARRAY(
+	SELECT http_route_id::varchar FROM ` + tableNameRoutesCertificate + ` WHERE certificate_id = $1
```
josephglanville: Is it necessary to cast to varchar here? I don't think it makes much difference either way just wondering. If so then there is a problem in the array handling of pgx.

josephglanville:
LGTM

router/api.go
```diff
@@ -204,6 +209,109 @@ func (api *API) DeleteRoute(ctx context.Context, w http.ResponseWriter, req *htt
 	w.WriteHeader(200)
 }
 
+func (api *API) CreateCert(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	log, _ := ctxhelper.LoggerFromContext(ctx)
+
+	var cert *router.RouteCert
+	if err := json.NewDecoder(req.Body).Decode(&cert); err != nil {
+		log.Error(err.Error())
+		httphelper.Error(w, err)
```
titanous: `httphelper.Error` should handle logging the error, is that not happening?

router/api.go
```diff
@@ -204,6 +209,109 @@ func (api *API) DeleteRoute(ctx context.Context, w http.ResponseWriter, req *htt
...
+		case ErrInvalid:
+			jsonError.Code = httphelper.ValidationErrorCode
+			jsonError.Message = "Invalid cert"
+		default:
+			log.Error(err.Error())
+			httphelper.Error(w, err)
+			return
+		}
+		httphelper.Error(w, jsonError)
+		return
```
titanous: This is unreachable code.

router/api.go
```diff
@@ -204,6 +209,109 @@ func (api *API) DeleteRoute(ctx context.Context, w http.ResponseWriter, req *htt
...
+}
+
+func (api *API) GetCert(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	log, _ := ctxhelper.LoggerFromContext(ctx)
+	params, _ := ctxhelper.ParamsFromContext(ctx)
+
+	l := api.router.HTTP.(*HTTPListener)
+	cert, err := l.GetCert(params.ByName("id"))
+	if err == ErrNotFound {
+		w.WriteHeader(404)
```
titanous: Use `httphelper.ObjectNotFoundError`

router/api.go
```diff
@@ -204,6 +209,109 @@ func (api *API) DeleteRoute(ctx context.Context, w http.ResponseWriter, req *htt
...
+func (api *API) DeleteCert(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	log, _ := ctxhelper.LoggerFromContext(ctx)
+	params, _ := ctxhelper.ParamsFromContext(ctx)
+
+	l := api.router.HTTP.(*HTTPListener)
+	err := l.RemoveCert(params.ByName("id"))
+	if err != nil {
+		switch err {
+		case ErrNotFound:
+			w.WriteHeader(404)
```
titanous: Ditto.

router/types/types.go
```diff
@@ -5,6 +5,22 @@ import (
...
+// RouteCert describes a TLS certificate for one or more routes
+type RouteCert struct {
+	// ID is the unique ID of this RouteCert
+	ID string `json:"id,omitempty"`
+	// Routes contains the IDs of routes assigned to this cert
+	Routes []string `json:"routes,omitempty"`
+	// TLSCert is the optional TLS public certificate. It is only used for HTTP routes.
+	TLSCert string `json:"tls_cert,omitempty"`
+	// TLSCert is the optional TLS private key. It is only used for HTTP routes.
+	TLSKey string `json:"tls_key,omitempty"`
```
titanous: I think `TLS` is redundant, remove it from the field names everywhere (just `cert`/`key`).

router/api.go
```diff
@@ -30,6 +30,11 @@ func apiHandler(rtr *Router) http.Handler {
 	r.GET("/routes", httphelper.WrapHandler(api.GetRoutes))
 	r.GET("/routes/:route_type/:id", httphelper.WrapHandler(api.GetRoute))
 	r.DELETE("/routes/:route_type/:id", httphelper.WrapHandler(api.DeleteRoute))
+	r.POST("/certs", httphelper.WrapHandler(api.CreateCert))
+	r.GET("/certs/:id", httphelper.WrapHandler(api.GetCert))
+	r.GET("/certs/:id/routes", httphelper.WrapHandler(api.GetCertRoutes))
+	r.DELETE("/certs/:id", httphelper.WrapHandler(api.DeleteCert))
+	r.GET("/certs", httphelper.WrapHandler(api.GetCerts))
```
titanous: I prefer expanding this to `certificate` everywhere.

router/schema.go
```diff
@@ -149,5 +149,61 @@ CREATE TRIGGER check_http_route_update
 		`ALTER TABLE tcp_routes ADD COLUMN leader boolean NOT NULL DEFAULT FALSE`,
 		`ALTER TABLE http_routes ADD COLUMN leader boolean NOT NULL DEFAULT FALSE`,
 	)
+	m.Add(5,
+		`CREATE TABLE certificates (
+			id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+			tls_cert text,
+			tls_key text,
+			tls_cert_sha256 text,
```
titanous: Shouldn't all of these be `NOT NULL`?

router/data_store.go
jvatic: Yeah, pgx won't scan it into a `[]string` without the casting.

router/api.go
jvatic: Probably is, the extra log statement is in all the other api handlers i looked at so i'll clean that up.

router/api.go
```diff
@@ -204,6 +209,107 @@ func (api *API) DeleteRoute(ctx context.Context, w http.ResponseWriter, req *htt
...
+	if err := json.NewDecoder(req.Body).Decode(&cert); err != nil {
+		log.Error(err.Error())
+		httphelper.Error(w, err)
+		return
+	}
+
+	l := api.router.HTTP.(*HTTPListener)
+	err := l.AddCert(cert)
+	if err != nil {
+		rjson, jerr := json.Marshal(&cert)
```
titanous: Since `cert` is already a pointer, there's no need to take a pointer to it.

router/api.go
```diff
@@ -204,6 +209,107 @@ func (api *API) DeleteRoute(ctx context.Context, w http.ResponseWriter, req *htt
...
+		if jerr != nil {
+			log.Error(jerr.Error())
+			httphelper.Error(w, jerr)
+			return
+		}
+		jsonError := httphelper.JSONError{Detail: rjson}
```
titanous: Is there a reason to provide the request JSON in the detail? Do we do it anywhere else? It may not be accurate because a round-trip through the decoder/encoder is not guaranteed to provide the exact same bytes.

router/data_store.go
```diff
@@ -89,38 +98,224 @@ INSERT INTO ` + tableNameTCP + ` (parent_ref, service, leader, port)
...
+		tx.Rollback()
+		return err
+	}
+	return tx.Commit()
+}
+
+func (d *pgDataStore) addCertWithTx(tx *pgx.Tx, c *router.Certificate) error {
+	tlsCertSHA256 := hex.EncodeToString(sha256.New().Sum([]byte(c.Cert)))
+	if err := tx.QueryRow(sqlSelectCert, tlsCertSHA256).Scan(&c.ID, &c.CreatedAt, &c.UpdatedAt); err != nil {
+		if err := tx.QueryRow(sqlAddCert, c.Cert, c.Key, tlsCertSHA256).Scan(&c.ID, &c.CreatedAt, &c.UpdatedAt); err != nil {
```
titanous: Is this a sort of upsert attempt? @josephglanville perhaps you can propose a better strategy?

router/data_store.go
```diff
@@ -89,38 +98,224 @@ INSERT INTO ` + tableNameTCP + ` (parent_ref, service, leader, port)
...
+			return err
+		}
+	}
+	for _, rid := range c.Routes {
+		if _, err := tx.Exec(sqlCleanupCertificates, rid); err != nil {
+			return err
+		}
+		if _, err := tx.Exec(sqlAddCertificate, rid, c.ID); err != nil && !postgres.IsUniquenessError(err, "") {
```
titanous: Ditto.

router/data_store.go
```diff
@@ -89,38 +98,224 @@ INSERT INTO ` + tableNameTCP + ` (parent_ref, service, leader, port)
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *pgDataStore) addCertificateWithTx(tx *pgx.Tx, r *router.Route) error {
```
titanous: What is the difference between `addCertWithTx` and `addCertificateWithTx`?

jvatic: The latter was `addRouteCertWithTx` (casualty of search/replace renaming `router.RouteCert` -> `router.Certificate`).

router/data_store.go
jvatic: Also, I believe a uniqueness error here would abort the transaction so ignoring it would cause an error elsewhere.

router/api.go
jvatic: This behavior was copied from `CreateRoute`.

router/data_store.go
jvatic: The goal here is to create a certificate record or get the id of an existing one.

router/data_store.go
```diff
@@ -45,10 +52,12 @@ type pgDataStore struct {
...
-	routeTypeHTTP = "http"
-	routeTypeTCP  = "tcp"
-	tableNameHTTP = "http_routes"
-	tableNameTCP  = "tcp_routes"
+	routeTypeHTTP              = "http"
+	routeTypeTCP               = "tcp"
+	tableNameHTTP              = "http_routes"
+	tableNameTCP               = "tcp_routes"
+	tableNameCertificates      = "certificates"
+	tableNameRoutesCertificate = "routes_certificate"
```
jvatic: This should probably be named `routes_certificates`.

titanous: Actually, `route_certificates` is better I think.

router/api.go
jvatic: I don't think it's necessary.

router/data_store.go
josephglanville: It's probably better to catch the conflict and continue rather than running the select first.
i.e:
```
if err := tx.QueryRow(sqlAddCert); err != nil && !postgres.IsUniquenessError(err, '") {
  // error out
}
```

jvatic: That's what I had at first, but subsequent queries in the transaction failed due to the error aborting the transaction.

jvatic: Here's the error:

```go
pgx.PgError{Severity:"ERROR", Code:"25P02", Message:"current transaction is aborted, commands ignored until end of transaction block", Detail:"", Hint:"", Position:0, InternalPosition:0, InternalQuery:
"", Where:"", SchemaName:"", TableName:"", ColumnName:"", DataTypeName:"", ConstraintName:"", File:"postgres.c", Line:1283, Routine:"exec_parse_message"}
```

for 

```go
        if err := tx.QueryRow(sqlAddCert, c.Cert, c.Key, tlsCertSHA256).Scan(&c.ID, &c.CreatedAt, &c.UpdatedAt); err != nil {
                if !postgres.IsUniquenessError(err, "") {
                        return err
                }
                if err := tx.QueryRow(sqlSelectCert, tlsCertSHA256).Scan(&c.ID, &c.CreatedAt, &c.UpdatedAt); err != nil {
                        return err
                }
        }
```

titanous: Yeah, you'll need to rollback the transaction and start a new one in that case.

josephglanville: Ahh, completely forgot that won't work in transaction.
Yeah the select is probably the way to go then. It should be safe/race free because you don't attempt to update the existing row, only insert if it doesn't exist.

jvatic: Hmm, that complicates things when coming from `addHTTP`.

titanous: Okay, let's leave it as-is then.

jvatic:
CI failures don't appear to be related to this change set.

router/schema.go
```diff
@@ -149,5 +149,61 @@ CREATE TRIGGER check_http_route_update
...
+
+				IF NOT FOUND THEN
+					INSERT INTO certificates (cert, key)
+					VALUES (http_route.tls_cert, http_route.tls_key)
+					RETURNING id INTO certificate_id;
+				END IF;
+
+				INSERT INTO route_certificates (http_route_id, certificate_id) VALUES(http_route.id, certificate_id);
+			END LOOP;
+		END $$`,
```
titanous: Have you tested this migration? It doesn't look like it populates `cert_sha256`. There is a good example of testing a migration here: https://github.com/flynn/flynn/blob/master/controller/migrate_test.go

jvatic: Fixed migration and added test in https://github.com/flynn/flynn/pull/2890/commits/0bb23c1e0a96142335c2fdee0dab1ba7337321c4

jvatic: (will squash before merge)

router/migrate_test.go
```diff
@@ -0,0 +1,86 @@
...
+	cert := tlsConfigForDomain("migrationtest.example.com")
+	r := &router.Route{
+		ParentRef:     "some/parent/ref",
+		Service:       "migrationtest.example.com",
+		Domain:        "migrationtest.example.com",
+		LegacyTLSCert: cert.Cert,
+		LegacyTLSKey:  cert.PrivateKey,
+	}
+	err := db.Exec(`INSERT INTO http_routes (parent_ref, service, domain, tls_cert, tls_key) VALUES ($1, $2, $3, $4, $5)`, r.ParentRef, r.Service, r.Domain, r.LegacyTLSCert, r.LegacyTLSKey)
+	c.Assert(err, IsNil)
```
titanous: This test should insert multiple routes with duplicate certificates and test that dedupe works as expected. It's probably worth adding some trailing and leading spaces+newlines (btw are we stripping that in the main insert codepath?).

jvatic: Added another route with the same cert to test that only one certificate record will be created per cert. Leading and trailing whitespace isn't currently handled anywhere, I'll add this next.

jvatic: Leading and trailing spaces and newlines are now stripped in both migration and main codepaths.

jvatic:
@titanous comments addressed

titanous:
LGTM

PR #2892, opened 2016-06-04
scheduler: Shuffle hosts when placing jobs

### LINKED ISSUE #2796: scheduler: Don't place multiple n=1 apps on the same host
The scheduler currently only distributes process types across all hosts on a per-app basis, and it also [picks hosts in sorted order](https://github.com/flynn/flynn/blob/master/controller/scheduler/scheduler.go#L693-L712) (likely to make the tests more predictable), which means multiple apps scaled to n=1 all get placed on the same host (the one that is first in the sorted list).

Possible fixes:

* Pick a host at random when they have the same amount of jobs and expect that to distribute better
* When no hosts have a job of a particular type, do some further counting to determine which host has the least jobs of any app / type

/cc @titanous

lmars:
This means that if multiple hosts have zero jobs of a given type, we pick one at random rather than always picking the first one, thus avoiding always scheduling n=1 apps on the same host.

Fixes #2796.


josephglanville:
LGTM

PR #2902, opened 2016-06-08
slugbuilder: Don't checkout origin/HEAD when cloning buildpacks

lmars:
The Go buildpack is currently failing to clone due to a missing .gitmodules file (see https://github.com/heroku/heroku-buildpack-go/pull/134).

Given we checkout an explicit version, there is no need to checkout origin/HEAD first.


josephglanville:
LGTM

PR #2912, opened 2016-06-11
host/cli: collect-debug-info improvements

josephglanville:
Namely always collects full log files now.
Unless it needs to switch to fallback it will only collect system logs.
If the total file size exceeds the gist limit then it will automatically fallback to tarball.


titanous:
LGTM

PR #2914, opened 2016-06-12
discoverd: Only log failed HTTP requests

josephglanville:
When logging all successful heartbeats etc the discoverd log can become overly verbose.
This change set removes logging of successful HTTP requests and redirects in favour of only logging failures.
We may need to add explicit logging for certain events to compensate for this if we find it necessary.

### DIFF
--- a/discoverd/server/handler.go
+++ b/discoverd/server/handler.go
@@ -9,8 +9,10 @@ import (
 	"strconv"
 	"strings"
 	"sync/atomic"
+	"time"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/julienschmidt/httprouter"
+	log "github.com/flynn/flynn/Godeps/_workspace/src/gopkg.in/inconshreveable/log15.v2"
 	"github.com/flynn/flynn/discoverd/client"
 	dt "github.com/flynn/flynn/discoverd/types"
 	hh "github.com/flynn/flynn/pkg/httphelper"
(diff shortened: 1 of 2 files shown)


discoverd/server/handler.go
```diff
@@ -22,6 +23,15 @@ import (
 // StreamBufferSize is the size of the channel buffer used for event subscription.
 const StreamBufferSize = 64 // TODO: Figure out how big this buffer should be.
 
+func loggerFn(handler http.Handler, logger log.Logger, clientIP string, rw *hh.ResponseWriter, req *http.Request) {
+	handler.ServeHTTP(rw, req)
+	switch rw.Status() {
+	case 200, 307, 0: // If the status code is 0 it means no response header was written and http.Server will set StatusOK
+	default:
+		logger.Info("request failed", "status", rw.Status())
```
titanous: This should log all of the same fields that the default logger does.

josephglanville:
@titanous corrected this so it matches normal logs, minus the fact that the starting request log message is omitted.

discoverd/server/handler.go
```diff
@@ -22,6 +24,16 @@ import (
 // StreamBufferSize is the size of the channel buffer used for event subscription.
 const StreamBufferSize = 64 // TODO: Figure out how big this buffer should be.
 
+func loggerFn(handler http.Handler, logger log.Logger, clientIP string, rw *hh.ResponseWriter, req *http.Request) {
+	start := time.Now()
+	handler.ServeHTTP(rw, req)
+	switch rw.Status() {
+	case 200, 307, 0: // 0 == 200 OK
+	default:
+		logger.Info("request completed", "status", rw.Status(), "duration", time.Since(start))
```
titanous: Since this doesn't log the request start, it should also log th request method/path/client ip.

josephglanville: Fixed.

titanous:
LGTM

PR #2916, opened 2016-06-12
discoverd: Deployment fixes

josephglanville:
Fixes a linearizability bug caused by using GetMeta in concert with event stream that may be connected to a different server.


titanous:
LGTM

PR #2918, opened 2016-06-13
script: Add --peers to start-flynn-host

josephglanville:
This makes it easier to test scenarios that require adding/removing
hosts from the cluster manually.

After bootstrapping a cluster with `bootstrap-flynn --size $SIZE` you can then add a new peer by providing it with the indices of the peers you want it to connect to. i.e `start-flynn-host --peers 0,1,2 3`.
This will start the new host at index 3 (`192.0.2.203`) connecting to the existing peers.


lmars:
LGTM

PR #2920, opened 2016-06-13
cli: Don't wait for job events if expected count is 0

### LINKED ISSUE #2901: Deleting an app that has been scaled to 0 times out waiting for scale down.
```
Timed out waiting for job events. Waited 20 seconds.
expected: map[]
actual: map[]
```

josephglanville:
Fixes deleting apps that have been scaled to 0.

Fixes #2901

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -219,6 +219,11 @@ func scaleToZero(appName string, client controller.Client) error {
 	}
 	processes := make(map[string]int)
 	expected := client.ExpectedScalingEvents(formation.Processes, processes, release.Processes, 1)
+
+	if expected.Count() == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


lmars:
LGTM (although we should probably be scaling down all formations for the app being deleted, not just the current release).

josephglanville:
That is a good point. We also need to handle deleting releases when we delete apps too.

lmars:
@josephglanville 

> We also need to handle deleting releases when we delete apps too.

True, but we should probably make releases belong to apps first (see https://github.com/flynn/flynn/pull/2769#issuecomment-216002816) to make it easier to know which releases to delete.

Perhaps open another issue / PR rather than blocking this PR?

josephglanville:
Indeed, we can tackle those 2 separately.

PR #2922, opened 2016-06-13
S3 support for blobstore

### LINKED ISSUE #2622: Add support for blobs in S3
If S3 is available, we should use it instead of Postgres to store blobs.

titanous:
This PR refactors the blobstore to support multiple storage backends, and adds support for S3 as a backend in addition to the default of Postgres. Files stored are always tracked in Postgres, backends only implement the storage of the actual bits.

To configure a new S3 backend and set it as the default, set these two environment variables:

```text
BACKEND_S3MAIN=backend=s3 region=us-east-1 access_key_id=xxx secret_access_key=xxx
DEFAULT_BACKEND=s3main
```

The name (s3main in this case) may be an arbitrary string. The bucket must already exist (with a lifecycle policy that cleans up incomplete multipart uploads after a few days) and the AWS credentials should have a policy like this:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:DeleteObject",
                "s3:GetObject",
                "s3:ListBucket",
                "s3:PutObject",
                "s3:ListMultipartUploadParts",
                "s3:AbortMultipartUpload",
                "s3:ListBucketMultipartUploads"
            ],
            "Resource": [
                "arn:aws:s3:::flynn-blobs",
                "arn:aws:s3:::flynn-blobs/*"
            ]
        }
    ]
}
```

After changing the default backend, existing artifacts in the old backend may be migrated by running:

```text
flynn -a blobstore run -e /bin/flynn-blobstore-migrate -- -delete
```

Which will move each blob that is not in the default backend to it and delete from the previous backend.

Closes #2622


test/test_blobstore.go
```diff
@@ -0,0 +1,84 @@
...
+	t.Assert(r.flynn("create", "blobstore-backend-test"), Succeeds)
+	t.Assert(r.git("push", "flynn", "master"), Succeeds)
+
+	// set default backend to S3
+	t.Assert(flynn(t, "/", "-a", "blobstore", "env", "set", "BACKEND_S3="+s3Config, "DEFAULT_BACKEND=s3"), Succeeds)
+
+	// test that downloading blob from postgres still works
+	t.Assert(r.flynn("run", "echo", "1"), Succeeds)
+
+	// check that artifact is now in S3
```
lmars: Is this comment valid?

test/test_blobstore.go
```diff
@@ -0,0 +1,84 @@
...
+	migration := flynn(t, "/", "-a", "blobstore", "run", "-e", "/bin/flynn-blobstore-migrate", "--", "-delete", "-prefix", u.Path)
+	t.Assert(migration, Succeeds)
+	t.Assert(migration, OutputContains, "Moving "+u.Path)
+	t.Assert(migration, OutputContains, "from postgres to s3")
+
+	// check that slug is now stored in S3
+	noRedirectsClient := &http.Client{
+		CheckRedirect: func(req *http.Request, via []*http.Request) error { return errors.New("no redirects") },
+	}
+	res, err := noRedirectsClient.Get(artifact.URI)
```
lmars: Is it worth using a retry dialer here?

lmars:
We'll need to update `TestReleaseImages` to not use the blobstore anymore (see [here](https://github.com/flynn/flynn/blob/master/test/test_release.go#L78-L86)), and expose `BLOBSTORE_S3_CONFIG` in CI, which I'm working on and will PR separately.

lmars:
@titanous I didn't see any multi part uploading, do users need the multi part upload related permissions? I guess it won't hurt in case we implement them in the future.

titanous:
The s3manager `Uploader` automatically handles the multi-part upload (we'll probably need to rip it out and replace with our own at some point, but it works well enough for now)

test/test_blobstore.go
titanous: I don't think so, do you see anything that would be eventually consistent?

test/test_blobstore.go
titanous: Fixed.

blobstore/backend/redirect_file_stream.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (s *redirectFileStream) Read(p []byte) (int, error) {
+	if s.body == nil {
+		res, err := http.Get(s.url)
+		if err != nil {
+			return 0, err
+		}
+		if res.StatusCode != 200 {
+			return 0, &url.Error{
```
imjorge: should close `res.Body`

titanous: Fixed, thanks.

blobstore/data/file_repo.go
```diff
@@ -0,0 +1,434 @@
...
+		}
+	}
+	return NewFileRepo(db, backends, defaultBackend), nil
+}
+
+func parseBackendInfo(s string) map[string]string {
+	info := make(map[string]string)
+	for _, token := range strings.Split(s, " ") {
+		kv := strings.SplitN(token, "=", 2)
+		info[kv[0]] = kv[1]
```
imjorge: this will panic with `index out of range` if the `s` is ill-formed (e.g `backend=s3 crash`)

blobstore/data/file_repo.go
```diff
@@ -0,0 +1,434 @@
...
+
+func (r *FileRepo) ListFilesExcludingDefaultBackend(prefix string) ([]BackendFile, error) {
+	rows, err := r.db.Query(
+		"SELECT file_id, file_oid, external_id, backend, name, type, size, sha512, updated_at FROM files WHERE backend != $1 AND NAME LIKE $2 || '%' AND deleted_at IS NULL",
+		r.defaultBackend.Name(), prefix,
+	)
+	if err != nil {
+		return nil, err
+	}
+
```
imjorge: `defer rows.Close()`

blobstore/data/file_repo.go
```diff
@@ -0,0 +1,434 @@
...
+	var res []BackendFile
+	for rows.Next() {
+		var info backend.FileInfo
+		var backendName string
+		var externalID *string
```
imjorge: does `externalID` needs to be `*string` or could it be just `string`? not seeing the difference for other cases here.

blobstore/data/file_repo.go
```diff
@@ -0,0 +1,434 @@
...
+		if rows.Scan(&info.ID, &info.Oid, &externalID, &backendName, &info.Name, &info.Type, &info.Size, &sha512, &info.ModTime); err != nil {
+			return nil, err
+		}
+		if externalID != nil {
+			info.ExternalID = *externalID
+		}
+		info.ETag = base64.StdEncoding.EncodeToString(sha512)
+		f := BackendFile{FileInfo: info}
+		f.Backend, _ = r.getBackend(backendName)
+		res = append(res, f)
```
imjorge: this may append `nil` to `res`, right?

blobstore/data/file_repo.go
```diff
@@ -0,0 +1,434 @@
...
+	// words, mimic either deleting a file or recursively deleting a
+	// directory)
+	rows, err := tx.Query(
+		"UPDATE files SET deleted_at = now() WHERE name ~ ('^' || $1 || '(/.*)?$') AND deleted_at IS NULL RETURNING file_id, external_id, backend, name", name,
+	)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+	backendFiles := make(map[string][]backend.FileInfo)
```
imjorge: `defer rows.Close()`

blobstore/data/file_repo.go
titanous: A pointer is used, as pgx requires it to decode a nullable field.

blobstore/data/file_repo.go
titanous: I don't think so, `f` will always be non-nil.

blobstore/data/file_repo.go
titanous: Fixed.

blobstore/data/file_repo.go
titanous: Fixed.

blobstore/data/file_repo.go
titanous: Fixed.

test/test_blobstore.go
lmars: So I had a test failure because the blobstore crashed which a retry here would have fixed, but a better fix is for the test to fail quicker (the deployment should wait for the blobstore service, not just an up event, I'll add that change).

lmars: d9ae20c

test/test_blobstore.go
```diff
@@ -0,0 +1,84 @@
...
+		// backend=s3 access_key_id=xxx secret_access_key=xxx bucket=blobstore-ci region=us-east-1
+		t.Skip("missing BLOBSTORE_S3_CONFIG env var")
+	}
+
+	r := s.newGitRepo(t, "http")
+	t.Assert(r.flynn("create", "blobstore-backend-test"), Succeeds)
+	t.Assert(r.git("push", "flynn", "master"), Succeeds)
+
+	// set default backend to S3
+	t.Assert(flynn(t, "/", "-a", "blobstore", "env", "set", "BACKEND_S3="+s3Config, "DEFAULT_BACKEND=s3"), Succeeds)
```
lmars: The `flynn` commands are echoed to the log, which means `s3Config` here gets printed.

test/test_blobstore.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	r := s.newGitRepo(t, "http")
+	t.Assert(r.flynn("create", "blobstore-backend-test"), Succeeds)
+	t.Assert(r.git("push", "flynn", "master"), Succeeds)
+
+	// set default backend to S3 without printing secrets
+	cmd := exec.Command("sh", "-c", args.CLI+` -a blobstore env set "BACKEND_S3=$BLOBSTORE_S3_CONFIG" DEFAULT_BACKEND=s3`)
+	cmd.Env = flynnEnv(flynnrc)
+	cmd.Env["BLOBSTORE_S3_CONFIG"] = s3Config
```
lmars: `flynnEnv` is a copy of `os.Environ()` just with `FLYNNRC` set, so it will already have `BLOBSTORE_S3_CONFIG`

blobstore/data/file_repo.go
imjorge: true for `f`. meant to say `f.Backend`.

lmars:
LGTM

temujin9:
Does this encrypt the blobs?

titanous:
@temujin9 We don't set any flags related to that, are you looking for [server-side encryption](https://docs.aws.amazon.com/AmazonS3/latest/dev/serv-side-encryption.html) in S3? If so, which of those options?

temujin9:
Gotcha. Lemme see if I care, first: there's every chance that I don't.

Issue #2921, opened 2016-06-13
Show current flynn version in dashboard


kwando:
It would be nice if the dashboard displayed the current version of flynn-host, bonus points for displaying a notification if you are not running the latest version.

PR #2923, opened 2016-06-13
controller: Add version endpoint

jvatic:
This will be used for determining which version of the controller client to use once there's more than one, and also for #2921.


titanous:
This information should already be exposed in the health status endpoint.

jvatic:
Ah, so it is. Missed that.

PR #2925, opened 2016-06-13
status: Make the docker-receive app optional

lmars:
Users restoring from a cluster backup may not have a docker-receive release, so docker-receive will not start as part of bootstrap, causing the final status check to fail.

Given not all users will be using docker-receive, we should just make it optional and have a mechanism to create a release and scale it up when pushing an image.


titanous:
LGTM

PR #2928, opened 2016-06-14
status: Add CORS response headers

jvatic:


josephglanville:
LGTM.

I assume this is to allow status endpoint to be accessed from Javascript on dashboard?

titanous:
I don't think that map operation is safe. I have a commit you can cherrypick on the go1.6 branch.

titanous:
LGTM

jvatic:
@josephglanville Yup. @titanous I've replaced e72c00b with 05b8f92.

PR #2931, opened 2016-06-15
Godeps: Update hashicorp/raft

josephglanville:
Minor update to bring in additional log message on transition to follower state.

Can make it easier to follow discoverd logs to find previous leader when something bad happens.


titanous:
LGTM

PR #2932, opened 2016-06-15
taffy: Added support for deploying from repos that require key auth.

maran:
As discussed on IRC my first attempt to get taffy to support deploying from repos that require key authentication.

### DIFF
--- a/taffy/main.go
+++ b/taffy/main.go
@@ -3,9 +3,11 @@ package main
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/flynn/flynn/Godeps/_workspace/src/github.com/flynn/go-docopt"
@@ -31,6 +33,12 @@ func parsePairs(args *docopt.Args, str string) (map[string]string, error) {
 	return item, nil
 }
 
+func ensureDir(dir string) {
+	if err := os.MkdirAll(filepath.Join(dir, ".ssh"), 0700); err != nil {
+		log.Fatal(err)
+	}
+}
+
 func main() {
 	usage := `
 Usage: taffy <app> <repo> <branch> <rev> [-e <var>=<val>]... [-m <key>=<val>]...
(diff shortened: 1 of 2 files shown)


taffy/main.go
```diff
@@ -54,6 +55,20 @@ Options:
...
+
+	if os.Getenv("SSH_CLIENT_KEY") != "" {
+		os.Mkdir(os.Getenv("HOME")+"/.ssh/", 0700)
+		ioutil.WriteFile(os.Getenv("HOME")+"/.ssh/id_rsa", []byte(os.Getenv("SSH_CLIENT_KEY")), 0600)
+	}
+
+	if os.Getenv("SSH_CLIENT_HOSTS") != "" {
+		os.Mkdir(os.Getenv("HOME")+"/.ssh/", 0700)
+		ioutil.WriteFile(os.Getenv("HOME")+"/.ssh/known_hosts", []byte(os.Getenv("SSH_CLIENT_HOSTS")), 0600)
+	}
```
jvatic: The slugbuilder already takes care of this part, so setting `SSH_CLIENT_KEY` and `SSH_CLIENT_HOSTS` on the flynn receiver job should be enough (it sets them on the slugbuilder job).

lmars:
@maran the test looks good, address the one comment so far and this should be good to merge.

taffy/main.go
lmars: This change is necessary because it is the `taffy` process which clones the private repo, not the receiver.

taffy/main.go
```diff
@@ -54,6 +55,20 @@ Options:
...
+
+	if os.Getenv("SSH_CLIENT_KEY") != "" {
+		os.Mkdir(os.Getenv("HOME")+"/.ssh/", 0700)
+		ioutil.WriteFile(os.Getenv("HOME")+"/.ssh/id_rsa", []byte(os.Getenv("SSH_CLIENT_KEY")), 0600)
+	}
+
+	if os.Getenv("SSH_CLIENT_HOSTS") != "" {
+		os.Mkdir(os.Getenv("HOME")+"/.ssh/", 0700)
+		ioutil.WriteFile(os.Getenv("HOME")+"/.ssh/known_hosts", []byte(os.Getenv("SSH_CLIENT_HOSTS")), 0600)
```
lmars: We need to check the errors for `os.Mkdir` and `ioutil.WriteFile`:

```go
if err := os.Mkdir(...); err != nil {
  log.Fatal(err)
}
if err := ioutil.WriteFile(...); err != nil {
  log.Fatal(err)
}
```

taffy/main.go
jvatic: Sorry about that!

taffy/main.go
lmars: The code would probably be more readable if you put the env vars in variables, and you should use `filepath.Join` and `os.MkdirAll`:

```go
home := os.Getenv("HOME")

if key := os.Getenv("SSH_CLIENT_KEY"); key != "" {
  if err := os.MkdirAll(filepath.Join(home, ".ssh"); err != nil {
    log.Fatal(err)
  }
  if err := ioutil.WriteFile(filepath.Join(home, ".ssh", "id_rsa"), []byte(key), 0600); err != nil {
    log.Fatal(err)
  }
}

maran:
Could this be reviewed again? I forced push a small rewrite.

lmars:
LGTM

lmars:
Thanks!

PR #2939, opened 2016-06-17
scheduler: Reset HostID when placing jobs

lmars:
This is particularly problematic if we fail to place an omni job, because subsequent placements will treat all hosts as having the requested number of omni jobs, and a host will be picked at random, leading to stopSurplusOmniJobs stopping the job just placed and leading to flapping jobs.


josephglanville:
Good find. LGTM

PR #2940, opened 2016-06-17
test: Ping new discoverd node until it responds

josephglanville:
Hardens the Promote/Demote test against `discoverd` taking a while to start up HTTP API .


lmars:
LGTM

PR #2942, opened 2016-06-17
test: Set BLOBSTORE_S3_CONFIG in CI

lmars:
I've also switched `TestReleaseImages` to use a simple file server rather than the blobstore.


titanous:
LGTM

PR #2945, opened 2016-06-18
discoverd: Guard against decoding nil metadata

josephglanville:
It's possible for our event stream to not yet reflect the creation of
the deployment metadata because we are connected to a follower that
hasn't yet replicated the service meta command.

### DIFF
--- a/discoverd/deployment/deployment.go
+++ b/discoverd/deployment/deployment.go
@@ -64,6 +64,7 @@ type Deployment struct {
 	meta    *discoverd.ServiceMeta
 }
 
+// update drains any pending events, updating the service metadata, it doesn't block.
 func (d *Deployment) update() error {
 	select {
 	case event, ok := <-d.events:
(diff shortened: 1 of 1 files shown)


discoverd/deployment/deployment.go
```diff
@@ -88,6 +89,21 @@ outer:
...
+		// If the metadata is nil then wait for another metadata event before re-evaluating
+		if d.meta == nil {
+			select {
+			case event, ok := <-d.events:
+				if !ok {
+					return fmt.Errorf("service stream closed unexpectedly: %s", d.stream.Err())
+				}
+				if event.Kind == discoverd.EventKindServiceMeta {
+					d.meta = event.ServiceMeta
+					continue outer
```
titanous: It seems a bit weird to call `update` again, which this appears to do. Shouldn't this break out if meta is non-nil?

josephglanville: The difference here is we are just making sure a service meta event has been received. `update` only ensures that the events channel has been drained. Ensuring that a service meta even has been received should ensure that decode shouldn't fail (as apart from when no deployment has started we never leave the metadata in a state that is not decodable by `decode`).

titanous:
LGTM

PR #2946, opened 2016-06-18
discoverd: Don't error if DNS isn't running during promote/demote

josephglanville:
If the DNS server hasn't been started yet it's likely we are
are waiting on the host network to be configured.
Because we hold the lock after waitHostDNSConfig() unblocks the DNS
configuration will block on us releasing the lock.
Once this happens the DNS server will be properly configured taking into
account the state after promotion/demotion.

This basically fixes a race where if a new node is added and promote is called before Flannel has come up and notified the host + discoverd has recognized and configured it's DNS server.


titanous:
LGTM

PR #2947, opened 2016-06-18
Use vendor/ instead of Godeps

titanous:
`GO15VENDOREXPERIMENT` is now enabled by default in Go 1.6 and godep supports the `vendor/` directory natively, so switch to it.


titanous:
@lmars I think `TestReleaseImages` is failing because the wrong version of Go is being used, let me know if you need any help fixing.

lmars:
LGTM

Issue #2082, opened 2015-11-04
Add TLS proxying support to router


rikur:
As a company, we are using client certificates to authenticate our users. We would like to use client certificates to limit access to some of our flynn applications as well.

https://golang.org/pkg/crypto/tls/ has support for them.
http://www.bite-code.com/2015/06/25/tls-mutual-auth-in-golang/ has a code sample.

titanous:
Yeah, this is interesting. Since the router terminates TLS connections, it would have to do the handshake with the client. How do you think the handshake details should be passed back from the router to the app?

titanous:
This should be implemented by adding a new route type that reads the SNI portion of the handshake, establishes a TCP connection to the backend and then proxies the full ClientHello back, splicing the connection in. This way the router is not party to the TLS handshake and the user app can use any TLS configuration it wants.

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

PR #2948, opened 2016-06-19
router: Add support for HTTP/2

### LINKED ISSUE #1050: Implement HTTP/2
The router should support HTTP/2.

titanous:
Go 1.6 has great support for HTTP/2 so this is entirely transparent and enables it for all sites hosted on Flynn.

Closes #1050


josephglanville:
LGTM

codyaray:
Does this router terminate the HTTP/2 connection and open an HTTP/1.1 connection to the app or does it proxy/passthrough the HTTP/2 connection?

titanous:
The former. Currently we don't send HTTP/2 requests to the backend instances.

codyaray:
Is there any plan or thoughts for proxying HTTP/2? We're looking for a PaaS which we can use in conjunction with gRPC (grpc.io) which requires HTTP/2 all the way to the backend.

titanous:
@codyaray We'll probably implement this by allowing opaque TLS streams to be proxied to backends without modification: #2082

codyaray:
Okay, thanks for letting me know. I subscribed to that issue for updates on support. :)

lmars:
@codyaray if you are happy to run on a non-standard port, you could use a TCP route:

```
$ flynn route add tcp --service my-app-web --port 8443
```

Then any connections to port 8443 will be proxied directly to a backend.

PR #2949, opened 2016-06-19
router: Re-enable upgrade path test

### LINKED ISSUE #2358: router: Go 1.4.3 request smuggling protection closes connections unnecessarily
Go 1.4.3 introduced some new security features into the HTTP stack, specifically new protections against request smuggling. These seem to be causing issues as seen in #2356. We need to create a minimal test case to bring to upstream and ensure the test it properly re-enabled when upgrading to Go 1.5.

josephglanville:
This test was broken by a bug in request smuggling protection in
Go 1.4.2. Re-enable the test that we have now upgraded to Go 1.6.2

Fixes #2358


titanous:
LGTM

PR #2951, opened 2016-06-19
Deploy and HTTP request cancellation

### LINKED ISSUE #94: Terminate interrupted builds right away
If I press `Ctrl-C` while a build triggered by a git push is running, the build should terminate before the app is deployed.

### LINKED ISSUE #958: router: use CloseNotify to cancel requests when clients disconnect
When a client disconnects, the router should also disconnect from backends (so they can stop doing unnecessary work). This was originally implemented in #926, but unfortunately that had to be put on hold due to incompatibilities in net/http between the `CloseNotifier` and `Hijacker` interfaces. That's documented in https://github.com/golang/go/issues/9763

Once that issue is resolved, it should be possible to alter the router so that it behaves well when clients go away.

### LINKED ISSUE #2852: Cancelling a "git push" doesn't cancel the deploy
When a buildpack is running after a `git push`, killing the local process won't stop it from running and deploying the new code. This makes it difficult to cancel a deploy if we realize that there's a problem (we'd have to kill the slugbuilder job manually). Is there a way to detect from within a `post-receive` hook whether the disconnection from the client was intentional or not?

titanous:
This PR implements support for cancelling in-progress deploys by terminating the git process that is handling the push (for example by pressing Ctrl-C). Also implemented is support in the router for cancelling in-progress HTTP requests when clients disconnect.

Closes #958
Closes #94
Closes #2852


josephglanville:
LGTM

test/test_git_deploy.go
```diff
@@ -268,3 +272,60 @@ func (s *GitDeploySuite) TestSlugbuilderLimit(t *c.C) {
...
+	t.Assert(err, c.IsNil)
+
+	done := make(chan struct{})
+	go func() {
+		select {
+		case <-done:
+		case <-time.After(30 * time.Second):
+			cmd.Process.Signal(syscall.SIGTERM)
+			cmd.Wait()
+			t.Fatal("git push timed out")
```
lmars: `t.Fatal` should be called in the main goroutine.

PR #2952, opened 2016-06-19
PostgreSQL 9.5

### LINKED ISSUE #2534: Postgres: Update to version 9.5
Postgres 9.5 adds a lot of cool features (most notable UPSERT) as well as a lot of other small improvements.

Release notes can be found here: http://www.postgresql.org/docs/9.5/static/release-9-5.html

There are a couple breaking changes in the release, but they seem to be pretty minimal.

josephglanville:
Upgrade to 9.5, making use of the new UPSERT functionality to clean up some controller queries.

Closes #2534

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -109,7 +109,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     curl --fail --silent https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
     sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ trusty-pgdg main" >> /etc/apt/sources.list.d/postgresql.list'
     sudo apt-get update
-    sudo apt-get install -y postgresql-9.4 postgresql-contrib-9.4
+    sudo apt-get install -y postgresql-9.5 postgresql-contrib-9.5
     sudo -u postgres createuser --superuser vagrant
     grep '^export PGHOST' ~/.bashrc || echo export PGHOST=/var/run/postgresql >> ~/.bashrc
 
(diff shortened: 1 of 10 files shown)


titanous:
It's probably worth adding a guard to the in-place updater that refuses to proceed so that the controller doesn't end up using broken queries.

josephglanville:
The main changes here are we are dropping config file statements that shouldn't be required any longer. 

`ssl_renegotion_limit` is deprecated and can only be set to 0. It will be removed entirely in the next PostgreSQL release.

The WAL checkpoint settings have changed into 2 new settings, `min_wal_size` and `max_wal_size` in the process though the defaults were raised considerably and should be sufficient for our usecase.

I have also lowered the amount of WAL segments kept. This setting was inherited from the upstream Manatee implementation and was probably designed for systems running with very large amounts of spare disk space. As most of our clusters are deployed with minimal storage consuming up to 16GB of WAL on each PostgreSQL job is likely slightly excessive. This could only be problematic in the case where a `pg_basebackup` takes long enough to run that the amount of writes exceeds the new setting of 2GB. Which should be unlikely especially with our recommendation the large production users move to using S3 for blob storage.

titanous:
The first commit message should have a `appliance/postgresql: ` subsystem prefix.

titanous:
The commit message for bdcdf67d0df512dda0dfe5e49d996de23d33cb56 should be updated to say `host/cli` instead of `cli`.

titanous:
Looks like the CI rootfs and Vagrantfiles need updates.

titanous:
LGTM pending comment

titanous:
Is anything blocking this?

PR #2955, opened 2016-06-19
pkg/tlsconfig: Remove two ciphersuites banned in HTTP/2

titanous:
`TLS_RSA_WITH_AES_128_GCM_SHA256` and `TLS_RSA_WITH_AES_256_GCM_SHA384` are banned for use with HTTP/2 in RFC 7540 Appendix A, but this is not reflected in the http2 package, so the server will incorrectly error when starting because it thinks that there are bad ciphersuites listed before good ones.

Remove them for now and re-add when the http2 package is fixed.


lmars:
LGTM

titanous:
Upstream fix at: https://go-review.googlesource.com/#/c/24214/

PR #2958, opened 2016-06-19
gitreceive: Wait for service event on initial push

### LINKED ISSUE #2884: Initial and subsequent deployments should use the same code path
Initial deployments of broken code can succeed, and then fail on subsequent releases.

Quoth @titanous on IRC: "The initial deployment goes through a different code path, it just sets the new processes as desired, subsequent deploys monitor the new processes so that the deploy can be rolled back before it completes to prevent downtime. There are almost certainly better ways to handle this, it's mostly an artifact of some changes we made."

This is confusing to users, as well as being less desirable from a code-cleanliness perspective.

lmars:
Fixes #2884.

### DIFF
--- a/gitreceive/receiver/flynn-receive.go
+++ b/gitreceive/receiver/flynn-receive.go
@@ -3,6 +3,7 @@ package main
 import (
 	"archive/tar"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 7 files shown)


titanous:
I think we should ensure that if the git push is rejected, the code doesn't end up running, so scale the formation back down if possible and if not log the fact that there was an unclean deploy.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #2959, opened 2016-06-19
logaggregator: Fix race in TestHostCursors

titanous:
Message processing happens in an independent goroutine, so we can’t just assume that they are processed after writing to the socket.


lmars:
LGTM

PR #2962, opened 2016-06-20
router: Store certificate SHA-256 as binary not hex

titanous:
The hex-encoded hash can be too wide for the index, it appears to be variable length before this patch. Store it as a shorter bytea field.

     ERROR: index row size 2792 exceeds maximum 2712 for index \"certificates_cert_sha256_idx\" (SQLSTATE 54000)


josephglanville:
LGTM

PR #2967, opened 2016-06-21
test/rootfs: Install libsasl2 in rootfs

josephglanville:
Required to build new `mgo` version with SASL authentication.


titanous:
Should this be added to the Vagrantfile as well?

josephglanville:
Ahh yep. Though I think we actually handle most of that in the packer phase, may be best to just keep it all there for now.

titanous:
LGTM

lmars:
LGTM

PR #2975, opened 2016-06-21
WIP: Cause ReleaseSuite.TestReleaseImages to fail

lmars:
This is a dummy PR to trigger `ReleaseSuite.TestReleaseImages` to fail so we can see why it is taking so long in CI (when I run it manually it takes ~5mins, but it is currently double that in CI).


lmars:
More recent builds seem to be quicker at running this, but in any case the test takes much less time than the whole test run, so we should aim to parallelise the other tests.

Issue #2895, opened 2016-06-07
App processes running old release after a successful update of environment variable


tjboring:
Experienced this issue tonight:
- Updated one environment variable in an app running on our Flynn cluster
- Flynn created a new release (I didn't check to see if the app was running the release, though)
- Updated a second environment variable in the same app
- We noticed that the change didn't have the effect we were expecting
- Tried restarting processes by scaling down then up
- Effect of the previous step was that instead of 4 processes (2 web, 2 workers), we now had 8 (4 web, 4 workers)
- Thinking it was an issue with the version, I pushed the latest from git repo to the cluster (i.e. > git push $remote $branch:master)
- Checked the output of `flynn -a $appname ps` and noticed the processes were not running the release that i just pushed, but one from 2 hours prior

I have debug logs to share; ping me on IRC.

lmars:
@tjboring please email the logs to me lewis@flynn.io (feel free to encrypt them with [my PGP key](https://keybase.io/lmars/key.asc)).

lmars:
@tjboring I have been unable to identify the issue here.

I have updated the debug logs so that they contain the scheduler leader's in-memory state (#2998), but you will need to be running on a cluster which has that change (it will most likely be in nightly tonight, but not in stable just yet).

In the mean time, please collect more debug logs as this continues to happen (preferably just after a deployment, and from each host in the cluster) as they may provide more insight.

tjboring:
@lmars thanks, lewis! i just upgraded to the latest stable today, but i'll keep an eye and grab debug logs as necessary.

tjboring:
So, this happened about in the last 24 hours. Here's what I know:

1. ~16h ago, we had a deploy from CircleCI, which was successful
1. In the last hour, I had a dev ask why his code wasn't running in the environment
1. checking the output of `flynn -a $app ps` i could see workers that were running a release from ~19h before

To resolve this issue, I did the following:

1. `flynn -a $app scale -r $release_id web=0 clock=0 worker=0
1. verified no processes running
1. `flynn -a $app scale web=N clock=N worker=N

I've collected the controller logs and emailed them as attachments.

lmars:
Looking at the logs from the latest incident, here is what happened:

* A deploy began at `2016-08-10T18:29:24` which determined that the old release scale was `worker:2 clock:1 web:2` but actual jobs running was `worker:1 clock:1 web:2` (i.e. a missing `worker` job)
[...]

PR #2998, opened 2016-06-27
scheduler: Add HTTP endpoint to dump internal state

lmars:
This will help debug issues like #2895.


lmars:
I realised that this sends the internal state to the HTTP client in the main scheduler loop so could block more important work that needs doing. I'm going to change this.

lmars:
Updated to copy the internal state inside the main loop and return to the HTTP handler to send on to the client, please review.

titanous:
LGTM

PR #3007, opened 2016-06-28
controller: Delete releases when deleting apps

### LINKED ISSUE #2900: Artifacts are not deleted when apps are deleted.
When an app is deleted slugs and repo caches are still retained in the blobstore. These should be deleted in order to reclaim space.

lmars:
Fixes #2900.


titanous:
LGTM

Issue #2559, opened 2016-03-12
controller doesn't stop serving HTTP requests during shutdown


titanous:
This results in requests failing because the database pool is closed but requests are still being served: https://gist.github.com/titanous/87c2b161fa0258b6736b

We should stop handling HTTP requests gracefully, finishing existing ones and stopping any streams while not accepting new requests.

PR #3014, opened 2016-06-28
controller: Shutdown fixes

lmars:
A bunch of fixes which make controller shutdown cleaner and also quicker (it currently takes 10s before being forcibly killed, increasing the time of all controller deployments in CI).


titanous:
Does this fully close #2559?

lmars:
@titanous I wasn't aware of #2559 but yes, closing the database pool waits for each checked out connection to be returned before closing, so in-flight requests will be completed before the pool closes, and the change here will make sure no new requests try to use the closed pool.

josephglanville:
LGTM

PR #3016, opened 2016-06-29
cli: Follow redirects when downloading slugs

### LINKED ISSUE #3012: Exporting an application fails with the S3 blobstore backend
Hi,
I just tried to export an application after having switched to the S3 Blobstore backend and I get this:

> unexpected status getting slug: 302

I still get a .tar with the files "app.json", "artifact.json", "formation.json", "release.json" and "routes.json", but the slug is missing.

It seems the downloader is not handling an HTTP redirect from S3?

lmars:
Fixes #3012.


titanous:
LGTM

PR #3030, opened 2016-07-01
host: Replace libvirt-lxc backend with libcontainer

### LINKED ISSUE #38: pinkerton cleanup "device or resource busy"
Sometimes when running `pinkerton cleanup` from the libvirt backend, it fails with this error:

```text
context.go:90: rename /var/lib/docker/aufs/mnt/tmp-flynn-discoverd-277ab3fe3b594e80ad1d40838c2f7b66 /var/lib/docker/aufs/mnt/tmp-flynn-discoverd-277ab3fe3b594e80ad1d40838c2f7b66-removing: device or resource busy
```

### LINKED ISSUE #2367: Flynn cluster broke after libvirt update
Hello all,

We had our 3 node flynn cluster go down today... here is one of the host logs: https://gist.github.com/anonymous/77673fd510ece2aa5098

Any help possibly identifying the problem from these logs would be appreciated ... we also have a hunch that some other host maintenance may have caused unintentional harm to the cluster (possibly a broken apt-get update, and other networking modifications).

We are just going to rebuild it ... if you find anything that could root-cause the failure in the log please post back, elsewise feel free to close... we are not in production yet, so its not a huge deal.

### LINKED ISSUE #2553: Error when starting flynn-host and libvirt-bin daemon is not running
> <imjorge> i get this on the flynn-host.log when starting flynn-host
> <imjorge> 15:22:16.447397 host.go:259: [Code-38] [Domain-7] Failed to connect socket to '/var/run/libvirt/libvirt-sock': No such file or directory
> <imjorge> installed with: sudo bash < <(curl -fsSL https://dl.flynn.io/install-flynn)
> <imjorge> am i missing someting?
> <imjorge> the /var/lib/libvirt folder exists but not the .sock he is trying to open
> <lmars> imjorge: what is the output of `sudo status libvirt-bin`?
> <imjorge> sure enough :) "libvirt-bin stop/waiting"
> <lmars> ok, is there anything interesting in `/var/log/libvirt/libvirtd.log`?
> -flynn-ci/#flynn- FAIL: [1 failure(s)] Push of flynn[refs/heads/master] by lmars: 9e2b71c => fdd80ae https://ci.flynn.io/builds/20160310150440-70fd72d2
> <imjorge> last line on libvirt log is
> <imjorge> 2016-01-25 17:27:13.369+0000: 1234: error : ebiptablesDriverTestCLITools:4308 : Testing of ip6tables command failed: Failure to execute command '$IPT -n -L
> <imjorge> FORWARD' : 'Another app is currently holding the xtables lock. Perhaps you want to use the -w option?'.
> <lmars> ok, do you have other programs running which potentially modify iptables?
> <imjorge> i dont think so
> <imjorge> docker does?
> <lmars> yes
> <lmars> so you have Docker running?
> <imjorge> yes
> <lmars> ok that is probably the issue
> <lmars> I haven't seen this before though
> <lmars> maybe try `sudo start libvirt-bin` to see if it was just a race
> <imjorge> so far it appears to be started
> <imjorge> let me up flynn-host again
> <lmars> ok
> <imjorge> flynn-host.log currently at "INFO[03-10|15:32:42] waiting for discoverd                    component=cluster-monitor fn=Run"
> <lmars> ok, I think thats ok, you should now bootstrap
> <lmars> unless this was an already bootstrapped cluster?
> <imjorge> The built-in dashboard can be accessed at ...
> <imjorge> and off he went
> <imjorge> lmars: tx
> <lmars> :)
> <lmars> do you mind opening an issue with the error you got?
> <imjorge> not a problem but not sure what to include there, mind pvt me to see what you want me to put on the issue (which logs and whatnot)?
> <lmars> just copy our IRC convo
> <lmars> I can update if necessary
> <lmars> just want to make sure we have a record

### LINKED ISSUE #2725: Missing /data in Postgres containers
Rarely we encounter ENOENT errors from within a Postgres container, complaining that `/data` does not exist.

Recently, this prevented the trigger from being written to promote a sync: https://gist.github.com/anonymous/35a52bd07a3a842c1d83e0b50a0029b6

The mounts within the container *look* correct: https://gist.github.com/85312afbf35aaac4a7bd409a0e27b957

And the files in the volume, look almost correct from the host mount namespace: https://gist.github.com/anonymous/0aa1f2832dc4e6096ac2f05f5e6fe786 (note the missing `instance_id` file, probably removed while cleaning up after a failed base backup pull).

So far I have no further data points explaining why this happens.

### LINKED ISSUE #2803: Status app shows mariadb and postgress as unhealthy
We are monitoring the output of the status app, and occasionally mariadb and postgres go into an unhealthy state. Here's the HTTP response from the status app when in this state:
```json
{
  "data": {
    "status": "unhealthy",
    "detail": {
      "blobstore": {
        "status": "healthy",
        "version": "v20160506.0"
      },
      "controller": {
        "status": "healthy",
        "version": "v20160506.0"
      },
      "controller-scheduler": {
        "status": "healthy",
        "version": "v20160506.0"
      },
      "controller-worker": {
        "status": "healthy",
        "version": "v20160506.0"
      },
      "dashboard": {
        "status": "healthy",
        "version": "v20160506.0"
      },
      "discoverd": {
        "status": "healthy",
        "version": "v20160506.0"
      },
      "flannel": {
        "status": "healthy",
        "version": "v20160506.0"
      },
      "gitreceive": {
        "status": "healthy",
        "version": "v20160506.0"
      },
      "logaggregator": {
        "status": "healthy",
        "version": "v20160506.0"
      },
      "mariadb": {
        "status": "unhealthy",
        "version": "v20160506.0"
      },
      "postgres": {
        "status": "unhealthy",
        "version": "v20160506.0"
      },
      "router": {
        "status": "healthy",
        "version": "v20160506.0"
      }
    },
    "version": "v20160506.0"
  }
}
```

Here's debug info gists for the nodes in our cluster while the unhealthy status is being reported:
https://gist.github.com/6adb0ba0604f08b4eb242f97e9394c5d
https://gist.github.com/a48cdc7b22ecaca9b9e67c8456fe5aa6
https://gist.github.com/7f5f61e22a8c49b633bd3d07f46dc1e3

They started reporting unhealthy today at 17:18 UTC. Another cluster had the same thing happen earlier today, but in that case it resolved itself after about 20 minutes.

Also, at the same time it started reporting unhealthy, another app returned 503 briefly. However, that resolved itself immediately. May or may not be related.

### LINKED ISSUE #3021: host: Use libcontainer to start containers
We should support running containers using [libcontainer](https://github.com/opencontainers/runc/tree/master/libcontainer) for the following reasons:

* More control over assigned resources (e.g. capabilities, currently hard-coded by libvirt-lxc)

* Do away with most of [containerinit](https://github.com/flynn/flynn/blob/master/host/containerinit/init.go) (containers will be direct children of `flynn-host` so FDs can be shared in a more conventional way rather than passing them over Unix sockets, and `libcontainer` does most of the other setup like networking and setting rlimits)

* Close our [longest open bug](https://github.com/flynn/flynn/issues/38) (well, hopefully :smile:)

* Drop `libvirt-lxc` as a dependency (this would break in-place updates, so we would either need to keep libvirt-lxc around or not allow users to do an in-place update)

* Support running Flynn in Flynn (which would make testing things like cluster backup / restore *much* easier)

I have made a start on this on the `host-libcontainer-runtime` branch: https://github.com/flynn/flynn/compare/host-libcontainer-runtime

lmars:
Opening as an early PR to see what happens in CI.

Closes #38 
Closes #3021
Closes #2553
Closes #2367
Closes #2725 
Closes #2803


lmars:
There are still a couple of test failures, but I think these are CI flakes present on master, so this is now ready for review.

An overview of how containers work with libcontainer:

* `flynn-host` exec's `flynn-init` with `libcontainer-init` as a command line argument, and the `_LIBCONTAINER_INITPIPE` env var pointing at a pipe file descriptor opened by `flynn-host`
* `flynn-init` receives binary netlink data followed by a JSON configuration payload from the pipe which it uses to create namespaces, setup networking etc. (this is handled transparently by `libcontainer`)
* `flynn-init` exec's `/.containerinit` inside the container (which is just a bind mount of `flynn-init` from the host), and then `containerinit` proceeds as it does currently

Some notes:

* we continue to use cgroup partitions as we did with libvirt, but these are now named `/flynn/{system,user,background}` (rather than libvirt's `/machine/{system,user,background}.partition`)
* starting a container no longer leaves a supervisor process in the root mount namespace, so we no longer need `nsumount` and I have removed it
* I have manually verified that mounts are no longer being leaked (so #38 is fixed)
* networking setup is now handled by libcontainer and has been removed from `containerinit`
* since containers are now started as direct children of `flynn-host`, we could stop passing FDs over unix sockets, but that would break restarting `flynn-host` or updating it (the containers will get re-parented to init, and we will lose the FD references).

host/flynn-init/main.go
```diff
@@ -1,6 +1,25 @@
...
+
+func init() {
+	if len(os.Args) > 1 && os.Args[1] == "libcontainer-init" {
+		runtime.GOMAXPROCS(1)
+		runtime.LockOSThread()
+		factory, _ := libcontainer.New("", libcontainer.Cgroupfs)
+		if err := factory.StartInitialization(); err != nil {
+			log.Fatal(err)
+		}
+	}
```
titanous: Is there a reason to put this in `init` instead of `main`?

host/libcontainer_backend.go
```diff
@@ -455,21 +401,110 @@ func (l *LibvirtLXCBackend) Run(job *host.Job, runConfig *RunConfig, rateLimitBu
...
-		return err
-	}
-
-	if err := bindMount(l.resolvConf, filepath.Join(rootPath, "etc/resolv.conf"), false, true); err != nil {
-		log.Error("error bind mounting resolv.conf", "err", err)
-		return err
+	config := &configs.Config{
+		Rootfs: rootPath,
+		Capabilities: []string{
+			"CAP_CHOWN",
```
titanous: Where was this capabilities list taken from?

host/libcontainer_backend.go
```diff
@@ -455,21 +401,110 @@ func (l *LibvirtLXCBackend) Run(job *host.Job, runConfig *RunConfig, rateLimitBu
...
+			{Type: configs.NEWNS},
+			{Type: configs.NEWUTS},
+			{Type: configs.NEWIPC},
+			{Type: configs.NEWPID},
+		}),
+		Cgroups: &configs.Cgroup{
+			Path: filepath.Join("/flynn", job.Partition, job.ID),
+			Resources: &configs.Resources{
+				AllowedDevices: configs.DefaultAllowedDevices,
+				Memory:         1 * units.GiB,
```
titanous: Is this the memory limit? If so, should it come from the job resource config?

host/libcontainer_backend.go
```diff
@@ -602,85 +629,60 @@ func (l *LibvirtLXCBackend) Run(job *host.Job, runConfig *RunConfig, rateLimitBu
...
-		OnPoweroff: "preserve",
-		OnCrash:    "preserve",
+			{
+				Type:              "veth",
+				Name:              "eth0",
+				Bridge:            l.bridgeName,
+				Address:           initConfig.IP,
+				Gateway:           initConfig.Gateway,
+				Mtu:               1500,
+				HostInterfaceName: "veth" + random.String(4),
```
titanous: This name is not random enough, it will encounter collisions (birthday paradox). What is libcontainer's behavior in this case? Perhaps we should use an atomically incrementing integer?

host/libcontainer_backend.go
titanous: Nevermind, found the setting below.

titanous:
This is awesome! Have you tried removing the vendored libvirt client code?

titanous:
Looks like we'll probably need some minor docs updates.

host/Tupfile
```diff
@@ -2,6 +2,5 @@ include_rules
 : |> sed 's/{{TUF-ROOT-KEYS}}/@(TUF_ROOT_KEYS)/g' cli/root_keys.go.tmpl > %o |> cli/root_keys.go
 : cli/root_keys.go |> !cgo |> bin/flynn-host
 : bin/flynn-host |> gzip -9 --keep bin/flynn-host |> bin/flynn-host.gz
-: |> !go ./flynn-init |> bin/flynn-init
+: |> !cgo ./flynn-init |> bin/flynn-init
```
titanous: This will break containerinit if the rootfs doesn't have libc. Perhaps we should put the libcontainer part in `flynn-host` instead?

host/libcontainer_backend.go
lmars: Ah yes.

So after digging around, there is no code in libcontainer / runc which sets `HostInterfaceName` (which is why I just went with this random name). Looking at Docker v.1.11.1, it actually uses [libnetwork's GenerateIfaceName](https://github.com/docker/docker/blob/v1.11.1/vendor/src/github.com/docker/libnetwork/drivers/bridge/bridge.go#L889-L893) to do this, which just [tries a few times](https://github.com/docker/docker/blob/v1.11.1/vendor/src/github.com/docker/libnetwork/netutils/utils_linux.go#L33-L50).

We already have libnetwork vendored, so we could use that same method, or we could use an incremental int (although we will need to sychronise that across an udpate / restart). WDYT?

titanous: The libnetwork function sounds fine.

host/libcontainer_backend.go
lmars: This is actually the full list of capabilities from `/usr/include/linux/capability.h` on Ubuntu trusty.

I did intend to strip it down but I forgot. We could stick with the same list we have been using with `libvirt-lxc` (which blacklisted `CAP_SYS_MODULE`, `CAP_SYS_TIME`, `CAP_MKNOD`, `CAP_AUDIT_CONTROL`, and `CAP_MAC_ADMIN`, see [here](https://github.com/libvirt/libvirt/blob/v1.2.2/src/lxc/lxc_container.c#L1741-L1745)), or do a review of each capability to determine if we want to drop it or not (`libcontainer` provides "good defaults" too, see [here](https://github.com/opencontainers/runc/blob/master/libcontainer/SPEC.md#security))

titanous: Let's take the list from Docker.

lmars:
@titanous comments addressed (in terms of docs, I just updated [this line](https://github.com/flynn/flynn/commit/c1c1c6a366565a36ed76b2df3f4f54bb0ac5e4eb#diff-1ecddd30aee43db4147f32e89570bad9L46) in the architecture docs).

host/libcontainer_backend.go
```diff
@@ -53,6 +53,27 @@ const (
 	defaultMemory     = 1 * units.GiB
 )
 
+// defaultCapabilities is a list of capabilities which are set inside a
+// container, taken from:
+// https://github.com/opencontainers/runc/blob/v1.0.0-rc1/libcontainer/SPEC.md#security
+var defaultCapabilities = []string{
+	"CAP_NET_RAW",
```
titanous: This list does not appear to match the link.

lmars: In what way do they differ? I just took the ones with `Enabled: 1` set.

titanous: Ah, I missed the second column. That makes sense.

host/libcontainer_backend.go
```diff
@@ -3,58 +3,100 @@ package main
...
+	"CAP_KILL",
+	"CAP_SYS_CHROOT",
+}
+
+func init() {
+	// when starting a container with libcontainer, we first exec the
+	// current binary with libcontainer-init as the first argument,
+	// which triggers the following code to initialise the container
+	// environment (namespaces, network etc.) then exec containerinit
+	if len(os.Args) > 1 && os.Args[1] == "libcontainer-init" {
```
titanous: Put this in the `main` function so it's obvious what's going on.

lmars:
Updates are not currently working because all containers are exiting when the old `flynn-host` process dies.

I initially thought `flynn-host` was killing them on shutdown but this also happens when it dies abruptly with `SIGKILL`, so it must be the kernel which kills the containers.

The likely suspect is the [ParentDeathSignal](https://github.com/opencontainers/runc/blob/v1.0.0-rc1/libcontainer/configs/config.go#L84-L86) parameter, but so far I have not been able to configure things so that the containers do not exit (I have tried modifying that config variable, removing all `prctl` calls, making a `prctl` call to reset it in `containerinit`, but none of those approaches prevent the containers from dying).

I am investigating further.

lmars:
The containers were previously exiting when the parent `flynn-host` daemon exited because they were getting a TTY hangup signal due to the fact that `flynn-host` opened the console, which subsequently closed when it died.

I have changed things so that only TTY jobs get a console (f5d6b84) which means those jobs will not survive an update to `flynn-host`, but those jobs are usually interactive jobs (e.g. `flynn run bash`) so I don't think we need to worry about them.

I have also fixed job resurrection so that it only happens for jobs which aren't actually running (c7d703d), and although that could be changed on master I am bundling it in here as I hope we can merge this.

The last few test runs have had some failures but they are all tests which fail intermittently on master, and are mostly controller deployment related (something which I will be refactoring soon).

@titanous @josephglanville this is ready for a review.

host/libcontainer_backend.go
```diff
@@ -602,85 +636,71 @@ func (l *LibvirtLXCBackend) Run(job *host.Job, runConfig *RunConfig, rateLimitBu
...
-					Type:   "mount",
-					Source: lt.FSRef{Dir: rootPath},
-					Target: lt.FSRef{Dir: "/"},
-				},
-				{
-					Type:   "ram",
-					Source: lt.FSRef{Usage: "65535"}, // 64MiB
-					Target: lt.FSRef{Dir: "/dev/shm"},
-				},
+	ifaceName, err := netutils.GenerateIfaceName("veth", 4)
```
titanous: Should this be moved into the else block below?

host/libcontainer_backend.go
```diff
@@ -602,85 +636,71 @@ func (l *LibvirtLXCBackend) Run(job *host.Job, runConfig *RunConfig, rateLimitBu
...
+	if job.Config.TTY {
+		console, err := process.NewConsole(0, 0)
+		if err != nil {
+			return err
+		}
+		// copy the console to /dev/null because if we don't
+		// do anything with the console, it will be destroyed
+		// by the garbage collector and the job will get a
+		// tty hangup.
+		go io.Copy(ioutil.Discard, console)
```
titanous: This is weird, can we disassociate the console from the flynn-host process somehow?

script/install-flynn.tmpl
```diff
@@ -127,7 +127,7 @@ main() {
   local packages=(
     "aufs-tools"
     "iptables"
-    "libvirt-bin"
+    "cgroup-lite"
```
titanous: Were you able to get rid of the cgroup-lite package by setting up the cgroups in flynn-host?

lmars: Ah yes forgot about this, see bea0afb.

lmars:
@titanous so I went on a small TTY adventure today and eventually realised the issue was that not assigning a console to the container meant that when `containerinit` [opened a PTY](https://github.com/flynn/flynn/blob/v20160707.1/host/containerinit/init.go#L673) to hand to the job, it ended up having that PTY as its controlling terminal, so then exec'ing the job with `Setctty` for that same terminal would fail with `EPERM` because terminals can only be controlling for a single session.

Long story short, updating `github.com/kr/pty` to include [this change](https://github.com/kr/pty/commit/3b1f6487b7fc649d5f146df04e623bd55ba1bf7f) fixed that, and means there is now no need to create consoles in `flynn-host` (218e56e), so therefore no TTY hang ups when `flynn-host` exits.

Ready for another review :smile:.

titanous:
LGTM!

josephglanville:
LGTM. 🎉

lmars:
woot!

temujin9:
Woot indeed. Can I get a head's up when this releases?

titanous:
The latest nightly has it.

PR #3032, opened 2016-07-01
dashboard: Added a link below the apps list to toggle the system apps

### LINKED ISSUE #2109: dashboard: Document how to view system apps
Adding `system=true` to the URL params will show all the system apps (e.g. https://dashboard.demo.localflynn.com/apps?system=true).

There should be a link somewhere in the UI to toggle this param.

philiplb:
fixes #2109

Signed-off-by: Philip Lehmann-Boehm <philip@philiplb.de>

### DIFF
--- a/dashboard/app/lib/javascripts/dashboard/views/apps-list.js.jsx
+++ b/dashboard/app/lib/javascripts/dashboard/views/apps-list.js.jsx
@@ -17,8 +17,15 @@ var AppsList = React.createClass({
 		return (
 			<ul className="items-list">
 				{apps.map(function (app) {
+					var classes = [];
+					if (assertEqual(app.id, selectedAppId)) {
+						classes.push("selected");
+					}
+					if (assertEqual(app.meta["flynn-system-app"], "true")) {
+						classes.push("system-application");
+					}
 					return (
-						<li key={app.id} className={assertEqual(app.id, selectedAppId) ? "selected" : ""}>
+						<li key={app.id} className={ classes.join(" ") }>
 							<RouteLink path={getAppPath(app.id)}>
 								{app.name}
 							</RouteLink>
--- a/dashboard/app/lib/javascripts/dashboard/views/apps.js.jsx
+++ b/dashboard/app/lib/javascripts/dashboard/views/apps.js.jsx
@@ -3,6 +3,7 @@ import AppsStore from '../stores/apps';
 import AppsListHeader from './apps-list-header';
 import AppsList from './apps-list';
 import App from './app';
+import RouteLink from './route-link';
 
 var Apps = React.createClass({
 	displayName: "Views.Apps",
@@ -16,6 +17,12 @@ var Apps = React.createClass({
 					{React.createElement(AppsList, extend({}, this.props.appsListProps, {
 						apps: this.state.apps
 					}))}
+
+					<section className="system-applications-link">
+						<RouteLink path={"/apps"} params={{system: !this.props.showSystemApps }}>
+							{this.props.showSystemApps ? 'Hide' : 'Show'} System Applications
+						</RouteLink>
+					</section>
 				</section>
 
 				<section className="panel app-panel full-height">
--- a/dashboard/app/lib/stylesheets/dashboard/apps.scss
+++ b/dashboard/app/lib/stylesheets/dashboard/apps.scss
@@ -12,3 +12,6 @@
   overflow-y: auto;
 }
 
+.system-applications-link {
+  text-align: right;
+}
--- a/dashboard/app/lib/stylesheets/dashboard/items-list.scss
+++ b/dashboard/app/lib/stylesheets/dashboard/items-list.scss
@@ -18,14 +18,19 @@
   > li {
     position: relative;
     border-bottom: 1px solid $grayBlueColor;
+    border-left: 0.5em solid transparent;
     &:last-of-type {
       border: 0;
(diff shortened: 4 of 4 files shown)


dashboard/app/lib/javascripts/dashboard/views/apps.js.jsx
```diff
@@ -16,6 +17,10 @@ var Apps = React.createClass({
 					{React.createElement(AppsList, extend({}, this.props.appsListProps, {
 						apps: this.state.apps
 					}))}
+
+					<section>
+						<SystemApplicationsLink />
```
jvatic: `this.props.showSystemApps` is [passed in from the router](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/routers/apps.js#L93-L127). Between that and passing in a `getAppsPath` prop function similar to the `getAppPath` one we should be able to inline the whole thing here fairly easily.

I'm thinking something like:

```jsx
<RouteLink path={ this.props.getAppsPath({ system: !this.props.showSystemApps }) }>
  { this.props.showSystemApps ? 'Hide' : 'Show' } System Applications
</RouteLink>
```

(The `<RouteLink/>` component is defined [here](https://github.com/flynn/flynn/blob/master/dashboard/app/lib/javascripts/dashboard/views/route-link.js.jsx).)

philiplb:
Good hint. I simplyfied the View a lot with the already existing flag. And I removed the handling of existing parameters. Reason: Actually weird behaviour if a system application was selected and then they were hidden again. The application still showed up on the right, but had no entry in the list anymore. Then it's better to not select anything IMO.
And I added a little marker on the system application list entries.

I tried the way you proposed but I like it a bit more if the little component is self contained and not another component used elsewhere is used. This could lead to side effects later.

dashboard/app/lib/javascripts/dashboard/views/apps-list.js.jsx
```diff
@@ -17,8 +17,16 @@ var AppsList = React.createClass({
 		return (
 			<ul className="items-list">
 				{apps.map(function (app) {
+					var classes = [];
+					if (assertEqual(app.id, selectedAppId)) {
+						classes.push("selected");
+					}
+					if (typeof app.meta["flynn-system-app"] !== undefined &&
```
jvatic: No need to test for `undefined`, the latter assertion alone is fine. (Also `app.meta.hasOwnProperty('flynn-system-app')` is more idiomatic.)

dashboard/app/lib/stylesheets/dashboard/items-list.scss
```diff
@@ -26,6 +26,9 @@
       background-color: $greenColor;
       color: $whiteColor;
     }
+
+    &.system-application {
+      border-left: 5px solid $darkerGrayBlueColor;
```
jvatic: should probably add a transparent border to all the other apps to align things properly, and `em` would be a better unit in this case (something like `0.25em`).

jvatic:
> Then it's better to not select anything IMO.

Sounds reasonable.

> I like it a bit more if the little component is self contained and not another component used elsewhere is used.

In this case it's better to use the `RouteLink` component. It's one of the building blocks of the entire application and is used everywhere else to override the default click behavior and use pushState instead.

> This could lead to side effects later.

It's more likely that re-implementing this logic everywhere will cause things to work inconsistently and also makes any issue related to this harder to track down. So long as the component has a clearly defined singular purpose side effects should be non-existent.

dashboard/app/lib/javascripts/dashboard/views/system-applications-link.js.jsx
```diff
@@ -0,0 +1,26 @@
...
+var SystemApplicationsLink = React.createClass({
+	displayName: "Views.SystemApplicationLink",
+
+	render: function () {
+		var props = extend({}, this.props);
+		props.href = this.__getHref();
+		props.className = "system-applications-link";
+		var label = (this.props.showSystemApps ? "Hide" : "Show") +
+			" System Applications";
+		return React.createElement('a', props, label);
```
jvatic: JSX is preferred over using the React API directly. This could be simplified to:

```jsx
return (
  <a href={this.__getHref()} className="system-applications-link">
    {this.props.showSystemApps ? 'Hide' : 'Show'} System Applications
  </a>
);
```

philiplb:
I see, I thought the RouteLink was just used for the route links of the applications. :)
If you like, I would visit it again and switch to it.
If the current solution is ok for you, go for a merge of course. :) I would later then open a separate PR then.

jvatic:
`RouteLink` should definitely be used here.

```jsx
import RouteLink from './route-link';
// ...
<RouteLink path={"/apps"} params={{system: !this.props.showSystemApps }}>
 ...
</RouteLink>
```

philiplb:
Done. :)

dashboard/app/lib/stylesheets/dashboard/apps.scss
```diff
@@ -12,3 +12,6 @@
   overflow-y: auto;
 }
 
+.system-applications-link {
+  float: right;
```
jvatic: Wrap in a `<div>` and set `text-align: right` on it (a clear fix is needed otherwise).

jvatic:
LGTM pending comment
Thanks!

philiplb:
Nice.
Comment addressed.

PR #3034, opened 2016-07-02
test: Use the default flynn-host backend

lmars:
This allows us to test changing the default backend (for example to libcontainer) without needing to modify the test runner.


titanous:
LGTM

Issue #3031, opened 2016-07-01
can't SSL to gitreceive (via git push) on fresh AWS installation


swaits:
Just installed using v20160701.0.

When I try to git push I get: `fatal: unable to access 'https://blah/blah.git/': SSL: certificate verification failed (result: 5)`

@jvatic has more context on this (via irc)

titanous:
Fixed by #3049.

Issue #3035, opened 2016-07-02
controller: domain migration doesn't create new route for all apps


jvatic:

jvatic:
This was fixed in #3049

PR #3036, opened 2016-07-02
installer: Add workaround for issue #3035

jvatic:


titanous:
Does this also apply to the `migrate-domain` command? Has it already been fixed in the domain migration worker code?

jvatic:
@titanous I believe it does, however I've yet to track down the source of the issue.

jvatic:
This also fixes #3031

titanous:
I'd like to see the root cause identified and fixed before merging this.

titanous:
Superceded by #3049.

PR #3049, opened 2016-07-05
router: Fix TLS migration

### LINKED ISSUE #2989: Restore a backup v20160512 to v20160624.1 doesn't restore the TLS certificates
Hi,
after upgrading from v20160512 to v20160624.1 from backup, all TLS routes have lost their certificate and are using some self signed ones now.
Currently, I'm restoring the routes manually by removing and adding them.

jvatic:
Closes #2989


titanous:
LGTM

PR #3055, opened 2016-07-05
util/packer,Vagrantfile: Only start Docker at boot in dev VM

titanous:
Before this change, it started in provisioned EC2 instances as well.


josephglanville:
LGTM

PR #3056, opened 2016-07-05
blobstore: Fix parsing of double space in backend config

titanous:
This fixes a case where the parser will error if there are two spaces in a row.


josephglanville:
LGTM

titanous:
@philiplb Thanks for reporting this!

PR #3057, opened 2016-07-05
test/rootfs: Bump rootfs size to 50G

titanous:
CI jobs occasionally fail with `ENOSPC`, increase from 32G to 50G to fix.


josephglanville:
LGTM

PR #3064, opened 2016-07-06
Sirenia resurrection/recovery improvements

### LINKED ISSUE #2877: host/fixer: When recovering Sirenia clusters we should terminate extra jobs that aren't participating in the cluster.
Currently the recovery logic only checks for the number of postgres jobs that are running when determining if it needs to start new jobs to revive the cluster. However if all postgres jobs were killed before a new primary and sync could be elected then the cluster will be in a state where none of the running jobs are assigned roles (or potentially are still running but unpromotable asyncs).

The fixer should examine the metadata against the list of running jobs and terminate any that are not assigned roles in the service metadata currently before attempting to recover the primary/sync to return the cluster to writable state.

josephglanville:
This improves the Sirenia recovery code to understand which instances are actually partaking in the cluster rather than just aiming to achieve a certain count of instances running.

This will help in cases where postgres instances have been respawned by the scheduler but never moved from unassigned into taking part in the cluster itself. This mostly happens in instances of partial cluster failure, OOM killer and other sorts of mayhem.

Closes #2877


titanous:
LGTM

PR #3072, opened 2016-07-07
test: Temporarily skip TestReleaseImages until new stable release

josephglanville:
Postgres 9.5 isn't backwards compatible to 9.4 so we have a check in `flynn-host update` to prevent updating from releases with 9.4. Until a new stable is released this will break `TestReleaseImages`.
Temporarily skip the test until we ship a new stable release with 9.5.

### DIFF
--- a/test/test_release.go
+++ b/test/test_release.go
@@ -106,6 +106,9 @@ func (s *ReleaseSuite) TestReleaseImages(t *c.C) {
 	if testCluster == nil {
 		t.Skip("cannot boot release cluster")
 	}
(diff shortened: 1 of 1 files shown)


lmars:
@josephglanville `TestReleaseImages` updates a cluster from the commit being built so it will now be using postgres 9.5, we just need to bump the hard coded version string in use (currently `v20150131.0-test`, see https://github.com/flynn/flynn/blob/master/test/test_release.go).

josephglanville:
Ohh I see, lets close this then and create PR to update version string.

PR #3078, opened 2016-07-07
controller: Log formation expansion error

titanous:
Instead of closing the connection with an error, log it and move on.

This avoids bricking the streaming endpoint permanently if there is an issue with referential integrity or decoding a specific field.


lmars:
LGTM

PR #2744, opened 2016-04-26
controller: Schedule one-off jobs via the scheduler

lmars:
This adds the concept of a "job request" which is persisted to the database, and then the scheduler fulfils job requests by starting jobs in the cluster.

This currently only affects jobs requested via `client.RunJob{Attached,Detached}`, however I have plans to change both slugbuilder and worker jobs to use this too, but I will do that in subsequent PRs.

Refs #2620.

PR #3080, opened 2016-07-08
worker: Don't rollback one-by-one deploy if cleanup fails

lmars:


josephglanville:
So with this change the new formation will stay scaled up but the deployment will be marked as failed?

lmars:
> So with this change the new formation will stay scaled up but the deployment will be marked as failed?

Correct, with the hope that the next time the deploy is attempted, it is marked as done.

I plan to rewrite deployments once #2744 is merged, so this is just a minor change to prevent the "failed" deployment nuking a controller release for no good reason.

josephglanville:
Sounds good. LGTM

titanous:
LGTM

PR #3084, opened 2016-07-08
script: Move FLANNEL_BACKEND var back to bootstrap-flynn

lmars:
`FLANNEL_BACKEND` only affects bootstrap, not starting `flynn-host` (the variable was moved by accident in commit f5a2a23).


titanous:
LGTM

PR #3085, opened 2016-07-09
test: Remove Sirenia deploy timeout

josephglanville:
Instead we rely on the controller to do the right thing.
Worst case the test will be terminated if it hangs.


titanous:
LGTM

PR #3087, opened 2016-07-10
router: Fix race when calling Set on new HTTP/TCP routes

### LINKED ISSUE #1034: router: unnecessary mutex lock in tcpRoute.Serve()?
I stumbled across the following mutex calls in the TCP router: https://github.com/flynn/flynn/blob/6369dd5675b215201d1f923b118642a1f729270c/router/tcp.go#L249-L251

```go
r.mtx.RLock()
go r.service.ServeConn(conn)
r.mtx.RUnlock()
```

I didn't dig any further as I didn't want to get sidetracked from what I'm working on, but it has to be either a race or completely unnecessary. /cc @benburkert @titanous

titanous:
Each time `Set` is called a new `*proxy.ReverseProxy` is built and saved into the route, but there is no mutex lock for reads. Fix by doing an atomic swap and using atomic reads.

Closes #1034

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/flynn/flynn/discoverd/cache"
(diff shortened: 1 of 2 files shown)


titanous:
I see another race in the HTTP code, will fix.

titanous:
Looking closer, there is no race here. A completely new route is built for each call of Set.

PR #3089, opened 2016-07-10
gitreceive: Show error when pushing to non-master branch

### LINKED ISSUE #2039: git push flynn fails if not in master branch
https://gist.github.com/rikur/11b0fc77778b5b34a36c

```
11:46 < rikurr> titanous: the pre-receive scripts aborts any pushes that are not from master branch: https://gist.github.com/rikur/11b0fc77778b5b34a36c
11:47 < rikurr> What could be the purpose for having that check there?
11:47 < rikurr> I'll remove it for now I guess, but if you want me to replace it with some other functionality, let me know and I will create a PR
11:48 < titanous> rikurr: hmm, I guess we should just save/ignore other branches?
11:48 < titanous> only pushes to master should deploy, but we cache the whole repo
11:49 < titanous> do pushes with a master branch and another branch work?
11:49 < titanous> `git push flynn heroku:master` should work in your case
11:59 < rikurr> why should we only deploy from master? That's a big limitation IMHO.
11:59 < rikurr> I want our devs to deploy any branch they're in
12:00 < rikurr> What kind of problems would that cause?
12:00 < titanous> yeah, you can, it just needs to be pushed to the master branch on the remote
12:00 < titanous> otherwise we don't know which branch to deploy (a push can contain multiple branches)
12:00 < titanous> so for example `git push flynn local-dev:master` would deploy the local 'local-dev' branch
12:02 < rikurr> ah well what a bout.. if only one branch if built, we use that. otherwise we give the user a helpful error message or even prompt to enter branch do deploy?
12:03 < rikurr> s/if only one branch if built/if only one branch if pushed/
12:03 < titanous> I think the solution is to have a good error message, it's too easy to push the wrong branch and have it deploy otherwise
12:04 < titanous> there are also git config tricks that will map a local branch to a different remote one
```

In my case
```git push flynn heroku:master```
should have worked.

I will try to create a PR to give user a helpful error message so they don't get stuck like I did.

titanous:
Closes #2039

### DIFF
--- a/gitreceive/server.go
+++ b/gitreceive/server.go
@@ -294,6 +294,7 @@ func (w writeFlusher) Write(p []byte) (int, error) {
 
 var prereceiveHook = []byte(`#!/bin/bash
 set -eo pipefail;
+
 git-archive-all() {
 	GIT_DIR="$(pwd)"
 	cd ..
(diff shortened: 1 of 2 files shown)


lmars:
LGTM

Issue #3095, opened 2016-07-11
Cluster completely broken


dottodot:
My apps seem to still be running but pretty much everything else doesn't work such as

flynn ps
backup
dashboard
can't collect debug info

even fix hosts end on an error of
```14:00:30.676978 host.go:136: already have enough instances, unable to fix```

is there anything I can do that will help fix it at least so that I can backup to a new cluster

**Comment by titanous** [comment]
This gist describes how to manually take a backup: https://gist.github.com/titanous/36754c1da91c632640a7

**Comment by dottodot** [comment]
so if I have the following in zfs list how do I know which one to take a snapshot of?

```
NAME                                                 USED  AVAIL  REFER  MOUNTPOINT
flynn-default                                       22.7G  13.2G    19K  /flynn-default
flynn-default/11542b19-b5ba-4fdd-a38f-f4c371ea14e9  18.7G  13.2G  18.7G  /var/lib/flynn/volumes/zfs/mnt/11542b19-b5ba-4fdd-a38f-f4c371ea14e9
flynn-default/3169f0b9-5337-4777-bd8b-552a7daac9d7  3.92G  13.2G  3.92G  /var/lib/flynn/volumes/zfs/mnt/3169f0b9-5337-4777-bd8b-552a7daac9d7
flynn-default/935d8c53-b17d-457a-b76a-6b94015a30dc    20K  13.2G    20K  /var/lib/flynn/volumes/zfs/mnt/935d8c53-b17d-457a-b76a-6b94015a30dc
flynn-default/c731e101-1399-4f5c-9d62-43b83f638d89   534K  13.2G   534K  /var/lib/flynn/volumes/zfs/mnt/c731e101-1399-4f5c-9d62-43b83f638d89
flynn-default/c7b286a8-771d-47ad-a82e-fa1b0090ae6f    20K  13.2G    20K  /var/lib/flynn/volumes/zfs/mnt/c7b286a8-771d-47ad-a82e-fa1b0090ae6f
```

titanous:
The largest one, `flynn-default/11542b19-b5ba-4fdd-a38f-f4c371ea14e9`.

dottodot:
Any idea how I get past 

error connecting to postgresFATAL: password authentication failed for user "postgres" (SQLSTATE 28P01)

titanous:
You may need to tweak the `pg_hba.conf`. I haven't seen that error before, was postgres already installed on the server?

dottodot:
[...]

PR #3096, opened 2016-07-11
host: Remove unnecessary CAP_AUDIT_{READ,WRITE}

titanous:
`CAP_AUDIT_READ` does not exist in kernel 3.13, and we don’t need/want these capabilities.

Refs #3095


lmars:
LGTM

PR #3100, opened 2016-07-12
all: WIP Support custom squashfs based images

lmars:
This is an experiment to test out using squashfs based images which have many benefits over the AUFS diffs currently in use:
- Easy to distribute
- Easy to mount
- Easy to checksum / verify
- Easy to share between jobs (mounted once, uses the kernel file-cache)

The aim is to also use squashfs for slugs so they benefit from the above too.
---
### image format

An image is a JSON manifest which contains a list of layers which should be mounted, e.g.:

``` json
{
  "version": "v1",
  "layers": [
    {
      "id": "3fdfac2f9355dfdb1e3b1af1afdf143acabd555dcd8fc6b80c1ea62423ac2122a634b119140663ee213c0d621d1e51de40ac9cdf8cbaa24ca7a27593ba7d05ff.squashfs",
      "type": "squashfs"
    },
    {
      "id": "8d4f80ddb31de6c9a5413c1f0ca1fb4585866e2b966591491ccc0363f2760a644d1512786eb8f5500a873d2da4f0636fa72202b4fcd68e823be8489ac543bd32.squashfs",
      "type": "squashfs"
    },
    {
      "id": "385d3eda8d4823312f6e0abcc7925795dd5b587ed333b1d29ebb83fa50dbbbbe1697dcb3e55e01eb9f4d85e76cbdbac3f907d8ff122dd09e27e56cdad8f573d5.squashfs",
      "type": "squashfs"
    }
  ]
}
```
---
### mounting an image

Images are mounted using overlayfs with a `lowerdir` for each of the squashfs layers from the manifest (requires Linux >= 3.19, see [here](http://www.phoronix.com/scan.php?page=news_item&px=MTg1Nzg)), and a temporary `upperdir`.
---
### storage

There is a root directory (hard-coded as `/var/lib/flynn/image`) with 4 sub-directories:
- `manifests` - image manifests
- `layers` - squashfs layer files
- `mnt` - mounted squashfs layers + mounted overlayfs job mounts
- `tmp` - temporary files

Images are stored as `SHA512.json`, layers as `SHA512.squashfs`, with images supporting "tagging" via symlinks (e.g. `manifests/cedarish.json` -> `manifests/1062edd28317eff4f58be2da2fde33c6fb8477bfe7b788f7c0ce10a83e8a297c.json`)
---
### convert docker -> squashfs

I've added `util/imagebuilder` which takes a reference to a Docker image and creates squashfs layers for each tagged layer in the image's history.

For example `flynn/slugrunner` is converted to three layers: ubuntu, cedarish and the slugrunner scripts, whereas `flynn/controller` becomes two layers: busybox + controller binaries.

The generated layers are tagged with the ID of the corresponding Docker image to avoid converting already converted images.

We probably want to support building squashfs layers from scratch, but converting from Docker images gives us a good starting point.
---

Flynn partially boots but I had some instances of files being missing, but thought it was worth getting feedback on what has been done so far.


lmars:
@titanous I didn't add `util/imagebuilder` to `tup` as I was getting the following error when referencing it from other Tupfiles:

```
Unable to use inputs from a generated directory
```

I didn't take the time to figure out how to do that, but it looks like we use `util/release` in the same fashion, I just couldn't wrap my head around the `tup` syntax 😛

titanous:
I'm going to think about the JSON manifest a bit today, we probably want to reorganize it a bit.

I'm not sure about the "tagging", what is the use case? Instead, we can put an image name in the manifest.

titanous:
The "generated directory" error from tup is because you are putting the binary into a `bin` directory, which tup doesn't like for some stupid reason. Instead, put the binary beside the source code, then you will be able to reference it from other Tupfiles.

host/libcontainer_backend.go
```diff
@@ -389,17 +397,47 @@ func (l *LibcontainerBackend) Run(job *host.Job, runConfig *RunConfig, rateLimit
...
-	// TODO(lmars): stream pull progress (maybe to the app log?)
-	imageID, err := l.pinkerton.PullDocker(artifactURI, ioutil.Discard)
-	if err != nil {
-		log.Error("error pulling image", "err", err)
-		return err
+
+	var rootPath string
+	switch job.ImageArtifact.Type {
+	case host.ArtifactTypeDocker:
+		log.Info("pulling image")
```
titanous: I assume the plan is to delete this code entirely? We should translate any images into Flynn format before running them.

lmars:
@titanous 

> I'm not sure about the "tagging", what is the use case? Instead, we can put an image name in the manifest.

Yes I'm not sold on it, I just wanted a quick way to prevent rebuilding all the squashfs layers from the Docker images, so having the symlink be the Docker image ID meant I could do a simple lookup before building.

> The "generated directory" error from tup is because you are putting the binary into a bin directory, which tup doesn't like for some stupid reason. Instead, put the binary beside the source code, then you will be able to reference it from other Tupfiles.

Ok that fixed it thanks, I would never have realised that!

host/libcontainer_backend.go
lmars: Absolutely, I just wanted to do the minimum amount of work in the short term.

lmars:
@titanous my current thinking is that we should add a `squashfs` volume provider so we can re-use the volume API for managing squashfs layers, and have the following operations:

* `Create` - given a directory tree (e.g. a tar stream), make a squashfs layer, generate sha512 and mount it
* `Import` - given a squashfs file, write it to disk and mount it
* `Export` - given an ID, return the squashfs file

The `Import` / `Export` operations would help when releasing existing layers and fetching new ones.

We could also have a `tmpfs` volume provider which provides the read-write layer for jobs, then we can have other jobs mount those same tmp volumes and pass them to the `squashfs` provider. This would then help to create slugs as squashfs layers by just running a bunch of jobs.

WDYT?

titanous:
Currently the volume API doesn't really match those primitives, but it does seem like using the same API to manage image layers, tmpfs, zfs, nfs, and so on makes sense. Any thoughts @josephglanville?

lmars:
@titanous I've also been thinking about replacing the current image artifacts with a more generic "mount spec" (I know we touched on this briefly before) which has a list of mounts, and any appearing at the same path get overlayed? For example:

```json
[
  {
    "type":   "squashfs",
    "id":     "xxx.squashfs",
    "target": "/"
  },
  {
    "type":   "squashfs",
    "id":     "yyy.squashfs",
    "target": "/"
  },
  {
    "type":   "squashfs",
    "id":     "zzz.squashfs",
    "target": "/"
  },
  {
    "type":   "tmpfs",
    "id":     "uuid",
    "target": "/"
  },
  {
    "type":   "zfs",
    "id":     "uuid",
    "target": "/data"
  }
]
```

With my previous volume proposal, the individual objects in a mount spec would just be volumes.

titanous:
What object would a "mountspec" be contained within?

lmars:
A job would have a mountspec, but we would need a way to generate that mountspec from one of our image manifests.

Perhaps we still have the image artifact which is a reference to the image manifest, and its purpose is to provide a set of mounts to stick in the mountspec?

I guess mountspec squashfs entries would need a URI so flynn-host can fetch any it doesn't have locally.

titanous:
Yeah, we need some concept of "images" which are static and can be retrieved and cryptographically verified. There is also a question if the source of defaults like entrypoint, env vars, etc. is sourced from the image like in most of the other existing container systems.

josephglanville:
I agree that we should look at trying to unify the storage backends in some way. The current volume provider API needs work anyway, it lacks appropriate metadata to implement persistence properly for instance. I was considering just add a Meta field to volume.Info but perhaps a broader rethink is necessary to properly accomodate NFS etc. It also feels like we are missing a scheduler subsystem for handling volume management. I attempted to work on this in the persistent volumes branch but ripped it out because it needs more thought.

I had some thoughts regarding something similar to the mountspec but perhaps at a different layer. The example above implies it's resolved when submitting the job to flynn-host, either by the scheduler or flynn-host itself. I was envisioning a construct that would extend the current process type description with more powerful configuration like what storage to attach to each job, it would perhaps be very similar to a "resolved/expanded" version of the mountspec but be exposed via the controller API. This would enable multiple volumes per process along with ability to mix and match storage data sources.

titanous:
@josephglanville That makes sense. There is a significant difference between a release process type and a job configuration, as a process type is template, and a job is an instance of that template. This means that it probably doesn't make sense to reference a read-write volume directly in the process type, instead we need some kind of other higher level representation that is host-agnostic.

titanous:
Also, from an API UX perspective, it'd be ideal if the mountspec resolution happened at a lower level so that consumers can just point at an "image" reference (the usual case) and have the system sort out what that means at a lower level.

josephglanville:
Yep, it would be similar but definitely very different to the resolved/expanded version that is actually used to start a job.

lmars:
@titanous 

> we need some concept of "images" which are static and can be retrieved and cryptographically verified

Maybe the mountspec can reference an image directly:

```json
[
  {
    "type":   "image",
    "id":     "uuid",
    "target": "/"
  },
  {
    "type":   "tmpfs",
    "id":     "uuid",
    "target": "/"
  },
  {
    "type":   "zfs",
    "id":     "uuid",
    "target": "/data"
  }
]
```

It would be expected that the given image has already been downloaded / verified / extracted locally (this is the case now, with images pulled from the TUF repository either by the initial `flynn-host download` or a `flynn-host update`), then `flynn-host` would just mount the appropriate `squashfs` layers from the manifest.

At a higher level, releases would still have an image artifact pointing at the image manifest, but rather than passing that in the job config it would create a mountspec as above.

> There is also a question if the source of defaults like entrypoint, env vars, etc. is sourced from the image like in most of the other existing container systems.

I'm not sure we need those since for internal images we can just put them in the bootstrap manifest / release config, and for Docker images we grab them with `docker inspect` and add them to the release. If we do need them though, I guess we could just add them to the image manifest?

@josephglanville I agree we will want to expose this at a higher layer, but I'm imagining having something which is compatible with the current high level view but provides enough flexibility to change things in the future.

lmars:
I also wonder whether we need a separate `tmpfs` layer, could just be a read-write `zfs` volume which gets garbage collected.

titanous:
> It would be expected that the given image has already been downloaded / verified / extracted locally

What about user-defined images and slugs?

titanous:
>  If we do need them though, I guess we could just add them to the image manifest?

Yeah, I think that information will be in the manifest, but I'm okay with a higher layer copying it into the job config to be very explicit.

titanous:
So thinking about this a bit more, I think that the image should be "dereferenced" before it shows up in the job. The Job would provide the direct URL to each layer, and cryptographic digest. The digest would be used for caching and verification.

We don't need to worry about platform architecture, signing/TUF, default entrypoints, etc. at the `flynn-host` level, other components will sort that out before building the job request.

lmars:
@titanous 

> the image should be "dereferenced" before it shows up in the job

So I assume we would need to store the image manifest in the database so we don't have to fetch it every time to determine what the layers are?

titanous:
> So I assume we would need to store the image manifest in the database so we don't have to fetch it every time to determine what the layers are?

Yeah, absolutely, this would be tracked in the controller, perhaps as an artifact.

lmars:
@titanous for bootstrap, do you think we should continue to reference the image manifest URL (and thus download the manifest at bootstrap time using the TUF db) or put the layer references in there directly? The bootstrap manifest itself is fetched from the TUF repo so we should be able to trust any direct layer references it contains.

The reason I ask is I'm just wondering how bootstrap will work in development when the layers exist on disk, but the image manifest won't be available at a URL, unless we maybe support `file://` URLs?

titanous:
I'm totally fine with putting the resolved layer details in the bootstrap manifest.

titanous:
FWIW, the controller will definitely end up caching the image details, so the bootstrap manifest references could also be in the format of whatever data structure that is.

lmars:
@titanous I'm thinking of adding `Artifact.Data []byte` which is a cache of the data available at the URL, WDYT?

lmars:
And maybe then the `Artifact` then doesn't even need a URL?

titanous:
One clarification: image manifests will be signed when served over the Internet, so they will necessarily be from a TUF repo, this means that the "URL" will actually be a reference to a specific TUF repo and image manifest within it.

I'm fine with the Artifact storing the entire manifest, it should probably be in a `jsonb` field called `manifest` or similar. The artifact should keep an optional reference to the source of the manifest, but it shouldn't be required (for example, uploading a manifest using the CLI tool).

lmars:
Some TODOs:

- [ ] cleanup mounts when a job exits
- [ ] maintain whiteouts when converting docker images

titanous:
I think the manifest format should be something like this.

For symmetry with TUF I've kept the `hashes` object, in practice we will require a SHA512 there. Allowing multiple entrypoints allows us to parse the Procfile ahead of time and drop each process type in the manifest. The special `_default` entrypoint (if specified) is used for everything else. The `linux_capabilities` are a strict subset of those that are permitted. `seccomp` can be ignored for now but I just wanted to see what it would look like.

One or more `rootfs` objects are specified, one per platform (we can add more specificity like CPU features and other variants if necessary). We need to figure out what the protocol for building URLs to the relevant files is, it should probably be compatible with TUF.

Let me know what you think.

```json
{
  "_type": "application/vnd.flynn.image.manifest.v1+json",
  "meta": {
    "name": "flynn/slugbuilder",
    "version": "v20160712.0",
    "arbitrary": "value"
  },
  "entrypoints": {
    "_default": {
      "env": [
        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
        "TERM=xterm"
      ],
    },
    "web": {
      "cwd": "/",
      "args": [
          "sh", "-c", "server $PORT"
      ],
      "env": [
        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
        "TERM=xterm"
      ],
      "linux_capabilities": [
        "CAP_AUDIT_WRITE",
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE"
      ],
      "seccomp": {
        "defaultAction": "SCMP_ACT_ALLOW",
        "architectures": [
          "SCMP_ARCH_X86"
        ],
        "syscalls": [
          {
            "name": "getcwd",
            "action": "SCMP_ACT_ERRNO"
          }
        ]
      }
    }
  },
  "rootfs": [
    {
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      },
      "layers": [
        {
          "type": "application/vnd.flynn.image.squashfs.v1",
          "length": 7682,
          "mountpoint": "/",
          "hashes": {
            "sha512": "0da4aa2735dcc289b6174d39fc20a0dbfe30cb65c36492c664a2031da2366811605617f91f0508d8d282fc475f6663fcd9883e8f2a3e909592ba1604408fd930"
          },
       },
       {
          "type": "application/vnd.flynn.image.squashfs.v1",
          "length": 2023,
          "mountpoint": "/",
          "hashes": {
            "sha512": "0da4aa2735dcc289b6174d39fc20a0dbfe30cb65c36492c664a2031da2366811605617f91f0508d8d282fc475f6663fcd9883e8f2a3e909592ba1604408fd930"
          }
        }
      ]
    }
  ]
}
```

lmars:
We will need to update the install script to check that overlayfs is present and supports multiple lower dirs.

lmars:
With regards to my previous comment about having a squashfs volume provider, I wonder whether we should put the squashfs files into zvols (e.g. using `dd if=squashfs of=/dev/zvol/...`) then just manage them using the the ZFS volume API? Caching would then be done by the ZFS ARC rather than kernel file cache, not sure if that would be a good thing or not.

I also think the read-write layer should also just be a ZFS filesystem layered on top of the read-only squashfs layers.

Having these managed by ZFS means we have one unified way of managing layers / job fs changes / slugs / job data.

@josephglanville WDYT?

josephglanville:
Using ZVOLs is an interesting idea, you would get ZFS data management but retain squashfs insane on-disk compression.

It also means ZFS won't be in the VFS chain when doing overlayfs shenanigans.

Can't think of any caveats other than the burst write throughput to the ZFS pool when you write out the squashfs images may be annoying to discoverd/postgres (but can probably be mitigated by  using a large block size).

Overall I think it's a good idea, could prove very useful for backups and stateful containers layer.

titanous:
Very interesting, definitely worth trying. If we can get throttling under control, I think it'd be great to store all data including images in ZFS.

We can try this setting to disable the ARC/page cache duplication:

```text
zfs set primarycache=metadata pool/zvol
```

lmars:
Made a few changes:

* Modelled the proposed image manifest in `controller/types/types.go`
* Added `Artifact.Manifest` which is a cache of the image manifest stored at the artifact's URI
* Added `Artifact.Meta["layer_url_template"]` which is a Go template to determine the layer URIs for the image manifest (uses `file://` URIs in dev as the layers are on disk, but will use `https://` URIs pointing at the TUF repo when released)
* Updated `util/imagebuilder` to build squashfs layers into `/var/lib/flynn/layer-cache` and image manifests into `image/` which are used to build the bootstrap manifest
* Added `job.Mountspecs` which is a slice of mountspecs which are passed to `libcontainer` to be mounted (with any mountspecs at the same path being mounted with OverlayFS)
* Added `ImportVolume` which takes a data stream, imports it into a zvol and mounts it
* Updated the `libcontainer` backend to fetch any missing layers and import them before starting the job

@titanous @josephglanville let me know if you are happy with the general direction here, and then maybe I'll make smaller PRs to a feature branch.

titanous:
The layer URL template concept is interesting, should it be an actual field on the Artifact? The main downside I see is that it is specific to Go, so if you want to write a program that uses it in a different language, the template language would have to be reimplemented.

lmars:
@titanous we could use an [RFC 6570 URI template](https://tools.ietf.org/html/rfc6570)?

titanous:
Yeah, perfect.

josephglanville:
RFC 6570 looks cool. Easy to implement and a cursory look seems there are already implementations for most languages. 👍

titanous:
After discussion in IRC, this is what we landed on for the read-write temporary filesystem layer:

An ext2 filesystem in a sparse file mounted via loop device for each temporary filesystem. The sparse file will be created from an empty ext2 sparse file template that is copied, [`SEEK_DATA`/`SEEK_HOLE`](http://man7.org/linux/man-pages/man2/lseek.2.html) can be used to copy the template quickly. The filesystem should be mounted with `noatime`.

This strategy gives us the following properties

- The layer has a hard quota in the form of the maximum size of the sparse file, but only consumes space when data is written.
- The layer has a O(1) destroy time via `unlink(2)`.



Other options considered:

- A ZFS dataset per layer. This gives us the quota capability and O(1) destroy, but ZFS is not currently supported by OverlayFS due to the lack of dentry revalidation.
- A ZFS zvol with an ext2 filesystem per layer. This would have the same quota and O(1) destroy capabilities, but the creation time was determined to be too slow. May be worth revisiting in the future.
- A directory per layer on a shared filesystem. This has the fastest setup time, but quotas can only be enforced at the UID level, which can be problematic for some use cases. The O(N) destroy time is also problematic.
- RAM-backed tmpfs. Some hosts will not have enough memory to support this. In the future we may allow choosing between a memory-backed tmpfs and the on-disk layer type based on application requirements.

This is a summary based on my understanding of the situation, please feel free to make any relevant corrections.

titanous:
Here's another take on the zvol strategy. What if we copy the ext2 metadata out of a sparse file into a zvol? That would mitigate the slow initialization time. The only question is: how long does it take to create a zvol?

lmars:
@titanous @josephglanville I decided to give ext2 on zvol (initialised from a sparse file) a try and the startup times seem reasonable: 221fb33

I will do some benchmarks tomorrow to see what the startup time is with this, but bootstrapping doesn't seem to be much slower (it did when creating an ext4 FS per job).

Side note: I realised `cp(1)` is "hole" aware so I assumed it too used `SEEK_DATA` / `SEEK_HOLE` but after observing with `strace`, as if by magic it knew the exact hole locations and just seeked the destination file. On closer inspection, I spotted:

```
ioctl(3, FS_IOC_FIEMAP, 0x7ffcbcd6e8e0) = 0
```

So it turns out `FS_IOC_FIEMAP` can be used to find the holes in a single syscall, and there is a [Go package](https://github.com/rancher/sparse-tools/blob/master/sparse/fiemap.go) too :smile:.

titanous:
You'll need to use a different fd for every parallel copy operation, easiest option is just to open and close the file for each template instantiation.

lmars:
I tracked how long it takes to open the sparse ext2 file, create the zvol and copy the data (i.e. from [here](https://github.com/flynn/flynn/blob/squashfs-images/host/libcontainer_backend.go#L813) to [here](https://github.com/flynn/flynn/blob/squashfs-images/host/libcontainer_backend.go#L831)), and also the total time to start the job (i.e. from [here](https://github.com/flynn/flynn/blob/squashfs-images/host/libcontainer_backend.go#L352) to [here](https://github.com/flynn/flynn/blob/squashfs-images/host/libcontainer_backend.go#L918)) and here is the data from starting 17 jobs on an EC2 m4.4xl instance with EBS gp2 root (the `percent` column is the percentage of total time spent creating the tmpfs):

```
tmpfs=67.301665ms total=124.201773ms percent=0.541874
tmpfs=63.094103ms total=144.964419ms percent=0.435239
tmpfs=63.006228ms total=170.772558ms percent=0.368948
tmpfs=63.942717ms total=203.407071ms percent=0.314358
tmpfs=99.415729ms total=219.158817ms percent=0.453624
tmpfs=79.536587ms total=178.995771ms percent=0.444349
tmpfs=67.201303ms total=188.499653ms percent=0.356506
tmpfs=81.062507ms total=271.373432ms percent=0.298712
tmpfs=83.608593ms total=315.719853ms percent=0.264819
tmpfs=73.639707ms total=438.163966ms percent=0.168064
tmpfs=68.64048ms  total=170.830383ms percent=0.401805
tmpfs=85.683993ms total=186.457087ms percent=0.459537
tmpfs=63.601868ms total=240.21612ms  percent=0.264769
tmpfs=69.269892ms total=283.677592ms percent=0.244185
tmpfs=71.700226ms total=188.194732ms percent=0.380990
tmpfs=66.570033ms total=227.93368ms  percent=0.292059
tmpfs=68.703992ms total=252.667574ms percent=0.271915
```

So it takes on average 72.7047ms to create the tmpfs, and that is on average 35% of the total time to start the job.

Do we think that is acceptable?

titanous:
I think that's fine for now. We can do all sorts of optimizations later, including storing the template in-memory, having a queue of prepared zvols, partitioning larger zvols, etc.

Great work!

lmars:
I've been trying to get slugs working with squashfs / overlayfs but mounting it is proving tricky.

My first approach was to mount the slug squashfs at `/app` *after* mounting the base squashfs + tmpfs overlayfs at `/`, but then this makes `/app` read-only (as it is a direct mount of the squashfs on top of the tmpfs).

I then instead overlayed the tmpfs over `/app`, but then that leaves the rest of the filesystem read-only. We could have a tmpfs per distinct mountpoint (i.e. one over `/` and one over `/app`) but that doesn't feel right.

I then tried to mount the slug squashfs as part of the root overlayfs by first creating a `TMPDIR`, bind mounting `TMPDIR/app` to the mounted slug and adding `TMPDIR` as a lowerdir in the overlayfs, but overlayfs does not follow the bind mount, it just shows an empty `/app` directory (see [here](http://askubuntu.com/questions/654508/how-can-i-mount-a-lowerdir-into-a-specific-subdirectory-with-overlayfs) for someone having a similar issue).

Another approach is to have the `/app` directory inside the squashfs layer, then it can just be overlay mounted at `/`, but this provides less flexibility.

So if we go with one of the above approaches, we either have:

* `/app` read-only but everything else read-write to the tmpfs
* `/app` read-write to the tmpfs but everything else read-only
* `/app` and `/` read-write with two distinct tmpfs overlayed over each
* have `/app` exist inside the squashfs and overlay mount it at `/`

@titanous @josephglanville thoughts?

titanous:
I think using the `/app` subdirectory in the squashfs image is the best option. Any immediate downsides to that approach?

lmars:
@titanous I don't think there any immediate downsides, I'll give that a go thanks.

lmars:
Adding `/app` to the squashfs layer seems to work fine, and so for now I have simplified things by expecting all layers in the mountspec to be mounted at `/` (we can modify this in the future, but we will need to deal with those mounts potentially being read-only).

The `slugbuilder` now creates an image artifact for the slug by creating a squashfs layer, uploading it to the blobstore and adding the layer to those of the slugrunner artifact, using `SLUG_ARTIFACT_ID` which is set by the receiver as the ID so the receiver knows what to add to `Release.ArtifactIDs`.

This does mean I have had to drop the `layer_url_template` because now the slug image has layers from multiple sources (TUF repo + blobstore), and I have instead added a URL field to layers, which isn't great because now the image manifest SHA will change depending on where you upload it to. 

Another approach is to have two separate image artifacts (one slugrunner, one slug) and relax the condition that releases have a single image artifact, but instead can have multiple image artifacts which are merged into one when populating `job.Mountspecs`. I didn't try this as it is quite a big refactor (lots of code expect there to be a single image artifact), but if we think that is the right approach I can do that.

titanous:
I think we definitely want to avoid putting the actual URLs in the manifest, the idea is that images should be content-addressable, the source should not matter. I wonder if the manifest should be able to incorporate the layers from another manifest/artifact by reference?

lmars:
If the manifest / artifact references other manifests / artifacts, then most of the code which expects a single image artifact (e.g. export / import and backup / restore) will need to be changed anyway to handle this relationship, so given releases already have many ordered artifacts, maybe we should just stack them that way?

titanous:
Yeah, I'm fine with that.

lmars:
I have updated docker-receive to convert Docker images to Flynn squashfs images (f1d7e7e).

Once an image is fully pushed, the image is pulled locally to the docker-receive job (into `/var/lib/docker`), the imagebuilder creates a squashfs layer for each layer in the Docker image (into `/var/lib/flynn`), the layers and manifest are uploaded to the blobstore and an artifact is created.

This approach was simple to implement (just add to the code which currently creates artifacts), but does mean that the image is effectively uploaded to the blobstore twice (once as the Docker diffs, and again as the squashfs layers), and the push takes some time to finalize after being pushed (all the squashfs layers are created at the end rather than as the individual layers are pushed). I feel like the simplicity of this approach outweighs the impact of extra blobstore storage / non-concurrent squashfs layer generation.

titanous:
Can we do the image conversion in a new job for each task? Also, what is `CAP_SYS_ADMIN` needed for?

lmars:
> Can we do the image conversion in a new job for each task?

Sure.

> what is CAP_SYS_ADMIN needed for?

Needed for the Docker AUFS driver (specifically [this line](https://github.com/docker/docker/blob/v1.9.1/daemon/graphdriver/aufs/aufs.go#L133-L135) which re-mounts `/var/lib/docker/aufs`, and mounting requires `CAP_SYS_ADMIN`).

titanous:
> Needed for the Docker AUFS driver

Can we use a different driver or none at all? We're just packing the individual layers, right?

lmars:
> Can we use a different driver or none at all? We're just packing the individual layers, right?

I started with the `vfs` driver (i.e. just plain directories) but it didn't seem to work, I'll experiment further.

lmars:
@titanous I added a very simple graph driver which just stores the diffs as local files, and also moved the artifact creation to a separate job: fdbc278

lmars:
Replaced by #3619.

PR #3103, opened 2016-07-12
appliance/mongodb: Scale up on first resource request

josephglanville:
Similar to MariaDB this has the MongoDB appliance only start up once the first resource is requested to be created.


pkg/sirenia/util/util.go
```diff
@@ -0,0 +1,135 @@
+package util
```
titanous: Can we call this something other than `util`?

titanous:
LGTM pending package rename.

pkg/sirenia/util/util.go
josephglanville: Renamed to `scale` instead.

titanous:
👍

PR #3109, opened 2016-07-13
appliance/mariadb: Fix waitReadWrite in integration tests

### LINKED ISSUE #3065: MariaDBSuite waitReadWrite/assertDownstream interaction is flawed
Unlike Postgres we aren't able to properly detect from transaction read-only state if the database is writeable. The master sets the database into read-write mode when it's configured with a downstream peer but it doesn't truely become writeable as writes will block if the semi-sync plugin doesn't recieve the appropriate number of acks. Because of this waitReadWrite probably needs to be updated to actually check for the peer status. This probably makes assertDownstream redundant in a few places.

josephglanville:
A more detailed explaination is why this change is necessary is
in the linked issue.

Closes #3065

### DIFF
--- a/appliance/mariadb/process_test.go
+++ b/appliance/mariadb/process_test.go
@@ -148,6 +148,7 @@ func insertRow(c *C, db *sql.DB, n int) {
 }
 
 func waitReadWrite(c *C, db *sql.DB) {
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

PR #3110, opened 2016-07-13
scheduler: Start persistence loop before syncing jobs

lmars:
If there are more than 1000 current jobs, the scheduler will deadlock as the `putJobs` channel will be full with no goroutine receiving from it.


josephglanville:
LGTM

PR #3112, opened 2016-07-13
cli: prevent empty environment variable keys

### LINKED ISSUE #3029: Flynn should not allow empty environment variable
While changing several environment variables, I must've fat fingered one and ended up created an env var named "". As a result of this env var, `git push` started failing like this:

```
remote: tar: env: Cannot open: File exists
remote: tar: Exiting with failure status due to previous errors
```

After removing the var:
```
> flynn -a $app env unset ''
```

I could push again.

slurms:
Fixes #3029. Not sure if this is at all correct... first attempt at go :).


controller/release.go
```diff
@@ -63,6 +63,15 @@ func (r *ReleaseRepo) Add(data interface{}) error {
 		release.ArtifactIDs = []string{release.LegacyArtifactID}
 	}
 
+	for key := range release.Env {
+		if key == "" {
```
lmars: Since `release.Env` is a map, you can query for the key directly:

```go
if _, ok := release.Env[""]; ok {
	// empty key exists
	return ...
}
```

slurms: Nice, I'll do that instead.

slurms:
Not sure what those test failures are caused by :/

josephglanville:
@slurms those failures are unrelated, your changes aren't to blame.
Unfortunately some flakiness has crept into the test suite, we are working on making it green again. 😄

controller/release.go
```diff
@@ -63,6 +63,13 @@ func (r *ReleaseRepo) Add(data interface{}) error {
 		release.ArtifactIDs = []string{release.LegacyArtifactID}
 	}
 
+	if _, ok := release.Env[""]; ok {
+		return ct.ValidationError{
+			Field:   "env",
+			Message: "you can't create an env var with an empty key",
```
lmars: Given this is usually done by mistake, it would be useful to include the value in the response, for example:

```
$ flynn env set =BAR
validation_error: env you can't create an env var with an empty key (tried to set ""="BAR")
```

slurms: Ah, yep -- seems like a more helpful message :+1:

controller/release.go
```diff
@@ -63,6 +63,13 @@ func (r *ReleaseRepo) Add(data interface{}) error {
 		release.ArtifactIDs = []string{release.LegacyArtifactID}
 	}
 
+	if _, ok := release.Env[""]; ok {
+		return ct.ValidationError{
+			Field:   "env",
+			Message: fmt.Sprintf("you can't create an env var with an empty key (tried to set \"\"=\"%s\")", release.env[""]),
```
lmars: Update the `if` statement so that it assigns the value to a variable and then use it here (rather than using `release.Env[""]` again):

```go
if value, ok := release.Env[""]; ok {
	// ... can reference `value` here
}
```

You can also use `%q` rather than `\"%s\"` to double-quote the variable in the output string.

slurms: Haha, sorry about the small issues (and thanks for reviewing)

PR #3114, opened 2016-07-13
appliance/mariadb: Improve logging in integration test

josephglanville:
Still tracking down sporadic failures, issues only occur in CI so this should help track them down.


lmars:
LGTM

PR #3117, opened 2016-07-13
test: Add pkg-config to rootfs

lmars:
This is required for building `libcontainer` and was previously pulled in when installing libvirt-bin.


titanous:
LGTM

PR #3120, opened 2016-07-13
cli: require typed confirmation for cluster delete

### LINKED ISSUE #3020: Require typed confirmation for installer cluster delete
The deletion confirmation should require typing the cluster name before proceeding to prevent accidental deletion.

slurms:
Closes #3020, although this still might require some test changes...

Signed-off-by: Nick Sandford <nick@sandford.id.au>

### DIFF
--- a/cli/cluster.go
+++ b/cli/cluster.go
@@ -221,6 +221,19 @@ func writeCACert(c controller.Client, name string) (string, error) {
 func runClusterRemove(args *docopt.Args) error {
 	name := args.String["<cluster-name>"]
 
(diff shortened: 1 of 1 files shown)


titanous:
@slurms That issue is actually about deletion of a provisioned cluster in the Web UI available from `flynn install`. The `flynn cluster remove` command just deletes a cluster config from `~/.flynnrc` which is not a dangerous operation.

slurms:
Oh... :laughing:

titanous:
🏆

PR #2366, opened 2016-01-12
util/packer: Pin kernel version to work around Docker/AUFS bug

titanous:
Ubuntu kernel 3.19.0-39 is does not trigger this bug: https://github.com/docker/docker/issues/18180

Verified with:

    docker run -it --rm akihirosuda/test18180

Fixes #2365

PR #3121, opened 2016-07-13
test,util/packer: Use latest Ubuntu xenial kernel

lmars:
The bug causing us to pin the kernel version (see #2366) has been fixed so we should be safe to upgrade, and we need a newer kernel version for the upcoming squashfs / overlayfs image changes.


titanous:
LGTM

PR #3024, opened 2016-06-30
script: Add dashboard-dev to help testing dashboard changes

lmars:
@jvatic this puts [your gist](https://gist.github.com/jvatic/df346cfdf1632ca0ccd3) into a script.

Workflow:

* `script/dashboard-dev compile` compile assets into `dashboard/app`
* `script/dashboard-dev run` updates bindata.go, builds and runs the dashboard on port 4457
* make code change, re-run `script/dashboard-dev compile`

Issue #1624, opened 2015-06-27
installer: Document development flow


jvatic:
The development flow for the installer should be documented in https://flynn.io/docs

When working on the installer app it's much faster to run it on it's own rather than running make and bootstrapping after each change.

Steps for getting it up and running:

1) Create or replace `installer/bindata.go` with the contents of https://gist.github.com/jvatic/9aa507d330b4c80e644a if working on the front-end (you'll need to repeat this if/after running `make` in the VM).

2) `cd installer/app && ENVIRONMENT=development go run compiler.go` (repeat after each modification to a JavaScript or CSS file).

3) From `{project root}/installer`, run `go build ../cli && ./cli install` (repeat after each modification to a go file, be sure not to commit the `cli` binary (you can delete it once the installer is running)).

PR #3128, opened 2016-07-14
Add dev script to compile/run the installer

jvatic:
Similar to the existing script for the dashboard (#3024).

Refs #1624

Workflow:

- `script/installer-dev compile` compiles the assets into `installer/app`
- `script/installer-dev run` updates `installer/bindata.go`, then builds and runs the installer on port 4455
- re-run the first step after each asset code change, and the second after changes to the Go code

### DIFF
--- a/installer/.gitignore
+++ b/installer/.gitignore
@@ -1,1 +1,2 @@
-bindata.go
+/bindata.go
+/cli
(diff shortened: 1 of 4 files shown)


lmars:
LGTM

josephglanville:
LGTM

PR #3131, opened 2016-07-15
host/cli: Show host Raft status in list

josephglanville:
If discoverd is available show the Raft status for each host, ie:
```
$ flynn-host list
ID     ADDR              RAFT STATUS
host1  192.0.2.201:1113  peer
host2  192.0.2.202:1113  peer
host0  192.0.2.200:1113  peer (leader)
host3  192.0.2.203:1113  proxy
```

### DIFF
--- a/host/cli/list.go
+++ b/host/cli/list.go
@@ -2,9 +2,11 @@ package cli
 
 import (
 	"errors"
+	"net"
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

PR #3135, opened 2016-07-16
appliance/mongodb: Stop peer before stopping heartbeater

josephglanville:
Prevents the peer state machine from running `replSetReconfig` after it should have been stopped. Further work is required to ensure safety in some failure cases but this will stabilise the tests in CI.


titanous:
LGTM

PR #3139, opened 2016-07-16
host: Simplify Upstart start on stanza

### LINKED ISSUE #3134: Flynn-host service prevents boot
I am running tests on the latest nightly `v20160714.0`. I am unable to get it to boot after running install-flynn. Here are some descriptions of the environment I am running Flynn in and the steps I have taken.

I am running the hosts in Redhat Enterprise Virtualization. Here is the software that RHEV reports for the VM hosts:
OS Version: RHEV Hypervisor - 7.2 - 20160413.0.el7ev
Kernel Version: 3.10.0 - 327.13.1.el7.x86_64
KVM Version: 2.3.0 - 31.el7_2.10
LIBVIRT Version: libvirt-1.2.17-13.el7_2.4
VDSM Version: vdsm-4.17.26-0.el7ev

The Flynn host install is directly from the latest Ubuntu 14.04.4 server ISO. 

The disks for the VMs are located on iSCSI storage pools managed by the RHEV cluster. I have created the VMs with 2 disks, one for the OS install and one that I create a zpool on directly using `sudo zpool create -f flynn-default /dev/vdb`. 

After creating the zpool I am able to reboot the hosts repeatedly with no issue. I then run `sudo bash ./install-flynn --channel nightly`. When I reboot the host after successful completion of the `install-flynn` script I the host locks on boot at this stage: flynn-boot-fail

I booted the host into recovery using the ISO and disabled the flynn-host service. I was then able to successfully boot. After it booted I ran `service flynn-host start` and it brought up the `flynn-host` daemon.

lmars:
The previous stanza was causing Upstart to hang on boot, perhaps some kind of circular dependency (I'm not sure).

The runlevel check has the same effect (see http://upstart.ubuntu.com/cookbook/#startup-process), and does not cause Upstart to hang.

Tested by installing Flynn in a VM, modifying `/etc/init/flynn-host.conf` and rebooting successfully multiple times.

Closes #3134

### DIFF
--- a/host/upstart.conf
+++ b/host/upstart.conf
@@ -1,6 +1,6 @@
-description "Flynn layer 0"
+description "Flynn host daemon"
 
-start on (local-filesystems and net-device-up IFACE!=lo and mounted MOUNTPOINT=/sys/fs/cgroup)
+start on runlevel [2345]
 stop on runlevel [!2345]
 respawn
 respawn limit 100 60


titanous:
LGTM

PR #3142, opened 2016-07-17
Sirenia stop peer heartbeater

josephglanville:
Mirrors the change made to the MongoDB appliance, this is more correct and may potentially close small windows where clients remain connected to the old primary for longer than necessary.


titanous:
LGTM

Issue #410, opened 2014-11-07
User accounts


temujin9:
We need user account management. A full-blown LDAP permission integration would be nice, but SSH keys taking the place of the current `flynn cluster add` stanza (which is currently the shared key to the whole cluster) would be good enough. 

Ideally, the keying should not be cluster-specific, to make migration between clusters easier.

kevinob11:
Am I correct that as of right now there is only one set of credentials to auth to a cluster when you add it via `flynn cluster add`? Is there a way to set up multiple sets of credentials or to reset / change them when someone on my team leaves?

temujin9:
@kevinob11 Currently, you are correct. This is the ticket for fixing that issue, and I believe it is fairly high on their queue. We definitely want it at our company, too.

kevinob11:
Thanks @temujin9 I got worried when I hadn't seen any action here since 2014. Happy to help in any way I can.

tjboring:
Any status on when this one might be done?

tjboring:
Until this is done, anyone have recommendations for dealing with adding new clusters? Currently what I'm doing is this (and it's a pain-in-the-ass):

- Document the cluster add command in an internal wiki
- Put the TLS pin and cluster key in LastPass
- Tell other engineers to get the pin/key from LastPass and run `flynn cluster add`

titanous:
I'm currently primarily working on this, I should have some news soon. That procedure sounds right, though only the key is secret, the TLS pin is not secret.

icarlosmendez:
Where do I acquire this elusive TLS pin? Is it the one that's listed in the documentation on https://flynn.io/docs/cli#adding-clusters? I simply don't see how it would have been created for me locally. I am running the demo in a vagrant box for context.

titanous:
@icarlosmendez It is printed as part of the `flynn cluster add` command, as well as stored in `~/.flynnrc` after running the command.

jokeyrhyme:
Have you considered partnering with something like GitLab for these purposes?

- https://about.gitlab.com/
- https://gitlab.com/gitlab-org/gitlab-ce/tree/master#gitlab

GitLab has:

- user directory and groups
- allows project owners to manage per-user and per-group permissions
- allows users to register and manage their SSH keys
- authenticates access to projects using these SSH keys

What if GitLab was installed in a container, just like the default control-plane's PostgreSQL service?

maxlazio:
[...]

PR #3143, opened 2016-07-17
pkg/mauth/compare: Add compare package

titanous:
I'm going to try to avoid one big PR for multi-user. Here's the first small component: a package that will be used for encoding comparison operators in macaroon caveats.

Refs #410

### DIFF
--- a/pkg/mauth/compare/compare.go
+++ b/pkg/mauth/compare/compare.go
@@ -1,0 +1,334 @@
+// Package compare provides various compare operators with binary codecs
+// useful for macaroon-based authorization.
+package compare
+
+import (
+	"encoding/binary"
+	"fmt"
+	"net"
+	"regexp"
+)
+
+type Type uint8
+
+const (
+	TypeFalse    Type = 0
+	TypeTrue     Type = 1
+	TypeIntegers Type = 2
+	TypeStrings  Type = 3
+	TypeRegexp   Type = 4
+	TypeCIDRs    Type = 5
+)
+
+// comparisons are encoded as a single type byte followed by zero or more bytes
+// that are specific to the comparison type.
+//
+// UINT8 - type
+// UINT8 0..n - bytes
+
+func UnmarshalBinary(b []byte) (interface{}, error) {
+	if len(b) < 1 {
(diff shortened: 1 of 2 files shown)


pkg/mauth/compare/compare.go
```diff
@@ -0,0 +1,311 @@
...
+		v := new(CIDRs)
+		return *v, v.UnmarshalBinary(b)
+	default:
+		return nil, fmt.Errorf("compare: unknown type %d", b[0])
+	}
+}
+
+type Bool bool
+
+func (want Bool) Compare(have bool) bool {
```
josephglanville: Perhaps `s/want/b`, not a material change but I had a double take when scanning function signatures because of the inconsistency.

titanous: Fixed.

titanous:
I have run go-fuzz against this package for a while and it found no crashes. (It did find two bugs in [json5](https://github.com/flynn/json5) though...)

titanous:
Updated the integers comparison to use varint encoding.

pkg/mauth/compare/compare.go
```diff
@@ -0,0 +1,334 @@
...
+
+func (b Bool) Compare(have bool) bool {
+	return bool(b) == have
+}
+
+func (b Bool) MarshalBinary() ([]byte, error) {
+	// type: 0x0 (false) OR 0x1 (true)
+
+	if bool(b) {
+		return []byte{byte(TypeTrue)}, nil
```
lmars: Would it make more sense for `Type` to be `byte` rather than `uint8` to avoid `byte(TypeXXX)` everywhere?

lmars:
LGTM pending comment.

pkg/mauth/compare/compare.go
titanous: `uint8` and `byte` are interchangeable, the type conversion is required to convert from `Type`.

lmars: Ah yes of course, I wasn't aware `uint8` and `byte` were interchangeable either thanks :smile:

PR #3153, opened 2016-07-19
appliance/mariadb: Wait for master_clients > 0

josephglanville:
The semi-sync plugin doesn't increment `master_clients` until the
connected slave has acked a message. Because of this `SLAVE_HOSTS`
can show a connected slave before `master_clients`.
Instead just use `SLAVE_HOSTS` for verifying the correct downstream
is present and use `master_clients` for determining read/write.

### DIFF
--- a/appliance/mariadb/process_test.go
+++ b/appliance/mariadb/process_test.go
@@ -166,17 +166,14 @@ func waitReadWrite(c *C, db *sql.DB) {
 	// Even if the database is read/write a slave must be connected
 	// for writes to be allowed.
 	err = queryAttempts.Run(func() error {
-		rows, err := db.Query("SHOW SLAVE HOSTS")
+		var discard interface{}
+		var masterClients int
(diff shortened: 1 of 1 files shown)


appliance/mariadb/process_test.go
```diff
@@ -166,18 +166,12 @@ func waitReadWrite(c *C, db *sql.DB) {
...
 	err = queryAttempts.Run(func() error {
-		rows, err := db.Query("SHOW SLAVE HOSTS")
-		if err != nil {
-			return err
-		}
-		for rows.Next() {
-			// We have a connected slave, thus the database should be writeable
+		var discard interface{}
+		var masterClients int
+		err = db.QueryRow("SHOW STATUS LIKE 'rpl_semi_sync_master_clients'").Scan(&discard, &masterClients)
```
titanous: `err` isn't checked.

titanous:
LGTM

PR #3159, opened 2016-07-20
cli: Fix service name for Docker processes

### LINKED ISSUE #3097: All apps deployed from Docker image via docker-receive showing "service unavailable"
On a fairly fresh cluster, except with the newer docker-receive per https://github.com/flynn/flynn/pull/3069#issuecomment-231137775, Flynn refuses to  route to apps, showing just a "Service Unavailable." 

I've tried even the barest bones Node.js Docker image (https://github.com/jlchereau/Docker-AWS), alongside 3 of our own apps -- all return the same error. Dashboard and other Flynn non-Docker released apps seem to load just fine; haven't tried a custom buildpack-built app on this cluster.

Apps are scaled and have the correct ports (I think). No abnormalities in the app logs. Restarting the (single in the case) Flynn host didn't fix anything. 

```Flynn host version v20160624.1.``` 

``flynn-host collect-debug-logs`` was actually to big for gist, so [here's an s3 link](https://s3.amazonaws.com/hbkapps-org/flynn-host-debug.tar.gz
) to the ``.tar.gz``

titanous:
The convention `APPNAME-web` is used for routes by default in the CLI, use that convention when naming the service registration configured for a process type in a release based on a Docker image.

Closes #3097

### DIFF
--- a/cli/docker.go
+++ b/cli/docker.go
@@ -236,7 +236,7 @@ func runDockerPush(args *docopt.Args, client controller.Client) error {
 			Port:  8080,
 			Proto: "tcp",
 			Service: &host.Service{
-				Name:   mustApp(),
+				Name:   mustApp() + "-web",
 				Create: true,
 			},
 		}}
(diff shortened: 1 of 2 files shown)


jvatic:
LGTM

josephglanville:
LGTM

PR #3160, opened 2016-07-20
controller: Fix event ordering assumption in TestStreamReleaseEvents

titanous:
We don't guarantee that events will be delivered in a specific order.


josephglanville:
LGTM

Issue #3161, opened 2016-07-20
CPU pegged at 100% on all cluster servers when builds are interrupted


kenaniah:
## Description

Tested on various clusters from `v20160623.0` to `v20160719.0`...

When performing a build via `git push flynn` and interrupting it build via `Ctrl+C`, the `SIGINT` is apparently not trapped correctly by the slugbuilder.

Attempting to push and interrupt the same build multiple times in succession will cause all servers in the cluster to hang with the CPU pegged at 100% usage, leaving the cluster completely inoperable and unrecoverable. 

## Steps to reproduce

`git push flynn`

Wait for slugbuilder to start...

Send `SIGINT` (via `Ctrl+C` for most terminals) once the build is in progress

Repeat a few times until the build system fails with:

```
remote: ERROR: Build failed: ratelimited: maximum concurrent AddJob calls running, try again later
```

rikur:
+1 we got hit by this too. @titanous mentioned that he wanted to replace the ruby script that parses the Procfile with a Go implementation -- that should fix it. At least in my case it was a runaway Ruby process that ate up all the CPU.

titanous:
What buildpack are you using? Is this specific to a certain app or buildpack?

rikur:
I was using a PHP buildpack and yet it was a defunct ruby process that brought the server down.

titanous:
@rikur Did you also cancel a build to trigger this?

kenaniah:
For the record, I was using multipack, which loaded both
```
https://github.com/heroku/heroku-buildpack-nodejs
https://github.com/heroku/heroku-buildpack-ruby
```

titanous:
So far I've been unable to reproduce this. Is there a specific phase of the deploy where cancelling it triggers the problem? Were you able to collect logs from the cluster after this happened?

titanous:
Also, do you know what process was pegging the CPU?

kenaniah:
I usually cancel immediately after the `-----> Building app_name...` part.

The servers are completely resource-constrained, so I can't collect logs, SSH, use the Flynn API, or do anything with them. 

It also might be interesting to note that in my case the git repo is roughly 385MB in size.

titanous:
Ahhh. That makes more sense. Are you on AWS? Have you moved blobs to S3?

titanous:
Also, did the issue resolve itself after waiting some period of time?

kenaniah:
[...]

PR #3164, opened 2016-07-21
gitreceive: Don't persist git repo if push failed

titanous:
When a push fails, there is no reason to upload the repo to the blobstore, as it should be unchanged.

Refs #3161


josephglanville:
LGTM

PR #3165, opened 2016-07-21
router: Don't allow malformed certificates/keys

titanous:
Before this patch the router would happily add broken and mismatched keys/certs to the database, which would result in sync breaking after a restart.


jvatic:
LGTM

titanous:
I've fixed the tests, please re-review.

jvatic:
LGTM

PR #3174, opened 2016-07-23
cli: Only use ENV from docker images as defaults

### LINKED ISSUE #3173: flynn app ENV vars get nuked by Dockerfile definitions during `flynn docker push`
Dockerfile:
```
...
ENV FOO "foo"
...
```

```
flynn -a dockerapp env set FOO=bar
flynn -a dockerapp docker push image
flynn -a dockerapp env
FOO=foo
```

I think the flynn ENV vars should take precedence here.

lmars:
Fixes #3173.


josephglanville:
LGTM

PR #3176, opened 2016-07-25
host: Volume garbage collection

### LINKED ISSUE #2976: flynn-host command for volume garbage collection
There should be a flynn-host command that can list and destroy ZFS volumes that are not currently in use.

josephglanville:
This implements a CLI command `flynn-host volume` that provides operators with administrative tools for listing, creating, deleting and automatically garbage collecting unused volumes.

Closes #2976

### DIFF
--- a/host/cli/volume.go
+++ b/host/cli/volume.go
@@ -1,0 +1,209 @@
+package cli
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"text/tabwriter"
+
+	"github.com/flynn/flynn/host/volume"
+	"github.com/flynn/flynn/pkg/cluster"
+	"github.com/flynn/go-docopt"
+)
+
+func init() {
+	Register("volume", runVolume, `
+usage: flynn-host volume list
+       flynn-host volume create [--provider=<provider>] <host>
+       flynn-host volume delete ID...
+       flynn-host volume gc
+
+Commands:
+    list    Display a list of all volumes of known Flynn hosts
+    create  Creates a volume on a host
+    delete  Deletes volumes, destroying any data stored on them
+    gc      Garbage collect currently unused volumes
+
+Examples:
+
+    $ flynn-host volume list
+
+    $ flynn-host volume create --provider default host0
+
+    $ flynn-host volume destroy 102fad07-07a3-4841-bded-d9e8a3eedbd6
+
+    $ flynn-host volume gc
+`)
+}
+
+func runVolume(args *docopt.Args, client *cluster.Client) error {
+	switch {
+	case args.Bool["list"]:
+		return runVolumeList(args, client)
+	case args.Bool["delete"]:
+		return runVolumeDelete(args, client)
+	case args.Bool["create"]:
+		return runVolumeCreate(args, client)
+	case args.Bool["gc"]:
+		return runVolumeGarbageCollection(args, client)
+	}
+	return nil
+}
+
+func runVolumeGarbageCollection(args *docopt.Args, client *cluster.Client) error {
+	// collect list of all volume ids currently attached to jobs
+	hosts, err := client.Hosts()
+	if err != nil {
+		return fmt.Errorf("could not list hosts: %s", err)
+	}
+	if len(hosts) == 0 {
+		return errors.New("no hosts found")
+	}
+
+	attached := make(map[string]struct{})
+	for _, h := range hosts {
+		jobs, err := h.ListJobs()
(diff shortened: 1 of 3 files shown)


josephglanville:
There is a potential for a race here between the scheduler creating a volume and placing a job if this command is run in that very narrow window and it manages to delete the volume before the job is placed.

This could cause a placement failure in the scheduler. Unfortunately I don't think we can currently close this race as the host API isn't capable of transactionally creating both the volume and the job at the same time currently (perhaps that an enhancement we should consider as the host daemon interface is improved).

However I think this solution is short term, once I get back to my persistent volumes work I think we can process volume garbage collection in the scheduler instead avoiding the problem and having this command only exist as a last resort option for operators.

lmars:
> There is a potential for a race here between the scheduler creating a volume and placing a job if this command is run in that very narrow window and it manages to delete the volume before the job is placed.

In that case, the job will fail and the scheduler will try again, so I don't think it is something we need to worry about.

host/cli/volume.go
```diff
@@ -0,0 +1,210 @@
...
+
+	"github.com/flynn/flynn/host/volume"
+	"github.com/flynn/flynn/pkg/cluster"
+	"github.com/flynn/go-docopt"
+)
+
+func init() {
+	Register("volume", runVolume, `
+usage: flynn-host volume list
+       flynn-host volume create [--provider] HOST
```
lmars: The provider flag should have an argument: `[--provider=PROVIDER]`

lmars: And it should have a default of `zfs` (I think you'll need an Options section for that).

host/cli/volume.go
```diff
@@ -0,0 +1,210 @@
...
+	// collect list of all volume ids currently attached to jobs
+	hosts, err := client.Hosts()
+	if err != nil {
+		return fmt.Errorf("could not list hosts: %s", err)
+	}
+	if len(hosts) == 0 {
+		return errors.New("no hosts found")
+	}
+
+	attached := make([]string, 0, 0)
```
lmars: Should this just be `var attached []string`?

host/cli/volume.go
```diff
@@ -0,0 +1,210 @@
...
+	for _, h := range hosts {
+		jobs, err := h.ListJobs()
+		if err != nil {
+			return err
```
lmars: I think we should just print this error and move on so one host in a bad state doesn't prevent all volume GC.

host/cli/volume.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return err
+	}
+
+	// iterate over list of all volumes, deleting any not found in the attached list
+	success := true
+outer:
+	for _, v := range volumes {
+		for _, a := range attached {
+			if a == v.Volume.ID {
```
lmars: This would be more efficient if `attached` were a `map[string]struct{}`

host/cli/volume.go
```diff
@@ -0,0 +1,210 @@
...
+					fmt.Printf("could not delete volume %s: %s\n", id, err)
+					continue outer
+				}
+				// delete the volume
+				fmt.Println(id, "deleted")
+				continue outer
+			}
+		}
+		success = false
+		fmt.Printf("could not delete volume %s: volume not found\n", id)
```
lmars: We should just try and destroy the volume and let the host determine if the volume doesn't exist.

host/cli/volume.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	if !success {
+		return errors.New("could not delete all volumes")
+	}
+	return nil
+}
+
+func runVolumeCreate(args *docopt.Args, client *cluster.Client) error {
+	hostId := args.String["HOST"]
```
lmars: We usually use angle bracket style arguments (i.e. `args.String["<host>"]`)

host/cli/volume.go
lmars: Oh in fact the default should be `default`

host/cli/volume.go
josephglanville: Unfortunately because Volume IDs don't encode the host we can't lookup the host directly, so if we can't find the volume in the list of volumes we were able to enumerate we can't attempt to delete it, as we don't know where it was/is located.

lmars: Ah ok I didn't realise that.

titanous:
LGTM

PR #3186, opened 2016-07-26
router: Use prepared queries

### LINKED ISSUE #3179: Router fails to load path-based routes
Restarting the router after adding a path-based route can sometimes result in this error and the route not getting loaded:

```text
Failed insert of path based route, consistency violation.
```

The problem appears to be that it is assumed that the path-based route will always be loaded after the parent bare domain route. This assumption does not hold, resulting in a failure to load the route.

As a very temporary workaround, the path-based route can be removed and re-added after restarting the router.

josephglanville:
Updates the router datastore to use prepared queries.

Closes #3179

### DIFF
--- a/router/setup_test.go
+++ b/router/setup_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/flynn/flynn/discoverd/testutil"
 	"github.com/flynn/flynn/pkg/postgres"
 	"github.com/flynn/flynn/pkg/testutils/postgres"
+	"github.com/flynn/flynn/router/schema"
 	"github.com/flynn/flynn/router/types"
 	. "github.com/flynn/go-check"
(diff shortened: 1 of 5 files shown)


titanous:
LGTM

router/setup_test.go
```diff
@@ -88,6 +90,13 @@ func (s *S) SetUpSuite(c *C) {
 	if err = migrateDB(db); err != nil {
 		c.Fatal(err)
 	}
+	db.Close()
+
+	// reconnect with prepared statements
+	pgxConfig.AfterConnect = schema.PrepareStatements
+	pgxpool, err = pgx.NewConnPool(pgxConfig)
```
titanous: err is unchecked

PR #3206, opened 2016-08-01
pkg/sirenia: Set timeout on HTTP client

josephglanville:
Sets the HTTP client timeout to avoid hanging the Sirenia client indefinitely.

### DIFF
--- a/pkg/sirenia/client/client.go
+++ b/pkg/sirenia/client/client.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/flynn/flynn/discoverd/client"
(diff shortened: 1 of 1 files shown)


lmars:
LGTM

lmars:
@josephglanville in fact, I wonder if it is worth also using the [Retry dialer](https://github.com/flynn/flynn/blob/master/pkg/dialer/retry.go#L15-L20) so there is a low timeout on actually trying a connection, but then we try multiple times?

josephglanville:
@lmars indeed, that is a good idea. I will add the Retry dialer.

titanous:
LGTM

PR #3207, opened 2016-08-01
installer: Fix race condition accessing cluster state

### LINKED ISSUE #2664: installer: Race condition accessing base cluster state
Running the installer with `-race` reported a race condition accessing `BaseCluster.State`. Some care needs to be taken when fixing this not to introduce a deadlock.

jvatic:
Closes #2664


titanous:
LGTM

PR #3221, opened 2016-08-03
discoverd: Allow setting service leader during SetServiceMeta

### LINKED ISSUE #3219: Sirenia doesn't retry discoverd SetLeader
v20160721.2

I'll email you the logs as there were too big.

`flynn ps` gives the error

> unknown_error: Something went wrong

Dashboard doesn't show any apps

Prior to the issues I had pushed an update and I got the application deployed message but then there was a mixture of old and new releases running at the same time, and more that there processes than there should have been i.e 4 not 3.

titanous:
This allows us to transactionally set the leader and the service meta at the same time, which is in turn used by Sirenia to prevent the metadata from getting out of sync with the leader entry.

Closes #3219

### DIFF
--- a/discoverd/client/service.go
+++ b/discoverd/client/service.go
@@ -175,6 +175,11 @@ func (s *service) Leaders(leaders chan *Instance) (stream.Stream, error) {
 type ServiceMeta struct {
 	Data json.RawMessage `json:"data"`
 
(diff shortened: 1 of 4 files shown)


josephglanville:
LGTM

PR #3223, opened 2016-08-03
test: Fix flaky HostSuite.TestAddFailingJob

titanous:
We don't guarantee the order that events are delivered in.

### DIFF
--- a/test/test_host.go
+++ b/test/test_host.go
@@ -63,7 +63,7 @@ func (s *HostSuite) TestAddFailingJob(t *c.C) {
 	t.Assert(h.AddJob(job), c.IsNil)
 
 	// check we get a create then error event
-	actual := make([]*host.Event, 0, 2)
+	actual := make(map[string]*host.Event, 2)
 loop:
 	for {
 		select {
@@ -71,7 +71,10 @@ loop:
 			if !ok {
 				t.Fatalf("job event stream closed unexpectedly: %s", stream.Err())
 			}
-			actual = append(actual, e)
+			if _, ok := actual[e.Event]; ok {
+				t.Fatalf("unexpected event: %v", e)
+			}
+			actual[e.Event] = e
 			if len(actual) >= 2 {
 				break loop
 			}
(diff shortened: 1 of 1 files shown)


test/test_host.go
```diff
@@ -63,26 +63,27 @@ func (s *HostSuite) TestAddFailingJob(t *c.C) {
...
 			}
 		case <-time.After(30 * time.Second):
 			t.Fatal("timed out waiting for job event")
 		}
 	}
-	t.Assert(actual, c.HasLen, 2)
-	t.Assert(actual[0].Event, c.Equals, host.JobEventCreate)
-	t.Assert(actual[1].Event, c.Equals, host.JobEventError)
-	jobErr := actual[1].Job.Error
+	t.Assert(actual[host.JobEventCreate], c.Not(c.IsNil))
```
lmars: Should this check that `actual[host.JobEventError]` is not nil, given that is what we use for further assertions below?

titanous: Done.

test/test_host.go
```diff
@@ -63,28 +63,31 @@ func (s *HostSuite) TestAddFailingJob(t *c.C) {
...
-	t.Assert(actual, c.HasLen, 2)
-	t.Assert(actual[0].Event, c.Equals, host.JobEventCreate)
-	t.Assert(actual[1].Event, c.Equals, host.JobEventError)
-	jobErr := actual[1].Job.Error
-	t.Assert(jobErr, c.NotNil)
-	t.Assert(*jobErr, c.Equals, `host: invalid job partition "nonexistent"`)
+	t.Assert(actual[host.JobEventCreate], c.NotNil)
+	e := actual[host.JobEventError]
+	c.Assert(e, c.NotNil)
+	c.Assert(e.Job, c.NotNil)
```
lmars: These need to be `t.Assert`

lmars:
LGTM pending comment

test/test_host.go
titanous: Fixed, thanks.

PR #3230, opened 2016-08-04
Fix DNS packet overflow

### LINKED ISSUE #3210: Discoverd returns invalid DNS responses
In some cases discoverd is returning DNS responses >512 bytes over UDP even when EDNS has not been requested.

It seems like the issue may be caused by the behavior of miekg/dns described at https://github.com/miekg/dns/issues/216.

titanous:
This ensures that responses to queries are always compressed. Sometimes responses from upstream servers will fit in the UDP limit of 512 bytes only when compressed, and the flag is only applied hop-by-hop.

I also bumped the DNS package and found that the DNS tests were not running, so I re-enabled them and fixed some test code that was broken.

Closes #3210


lmars:
LGTM

PR #3233, opened 2016-08-06
gitreceive,taffy,redis: Use artifact IDs rather than URIs

lmars:
Extracted from #3100.

### DIFF
--- a/updater/updater.go
+++ b/updater/updater.go
@@ -12,6 +12,7 @@ import (
 	"github.com/flynn/flynn/controller/client"
 	ct "github.com/flynn/flynn/controller/types"
 	"github.com/flynn/flynn/discoverd/client"
+	"github.com/flynn/flynn/host/types"
 	"github.com/flynn/flynn/pkg/status"
 	"github.com/flynn/flynn/pkg/version"
 	"github.com/flynn/flynn/updater/types"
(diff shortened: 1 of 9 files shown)


updater/updater.go
```diff
@@ -234,15 +251,19 @@ func deployApp(client controller.Client, app *ct.App, uri string, updateFn updat
...
+		"SLUGBUILDER_IMAGE_URI": slugBuilder.URI,
+		"SLUGRUNNER_IMAGE_URI":  slugRunner.URI,
 	}
 	updated := false
-	for key, uri := range uris {
-		if v, ok := env[key]; ok && v != uri {
-			env[key] = uri
+	for key, val := range updates {
+		if v, ok := env[key]; ok && v != val {
+			env[key] = val
```
titanous: Is this code correct? Don't we want to set the new env vars? Currently it looks like they won't be set if because of the `ok` check.

lmars: I think this is correct yes, we only want to set the env var if it already exists but has a different value.

lmars: Also note there is test coverage of this in TestReleaseImages [here](https://github.com/flynn/flynn/blob/artifact-ids/test/test_release.go#L242-L256).

lmars:
@titanous comments addressed

titanous:
LGTM

PR #3237, opened 2016-08-07
controller: Fix race in TestRunJobAttached

titanous:
This test was racing with the one-sided close detection in the controller handler.


lmars:
LGTM

PR #3248, opened 2016-08-09
script: Require --clean when re-running the install script

### LINKED ISSUE #3211: install-script: Just print message if Flynn is already installed
If Flynn is already installed and running on a host, the install-script should exit with a message advising to either `flynn-host update` or use `--clean` rather than proceeding to download the latest version (which leaves the host in an odd state).

lmars:
Fixes #3211.


titanous:
LGTM

PR #3275, opened 2016-08-11
pkg/sirenia/client: Remove omitempty tags on DatabaseInfo

titanous:
These fields should always be emitted in the JSON.


lmars:
LGTM

PR #3281, opened 2016-08-12
host: Fix checking minimum update version

lmars:
Updating from versions prior to v20160711.0 is currently broken as they try and download the nsumount binary which no longer exists.

We actually decided not to support in-place updates from libvirt-lxc to libcontainer so we just need to tell users to do a backup/restore instead, so I have bumped the minVersion and also fixed it so that it's compared against the version of hosts running in the cluster (rather than checking its own version, which will never fail).

### DIFF
--- a/host/cli/update.go
+++ b/host/cli/update.go
@@ -41,17 +41,21 @@ Options:
 Update Flynn components`)
 }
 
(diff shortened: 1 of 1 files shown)


titanous:
LGTM

PR #3294, opened 2016-08-15
Bump asset matrix

jvatic:
See https://github.com/jvatic/asset-matrix-go/pull/9 for details.


jvatic:
It seems this change causes compilation within the container to be slower, although doing so outside the container is around 6x as fast (~1m vs ~6m, used to be ~2m if I remember correctly). I'll investigate further.

titanous:
@jvatic What is the status of this PR?

jvatic:
This needs a bit more work, closing for now (will reopen another PR when ready)

PR #3299, opened 2016-08-16
discoverd/client: Ensure that AddService is always retried

titanous:
Move the logic from `maybeAddService` to `AddService` so that direct callers of `AddService` get retries as well. Fixes a intermittent failure while bootstrapping where the flannel wrapper errors out while discoverd is starting.

Thanks to @tianx2 for reporting this.


lmars:
LGTM

PR #3308, opened 2016-08-17
Fix bootstrapping from older backups

### LINKED ISSUE #2988: Status Page on the Dashboard not working due to an HTTP 401
Hi,
I just upgraded to the new stable version v20160624.1 and the status page of the dashboard refuses to work due to an HTTP 401 to the status endpoint. Have a look:
screen shot 2016-06-25 at 12 51 53

### LINKED ISSUE #3198: Restoring from backup pre-mongo doesn't create mongodb provider
Restoring from a backup pre-dating the mongodb appliance causes it not to be there. This in turn breaks database related views in the dashboard.

Thank you @swaits for reporting this.

jvatic:
- Ensures that MariaDB, MongoDB, and docker-receive are deployed,
and that STATUS_KEY is set in the dashboard
- Add mongodb to status app

<del>This isn't ready to merge just yet as MariaDB isn't healthy when it scales up. @josephglanville any input you can provide here would be appreciated.</del>
<br/>

Closes #3198
Closes #2988

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -184,17 +184,11 @@ func (m Manifest) RunWithState(ch chan<- *StepInfo, state *State) (*State, error
 	return state, nil
 }
 
-func Run(manifestData []byte, ch chan<- *StepInfo, cfg Config, only []string) error {
-	defer close(ch)
+func UnmarshalManifest(manifestData []byte, only []string) (Manifest, error) {
 	var manifest Manifest
 	var steps []json.RawMessage
 	if err := json.Unmarshal(manifestData, &steps); err != nil {
-		return err
-	}
-
-	cfg.Singleton = cfg.MinHosts == 1
-	if s := os.Getenv("SINGLETON"); s != "" {
-		cfg.Singleton = s == "true"
+		return nil, err
 	}
 
 	skipStep := func(step Step) bool {
@@ -212,24 +206,38 @@ func Run(manifestData []byte, ch chan<- *StepInfo, cfg Config, only []string) er
 	for _, s := range steps {
 		var step Step
 		if err := json.Unmarshal(s, &step.StepMeta); err != nil {
-			return err
+			return nil, err
 		}
 		if skipStep(step) {
 			continue
 		}
 		actionType, ok := registeredActions[step.StepMeta.Action]
 		if !ok {
-			return fmt.Errorf("bootstrap: unknown action %q", step.StepMeta.Action)
+			return manifest, fmt.Errorf("bootstrap: unknown action %q", step.StepMeta.Action)
 		}
 		step.Action = reflect.New(actionType).Interface().(Action)
 
 		if err := json.Unmarshal(s, &step.Action); err != nil {
-			return err
+			return nil, err
 		}
 		manifest = append(manifest, step)
 	}
+	return manifest, nil
+}
 
-	_, err := manifest.Run(ch, cfg)
+func Run(manifestData []byte, ch chan<- *StepInfo, cfg Config, only []string) error {
+	defer close(ch)
+	manifest, err := UnmarshalManifest(manifestData, only)
+	if err != nil {
+		return err
+	}
+
+	cfg.Singleton = cfg.MinHosts == 1
+	if s := os.Getenv("SINGLETON"); s != "" {
+		cfg.Singleton = s == "true"
+	}
+
+	_, err = manifest.Run(ch, cfg)
 	return err
 }
 
--- a/host/cli/bootstrap.go
+++ b/host/cli/bootstrap.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -22,6 +23,7 @@ import (
 	"github.com/flynn/flynn/host/types"
 	"github.com/flynn/flynn/pkg/exec"
 	"github.com/flynn/flynn/pkg/random"
+	"github.com/flynn/flynn/pkg/tlscert"
 	"github.com/flynn/go-docopt"
 )
 
@@ -256,6 +258,14 @@ $function$;
 			}
 		}
 	}
+	manifestStepMap := make(map[string]bootstrap.Step, len(manifestSteps))
+	steps, err := bootstrap.UnmarshalManifest(manifest, nil)
+	if err != nil {
+		return fmt.Errorf("error decoding manifest json: %s", err)
+	}
+	for _, step := range steps {
+		manifestStepMap[step.StepMeta.ID] = step
+	}
 
 	artifactURIs := make(map[string]string)
 	updateProcArgs := func(f *ct.ExpandedFormation, step *manifestStep) {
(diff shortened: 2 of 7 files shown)


bootstrap/bootstrap.go
```diff
@@ -184,17 +184,11 @@ func (m Manifest) RunWithState(ch chan<- *StepInfo, state *State) (*State, error
...
 	var manifest Manifest
 	var steps []json.RawMessage
 	if err := json.Unmarshal(manifestData, &steps); err != nil {
-		return err
-	}
-
-	cfg.Singleton = cfg.MinHosts == 1
-	if s := os.Getenv("SINGLETON"); s != "" {
-		cfg.Singleton = s == "true"
+		return manifest, err
```
lmars: Should these error returns not be `return nil, err`?

host/cli/bootstrap.go
```diff
@@ -223,6 +225,17 @@ $function$;
...
+	manifestStepMap := make(map[string]bootstrap.Step, len(manifestSteps))
+	{
+		steps, err := bootstrap.UnmarshalManifest(manifest, nil)
+		if err != nil {
+			return fmt.Errorf("error decoding manifest json: %s", err)
+		}
+		for _, step := range steps {
+			manifestStepMap[step.StepMeta.ID] = step
+		}
+	}
```
lmars: Why the braces?

host/cli/bootstrap.go
```diff
@@ -549,6 +603,36 @@ WHERE release_id = (SELECT release_id FROM apps WHERE name = 'discoverd')
...
+					PrivateKey: r.Certificate.Key,
+				}
+				break
+			}
+		}
+		dockerReceiveRouteStep := manifestStepMap["docker-receive-route"]
+		steps := bootstrap.Manifest{
+			manifestStepMap["docker-receive-secret"],
+			manifestStepMap["docker-receive"],
+			dockerReceiveRouteStep,
```
lmars: Does this need to be in a variable?

host/cli/bootstrap.go
```diff
@@ -504,11 +530,37 @@ WHERE release_id = (SELECT release_id FROM apps WHERE name = 'discoverd')
...
-	// get blobstore config
-	client, err := controller.NewClient("http://"+controllerInstances[0].Addr, data.Controller.Release.Env["AUTH_KEY"])
+	controllerKey := data.Controller.Release.Env["AUTH_KEY"]
+	state.SetControllerKey(controllerKey)
+	state.StepData["controller-key"] = &bootstrap.RandomData{controllerKey}
+	if shouldDeployMariaDB {
+		steps := bootstrap.Manifest{
+			manifestStepMap["mariadb-password"],
+			manifestStepMap["mariadb"],
+			manifestStepMap["add-mysql-provider"],
```
josephglanville: Might want to add the `mariadb-wait` step here.

host/cli/bootstrap.go
```diff
@@ -504,11 +530,37 @@ WHERE release_id = (SELECT release_id FROM apps WHERE name = 'discoverd')
...
+		if _, err := steps.RunWithState(ch, state); err != nil {
+			return fmt.Errorf("error deploying mariadb: %s", err)
+		}
+	}
+
+	if shouldDeployMongoDB {
+		steps := bootstrap.Manifest{
+			manifestStepMap["mongodb-password"],
+			manifestStepMap["mongodb"],
+			manifestStepMap["add-mongodb-provider"],
```
josephglanville: As above, probably want the `mongodb-wait` step.

host/cli/bootstrap.go
jvatic: Done.

host/cli/bootstrap.go
jvatic: Done.

jvatic:
@lmars @josephglanville comments addressed and both databases are now working as expected (I think my vagrant box being out of date was causing the mariadb issue).

One remaining issue is the mongodb status never shows up as healthy (both with clean bootstrap and from backup). I may just move that into another PR if it's not any easy fix.

jvatic:
The mongodb unhealthy status isn't due incorrect configuration in the status app as I thought it might have been.

titanous:
@jvatic Does mongodb become healthy after running `flynn resource add mongodb`?

jvatic:
@titanous Nope.

titanous:
Okay, please post the logs.

jvatic:
https://gist.github.com/e3e735c07dfdfee9327a445348457d9a

jvatic:
I've reproduced the mongodb status issue on the latest stable release: #3316

jvatic:
I can confirm #3317 fixes the mongodb status issue (#3316).

titanous:
Since we now have a way of testing backup restores, does it make sense to add some tests? /cc @lmars

jvatic:
Yeah, seems like a good idea

jvatic:
Haven't tested the test yet (0fcc251)

script/generate-backups
```diff
@@ -65,6 +65,16 @@ main() {
...
   done
+
+  # generate backup without MySQL and MongoDB, and with STATUS_KEY missing from the dashboard
+  local version="v20160814.0"
+  bootstrap_cluster "${version}"
+  git_deploy_app
+  add_redis_resource
+  remove_mysql
+  remove_mongodb
+  remove_dashboard_status_key
```
lmars: Rather than modifying the latest release, we should use a version where these cases were true.

Version `v20160309.0` (included above) satisfies all of these cases (MariaDB was added in `v20160311.1`, MongoDB in `v20160706.0` and dashboard `STATUS_KEY` in `v20160620.1`).

test/test_zz_backup.go
```diff
@@ -159,4 +160,18 @@ func (s *ZZBackupSuite) testClusterBackup(t *c.C, index int, path string) {
 	})
 	t.Assert(err, c.IsNil)
 	t.Assert(redisResult, SuccessfulOutputContains, "PONG")
+
+	if strings.HasSuffix(filepath.Base(path), "missing-dbs-and-dashboard-status.tar") {
```
lmars: We should do these checks for all the backups.

vendor/github.com/graphql-go/graphql
```diff
@@ -0,0 +1 @@
+Subproject commit 1d9282bf75ff1bb4518590536ef930d22fab83e8
```
lmars: Drop this change

lmars:
FYI I have added d9ef584 which fixes the fact that changes to in-memory releases is not reflected in the database and could lead to unexpected issues in the future.

titanous:
There appears to be a bug in this:

```text
14:10:28.928004 deploy-app mariadb error: object_exists: application "mariadb" already exists
14:10:28.928062 host.go:154: error deploying mariadb: object_exists: application "mariadb" already exists
```

When trying to restore a backup that was taken from a cluster with the mariadb app but not scaled up (the default configuration) so no database dump was not stored in the backup.

lmars:
I've pushed some more changes to fix the mariadb app existence and to also add tests for mysql / mongodb backups.

Unfortunately mysql restores seem to be broken:

```
++ 15:57:25.175 checking mysql resource
++ 15:57:25.175 /home/ubuntu/go/src/github.com/flynn/flynn/cli/bin/flynn mysql console -- -e SELECT * FROM foos
ERROR 1045 (28000): Access denied for user '0dba71ca1b751a3a96f2da7c4f492c16'@'100.101.100.22' (using password: YES)
```

This seems to be broken on master, you can verify by running:

```
$ script/bootstrap-flynn --from-backup https://s3.amazonaws.com/flynn-test/backups/cluster/v20160814.0-nodejs-mysql.tar
```

and then trying to get a console:

```
$ flynn -a nodejs mysql console
ERROR 1045 (28000): Access denied for user '0dba71ca1b751a3a96f2da7c4f492c16'@'100.100.88.22' (using password: YES)
```

For now I am not putting the mysql backup on the CI machine (so the tests will pass), but we need to fix this.

PR #3317, opened 2016-08-20
appliance/mongodb: Fix UserExists check

### LINKED ISSUE #3316: mongodb: Reports unhealthy
```
$ flynn resource add mongodb
Created resource b563fbb0-09e4-4f5d-b09d-c8f69d5b7dcc and release a4e457d4-9905-4244-b851-975138d89d0e.
```

```
$ flynn run curl -i mongodb.discoverd:27018/.well-known/status
HTTP/1.1 500 Internal Server Error
Content-Type: application/json
Date: Fri, 19 Aug 2016 21:34:56 GMT
Content-Length: 76

{
  "data": {
    "status": "unhealthy",
    "version": "v20160814.0"
  }
}
```

https://gist.github.com/9becbb975eb555be443dd54650916976

josephglanville:
Closes #3316

### DIFF
--- a/appliance/mongodb/process.go
+++ b/appliance/mongodb/process.go
@@ -764,13 +764,28 @@ func (p *Process) userExists() (bool, error) {
 	}
 	defer session.Close()
 
-	var entry struct {
-		Retval bson.M `bson:"retval"`
+	type user struct {
+		ID       string `bson:"_id"`
+		User     string `bson:"user"`
+		Database string `bson:"db"`
 	}
-	if err := session.Run(bson.M{"usersInfo": bson.M{"user": "flynn", "db": "admin"}}, &entry); err != nil {
+
+	var userInfo struct {
+		Users []user `bson:"users"`
+		Ok    int    `bson:"ok"`
+	}
+
+	if err := session.DB("admin").Run(bson.D{{"usersInfo", bson.M{"user": "flynn", "db": "admin"}}}, &userInfo); err != nil {
 		return false, err
 	}
-	return entry.Retval != nil, nil
+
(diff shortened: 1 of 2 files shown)


appliance/mongodb/process.go
```diff
@@ -770,7 +770,7 @@ func (p *Process) userExists() (bool, error) {
 	if err := session.Run(bson.M{"usersInfo": bson.M{"user": "flynn", "db": "admin"}}, &entry); err != nil {
 		return false, err
 	}
-	return entry.Retval != nil, nil
+	return entry.Retval == nil, nil
```
lmars: So it returns `nil` when you ask for the user info? That seems odd!

josephglanville: Actually you might be right.. I am going to take a closer look at the query because something isn't quite right.

josephglanville:
@lmars I think the problem was some of this was leftover from before mgo upgrade, I fixed this up so it actually parses the response.

bootstrap/manifest_template.json
```diff
@@ -412,6 +412,7 @@
     "release": {
       "env": {
         "FLYNN_MONGO": "mongodb",
+        "MONGO_DATABASE": "admin",
```
lmars: Do we also need a migration to update existing mongo apps to have this env var?

josephglanville: Shouldn't be necessary, this is just a convenience when debugging and consistency with the other appliances. I don't think it's likely to be used by users as it is only useful for examining stuff like replica set configuration.

titanous:
LGTM

PR #3325, opened 2016-08-25
blobstore: Add Google Cloud Storage backend

titanous:
Also did various refactoring to not just assume S3 as the only alternative.


josephglanville:
LGTM

lmars:
I have updated the CI init script so the env var will be set on the next CI restart. We will need to rebuild the rootfs to pickup the `SetEnv` change.

LGTM

PR #3336, opened 2016-08-27
pkg/postgres: Switch to using EXCLUSIVE locking on schema_migrations

josephglanville:
Previously an ACCESS EXCLUSIVE lock was held while performing database
migrations. This lock however cannot be attained while an ACCESS SHARE
lock is being held on the table, which is the case when `pg_dump` is
currently in the progress of dumping the database.

Advisory locks were also considered as they would be faster but they
are somewhat more complex and not the best fit as they don't honor
transaction semantics.


titanous:
LGTM

PR #3347, opened 2016-08-31
worker: Fix final scale down in one-by-one deploy

lmars:
The `oldScale` map will most likely contain types with a zero count (i.e. ones that have just been scaled down) and there is no need to add them to the diff.

### DIFF
--- a/controller/worker/deployment/one_by_one.go
+++ b/controller/worker/deployment/one_by_one.go
@@ -111,7 +111,9 @@ func (d *DeployJob) deployOneByOneWithWaitFn(waitJobs WaitJobsFn) error {
 	log.Info("ensuring old formation is scaled down to zero")
 	diff := make(ct.JobEvents, len(oldScale))
 	for typ, count := range oldScale {
-		diff[typ] = ct.JobDownEvents(count)
+		if count > 0 {
+			diff[typ] = ct.JobDownEvents(count)
+		}
 	}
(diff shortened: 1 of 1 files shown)


titanous:
Does this change the deploy behavior at all?

lmars:
Well, I saw a failed controller deploy which had this at the end:

```
t=2016-08-31T02:22:29+0000 lvl=info msg="ensuring old formation is scaled down to zero" app=worker fn=deployOneByOne
t=2016-08-31T02:22:29+0000 lvl=info msg="waiting for 1 job down event(s)" app=worker fn=deployOneByOne
t=2016-08-31T02:22:29+0000 lvl=info msg="got job event" app=worker fn=deployOneByOne job.id=4663b5b6-369d414a-c7a8-4bc5-a8b6-ef93bda86181 job.type=web job.state=down
t=2016-08-31T02:24:30+0000 lvl=eror msg="error waiting for job down events" app=worker fn=deployOneByOne err="timed out waiting for job events: map[web:map[down:1] worker:map[down:0]]"
```

so it got the one down event it was waiting for, but it was also waiting for zero worker down events (I think the event waiting does exact map equality, so no entry compared to a zero entry is not treated as being equal).

I have a bigger refactor in mind to make this process more robust after merging #2744, but this fixes the observed issue.

titanous:
LGTM

PR #3371, opened 2016-09-09
host: Allow non-TTY jobs to use open(2) on /dev/stdout & /dev/stderr

lmars:
Some containerised apps expect to be able to use `/dev/stdout` and `/dev/stderr` to write data to stdout / stderr (e.g. [the official Docker nginx app](https://github.com/nginxinc/docker-nginx/blob/1.11.1/mainline/jessie/Dockerfile#L21-L23)), but that doesn't work if those paths point to sockets (in which case `open(2)` fails with `ENXIO`, see http://marc.info/?l=ast-users&m=120978595414993).

My solution here is to provide jobs with a standard pipe, continue to provide flynn-host with a socket and copy data between the two. This isn't ideal as the data is now being copied through containerinit's memory. If we think this is problematic we could provide a flag to enable this just for jobs which require the use of `/dev/stdout` and `/dev/stderr`.

### DIFF
--- a/host/containerinit/init.go
+++ b/host/containerinit/init.go
@@ -609,6 +609,12 @@ func containerInitApp(c *Config, logFile *os.File) error {
 			return err
 		}
 
+		log.Info("creating FD proxies")
+		if err := createFDProxies(cmd); err != nil {
+			log.Error("error creating FD proxies", "err", err)
+			return err
+		}
+
 		if c.OpenStdin {
 			// Can't use cmd.StdinPipe() here, since in Go 1.2 it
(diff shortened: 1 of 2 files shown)


titanous:
Dumb idea: can we just bind the proxy pipe to `/dev/std{out,err}` directly so that programs using the FDs that we pass in don't go through the proxy, only wacky apps like nginx do?

lmars:
@titanous that would be better yes, I'll have a play around with it.

lmars:
@titanous comments addressed

titanous:
Very nice! LGTM

PR #3372, opened 2016-09-11
test: Test restoring historic cluster backups

lmars:
This adds `script/generate-backups` which bootstraps a Flynn cluster at various versions, deploys the Node.js example app, adds a Redis resource and backs up the cluster to a file.

There is also a test which iterates through those backups, bootstraps a cluster then checks the app came up with Redis correctly.

We probably want to expand this a little going further (e.g. use docker push, provision different resources etc.) but this at least gives us a good starting point to start avoiding cluster restore regressions.

### DIFF
--- a/host/cli/bootstrap.go
+++ b/host/cli/bootstrap.go
@@ -223,6 +223,40 @@ $function$;
 		return fmt.Errorf("error decoding manifest json: %s", err)
 	}
 
+	// add new env vars to system apps
+	for _, step := range manifestSteps {
+		var release *ct.Release
+		switch step.ID {
+		case "discoverd":
+			release = data.Discoverd.Release
+		case "flannel":
+			release = data.Flannel.Release
+		case "postgres":
+			release = data.Postgres.Release
+		case "mariadb":
+			if data.MariaDB != nil {
+				release = data.MariaDB.Release
+			}
+		case "mongodb":
+			if data.MongoDB != nil {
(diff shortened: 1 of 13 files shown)


titanous:
Amazing! LGTM

josephglanville:
Woah! LGTM

lmars:
@titanous @josephglanville I made some changes to make this more reliable:

* Mount the backups dir as a 9P filesystem inside the VMs at `/mnt/backups`
* Use different flannel networks for each backup test to avoid conflicts
* Add new env vars to system apps during restore (e.g. the new `NETWORK` var for flannel)
* Add some retries the the HTTP / Redis checks as user services aren't guaranteed to be up when the restore finishes

I have tested this 20 times in custom CI without error.

host/cli/bootstrap.go
```diff
@@ -223,6 +223,40 @@ $function$;
...
+		case "controller":
+			release = data.Controller.Release
+		}
+		if release == nil {
+			continue
+		}
+		for key, val := range step.Release.Env {
+			if _, ok := release.Env[key]; !ok {
+				if len(release.Env) == 0 {
+					release.Env = make(map[string]string, 1)
```
titanous: No need to supply a map size hint here.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #3377, opened 2016-09-13
test: Don't run BackupSuite concurrently

lmars:
It is proving disruptive to other tests.

### DIFF
--- a/host/volume/api/http.go
+++ b/host/volume/api/http.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"sync/atomic"
 
+	"github.com/flynn/flynn/discoverd/client"
 	"github.com/flynn/flynn/host/volume"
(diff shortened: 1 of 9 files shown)


titanous:
LGTM

lmars:
@titanous I had to add a few more fixes due to the fact I moved the default cluster to listen on just `eth0` (rather than `0.0.0.0` which prevented the backup clusters from booting).

host/volume/api/http.go
```diff
@@ -17,15 +19,21 @@ import (
 const snapshotContentType = "application/vnd.zfs.snapshot-stream"
 
 type HTTPAPI struct {
-	cluster *cluster.Client
-	vman    *volumemanager.Manager
+	vman *volumemanager.Manager
+
+	cluster    *cluster.Client
+	clusterMtx sync.Mutex
```
titanous: Can this be a `sync.Atomic` instead?

titanous:
LGTM

PR #3380, opened 2016-09-13
test: Use Let's Encrypt for CI interface

titanous:
This will require changes to the runner flags:

- Remove `-tls-key` and `-tls-cert`
- Add `-tls-dir` pointing at an empty directory to persist the certificate cache.
- Add `-domain ci.flynn.io`


lmars:
We should update the runner flags in the sample Upstart script: https://github.com/flynn/flynn/blob/master/test/scripts/upstart.conf#L40-L49

titanous:
> We should update the runner flags in the sample Upstart script

Done.

PR #3381, opened 2016-09-13
all: Add Flynn image manifests and imagebuilder 

lmars:
This is an intermediate pull request to address #2422 and replace #3100.

This adds the Flynn image manifest types to the controller and adds an imagebuilder which converts Docker images to Flynn images.

### DIFF
--- a/util/imagebuilder/main.go
+++ b/util/imagebuilder/main.go
@@ -1,0 +1,119 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"syscall"
+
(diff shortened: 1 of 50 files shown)


util/imagebuilder/main.go
```diff
@@ -0,0 +1,119 @@
...
+		return err
+	}
+	return json.NewEncoder(os.Stdout).Encode(manifest)
+}
+
+type layerStore struct {
+	root string
+}
+
+func (l *layerStore) Lock(id string, fn func() error) error {
```
titanous: Should this be called `DoLocked` or similar?

lmars: Updated in bfeefb8.

PR #3393, opened 2016-09-15
Add basic telemetry submission

titanous:


host/cli/bootstrap.go
```diff
@@ -296,6 +296,23 @@ $$;`, step.Artifact.URI, step.ID))
 		data.MongoDB.ImageArtifact.URI = artifactURIs["mongodb"]
 	}
 
+	// set TELEMETRY_CLUSTER_ID
+	telemetryClusterID := random.UUID()
+	sqlBuf.WriteString(fmt.Sprintf(`
+UPDATE releases SET env = jsonb_set(env, '{TELEMETRY_CLUSTER_ID}', '%q')
+WHERE release_id = (SELECT release_id FROM apps WHERE name = 'controller');
```
lmars: Is the intention to explicitly change this value when restoring from a backup, or should there be a `WHERE env->>'TELEMETRY_CLUSTER_ID' IS NULL` check?

titanous: Yes, the `TELEMETRY_CLUSTER_ID` is new for every restore, and the `TELEMETRY_BOOTSTRAP_ID` is generated once and stays the same across multiple restores.

lmars: Ah ok, I had those two the other way round in my head.

titanous: Yeah, they are definitely ambiguous, but this made the most sense to me.

PR #3395, opened 2016-09-15
worker: Don't wait for old jobs to actually stop

### LINKED ISSUE #3285: worker: Still SetAppRelease if scaling old release down fails
If scaling down the old release during a deployment fails, we treat it as a failed deploy but we skip doing a rollback as that would involve stopping the new, functioning release (see [here](https://github.com/flynn/flynn/blob/v20160812.0/controller/worker/deployment/all_at_once.go#L68-L88) and [here](https://github.com/flynn/flynn/blob/v20160812.0/controller/worker/deployment/one_by_one.go#L106-L129)).

This then means though that we don't [set the app release](https://github.com/flynn/flynn/blob/v20160812.0/controller/worker/deployment/context.go#L114-L123) so the old (now scaled down) release is still the app's release rather than the new, functioning release, meaning subsequent deploys think the app has no scale and scales it to `web=1`, effectively "losing" the new release but leaving it scaled up.

I think the solution here is to just log the error when scaling the old formation fails but not return any error so that the deploy is marked as a success.

@titanous wdyt?

lmars:
If a formation change is accepted by the controller, we can then leave it up to the scheduler to actually stop the jobs. Failing the deployment if the events aren't received (e.g. because they are already down we just didn't know it yet) won't make any difference.

Fixes #3285.

### DIFF
--- a/controller/worker/deployment/all_at_once.go
+++ b/controller/worker/deployment/all_at_once.go
@@ -65,28 +65,22 @@ func (d *DeployJob) deployAllAtOnce() error {
 		}
 	}
 
-	// the new jobs have now started and they are up, so return
-	// ErrSkipRollback from here on out if an error occurs (rolling
-	// back doesn't make a ton of sense because it involves
-	// stopping the new working jobs).
 	log = log.New("release_id", d.OldReleaseID)
 	log.Info("scaling old formation to zero")
 	if err := d.client.PutFormation(&ct.Formation{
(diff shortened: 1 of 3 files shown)


lmars:
@titanous a few tests failed as they expect the down events before the deploy complete event, just get rid of the down event waiting or move to after the deploy completes?

titanous:
I think moving it to after makes sense.

PR #3409, opened 2016-09-21
host/containerinit: Wait 2 seconds before forwarding SIGTERM/INT

josephglanville:
We also now close the heartbeaters synchronously.

The purpose of this change is to give the router additional grace period to remove backends that will be shutting down shortly from it's routing table.

This ensures that new client requests won't be sent on dead or soon to be dead connections during scale down events, like during a deploy for instance.

### DIFF
--- a/host/containerinit/init.go
+++ b/host/containerinit/init.go
@@ -513,7 +513,8 @@ func babySit(process *os.Process, hbs []discoverd.Heartbeater) int {
 				continue
 			}
 			if sig == syscall.SIGTERM || sig == syscall.SIGINT {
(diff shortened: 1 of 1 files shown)


titanous:
The commit message should also describe a bit more why we are doing this, as it's not obvious.

josephglanville:
Updated to indicate reasoning for change.

titanous:
The commit message is still wrong, it should be `INT`, not `KILL`.

josephglanville:
Eyes somehow completed missed that, clearly glasses aren't enough.

PR #3421, opened 2016-09-24
scheduler,router: Wait for router drain when stopping jobs

lmars:
This is an attempt to prevent 503s during deployments by waiting for requests to finish to backends before stopping jobs.


josephglanville:
Tracking requests rather than attempting to attack the connection management is a cool idea, connection management is a gnarly proposition in comparison.
I wonder if adding a mutex acquire to every request has visible performance overhead though.

controller/scheduler/router.go
```diff
@@ -0,0 +1,106 @@
...
+			select {
+			case <-r.stop:
+				return
+			default:
+			}
+
+			if err := connect(); err == nil {
+				break retryLoop
+			}
+			time.Sleep(100 * time.Millisecond)
```
titanous: I think this could be deleted and moved up to the `select` as a `time.After`.

controller/scheduler/scheduler.go
```diff
@@ -1181,6 +1226,49 @@ func (s *Scheduler) HandleHostEvent(e *discoverd.Event) {
...
+}
+
+func (s *Scheduler) HandleRouterDrainEvent(e *RouterDrainEvent) {
+	drain, ok := s.routerDrains[e.Backend.JobID]
+	if !ok {
+		return
+	}
+	drain.drained[e.RouterID] = struct{}{}
+	s.logger.Info("job drained from router", "job.id", e.Backend.JobID, "router.id", e.RouterID, "count", len(drain.drained), "total", len(s.routers))
+	if len(drain.drained) == len(s.routers) {
```
titanous: Should we do a diff comparison, as routers might have been added/removed?

controller/scheduler/scheduler.go
```diff
@@ -1359,6 +1447,9 @@ func (s *Scheduler) handleJobStatus(job *Job, status host.JobStatus) {
 			formation = s.handleFormation(ef)
 		}
 		job.Formation = formation
+		if proc, ok := formation.Release.Processes[job.Type]; ok {
+			job.Service = proc.Service
```
titanous: I think this needs to be more complicated, we should only set this if there is a corresponding route. Also, if the route is later removed the scheduler will wait indefinitely for the backend to drain. I guess it comes down to needing to know if the backend is currently tracked by the router.

host/http.go
```diff
@@ -215,6 +227,15 @@ func (h *jobAPI) StopJob(w http.ResponseWriter, r *http.Request, ps httprouter.P
 	w.WriteHeader(200)
 }
 
+func (h *jobAPI) DeregisterJob(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
titanous: Maybe this should be called `DiscoverdDeregisterJob` or something like that? It's not clear what it is being deregistered from.

controller/scheduler/scheduler.go
```diff
@@ -1512,11 +1603,32 @@ func (s *Scheduler) stopJob(job *Job) error {
...
+			if err := host.client.DeregisterJob(job.JobID); err == nil {
+				log.Info("waiting for routers to stop sending requests")
+				select {
+				case <-drain.done:
+				case <-time.After(routerDrainTimeout):
+					log.Warn("timed out waiting for routers to stop sending requests")
+				}
+			} else {
+				log.Error("error signalling job to deregister from service discovery", "err", err)
+			}
```
titanous: Should the drain be cleared out in the deregister error and timeout branches?

router/http.go
```diff
@@ -482,15 +480,79 @@ type httpRoute struct {
...
+		reqs: make(map[string]int64),
+		cond: sync.NewCond(&sync.Mutex{}),
+	}
+	s.watchBackends()
+	return s
+}
+
+func (s *service) TrackRequestStart(backend string) {
+	s.cond.L.Lock()
+	defer s.cond.L.Unlock()
```
titanous: Changing this to not use `defer` will speed it up a bit.

router/http.go
```diff
@@ -482,15 +480,79 @@ type httpRoute struct {
...
+	s.reqs[backend]++
+}
+
+func (s *service) TrackRequestDone(backend string) {
+	s.cond.L.Lock()
+	defer s.cond.L.Unlock()
```
titanous: It's worth changing this to not use `defer` as well.

controller/scheduler/router.go
lmars: The reason I put this here is so that a retry is tried straight away, and we sleep only if the retry fails.

titanous: Ah, that makes sense.

controller/scheduler/router.go
```diff
@@ -0,0 +1,106 @@
...
+		// keep trying to reconnect, unless we are told to stop
+	retryLoop:
+		for {
+			select {
+			case <-r.stop:
+				return
+			default:
+			}
+
+			if err := connect(); err == nil {
```
imjorge: we are not closing stream between connect, right? would that be problem?

controller/scheduler/scheduler.go
```diff
@@ -278,6 +288,34 @@ func (s *Scheduler) streamHostEvents() error {
 	}
 }
 
+func (s *Scheduler) streamRouterEvents() {
+	log := s.logger.New("fn", "streamRouterEvents")
+
+	var events chan *discoverd.Event
+	var stream stream.Stream
```
imjorge: it this gets assigned, or if `connect` is called again (below) should the stream be closed?

controller/scheduler/router.go
lmars: This code path is only hit when the stream has closed (and so the `events` channel is also closed), so there is no need to explicitly close the stream.

controller/scheduler/scheduler.go
lmars: The stream is closed in a deferred function call [here](https://github.com/flynn/flynn/blob/scheduler-graceful-shutdown/controller/scheduler/router.go#L78).

lmars:
@titanous @josephglanville @imjorge I have updated things as follows:

* Added a `--drain-backends` flag when adding routes which enables backend request tracking and up / down / drained events
* Updated the scheduler to watch for router backend events and only wait for drain events for jobs which previously were marked as up by the router (i.e. only if the route has backend draining enabled)
* Added `SchedulerSuite.TestGracefulShutdown` which makes some long-lived HTTP requests and makes sure they are drained before the scheduler stops the backend jobs

I also ran `wrk` for 30s and performed multiple deployments (`flynn env set` in a loop).

Without this change:

```
$ wrk -t10 -c10 -d30s http://http-test.1.localflynn.com
Running 30s test @ http://http-test.1.localflynn.com
  10 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.33ms    1.45ms  26.92ms   90.24%
    Req/Sec     0.94k   130.66     1.72k    67.33%
  280384 requests in 30.05s, 31.82MB read
  Non-2xx or 3xx responses: 107
Requests/sec:   9329.55
Transfer/sec:      1.06MB

# many errors in the router logs
$ flynn -a router log -r | grep -F 'lvl=eror' | wc -l
470

# mostly ECONNREFUSED and ECONNRESET
$ flynn -a router log -r | grep -F 'lvl=eror' | grep -F 'connection refused' | wc -l
362
$ flynn -a router log -r | grep -F 'lvl=eror' | grep -F 'connection reset by peer' | wc -l
88
```

With this change:

```
$ wrk -t10 -c10 -d30s http://http-test.1.localflynn.com
Running 30s test @ http://http-test.1.localflynn.com
  10 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.35ms    1.60ms  32.41ms   91.21%
    Req/Sec     0.94k   140.51     1.43k    64.60%
  281359 requests in 30.04s, 31.93MB read
Requests/sec:   9364.86
Transfer/sec:      1.06MB

# No errors in the logs
$ flynn -a router log -r | grep -F 'lvl=eror' | wc -l
0
```

This was for the test HTTP app (`test/apps/http`).

I also did the same for a backend with more realistic response times (random between 50ms and 2000ms) and I found the same result, just with much smaller throughput so less errors (increasing the threads / connections leads to port exhaustion pretty quickly).

titanous:
Is there a downside to defaulting the `--drain-backends` flag to true?

lmars:
@titanous no downside, I have flipped the flag to be `--no-drain-backends`.

router/schema.go
```diff
@@ -210,6 +210,10 @@ CREATE TRIGGER notify_route_certificates_update
 	AFTER INSERT OR UPDATE OR DELETE ON route_certificates
 	FOR EACH ROW EXECUTE PROCEDURE notify_route_certificates_update()`,
 	)
+	migrations.Add(6,
+		`ALTER TABLE tcp_routes ADD COLUMN drain_backends boolean NOT NULL DEFAULT FALSE`,
+		`ALTER TABLE http_routes ADD COLUMN drain_backends boolean NOT NULL DEFAULT FALSE`,
```
titanous: Should we default these to true?

lmars:
@titanous comments addressed

PR #3425, opened 2016-09-25
Sirenia Tunables

josephglanville:
This changeset adds dynamic reconfiguration of Sirenia databases without the need to do a deployment.
Changes to the tunables are persisted through the Sirenia state machine.
The primary is responsible for applying the change to the discoverd metadata and each of the cluster peers pick up the change like they would a change to the topology.
Each database now supplies a default set of tunables to the peer state machine on initialisation.
Tunable sets are versioned to allow easy comparison and rejection of out of date requests to update tunables.


josephglanville:
@lmars @titanous this is ready for review.

titanous:
How does this interact with backups? It looks like tunables are only persisted in discoverd, is that correct?

Parameters operate on a whitelist basis, are you planning to add more before merging? (or some kind of whitelist override?)

josephglanville:
Currently they are not backed up that is correct.

The reason for the whitelist is that some parameters require more complicated infrastructure to change online, namely pg max connections. We can definitely whitelist more settings however, these were just the ones I know people are going to want to change.

titanous:
I was thinking they should be included in the cluster backup file as JSON and loaded into discoverd before/while booting the appliances, does that make sense?

josephglanville:
Yeah that makes sense, all add that to this PR.

lmars:
This is outdated, closing.

PR #3427, opened 2016-09-25
appliance/postgres: Refactor Sirenia wrapper

josephglanville:
Refactors the wrapper code in the Postgres appliance to better match convention established since building the other appliances.
This is a prerequisite step towards sharing more code between the appliances.


appliance/postgresql/handler.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	httphelper.JSON(w, 200, status)
+}
+
+func (h *Handler) handlePostStop(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	if err := h.Peer.Stop(); err != nil {
+		httphelper.Error(w, err)
+		return
+	}
+	if err := h.Heartbeater.Close(); err != nil {
```
imjorge: shouldn't the heartbeater be stopped before stopping the peer?

josephglanville: It's stopped after the peer has been instructed to shut down to avoid processing additional changes to the topology caused by removing ourselves from discoverd, it's a very narrow race but one nonetheless.

imjorge: Was thinking that, because HB is not closed right away, the service is marked as up when it really is not longer up.

PR #3429, opened 2016-09-27
discoverd: Return ServiceUnavailable when no current leader

### LINKED ISSUE #3318: discoverd returning 404 for path /services/cluster-monitor/meta during bootstrap
Was restoring a cluster from backup. Seems like `discoverd` started up just fine. `flannel` seems like it had an issue getting info out of `discoverd`.

Here is the gist - https://gist.github.com/anonymous/63637481e0bc1b3629d5dda9cc613f31

### LINKED ISSUE #3423: Launch of flynn cluster failed
We had a failed cluster creation at Bazaarvoice today 

https://gist.github.com/anonymous/63a16e6ef39f872f26c12250cd9f5c8d

Will be trying to get a new cluster up quickly as opposed to diagnosing this cluster.

josephglanville:
Requests to non-leader nodes that can't find a current leader will now be retried.

Closes #3423
Closes #3318

### DIFF
--- a/discoverd/server/handler.go
+++ b/discoverd/server/handler.go
@@ -426,7 +426,7 @@ func (h *Handler) serveStream(w http.ResponseWriter, params httprouter.Params, k
 func (h *Handler) serveGetRaftLeader(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
 	leader := h.Store.Leader()
 	if leader == "" {
-		hh.Error(w, ErrNoKnownLeader)
+		hh.ServiceUnavailableError(w, ErrNoKnownLeader.Error())
 		return
 	}
 
@@ -475,7 +475,7 @@ func (h *Handler) redirectToLeader(w http.ResponseWriter, r *http.Request) {
 	// Find the current leader.
 	leader := h.Store.Leader()
 	if leader == "" {
-		hh.Error(w, ErrNoKnownLeader)
+		hh.ServiceUnavailableError(w, ErrNoKnownLeader.Error())
 		return
 	}
 


lmars:
Do we not need to update the client to actually do the retry for retryable requests?

josephglanville:
@lmars the client is already setup to retry on `ServiceUnavailable` here: https://github.com/flynn/flynn/blob/master/discoverd/client/client.go#L260

It was put in place to retry requests from servers that are currently shutting down (https://github.com/flynn/flynn/blob/master/discoverd/server/handler.go#L124), not also using it for the ErrNoKnownLeader case was an oversight.

lmars:
@josephglanville ah ok I was looking for a call to [IsRetryableError](https://github.com/flynn/flynn/blob/master/pkg/httphelper/httphelper.go#L82-L86), should we be using that instead?

Also, looking a little closer at the code it seems the retry is done by cycling through the configured clients and trying them once each, but if the error is `ErrNoKnownLeader` is it not likely that each server will return that in quick succession? Should we not cycle through them a few times if the error is retryable (e.g. try for 10s [like the controller](https://github.com/flynn/flynn/blob/master/controller/client/v1/client.go#L795-L800))?

josephglanville:
Yeah that is a good point, we should use `IsRetryableError` instead.

Agree on the retry front. Currently we only try each server once.
I think the behavior should be to try all servers once, then pause and try them all again.
Thoughts?

josephglanville:
@lmars incorporated your suggestions to improve the client in #3432

PR #3432, opened 2016-09-27
discoverd/client: Improve client retry behavior

josephglanville:
* Don't shadow the errors module
* Retry until RequestDeadline is reached.
* Retry all servers again after RetryInterval.

### DIFF
--- a/discoverd/client/client.go
+++ b/discoverd/client/client.go
@@ -11,12 +11,18 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	dt "github.com/flynn/flynn/discoverd/types"
 	"github.com/flynn/flynn/pkg/httpclient"
 	hh "github.com/flynn/flynn/pkg/httphelper"
 	"github.com/flynn/flynn/pkg/stream"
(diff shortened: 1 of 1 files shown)


discoverd/client/client.go
```diff
@@ -23,6 +24,9 @@ var ErrTimedOut = errors.New("discoverd: timed out waiting for instances")
 
 var defaultLogger = log15.New("component", "discoverd")
 
+var RequestDeadline = 10 * time.Second
+var RetryInterval = 100 * time.Millisecond
```
titanous: Should these be constants or exposed publicly? Normally exposing global variables from a package is a bad idea. If you want them to be configurable, they should be stored as client configuration.

discoverd/client/client.go
```diff
@@ -23,6 +24,9 @@ var ErrTimedOut = errors.New("discoverd: timed out waiting for instances")
 
 var defaultLogger = log15.New("component", "discoverd")
 
+var requestDeadline = 10 * time.Second
+var retryInterval = 100 * time.Millisecond
```
titanous: Can these be constants?

PR #3452, opened 2016-09-30
demo: fix 'make dashboard'

### LINKED ISSUE #3450: `make dashboard` in demo project errors out for me on Ubuntu
I'm just trying out the demo project and I was able to run `make init`, install the CLI, start a cluster, and go to the dashboard and it worked.  However, after I shut it down and ran it again, (`make down; make up`) I tried the command it gave me to open the dashboard `make dashboard` and got this error:

```
$ make dashboard
Couldn't get a file descriptor referring to the console
Makefile:77: recipe for target 'dashboard' failed
make: *** [dashboard] Error 1
```

I tried make reset, which seemed to work fine, but I got the same error when running `make dashboard`.  I can still access the dashboard, just have to be sure to store the login token somewhere, since I can't access it via make anymore.

mltsy:
This PR fixes #3450 by trying `xdg-open` first, which is the Linux equivalent of OS X `open`
Then, if that fails, use `open`... I don't know if there's a better way to do this, but this works for me! :)

Signed-off-by: Joe Marty <jmarty@iexposure.com>

### DIFF
--- a/demo/Makefile
+++ b/demo/Makefile
@@ -74,7 +74,7 @@ token:
 
 # Open dashboard in browser
 dashboard:
-	@open http://dashboard.${CLUSTER_DOMAIN}?token=${$(MAKE) token}
+	@URL=http://dashboard.${CLUSTER_DOMAIN}?token=${$(MAKE) token} && xdg-open $$URL || open $$URL
 
 # Run a command in the VM via SSH
 ssh-cmd:
--- a/demo/Makefile
+++ b/demo/Makefile
@@ -74,7 +74,8 @@ token:
 
 # Open dashboard in browser
 dashboard:
-	@URL=http://dashboard.${CLUSTER_DOMAIN}?token=${$(MAKE) token} && xdg-open $$URL || open $$URL
+	@CMD=open && which xdg-open > /dev/null && CMD=xdg-open; \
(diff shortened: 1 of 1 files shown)


demo/Makefile
```diff
@@ -74,7 +74,7 @@ token:
 
 # Open dashboard in browser
 dashboard:
-	@open http://dashboard.${CLUSTER_DOMAIN}?token=${$(MAKE) token}
+	@URL=http://dashboard.${CLUSTER_DOMAIN}?token=${$(MAKE) token} && xdg-open $$URL || open $$URL
```
lmars: It would be cleaner to have a variable like `$(OPEN)` which is set conditionally. I don't know the exact syntax, but something like:

```
OPEN := open
if $(which xdg-open)
  OPEN = xdg-open
endif
```

mltsy: As far as I know, it's not possible to set a make variable based on executable bash conditions.  You can only set shell & environment variables in that context, but I'll try that...

mltsy:
I'm not sure if I did something wrong here... there seem to be 3 unrelated tests failing.  At least from what I can tell they're unrelated... ?

lmars:
@mltsy we have some tests which fail intermittently in CI (I have pushed some fixes in the last couple of days), these specific failures are unrelated to your change.

lmars:
Thanks!

PR #3463, opened 2016-10-03
demo: depend on install-cli target on Makefile where required

### LINKED ISSUE #3459: Vagrant installer: flynn command not found
seems that `flynn` must be installed prior the execution of vagrant installer

```
Connection to 127.0.0.1 closed.
make[2]: uscita dalla directory "/home/ale/flynn/demo"
bash: riga 1: flynn: comando non trovato
Makefile:35: set di istruzioni per l'obiettivo "bootstrap" non riuscito
make[1]: *** [bootstrap] Errore 127
make[1]: uscita dalla directory "/home/ale/flynn/demo"
Makefile:22: set di istruzioni per l'obiettivo "init" non riuscito
make: *** [init] Errore 2
```

i've destroyed the vagrant VM, manually installed the `flynn` cli and started again the vagrant installed and everything went ok

So, in docs, you should say that `flynn` cli must be installed prior running `make init` or `make init` should install the cli during the process

imjorge:
closes #3459 /cc @lmars

### DIFF
--- a/demo/Makefile
+++ b/demo/Makefile
@@ -14,7 +14,7 @@ CLI_INSTALL_BIN := /usr/local/bin/flynn
 VAGRANT_ENV := CLUSTER_DOMAIN="${CLUSTER_DOMAIN}" SSH_AUTH_SOCK="" FLAG_BOOTSTRAP=false
 
 
-.PHONY: init install-cli bootstrap up down halt update ssh reset token destroy dashboard ssh-cmd
+.PHONY: init bootstrap up down halt update ssh reset token destroy dashboard ssh-cmd
 
 
 # Initialize a new cluster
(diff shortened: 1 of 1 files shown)


demo/Makefile
```diff
@@ -31,7 +31,7 @@ install-cli:
 # Bootstrap the cluster
 # This should only be called once per cluster
 # Runs bootstrap in VM and cluster add on host
-bootstrap:
+bootstrap: install-cli
```
lmars: We should update `install-cli` so it only installs the CLI if it doesn't exist (probably easiest to rename `install-cli` to `$(CLI_INSTALL_BIN)` so the `make` will do that).

imjorge: @lmars done

lmars:
Thanks!

PR #3467, opened 2016-10-04
discoverd: Prevent deadlock when restoring from snapshot

### LINKED ISSUE #3451: flynn-host fails to fully restart after node reboot
To test automatic healing, I rebooted one node of a three-node cluster. When the node came back online, flynn-host started as expected: `sudo status flynn-host` shows 'running'. However, the local webserver on :1111 is only returning empty responses. This breaks most of the flynn-host commands available on that node and causes the cluster's discoverd process to log a continuous stream of connection failures.

On the restarted node:
```
ubuntu@ip-10-210-18-37:~$ flynn-host list
19:09:06.171373 host.go:154: Error sending HTTP request, errors: Get http://127.0.0.1:1111/services/flynn-host/instances: EOF
```

In the discoverd process logs, this message repeats several times per second:
```
[2016-09-30T19:12:30.114734548Z] 2016/09/30 19:12:30 [DEBUG] raft: Failed to contact 10.210.18.37:1111 in 2h45m59.891473048s
```

Just to be certain, I have confirmed that the restarted node is reachable by both other nodes, but attempts to `curl` the restarted node on port 1111 result in the same empty response.
```
ubuntu@ip-10-210-18-54:~$ curl http://10.210.18.37:1111/services/flynn-host/instances
curl: (52) Empty reply from server
```

lmars:
Fixes #3451.

### DIFF
--- a/discoverd/server/store.go
+++ b/discoverd/server/store.go
@@ -197,10 +197,23 @@ func (s *Store) Open() error {
 	}
 
 	// Create raft log.
+	//
+	// The mutex must be unlocked as initializing the raft store may
+	// call back into methods which acquire the lock (e.g. Restore)
+	s.mu.Unlock()
 	r, err := raft.NewRaft(config, s, cacheStore, stableStore, ss, s.peerStore, s.transport)
+	s.mu.Lock()
 	if err != nil {
 		return fmt.Errorf("raft: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


ghkora:
Quoting Joseph Glanville <notifications@github.com>:

PR #3471, opened 2016-10-05
host: Use SquashFS + OverlayFS via Mountspecs 

lmars:
Summary of changes:

* Support importing raw (possibly sparse) filesystem images into zvols and then mounting them
* Add `host.Mountspec` having `URL` and `Hashes` which reference a SquashFS filesystem, allowing the host to download, verify and import it into a volume
* Add `job.Mountspecs` which is a list of SquashFS mountspecs which are layered at the root directory of the container using OverlayFS, with a read-write Ext2 volume on top
* Add `resource.TypeDisk` which controls how big the read-write Ext2 volume is for each job

Things to note:

* I have made the default disk limit 100MB because most jobs should not need to write to the filesystem, and if they do they should provision a data volume and use that, but in case this is problematic I've added support to make it configurable in a future PR (i.e. via `flynn limit set disk=1G`)
* I don't expect this to pass CI (or to even build) but this should give a good idea of the changes needed in other components (i.e. providing `job.Mountspecs` instead of `job.ImageArtifact` / `job.FileArtifacts`)

### DIFF
--- a/host/libcontainer_backend.go
+++ b/host/libcontainer_backend.go
@@ -2,14 +2,20 @@ package main
 
 import (
 	"bytes"
+	"crypto/sha256"
+	"crypto/sha512"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash"
 	"io"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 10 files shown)


host/libcontainer_backend.go
```diff
@@ -1386,3 +1555,45 @@ func createCGroupPartition(name string, cpuShares int64) error {
...
+	if err := f.Truncate(size); err != nil {
+		return "", err
+	}
+
+	cmd := exec.Command("mkfs.ext2", "-F", "-L", "rootfs", "-m", "0", f.Name())
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return "", fmt.Errorf("error creating ext2 filesystem: %s: %s", err, out)
+	}
+
+	t.cache[size] = f.Name()
```
titanous: I think we should only cache  the default size, to prevent unbounded cache growth. In the future we can evaluate common sizes and add them or do some heuristics.

host/resource/resource.go
```diff
@@ -33,6 +33,10 @@ const (
 	// practice, a 1000 milliCPU limit is equivalent to 1024 CPU shares.
 	TypeCPU Type = "cpu"
 
+	// TypeDisk specifies the available disk space in bytes inside the
+	// container.
+	TypeDisk = "disk"
```
titanous: This should probably be `TypeWritableOverlayVolume` or `TypeTempDisk` or something along those lines so it doesn't get confused with the data volume support.

lmars:
@titanous comments addressed (I renamed the disk limit constant to `TypeTempDisk`)

PR #3472, opened 2016-10-05
adding a note about using external blobstore is still a good idea eve…

kckrinke:
…n if not within AWS or GC


lmars:
@kckrinke please add a `docs:` prefix and a [DCO sign-off](https://github.com/flynn/flynn/blob/master/CONTRIBUTING.md#developers-certificate-of-origin) to your commit message.

PR #3481, opened 2016-10-06
all: Add Flynn image artifacts

lmars:
Summary:

* Add `ArtifactTypeFlynn` with new db columns `Manifest`, `Hashes` and `LayerURLTemplate`
* Expect `release.Artifacts` to just be a list of Flynn image artifacts which are overlayed in the containers root directory
* Deprecate "docker" and "file" artifacts (they will be removed in a further PR before merging to master)
* Add `utils.SetupMountspecs` which takes a list of Flynn image artifacts and populates `job.Mountspecs`
* Update manifest interpolater to convert `$image_artifact` into a Flynn image artifact and use in the bootstrap manifest


util/release/manifest.go
```diff
@@ -35,45 +35,54 @@ func manifest(args *docopt.Args) {
...
+				Type:     ct.ArtifactTypeFlynn,
+				URI:      fmt.Sprintf("%s?target=/images/%s.json", imageRepository, image.ID()),
+				Manifest: image,
+				Hashes:   image.Hashes(),
+				Meta:     map[string]string{"flynn.component": name},
+			}
+			if version.Dev() {
+				artifact.LayerURLTemplate = "file:///var/lib/flynn/layer-cache/{id}.squashfs"
+			} else {
+				artifact.LayerURLTemplate = fmt.Sprintf("%s?target=/layers/{id}.squashfs", imageRepository)
```
titanous: Should this path include the release version?

lmars:
@titanous I have moved the image verification logic into `pkg/verify` and updated `TestFlynnArtifact` to be much more exhaustive.

util/release/manifest.go
lmars: I decided not to put the version in the URLs because there doesn't seem to be a need to. There will still be an image manifest per version (e.g. `/vXXX/images.json`) but it will just have URLs to images like `/image/SHA.json` containing layers with URLs like `/layer/SHA.layer`.

pkg/verify/verify.go
```diff
@@ -0,0 +1,103 @@
...
+		h := &Hash{algorithm: algorithm, expected: value}
+		switch algorithm {
+		case "sha256":
+			h.hash = sha256.New()
+		case "sha512":
+			h.hash = sha512.New()
+		case "sha512_256":
+			h.hash = sha512.New512_256()
+		default:
+			return nil, &ErrUnknownAlgorithm{algorithm}
```
titanous: This error shouldn't exist, as it will prevent adding new hashes to manifests in the future.

pkg/verify/verify.go
```diff
@@ -0,0 +1,103 @@
...
+
+type Verifier struct {
+	hashes []*Hash
+	size   int64
+	reader *io.LimitedReader
+}
+
+func NewVerifier(hashes map[string]string, size int64) (*Verifier, error) {
+	if len(hashes) == 0 {
+		return nil, ErrNoHashes
```
titanous: This check should run at the end so that unknown hash algorithms are not counted.

util/release/manifest.go
titanous: Do you think it makes sense to include it anyway for debugging and human readability purposes?

controller/artifact.go
```diff
@@ -31,33 +35,49 @@ func (r *ArtifactRepo) Add(data interface{}) error {
...
+		}
+	}
+
 	tx, err := r.db.Begin()
 	if err != nil {
 		return err
 	}
 
-	err = tx.QueryRow("artifact_insert", a.ID, string(a.Type), a.URI, a.Meta).Scan(&a.CreatedAt)
+	err = tx.QueryRow("artifact_insert", a.ID, string(a.Type), a.URI, a.Meta, a.Manifest, a.Hashes, a.Size, a.LayerURLTemplate).Scan(&a.CreatedAt)
```
titanous: Should we persist the raw JSON manifest instead of the round-tripped manifest so that the hashes continue to match?

lmars:
@titanous comments addressed

PR #3482, opened 2016-10-07
pkg/backup: Run database dumps in background partition

josephglanville:
This runs the database dumps on the cluster in the `background` partition so they run at reduced priority to user jobs.

Also adds capability to scheduler jobs with custom partition via the controller.

### DIFF
--- a/controller/types/types.go
+++ b/controller/types/types.go
@@ -250,6 +250,14 @@ func JobDownEvents(count int) map[JobState]int {
 	return map[JobState]int{JobStateDown: count}
 }
 
+type PartitionType string
+
+const (
+	PartitionTypeBackground PartitionType = "background"
+	PartitionTypeSystem     PartitionType = "system"
+	PartitionTypeUser       PartitionType = "user"
+)
+
 type NewJob struct {
 	ReleaseID  string             `json:"release,omitempty"`
 	ArtifactID string             `json:"artifact,omitempty"`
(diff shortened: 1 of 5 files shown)


schema/controller/new_job.json
```diff
@@ -54,6 +54,9 @@
     },
     "resources": {
       "$ref": "/schema/controller/common#/definitions/resources"
+    },
+    "partition": {
+      "$ref": "/schema/controller/common#/definitions/parition"
```
lmars: s/parition/partition/

pkg/backup/backup.go
```diff
@@ -71,6 +73,7 @@ func Run(client controller.Client, out io.Writer, progress ProgressBar) error {
 				"MONGO_PWD": mongodbRelease.Env["MONGO_PWD"],
 			},
 			DisableLog: true,
+			Partition:  "background",
```
lmars: Is it worth having constants for these (e.g. `ct.PartitionTypeBackground` etc.)?

josephglanville: Yeah probably a good idea even if we do add custom partition support.

controller/types/types.go
```diff
@@ -243,6 +243,12 @@ func JobDownEvents(count int) map[JobState]int {
 	return map[JobState]int{JobStateDown: count}
 }
 
+const (
+	PartitionTypeBackground = "background"
+	PartitionTypeSystem     = "system"
+	PartitionTypeUser       = "user"
```
lmars: Make these an explicit type (e.g. `type PartitionType string`) then we can't slip up and assign random strings to `NewJob.Partition`

titanous:
@josephglanville Is this ready for @lmars to review again?

josephglanville:
Yep, will rebase it.

josephglanville:
@lmars ready for review.

PR #3496, opened 2016-10-14
Migrate slugs and docker-receive releases to Flynn images

lmars:
This includes the following changes:

* Add `Uid` / `Gid` to `ImageEntrypoint` which controls the User / Group used for the job process
* Build slugs as Flynn images containing `/app` and `/etc/{passwd,group}` with a `flynn` user
* Convert Docker images to Flynn images when pushed to `docker-receive`
* Update app import / export to support Flynn images
* Delete images from `docker-receive` based apps when deleting releases

### DIFF
--- a/cli/export.go
+++ b/cli/export.go
@@ -117,14 +117,88 @@ func runExport(args *docopt.Args, client controller.Client) error {
 		}
 	}
 
-	var artifact *ct.Artifact
-	if artifactID := release.ImageArtifactID(); artifactID != "" {
-		artifact, err = client.GetArtifact(artifactID)
+	blobstoreRelease, err := client.GetAppRelease("blobstore")
+	if err != nil {
+		return fmt.Errorf("error getting blobstore release: %s", err)
+	}
+	download := func(name, url string, release *ct.Release) error {
+		reqR, reqW := io.Pipe()
+		config := runConfig{
+			App:        mustApp(),
+			Release:    release.ID,
+			Artifacts:  release.ArtifactIDs[:1],
+			DisableLog: true,
+			Args:       []string{"curl", "--include", "--location", "--raw", url},
+			Stdout:     reqW,
+			Stderr:     ioutil.Discard,
+		}
+		if bar != nil {
+			config.Stdout = io.MultiWriter(config.Stdout, bar)
+		}
+		go func() {
+			if err := runJob(client, config); err != nil {
+				shutdown.Fatalf("error downloading %s: %s", name, err)
+			}
+		}()
+		req := bufio.NewReader(reqR)
+		var res *http.Response
+		maxRedirects := 5
+		for i := 0; i < maxRedirects; i++ {
+			res, err = http.ReadResponse(req, nil)
+			if err != nil {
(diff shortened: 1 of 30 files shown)


josephglanville:
Does the conversion stage handle AUFS whiteouts?

lmars:
@josephglanville yes we convert AUFS whiteouts to OverlayFS whiteouts in the `imagebuilder`:

https://github.com/flynn/flynn/blob/flynn-images/pkg/imagebuilder/builder.go#L138-L181

cli/export.go
```diff
@@ -531,26 +571,96 @@ func runImport(args *docopt.Args, client controller.Client) error {
...
+		}
+
+		slugImageID := random.UUID()
+		config := runConfig{
+			App:        app.ID,
+			Release:    gitreceiveRelease.ID,
+			ReleaseEnv: true,
+			Artifacts:  []string{slugBuilderID},
+			DisableLog: true,
+			Args:       []string{"/bin/convert-legacy-slug.sh"},
```
titanous: What will happen if the import command is called against a cluster that doesn't have this code?

docker-receive/artifact/main.go
```diff
@@ -0,0 +1,153 @@
...
+type layerStore struct{}
+
+func (l *layerStore) Load(id string) (*ct.ImageLayer, error) {
+	res, err := http.Get(jsonURL(id))
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	if res.StatusCode != http.StatusOK {
+		return nil, nil
```
lmars: It returns `nil` to indicate to the `imagebuilder` that the layer doesn't exist so needs to be built.

I've made this more specific (i.e. just for `404`) and added a comment.

lmars:
@titanous comments addressed.

controller/jobs.go
```diff
@@ -243,6 +243,21 @@ func (c *controllerAPI) RunJob(ctx context.Context, w http.ResponseWriter, req *
 		return
 	}
 
+	artifacts := make([]*ct.Artifact, len(artifactIDs))
+	artifactList, err := c.artifactRepo.ListIDs(artifactIDs...)
+	if err != nil {
+		respondWithError(w, err)
+		return
+	}
+	for i, id := range artifactIDs {
```
imjorge: why do need to make a copy of `artifactList` into `artifacts`. why not just use `artifactList`?

cli/export.go
```diff
@@ -117,14 +117,88 @@ func runExport(args *docopt.Args, client controller.Client) error {
...
+		maxRedirects := 5
+		for i := 0; i < maxRedirects; i++ {
+			res, err = http.ReadResponse(req, nil)
+			if err != nil {
+				return fmt.Errorf("error reading HTTP response: %s", err)
+			}
+			if res.StatusCode != http.StatusFound {
+				break
+			}
+		}
```
lmars: Whilst we are reading multiple responses, they are all coming over the same TCP connection (i.e. from `curl`), so closing the response would actually break reading the body later on.

PR #3497, opened 2016-10-14
host/cli: Fix volume garbage collection

josephglanville:
Volumes from stopped jobs are still considered attached.
This fixes volume gc so that unless the volume is attached to a running
or starting job it will be collected.


lmars:
We should probably add some test coverage for this, although I don't think we should block merging this fix.

josephglanville:
Agreed, I will look at adding some tests for it.

PR #3519, opened 2016-10-16
cli: Add --jobs to commands that invoke pg_restore

josephglanville:
Allows altering the concurrency of the restore process.
This change also changes the default behaviour to use 4 jobs.

### DIFF
--- a/cli/pg.go
+++ b/cli/pg.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 
 	"github.com/cheggaaa/pb"
 	"github.com/docker/docker/pkg/term"
@@ -16,11 +17,12 @@ func init() {
 	register("pg", runPg, `
 usage: flynn pg psql [--] [<argument>...]
        flynn pg dump [-q] [-f <file>]
-       flynn pg restore [-q] [-f <file>]
+       flynn pg restore [-q] [-j <jobs>] [-f <file>]
 
 Options:
 	-f, --file=<file>  name of dump file
 	-q, --quiet        don't print progress
+	-j, --jobs=<jobs>  number of pg_restore jobs to use [default: 1]
 
 Commands:
 	psql     Open a console to a Flynn postgres database. Any valid arguments to psql may be provided.
@@ -35,7 +37,7 @@ Examples:
 
     $ flynn pg dump -f db.dump
 
-    $ flynn pg restore -f db.dump
+    $ flynn pg restore -j 8 -f db.dump
 `)
 }
 
@@ -131,6 +133,10 @@ func pgDump(client controller.Client, config *runConfig) error {
 }
 
 func runPgRestore(args *docopt.Args, client controller.Client, config *runConfig) error {
+	jobs, err := strconv.Atoi(args.String["--jobs"])
+	if err != nil {
+		return err
+	}
 	config.Stdin = os.Stdin
 	var size int64
 	if filename := args.String["--file"]; filename != "" {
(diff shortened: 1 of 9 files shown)


cli/pg.go
```diff
@@ -160,11 +165,12 @@ func runPgRestore(args *docopt.Args, client controller.Client, config *runConfig
...
 		config.Stdin = bar.NewProxyReader(config.Stdin)
 	}
-	return pgRestore(client, config)
+
+	return pgRestore(client, config, args.String["--jobs"])
 }
 
-func pgRestore(client controller.Client, config *runConfig) error {
-	config.Args = []string{"pg_restore", "-d", config.Env["PGDATABASE"], "--clean", "--if-exists", "--no-owner", "--no-acl"}
+func pgRestore(client controller.Client, config *runConfig, jobs string) error {
```
lmars: Should jobs be an `int`?

josephglanville: I did that originally but it needs to be passed as a `string` anyway so instead just opted to validate it was a valid integer by not coerce it to `int` only to need to convert it back when invoking the job.

lmars:
@josephglanville looks like some genuine test failures:

```
++ 13:55:55.077 /home/ubuntu/go/src/github.com/flynn/flynn/test/../cli/bin/flynn pg restore -f /tmp/check-8995016276575641803/1/db.dump
strconv.ParseInt: parsing "": invalid syntax
```

josephglanville:
@lmars fixed, made mistake with docopt.

However I didn't look closely enough, appears `--jobs` is ignored when reading stdin.
Which makes sense because parallel restore requires to be able to seek to different tables.
I have modified to upload to a temp file but I think this should probably write to a data volume.

@titanous thoughts?

titanous:
Yeah, a data volume sounds right. That will allow us to move towards accepting database dumps from URLs, and resumable uploads.

josephglanville:
@titanous updated to use data volume, we can add script to container later to support fetching from URL etc.
Small changes were required to support data volumes for one-off jobs via the controller.

titanous:
Is this backwards-compatible with clusters that don't have those changes?

cli/pg.go
```diff
@@ -160,11 +166,22 @@ func runPgRestore(args *docopt.Args, client controller.Client, config *runConfig
...
-func pgRestore(client controller.Client, config *runConfig) error {
-	config.Args = []string{"pg_restore", "-d", config.Env["PGDATABASE"], "--clean", "--if-exists", "--no-owner", "--no-acl"}
+func pgRestore(client controller.Client, config *runConfig, jobs int) error {
+	// Provision a volume at /data to stream the dump to.
+	if jobs > 1 {
+		config.Data = true
+		config.Args = []string{
+			"bash",
+			"-c",
+			fmt.Sprintf("set -o pipefail; cat > /data/temp.dump && pg_restore -d %s -n public --clean --if-exists --no-owner --no-acl --jobs %d /data/temp.dump", config.Env["PGDATABASE"], jobs),
```
lmars: I assume we have to manually write the file in `/data` because `pg_restore` will create its own temp file in `/tmp`? Would setting `TMPDIR=/data` work?

controller/jobs.go
```diff
@@ -308,6 +309,11 @@ func (c *controllerAPI) RunJob(ctx context.Context, w http.ResponseWriter, req *
 		}
 	}
 
+	// provision data volume if required
+	if newJob.Data {
+		utils.ProvisionVolume(client, job)
```
lmars: We should check the error

cli/pg.go
josephglanville: I don't believe `pg_restore` creates temporary files (though I do believe `pg_dump` does if you use certain formats and stream to stdout).

We are writing the file to `/data` because it may be too big for the container root filesystem and `/tmp` may be on `tmpfs` and the dump could be bigger than available RAM.

controller/jobs.go
josephglanville: Yep will fix.

cli/pg.go
lmars: I'm confused, why do we need to write the file to disk at all? If `pg_restore` doesn't try to cache it on disk, then can't we just pass it directly via stdin? Or does it need to be a seekable file with `--jobs`?

lmars: Ah I missed your earlier comment, I get it.

josephglanville:
Updated with a compatibility check as the failure mode when the cluster didn't support the data volume was very unintuitive (the controller would accept the job, but not create the data volume, so job would fail on reading/writing to `/data`).

Required adding `Version` function to the controller client interface, was mentioned should see how this will interact with what we are doing with GraphQL.

/cc @lmars @jvatic

cli/pg.go
```diff
@@ -160,11 +166,28 @@ func runPgRestore(args *docopt.Args, client controller.Client, config *runConfig
...
 }
 
-func pgRestore(client controller.Client, config *runConfig) error {
-	config.Args = []string{"pg_restore", "-d", config.Env["PGDATABASE"], "--clean", "--if-exists", "--no-owner", "--no-acl"}
+func pgRestore(client controller.Client, config *runConfig, jobs int) error {
+	// Provision a volume at /data to stream the dump to.
+	if jobs > 1 {
+		// Check if controller supports data volumes for one-off jobs
+		compatMessage := "Cluster versions prior to %s don't support the --jobs argument for parallel restore."
+		if err := compatCheck(client, "v20161018.0", compatMessage); err != nil {
```
josephglanville: This is actually a dependency on the controller version (I added support for data volumes for one off jobs in this PR), but yeah, it would be nice if the Postgres version was also exposed in the status details.

The GraphQL stuff sounds good, if the error returned indicated which field was the problem that would probably solve most of the backwards compatible error handling.

PR #3529, opened 2016-10-18
bootstrap: Wait for Sirenia databases to become read/write

### LINKED ISSUE #3527: Unable to restore cluster from backup on digitalocean, stuck on "restore: start"
I tried 4 times to restore my cluster from backup on digitalocean (3 nodes) using the "flynn install" and it's get stuck on: "restore: start".

Local flynn version: v20161018.0
Node flynn-host version: v20161017.1 

here is my debug info of 1 of the nodes:
https://gist.github.com/d51f9025b98f0dd81b7fe16422fec5fe 

this is the installation log:

<script src="https://gist.github.com/ron137/ace4939bc2913a7f695e2a6ea61a47b4.js"></script>

josephglanville:
This fixes a race condition where attempting to restore the MariaDB database would deadlock with the initial replication backup transfer. However it's also correct to use this for Postgres and MongoDB so updated both of those to use this new wait step too.

Fixes #3527

### DIFF
--- a/bootstrap/sirenia_wait_action.go
+++ b/bootstrap/sirenia_wait_action.go
@@ -1,0 +1,41 @@
+package bootstrap
+
+import (
+	"time"
+
+	"github.com/flynn/flynn/discoverd/client"
+	"github.com/flynn/flynn/pkg/attempt"
+	"github.com/flynn/flynn/pkg/sirenia/client"
+)
+
+type SireniaWaitAction struct {
+	Service string `json:"service"`
+}
+
+func init() {
+	Register("sirenia-wait", &SireniaWaitAction{})
+}
+
+func (a *SireniaWaitAction) Run(s *State) error {
+	// use discoverd client to lookup leader
+	d, err := s.DiscoverdClient()
+	if err != nil {
+		return err
+	}
+
+	var leader *discoverd.Instance
+	err = attempt.Strategy{
+		Min:   5,
+		Total: 5 * time.Minute,
+		Delay: 500 * time.Millisecond,
+	}.Run(func() error {
+		leader, err = d.Service(a.Service).Leader()
+		return err
+	})
+	if err != nil {
+		return err
+	}
+
+	// connect using sirenia client and wait until database reports read/write
+	return client.NewClient(leader.Addr).WaitForReadWrite(5 * time.Minute)
+}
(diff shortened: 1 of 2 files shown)


josephglanville:
The `sirenia` name is the generic one, it's not specific to MariaDB, it's just the name of the state machine package.

bootstrap/sirenia_wait_action.go
```diff
@@ -0,0 +1,51 @@
...
+	fail:
+		select {
+		case <-timeout:
+			return fmt.Errorf("bootstrap: timed out waiting for discoverd leader for %s", a.Service)
+		case <-time.After(waitInterval):
+		}
+	}
+
+	// connect using sirenia client and execute WaitForReadWrite with 5 minute timeout
+	if err := client.NewClient(leader.Addr).WaitForReadWrite(5 * time.Minute); err != nil {
```
jvatic: should this use `waitMax` also?

josephglanville: Yep, good call.

bootstrap/sirenia_wait_action.go
```diff
@@ -0,0 +1,51 @@
...
+			goto fail
+		}
+		break
+	fail:
+		select {
+		case <-timeout:
+			return fmt.Errorf("bootstrap: timed out waiting for discoverd leader for %s", a.Service)
+		case <-time.After(waitInterval):
+		}
+	}
```
lmars: This could use `pkg/attempt`

bootstrap/sirenia_wait_action.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// connect using sirenia client and wait until database reports read/write
+	if err := client.NewClient(leader.Addr).WaitForReadWrite(waitMax); err != nil {
+		return err
+	}
+	return nil
```
lmars: Return the error directly.

josephglanville:
@lmars comments addressed.

bootstrap/sirenia_wait_action.go
```diff
@@ -0,0 +1,38 @@
...
+
+	var leader *discoverd.Instance
+	err = attempt.Strategy{
+		Min:   5,
+		Total: 5 * time.Minute,
+		Delay: 500 * time.Millisecond,
+	}.Run(func() error {
+		leader, err = d.Service(a.Service).Leader()
+		return err
+	})
```
lmars: Check the `err`

PR #3536, opened 2016-10-19
slugrunner: Allow user to write to /dev/std{out,err}

### LINKED ISSUE #3535: slugrunner: User cannot write to /dev/stdout
Slug based jobs run as the `nobody` user (see [here](https://github.com/flynn/flynn/blob/master/slugrunner/runner/init#L59)) but that user does not have write permissions to `/dev/stdout` as it is either owned by root:

```
# use a non-tty stdin so the job has no TTY
$ flynn run ls -l /dev/stdout < /dev/null
prw-r--r-- 1 root root 0 Oct 19 00:00 /dev/stdout

$ flynn run bash <<< 'echo > /dev/stdout'
bash: line 1: /dev/stdout: Permission denied
```

or is a symlink to the root owned `/dev/pts/0` for TTY jobs:

```
$ flynn run ls -l /dev/stdout
lrwxrwxrwx 1 root root 15 Oct 19 00:01 /dev/stdout -> /proc/self/fd/1

$ flynn run ls -l /proc/self/fd/1
lrwx------ 1 nobody nogroup 64 Oct 19 00:01 /proc/self/fd/1 -> /dev/pts/0

$ flynn run ls -l /dev/pts/0
crw--w---- 1 root tty 136, 0 Oct 19 00:02 /dev/pts/0

$ flynn run bash -c 'echo foo > /dev/stdout'
bash: /dev/stdout: Permission denied
```

We should `chown nobody /dev/std{in,out,err}` and add the `nobody` user to the `tty` group.

lmars:
Fixes #3535.

### DIFF
--- a/slugrunner/runner/init
+++ b/slugrunner/runner/init
@@ -53,6 +53,9 @@ case "$1" in
     ;;
 esac
 
+# allow the user to access /dev/std{out,err}
+chown nobody:nogroup /dev/std{out,err}
+
 ## Run!
 
 chown -R nobody:nogroup .
--- a/test/test_git_deploy.go
+++ b/test/test_git_deploy.go
@@ -16,9 +16,11 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/docker/docker/pkg/term"
 	ct "github.com/flynn/flynn/controller/types"
 	"github.com/flynn/flynn/pkg/attempt"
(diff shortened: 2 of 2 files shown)


lmars:
@josephglanville yeh I already did a hack for Docker images in #3371, but they generally run as root, so another hack needed for non-root slugs 😛

PR #3542, opened 2016-10-20
3532 cli enhancements

### LINKED ISSUE #3532: Allow `flynn ps kill` to receive a list of jobs IDs
Right now `flynn ps kill` only receives one job ID. Since the command takes some seconds to finish, It would be great to allow `flynn ps kill` to receive a list of jobs IDs like `docker ps kill id1 id2 id3 ...`, that way the user won't have to wait a lot of time if he/she needs to kill multiple  jobs.

If it's ok I can work on the implementation.

simon0191:
- Resolves #3532 
- Add --quiet flag to `flynn ps`
- Add --type flag to `flynn ps`
- Let `flynn kill` receive multiple job IDs

### DIFF
--- a/cli/kill.go
+++ b/cli/kill.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"log"
 
 	"github.com/flynn/flynn/controller/client"
@@ -9,16 +10,23 @@ import (
 
 func init() {
 	register("kill", runKill, `
-usage: flynn kill <job>
+usage: flynn kill <job>...
 
-Kill a job.`)
+Kill running jobs.`)
 }
 
 func runKill(args *docopt.Args, client controller.Client) error {
-	job := args.String["<job>"]
-	if err := client.DeleteJob(mustApp(), job); err != nil {
-		return err
+	success := true
+	for _, job := range args.All["<job>"].([]string) {
+		if err := client.DeleteJob(mustApp(), job); err != nil {
+			success = false
+			log.Printf("ERROR: could not kill job %s: %s\n", job, err)
+			continue
+		}
+		log.Printf("Job %s killed.", job)
 	}
-	log.Printf("Job %s killed.", job)
+	if !success {
+		return errors.New("Could not kill all jobs.")
+	}
 	return nil
 }
--- a/cli/ps.go
+++ b/cli/ps.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"sort"
 	"strings"
 	"time"
@@ -13,13 +14,14 @@ import (
 
 func init() {
 	register("ps", runPs, `
-usage: flynn ps [-a] [-c]
+usage: flynn ps [-a] [-c] [-q]
 
 List flynn jobs.
 
 Options:
   -a, --all      Show all jobs (default is running and pending)
   -c, --command  Show command
+	-q, --quiet    Only display IDs
 
 Example:
 
@@ -35,6 +37,12 @@ Example:
        host0-205595d8-206a-46a2-be30-2e98f53df272  web   up     About a minute ago  cf39a906-38d1-4393-a6b1-8ad2befe842	  /runner/init start web
        host0-0f34548b-72fa-41fe-a425-abc4ac6a3857  web   up     About a minute ago  cf39a906-38d1-4393-a6b1-8ad2befe842	  /runner/init start web
        host0-129b821f-3195-4b3b-b04b-669196cfbb03  run   down   5 seconds ago       cf39a906-38d1-4393-a6b1-8ad2befe842	  /runner/init /bin/bash
+
+			 $ flynn ps --all --quiet
+       host0-52aedfbf-e613-40f2-941a-d832d10fc400
+       host0-205595d8-206a-46a2-be30-2e98f53df272
+       host0-0f34548b-72fa-41fe-a425-abc4ac6a3857
+       host0-129b821f-3195-4b3b-b04b-669196cfbb03
 `)
 }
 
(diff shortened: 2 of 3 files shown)


imjorge:
@simon0191 what do you think about using a job predicate. Basically, an inline function that receives a job and says if it should be printed or not. The job predicate can encode all the conditions (only running jobs or all, given type, ...). Thus we could use the simple linear algorithm the code had instead of the jumps between functions and required pre-steps.

imjorge:
@simon0191 also not sure if you need to check both short and long args with go-docopt. I think the longer suffice and the short are not even on the `Bool` map? worth the investigation.

cli/ps.go
```diff
@@ -43,39 +56,64 @@ func runPs(args *docopt.Args, client controller.Client) error {
 	if err != nil {
 		return err
 	}
+
+	jobs = prepareAndFilterJobs(jobs, args.Bool["--all"] || args.Bool["-a"], args.String["<type>"])
```
josephglanville: You only need to check the long form flag, i.e: `--all`

cli/ps.go
```diff
@@ -43,39 +56,64 @@ func runPs(args *docopt.Args, client controller.Client) error {
...
 	sort.Sort(sortJobs(jobs))
 
+	if args.Bool["--quiet"] || args.Bool["-q"] {
+		printJobsQuiet(jobs)
+	} else {
+		printJobs(jobs, args.Bool["--command"] || args.Bool["-c"])
```
josephglanville: As above.

cli/kill.go
```diff
@@ -9,16 +10,23 @@ import (
 
 func init() {
 	register("kill", runKill, `
-usage: flynn kill <job>
+usage: flynn kill ID...
```
lmars: Keep the angle brackets for consistency with other usage (i.e. `<job>...`)

cli/kill.go
```diff
@@ -9,16 +10,23 @@ import (
...
 
 func runKill(args *docopt.Args, client controller.Client) error {
-	job := args.String["<job>"]
-	if err := client.DeleteJob(mustApp(), job); err != nil {
-		return err
+	success := true
+	for _, id := range args.All["ID"].([]string) {
+		if err := client.DeleteJob(mustApp(), id); err != nil {
+			success = false
+			log.Printf("could not kill job %s: %s\n", id, err)
```
lmars: Prefix this log message with `ERROR: ` so it is more obvious to the user that the job did not get killed.

cli/ps.go
```diff
@@ -43,39 +56,64 @@ func runPs(args *docopt.Args, client controller.Client) error {
...
-		if j.Type == "" {
-			j.Type = "run"
-		}
-		if !args.Bool["--all"] && j.State != ct.JobStateUp && j.State != ct.JobStatePending {
-			continue
-		}
-		id := j.ID
-		if id == "" {
-			id = j.UUID
-		}
```
lmars: Rather than splitting things out into separate functions, it would be clearer just to do the filtering in the existing loop:

```go
if args.String["--type"] != "" && j.Type != args.String["--type"] {
	continue
}
if args.Bool["--quiet"] {
	fmt.Println(id)
	continue
}
```

simon0191:
- Now I'm only checking the long flag of docopt in both `flynn kill` and `flynn ps`
- Replaced `ID...` with `<job>...` for consistency in `flynn kill`
- Add ERROR prefix to error message in `flynn kill`

- Regarding *"Rather than splitting things out into separate functions..."* and *"...we could use the simple linear algorithm the code had instead of the jumps between functions..."* here's the version with everything in one function and below my thoughts:

```
func runPs(args *docopt.Args, client controller.Client) error {
	jobs, err := client.JobList(mustApp())
	if err != nil {
		return err
	}
	sort.Sort(sortJobs(jobs))
	w := tabWriter()
	defer w.Flush()
	headers := []interface{}{"ID", "TYPE", "STATE", "CREATED", "RELEASE"}
	if args.Bool["--command"] {
		headers = append(headers, "COMMAND")
	}
	listRec(w, headers...)
	for _, j := range jobs {
		if !args.Bool["--all"] && j.State != ct.JobStateUp && j.State != ct.JobStatePending {
			continue
		}
		if j.Type == "" {
			j.Type = "run"
		}
		if args.String["<type>"] != "" && j.Type != args.String["<type>"] {
			continue
		}
		id := j.ID
		if id == "" {
			id = j.UUID
		}
		if args.Bool["--quiet"] {
			fmt.Println(id)
			continue
		}
		var created string
		if j.CreatedAt != nil {
			created = units.HumanDuration(time.Now().UTC().Sub(*j.CreatedAt)) + " ago"
		}
		fields := []interface{}{j.ID, j.Type, j.State, created, j.ReleaseID}
		if args.Bool["--command"] {
			fields = append(fields, strings.Join(j.Args, " "))
		}
		listRec(w, fields...)
	}
	return nil
}
```



  - With the *"just one function"* implementation filtering logic, printing logic and job preparation logic (fixing type and ID) is mixed up.  
  - With the *"just one function"* implementation we are doing 3 things that are not needed if the `--all` flag is on:

```
	w := tabWriter()
	defer w.Flush()
	headers := []interface{}{"ID", "TYPE", "STATE", "CREATED", "RELEASE"}
```
  - With the current implementation, `prepareAndFilterJobs` function will decrease jobs slice size
   and `sort` and further functions will process less jobs.
  - With the current implementation it's easier to read the code and to make modifications
  because the logic is splited in functions with different responsibilities.


So, What do you think?
I have the *"just one function"* commit ready to be pushed in case you still want that implementation.

lmars:
@simon0191 lets go with having all the filtering in the existing loop.

simon0191:
@lmars, Done

cli/ps.go
```diff
@@ -44,37 +57,41 @@ func runPs(args *docopt.Args, client controller.Client) error {
...
+		if args.Bool["--quiet"] {
+			fmt.Println(id)
+			continue
+		}
 		var created string
 		if j.CreatedAt != nil {
 			created = units.HumanDuration(time.Now().UTC().Sub(*j.CreatedAt)) + " ago"
 		}
-		fields := []interface{}{id, j.Type, j.State, created, j.ReleaseID}
+		fields := []interface{}{j.ID, j.Type, j.State, created, j.ReleaseID}
```
lmars: This should still use `id` rather than `job.ID`

simon0191:
@lmars Some tests failed but I think that they don't have to do with the changes I did:

```
15:22:43.418 FAIL: test_release.go:106: ReleaseSuite.TestReleaseImages
15:36:33.397 FAIL: test_scheduler.go:638: SchedulerSuite.TestDeployController
```

lmars:
Thanks!

PR #3544, opened 2016-10-20
host: Log OOM kills to the job's init log

### LINKED ISSUE #3375: Log OOM kills to the containerinit log
When the OOM killer kills a task, it is currently only logged to dmesg. There is a notification system that allows a userspace process to get a notification, and we should hook it up in containerinit so that OOM events are visible (in concert with #2669).

https://github.com/torvalds/linux/blob/9395452b4aab7bc2475ef8935b4a4fb99d778d70/Documentation/cgroup-v1/memory.txt#L729
https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Resource_Management_Guide/sec-Using_the_Notification_API.html

lmars:
As well as logging the OOM notifications to the init log, I have introduced a `MsgID` type which encapsulates the "IDx" message IDs used in raw syslog messages, plus a cleanup of the unused `ActiveJob.ContainerID`.

Closes #3375.

### DIFF
--- a/host/libcontainer_backend.go
+++ b/host/libcontainer_backend.go
@@ -145,14 +145,16 @@ type LibcontainerBackend struct {
 }
 
 type Container struct {
-	ID       string `json:"id"`
-	RootPath string `json:"root_path"`
-	IP       net.IP `json:"ip"`
+	ID        string         `json:"id"`
+	RootPath  string         `json:"root_path"`
+	IP        net.IP         `json:"ip"`
+	MuxConfig *logmux.Config `json:"mux_config"`
 
 	container libcontainer.Container
 	job       *host.Job
 	l         *LibcontainerBackend
 	done      chan struct{}
+
 	*containerinit.Client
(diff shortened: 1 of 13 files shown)


lmars:
I'm wondering how helpful this is as implemented. Most users won't know to pass the `--init` flag to `flynn log` so will miss this information.

I'm thinking of adding a separate stream (e.g. `alert`) which would always appear in `flynn log`, and perhaps colourise it in terminals.

Alternatively we could always include the init log in `flynn log` but perhaps support log level filters, so for example the critical OOM alert would always appear but info logs like "starting the job", "adding healthcheck" etc. would only appear with an explicit `--init-level=info`.

@titanous @josephglanville @jvatic thoughts?

titanous:
I think this is sufficient as a first pass, we should probably consider all of the log lines and what should be exposed by default but I don't think we need to do that now.

PR #3406, opened 2016-09-20
blobstore: merge blobstore commands into single binary

imjorge:
Closes #3353 

Any documentation needing update?

/cc @titanous @lmars

PR #3548, opened 2016-10-21
Fix syntax for blobstore migrate command in production docs

mfclarke:
Ran into this while migrating my blobs to s3. Was just missing the ```-```.


lmars:
@mfclarke what version of Flynn are you using? The latest version just has a single `flynn-blobstore` binary (see #3406).

We should leave the command as it is but add a note underneath that old clusters should use `/bin/flynn-blobstore-migrate`.

lmars:
Thanks!

PR #3554, opened 2016-10-22
Document the flynn-host fix command

nicatronTg:
While attempting to recover from every node in my flynn cluster collapsing, I needed some way to try to automatically recover the cluster. I had read a few issues on here before starting and happened upon `flynn-host fix` which did indeed recover the cluster for me, but it's conspicuously absent from the docs. I only found it referenced [on a blog post](https://flynn.io/blog/2015-11-16-this-week-in-flynn) and in issues on here -- never on the actual documentation.

Whether or not this is the right spot for it is up for debate, but at least right now, it seems like this should be noted _somewhere_. If this command isn't intended to be used, that's also a fair note, but it did solve most of my cluster problems, which I think makes its value worthy of being documented.

### DIFF
--- a/docs/content/production.html.md
+++ b/docs/content/production.html.md
@@ -347,6 +347,20 @@ The `controller`, `router`, and `blobstore` components store data in
 a PostgreSQL cluster managed by Flynn. Their databases may be accessed by
 running `flynn -a $APP_NAME pg psql`.
 
+### Recovery
+
+In the event that a cluster is having difficulties with apps, either internal
+or external, not starting after a cluster reboots, it may be necessary to run 
+the automated cluster repair tool. 
+
+Before attempting recovery, [perform a backup](#cluster-backup) and [collect the logs](#retrieving-logs).
+Send the logs to [logs@flynn.io](mailto:logs@flynn.io) to debug the cluster
+and help determine what caused it to fail.
+
+Next, run `flynn-host fix --min-hosts=N` from any node in the cluster to start
+an automated cluster health check which can fix several common problems, where
+`N` is the minimum number of expected hosts in the cluster.
+
 ## Updating
 
 There are two ways to update Flynn: in-place and backup/restore. The in-place


docs/content/production.html.md
```diff
@@ -273,6 +273,13 @@ from the exported app. If you'd like to provide a new name for the app, the
 app name and the cluster domain. To import the old routes in addition to the new
 route, add the `--routes` flag.
 
+### Recovery
+
+In the event that a cluster is having difficulties with apps, either internal
+or external, not starting after a cluster reboots, it may be necessary to run 
+the automated cluster repair tool. Run `flynn-host fix` from any node in the cluster
+to start an automated cluster health check which can fix several common problems.
+
```
lmars: This belongs under the "Debugging" section.

docs/content/production.html.md
```diff
@@ -273,6 +273,13 @@ from the exported app. If you'd like to provide a new name for the app, the
 app name and the cluster domain. To import the old routes in addition to the new
 route, add the `--routes` flag.
 
+### Recovery
+
+In the event that a cluster is having difficulties with apps, either internal
+or external, not starting after a cluster reboots, it may be necessary to run 
+the automated cluster repair tool. Run `flynn-host fix` from any node in the cluster
```
lmars: Before running `flynn-host fix`, we should advise users to try and create a backup first if possible, and also post the gist created by `sudo flynn-host collect-debug-info` in a GitHub issue (or send to logs@flynn.io) so we can diagnose what caused the cluster to enter the failing state.

nicatronTg:
@lmars moved to the debugging section per your request and added a note about collecting logs & backing up. I'm not 100% on relative anchor links in markdown with how this is formatted, though -- so can you let me know if those need to change to be something else (or removed entirely?)?

docs/content/production.html.md
```diff
@@ -347,6 +347,19 @@ The `controller`, `router`, and `blobstore` components store data in
 a PostgreSQL cluster managed by Flynn. Their databases may be accessed by
 running `flynn -a $APP_NAME pg psql`.
 
+### Recovery
+
+In the event that a cluster is having difficulties with apps, either internal
+or external, not starting after a cluster reboots, it may be necessary to run 
+the automated cluster repair tool. 
+
+Before attempting recovery, [perform a backup](#backup/restore) and [collect the logs](#retrieving-logs).
```
lmars: "perform a backup" should link to `#cluster-backup`.

docs/content/production.html.md
```diff
@@ -347,6 +347,19 @@ The `controller`, `router`, and `blobstore` components store data in
...
+Send the logs to [logs@flynn.io](mailto:logs@flynn.io) to debug the cluster
+and help determine what caused it to fail.
+
+Next, run `flynn-host fix` from any node in the cluster to start an automated
```
lmars: We should advise users to specify `--min-hosts=N` (the minimum expected number of hosts)

nicatronTg:
@lmars okay, I've added the minimum hosts arg and I've changed the anchor link to point to the cluster-backup section instead of backup/restore.

nicatronTg:
@lmars for the record, I sent this PR because I felt the documentation was needed on the website. If you plan to implement it or have already added it, then feel free to close this PR. If not, do you need additional changes before this can be merged?

PR #3559, opened 2016-10-23
host/cli: Collect Sirenia metadata in debug info

josephglanville:
Can be helpful in determining state of Sirenia clusters quickly without having to trawl through lots of logs.

### DIFF
--- a/host/cli/collect-debug-info.go
+++ b/host/cli/collect-debug-info.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


host/cli/collect-debug-info.go
```diff
@@ -182,3 +189,29 @@ func captureSchedulerState(gist *Gist) error {
...
+			State: &state.State{},
+		}
+		if len(meta.Data) > 0 {
+			err := json.Unmarshal(meta.Data, dstate.State)
+			if err != nil {
+				return err
+			}
+			body, err := json.MarshalIndent(dstate, "", "    ")
+			if err != nil {
+				return err
```
lmars: We should just continue on these errors so we try and get the other states (maybe log the error).

PR #3585, opened 2016-10-27
test: Update mariadb apt key

lmars:
Same change as #3530.


titanous:
This needs to be applied to the Vagrantfile as well and we should switch back to long key ids everywhere.

PR #3598, opened 2016-10-31
controller: Delete data volumes when one-off jobs stop 

lmars:
I think eventually we want volume GC at a higher level than flynn-host, but for now this prevents leaking data volumes which we know are ephemeral.

### DIFF
--- a/test/test_host.go
+++ b/test/test_host.go
@@ -67,7 +67,7 @@ func (s *HostSuite) TestAddFailingJob(t *c.C) {
 	t.Assert(h.AddJob(job), c.IsNil)
 
 	// check we get a create then error event
-	actual := make(map[string]*host.Event, 2)
+	actual := make(map[host.JobEventType]*host.Event, 2)
 loop:
 	for {
 		select {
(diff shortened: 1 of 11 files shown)


titanous:
Should this have a test?

lmars:
@titanous test added

test/test_host.go
```diff
@@ -471,6 +473,41 @@ func (s *HostSuite) TestNotifyOOM(t *c.C) {
...
+		t.Assert(err, c.IsNil)
+
+		// run the job
+		cmd := exec.JobUsingCluster(s.clusterClient(t), exec.DockerImage(imageURIs["test-apps"]), job)
+		out, err := cmd.CombinedOutput()
+		t.Assert(err, c.IsNil)
+		t.Assert(string(out), c.Equals, "/foo\n")
+
+		// check if the volume was deleted or not
+		vol, err = h.GetVolume(vol.ID)
```
titanous: This may need to be wrapped in retries as I believe the cleanup is async.

lmars:
@titanous I added a "cleanup" job event

PR #3606, opened 2016-11-02
Log Sink support.

### LINKED ISSUE #1422: Support remote syslog servers
In our current environment, where we are running Docker containers directly, we use a sidekick container on each machine to ship syslog messages to a remote syslog server (Loggly in our case). After moving to Flynn, we would like to preserve that ability since the searching/visualization/alerting has become an integral part of our workflow.

It would be great if there was a way to hook into the logaggregator to process messages and send them to a remote syslog server. There was mention in the IRC room of log drains possibly being developed. That sounds like the type of thing that would permit this.

One simple solution that could be used in the meantime is to setup a file monitor on the file that the logaggregator dumps snapshots to, although I doubt that will scale very far.

### LINKED ISSUE #3335: Make on-disk logs conform to RFC 5424
Two part problem, but the parts feel closely related:

First, the core components are almost conforming to https://tools.ietf.org/html/rfc5424.html but with some bugginess that rsyslog doesn't like. https://tools.ietf.org/html/rfc5424.html#section-6.2.1 describes the PRI field that (according to @tianx2) prepends a zero where it shouldn't.

Second, having user apps conform to that same standard (by decorating each log line before it hits disk, and using the app name for the service field) would make processing messages on a per-application basis much easier.

josephglanville:
Adds support for connecting Flynn to external logging services.

This first pass adds support for a flexible SyslogSink backend that can connect to services that support RFC 6587 framing. Additionally users can specify a prefix that can be templated with variables on a per job basis. This can be used to provide token authentication or make use of metadata in the job for tagging etc.
It also supports rudimentary TLS using system trusted roots, custom roots and insecure modes to follow.

Closes #1422 
Closes #3335

### DIFF
--- a/controller/schema.go
+++ b/controller/schema.go
@@ -462,6 +462,19 @@ $$ LANGUAGE plpgsql`,
 		`CREATE INDEX ON releases (app_id) WHERE deleted_at IS NULL`,
 		`ALTER TABLE releases ADD CHECK (app_id IS NOT NULL OR deleted_at IS NOT NULL)`,
 	)
+	migrations.Add(28,
+		`CREATE TABLE sink_kinds (name text PRIMARY KEY)`,
+		`INSERT INTO sink_kinds (name) VALUES ('syslog')`,
+		`INSERT INTO event_types (name) VALUES ('sink'), ('sink_deletion')`,
+		`CREATE TABLE sinks (
+			sink_id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+			kind text NOT NULL REFERENCES sink_kinds,
+			config jsonb NOT NULL,
+			created_at timestamptz NOT NULL DEFAULT now(),
+			updated_at timestamptz NOT NULL DEFAULT now(),
+			deleted_at timestamptz
+		)`,
+	)
 }
 
 func migrateDB(db *postgres.DB) error {
--- a/controller/sink.go
+++ b/controller/sink.go
@@ -1,0 +1,273 @@
+package main
+
+import (
+	"encoding/json"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/flynn/flynn/controller/schema"
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/flynn/pkg/ctxhelper"
+	"github.com/flynn/flynn/pkg/httphelper"
+	"github.com/flynn/flynn/pkg/postgres"
+	"github.com/flynn/flynn/pkg/random"
+	"github.com/flynn/flynn/pkg/sse"
+	"github.com/jackc/pgx"
+	"golang.org/x/net/context"
+)
+
+type SinkRepo struct {
+	db *postgres.DB
+}
+
+func NewSinkRepo(db *postgres.DB) *SinkRepo {
+	return &SinkRepo{
+		db: db,
+	}
+}
+
+func (r *SinkRepo) Add(s *ct.Sink) error {
+	if s.ID == "" {
+		s.ID = random.UUID()
+	}
+	tx, err := r.db.Begin()
+	if err != nil {
+		return err
+	}
+	err = tx.QueryRow("sink_insert", s.ID, s.Kind, []byte(s.Config)).Scan(&s.CreatedAt, &s.UpdatedAt)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+	// create sink event
+	if err := createEvent(tx.Exec, &ct.Event{
+		AppID:      "",
+		ObjectID:   s.ID,
+		ObjectType: ct.EventTypeSink,
+	}, s); err != nil {
+		tx.Rollback()
+		return err
+	}
+	return tx.Commit()
+}
+
+func scanSinks(rows *pgx.Rows) ([]*ct.Sink, error) {
+	var sinks []*ct.Sink
+	for rows.Next() {
+		sink, err := scanSink(rows)
+		if err != nil {
+			rows.Close()
+			return nil, err
+		}
+		sinks = append(sinks, sink)
+	}
+	return sinks, rows.Err()
+}
+
+func scanSink(s postgres.Scanner) (*ct.Sink, error) {
+	sink := &ct.Sink{}
+	err := s.Scan(&sink.ID, &sink.Kind, &sink.Config, &sink.CreatedAt, &sink.UpdatedAt)
+	if err != nil {
+		if err == pgx.ErrNoRows {
+			err = ErrNotFound
+		}
+		return nil, err
+	}
+	return sink, err
+}
+
+func (r *SinkRepo) Get(id string) (*ct.Sink, error) {
+	row := r.db.QueryRow("sink_select", id)
+	return scanSink(row)
+}
+
+func (r *SinkRepo) List() ([]*ct.Sink, error) {
+	rows, err := r.db.Query("sink_list")
+	if err != nil {
+		return nil, err
+	}
+	return scanSinks(rows)
+}
+
+func (r *SinkRepo) ListSince(since time.Time) ([]*ct.Sink, error) {
+	rows, err := r.db.Query("sink_list_since", since)
+	if err != nil {
+		return nil, err
+	}
+	return scanSinks(rows)
+}
+
+func (r *SinkRepo) Remove(id string) error {
+	sink, err := r.Get(id)
+	if err != nil {
+		return err
+	}
+	tx, err := r.db.Begin()
+	if err != nil {
+		return err
+	}
+	err = tx.Exec("sink_delete", sink.ID)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+	// create sink remove event
+	if err := createEvent(tx.Exec, &ct.Event{
+		AppID:      "",
+		ObjectID:   sink.ID,
+		ObjectType: ct.EventTypeSinkDeletion,
+	}, sink); err != nil {
+		tx.Rollback()
+		return err
+	}
+	return tx.Commit()
+}
+
+// Create a new sink
+func (c *controllerAPI) CreateSink(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	var sink ct.Sink
+	if err := httphelper.DecodeJSON(req, &sink); err != nil {
+		respondWithError(w, err)
+		return
+	}
+
+	if err := schema.Validate(&sink); err != nil {
+		respondWithError(w, err)
+		return
+	}
+
+	if err := c.sinkRepo.Add(&sink); err != nil {
+		respondWithError(w, err)
+		return
+	}
+	httphelper.JSON(w, 200, &sink)
+}
+
+// Get a sink
+func (c *controllerAPI) GetSink(ctx context.Context, w http.ResponseWriter, req *http.Request) {
+	params, _ := ctxhelper.ParamsFromContext(ctx)
+
+	sink, err := c.sinkRepo.Get(params.ByName("sink_id"))
+	if err != nil {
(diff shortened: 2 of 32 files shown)


controller/sink.go
```diff
@@ -0,0 +1,273 @@
...
+func NewSinkRepo(db *postgres.DB) *SinkRepo {
+	return &SinkRepo{
+		db: db,
+	}
+}
+
+func (r *SinkRepo) Add(s *ct.Sink) error {
+	if s.ID == "" {
+		s.ID = random.UUID()
+	}
```
titanous: Is there a reason to not always generate a random ID?

host/logmux/sink.go
```diff
@@ -0,0 +1,604 @@
...
+func (s *LogAggregatorSink) Shutdown() {
+	s.shutdownOnce.Do(func() { close(s.shutdownCh) })
+}
+
+func (s *LogAggregatorSink) ShutdownCh() chan struct{} {
+	return s.shutdownCh
+}
+
+// SyslogSink is a flexible sink that can connect to TCP/TLS endpoints that use syslog framing.
+// The body of the message can be customised using a template.
```
titanous: s/body of the message/message prefix/

host/logmux/sink.go
```diff
@@ -0,0 +1,604 @@
...
+
+func NewSyslogSink(sm *SinkManager, info *SinkInfo) (sink *SyslogSink, err error) {
+	cfg := &ct.SyslogSinkConfig{}
+	if err := json.Unmarshal(info.Config, cfg); err != nil {
+		return nil, err
+	}
+	var t *template.Template
+	var cache *lru.Cache
+	if cfg.Prefix != "" {
+		// Initialse the template cache
```
titanous: Initialize

host/logmux/sink.go
```diff
@@ -0,0 +1,604 @@
...
+				prefix = p
+			}
+		}
+		// If not in the cache execute the template and cache the result
+		if len(prefix) == 0 {
+			jobID, _ := parseProcID(m.Message.ProcID)
+			job := s.sm.state.GetJob(jobID)
+			if job != nil && job.Job != nil {
+				var buf bytes.Buffer
+				if err := s.template.Execute(&buf, job.Job); err != nil {
```
titanous: Do we want to provide full access to the job? Should we instead have a struct with the information that we think is relevant?

host/logmux/sink.go
```diff
@@ -0,0 +1,604 @@
...
+	if s.template != nil {
+		// Lookup rendered template from cache
+		var prefix []byte
+		if cached, ok := s.cache.Get(m.Message.ProcID); ok {
+			if p, ok := cached.([]byte); ok {
+				prefix = p
+			}
+		}
+		// If not in the cache execute the template and cache the result
+		if len(prefix) == 0 {
```
titanous: This should probably check if the prefix is nil, as a template could return a zero-length result.

host/logmux/sink.go
```diff
@@ -0,0 +1,604 @@
...
+		return nil
+	}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (sm *SinkManager) persistSinks() {
+	sm.logger.Info("starting sink persistence routine")
+	ticker := time.NewTicker(60 * time.Second)
```
titanous: Should we persist a bit more frequently? Every 10 seconds?

host/logmux/sink.go
```diff
@@ -0,0 +1,604 @@
...
+		close(sm.shutdownCh)
+	})
+
+	sm.mtx.Lock()
+	defer sm.mtx.Unlock()
+
+	for _, s := range sm.sinks {
+		s.Shutdown()
+	}
+
```
titanous: Should we persist one last time before shutting down?

test/test_host.go
```diff
@@ -679,3 +681,94 @@ func (s *HostSuite) TestUpdateTags(t *c.C) {
...
+		AppID:     logApp.ID,
+		ReleaseID: release.ID,
+		Processes: map[string]int{"app": 1},
+	}), c.IsNil)
+	defer client.DeleteApp(logApp.ID)
+
+	// add sink to controller
+	config, err := json.Marshal(&ct.SyslogSinkConfig{
+		URL: fmt.Sprintf("tcp://%s.discoverd:%d", logApp.Name, syslogPort),
+	})
```
titanous: Should this test the prefix template? Also, the scheme is `tcp://` but I think it should be `syslog://`.

test/test_host.go
```diff
@@ -679,3 +681,94 @@ func (s *HostSuite) TestUpdateTags(t *c.C) {
...
+		case msg := <-msgs:
+			received[msg.HostID] = struct{}{}
+			if len(received) == len(hosts) {
+				return
+			}
+		case <-time.After(30 * time.Second):
+			t.Fatal("timed out waiting for log messages")
+		}
+	}
+}
```
titanous: This should test that a sink removal is propagated.

cli/log_sink.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/flynn/flynn/controller/client"
+	ct "github.com/flynn/flynn/controller/types"
+	"github.com/flynn/go-docopt"
+)
+
+func init() {
+	register("log-sink", runLogSink, `
+usage: flynn log-sink
+       flynn log-sink add syslog <url> [<prefix>]
+       flynn log-sink remove <id>
```
titanous: Should these commands be scoped to the cluster (under the `flynn cluster` subcommand or via a mandatory flag) to allow room for per-app sinks?

host/logmux/sink.go
```diff
@@ -0,0 +1,608 @@
...
+				prefix = buf.Bytes()
+				s.cache.Add(jobID, prefix)
+			}
+		}
+		// If the generated/cached prefix isn't 0 length then modify the message body
+		if len(prefix) != 0 {
+			m.Message.Msg = bytes.Join([][]byte{prefix, m.Message.Msg}, msgSep)
+		}
+	}
+	_, err := s.conn.Write(rfc6587.Bytes(m.Message))
```
titanous: Should we swap the app ID for the app name all the time or optionally?

cli/log_sink.go
josephglanville: Yeah that is probably a good idea. Per app sinks don't seem super useful but no need to rule out the possibility.

host/logmux/sink.go
josephglanville: I think making it available to the prefix formatter is probably the best idea. That way it can be used to populate tags etc for system that support structured parsing. Right now it is accessed via the job meta but replacing the job with something that just contains meta, app name, env is probably what I am going to do (actually have already done just haven't rebased/pushed yet).

titanous: I feel like most log aggregators will read the APP-NAME from the syslog message, and the ID isn't particularly useful externally.

josephglanville:
Ready for review.

host/logmux/sink.go
```diff
@@ -0,0 +1,640 @@
...
+		if err := sinkBucket.ForEach(func(k, v []byte) error {
+			sinkInfo := &SinkInfo{}
+			if err := json.Unmarshal(v, sinkInfo); err != nil {
+				return fmt.Errorf("failed to deserialize sink info: %s", err)
+			}
+			err := sm.addSink(string(k), sinkInfo, false)
+			if err != nil {
+				return err
+			}
+			return nil
```
lmars: `return sm.addSink(string(k), sinkInfo, false)`

host/logmux/sink.go
```diff
@@ -0,0 +1,640 @@
...
+	defer ticker.Stop()
+	for {
+		select {
+		case <-sm.shutdownCh:
+			return
+		case <-ticker.C:
+			sm.mtx.Lock()
+			for id := range sm.sinks {
+				if err := sm.persistSink(id); err != nil {
+					log.Warn("error persisting sink", "sink.id", id)
```
lmars: Include the error in the log message

host/logmux/sink.go
```diff
@@ -0,0 +1,640 @@
...
+	// Connect TCP connection to aggregator
+	conn, err := dialer.Retry.Dial("tcp", s.addr)
+	if err != nil {
+		return err
+	}
+	// Connect to logaggregator HTTP endpoint
+	host, _, _ := net.SplitHostPort(s.addr)
+	c, err := client.New("http://" + host)
+	if err != nil {
+		if conn != nil {
```
lmars: Isn't `conn` always non-nil here?

test/test_host.go
```diff
@@ -679,3 +683,140 @@ func (s *HostSuite) TestUpdateTags(t *c.C) {
...
+	}()
+
+	// deploy an omni job
+	t.Assert(client.PutFormation(&ct.Formation{
+		AppID:     app.ID,
+		ReleaseID: release.ID,
+		Processes: map[string]int{"omni": 1},
+	}), c.IsNil)
+
+	// wait for syslog message from each host
```
lmars: We should boot a 3 node cluster for this test so we know there are multiple hosts (and also prevents sending logs for other apps to the test aggregator). You'll need to use a cluster proxy to connect to the aggregator (see [here](https://github.com/flynn/flynn/blob/master/test/test_logaggregator.go#L81-L84)).

controller/schema.go
```diff
@@ -436,6 +436,19 @@ $$ LANGUAGE plpgsql`,
 		$$ LANGUAGE plpgsql`,
 		`CREATE TRIGGER check_artifact_manifest AFTER INSERT ON artifacts FOR EACH ROW EXECUTE PROCEDURE check_artifact_manifest()`,
 	)
+	migrations.Add(27,
+		`CREATE TABLE sink_kinds (name text PRIMARY KEY)`,
+		`INSERT INTO sink_kinds (name) VALUES ('syslog')`,
+		`INSERT INTO event_types (name) VALUES ('sink'), ('sink_deletion')`,
```
lmars: These need to be added to the [event schema](https://github.com/flynn/flynn/blob/master/schema/controller/event.json#L16-L36).

josephglanville:
@lmars comments addressed.

lmars:
:smile:

RobinDaugherty:
This is a great feature! But it's not clear to me whether this has been released. I just set up a new cluster with v20161218.1, and this is what I'm seeing:

```
# flynn-host log-sink add --help
usage: flynn-host log-sink list <host>

Commands:
    list    Display a list of sinks configured for a host

Examples:

    $ flynn-host log-sink list host1
```

Am I missing something?

lmars:
@RobinDaugherty you should add log sinks with the CLI via `flynn cluster log-sink add` rather than `flynn-host`, see:

```
$ flynn help cluster
...
    log-sink
        With no arguments, prints a list of registerd log-sinks for this cluster

    log-sink add syslog
        Creates a new syslog log sink with specified <url> and optionally <prefix> template.
        Supported schemes are syslog and syslog+tls

        options:
            --use-ids  Use app IDs instead of app names in the syslog APP-NAME field

        examples:
                        $ flynn cluster log-sink add syslog syslog+tls://rsyslog.host:514/

    log-sink remove
        Removes a log sink with <id>
```

So for example:

```
$ flynn cluster log-sink add syslog syslog+tls://rsyslog.host:514/
```

RobinDaugherty:
Thanks @lmars. I actually realized my mistake moments after submitting my comment, and I felt like a real fool. (I removed my comment right away, I was hoping nobody would see my foolishness. For posterity, I was running `flynn-host` on the Flynn node instead of the `flynn` CLI. `flynn-host` will list log sinks, but does not provide the functionality that `flynn cluster log-sink` provides.)

PR #3609, opened 2016-11-02
all: Release Flynn images

lmars:
Summary:

* `images.json` contains a map of name to Flynn image
* exporting images copies layers from `/var/lib/flynn/layer-cache` to `/VERSION/layers/ID.squashfs`, images to `/VERSION/images/ID.json` and `images.json` to `/VERSION/images.json.gz` in the TUF repo
* `flynn-host download` now opens the volume database and downloads Flynn images from the TUF repo directly into ZFS volumes
* the install script checks the kernel supports mounting OverlayFS with multiple lower directories (i.e. checks the kernel version is >= 3.19)


updater/types/types.go
```diff
@@ -31,8 +37,8 @@ var SystemApps = []SystemApp{
 	{Name: "controller"},
 	{Name: "logaggregator"},
 	{
-		Name:  "postgres",
-		Image: "flynn/postgresql",
+		Name:      "postgres",
+		ImageName: "postgresql",
```
titanous: Should we change the image name to match while we're here?

lmars: The image name is derived from the directory, so happy to move the code to `appliance/postgres`? Or we can just have an explicit `build-image --name postgres`.

titanous: Let's do the explicit flag, renaming will partially break git history.

lmars:
@titanous comments addressed

PR #3605, opened 2016-11-01
bootstrap: Convert slugs and Docker images on cluster restore 

lmars:
Some small updates plus the addition of slug and docker-receive migrators which convert to Flynn images on cluster restore.

The migrators connect directly to the controller database so that it can update the release artifacts in place and in a transaction, rather than creating new releases via the controller API.

I've also moved the various artifact update SQL to run after the controller has started so that we know the artifacts have the new manifest columns.

PR #3619, opened 2016-11-04
all: Switch to SquashFS + OverlayFS images

lmars:
This merges all the Flynn image changes into master:

#3381 - all: Add Flynn image manifests and imagebuilder
#3471 - host: Use SquashFS + OverlayFS via Mountspecs
#3481 - all: Add Flynn image artifacts
#3496 - Migrate slugs and docker-receive releases to Flynn images
#3605 - bootstrap: Convert slugs and Docker images on cluster restore
#3609 - all: Release Flynn images

Although there a lot of changes, they have already been reviewed so I have squashed into two commits to make this easier to revert if we need to.

Replaces #3100.


lmars:
:smile:

PR #3628, opened 2016-11-06
docs: Also rotate status app AUTH_KEY

josephglanville:
Must be updated in order to properly authenticate requests to the controller.

### DIFF
--- a/docs/content/production.html.md
+++ b/docs/content/production.html.md
@@ -453,6 +453,7 @@ To rotate an authentication key:
     flynn -a redis env set CONTROLLER_KEY=$NEW_KEY
     flynn -a mariadb env set CONTROLLER_KEY=$NEW_KEY
     flynn -a mongodb env set CONTROLLER_KEY=$NEW_KEY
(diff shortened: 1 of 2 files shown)


lmars:
Should we update [TestKeyRotation](https://github.com/flynn/flynn/blob/master/test/test_controller.go#L178-L212) too?

josephglanville:
Yep, looks like it should be updated to rotate keys for MongoDB/MariaDB too.

docs/content/production.html.md
```diff
@@ -453,6 +453,7 @@ To rotate an authentication key:
     flynn -a redis env set CONTROLLER_KEY=$NEW_KEY
     flynn -a mariadb env set CONTROLLER_KEY=$NEW_KEY
     flynn -a mongodb env set CONTROLLER_KEY=$NEW_KEY
+    flynn -a status env set AUTH_KEY=$NEW_KEY
```
titanous: The status `AUTH_KEY` is intentionally not the same as the controller key, as it is used in monitoring systems that don't need full access to Flynn. This line should not be added to the documentation or tested below.

PR #3579, opened 2016-10-26
vendor: Update github.com/docker/docker to v1.12.2

lmars:
This changes the on-disk layout for storing images, and will require us to update to Docker 1.12.2 in our dev envs, CI and the release machine.

Closes #2459 (and helps with #2789).

PR #3637, opened 2016-11-07
vendor: Update github.com/docker/docker to v1.12.2

lmars:
Another attempt at #3579 now that Flynn relies less on Docker at runtime.

This includes a WIP commit to upgrade Docker before building in CI, that commit needs to be dropped pre-merge.


lmars:
This is again proving difficult, this time it is the changes to the imagebuilder to load the AUFS layers which isn't working.

I am going to give up on this and instead just remain on Docker 1.9 for the foreseeable future. We know it works, and it is now just a build dependency. We will need to install it manually though when updating to 16.04 as there is no released xenial deb package.

PR #3639, opened 2016-11-08
bootstrap: Only migrate active slugs / Docker images

lmars:


titanous:
Can we migrate the current app releases as well? That would allow scaling up apps that are scaled down at the time of the backup.

lmars:
@titanous comments addressed

lmars:
Oh in fact I need to update `host/cli/bootstrap.go`...

lmars:
@titanous ok I updated the bootstrap code to just check if there are any legacy slugs / Docker images before running the relevant migrators.

PR #3644, opened 2016-11-08
host: Prevent deadlock when restoring backend state

lmars:


host/libcontainer_backend.go
```diff
@@ -1404,7 +1404,11 @@ func (l *LibcontainerBackend) UnmarshalState(jobs map[string]*host.ActiveJob, jo
 		if state.NetworkConfig != nil && state.NetworkConfig.JobID != "" {
 			if _, ok := readySignals[state.NetworkConfig.JobID]; ok {
 				log.Info("using stored network config", "job.id", state.NetworkConfig.JobID)
-				l.host.ConfigureNetworking(state.NetworkConfig)
+
+				// run ConfigureNetworking in a goroutine to avoid deadlock
+				// between state.Restore and PersistGlobalState which both
+				// access the state database
+				go l.host.ConfigureNetworking(state.NetworkConfig)
```
titanous: Is there anything that assumes this will be complete once this method returns?

lmars: I've taken a close look and I'm pretty sure nothing expects these calls to be completed.

PR #3658, opened 2016-11-10
dashboard: Mark deleted releases as deleted in app history

### LINKED ISSUE #3005: dashboard: Deleted releases still appear in the history of an application
Hi,
yesterday, I deleted some releases of an application via the CLI. The CLI doesn't show them anymore, so I assume it worked. But the dashboard still shows them.

jvatic:
Closes #3005

screen shot 2016-11-10 at 12 42 41 pm


jvatic:
There's a few cases in which this will break. I don't like having a request for each release event to see if it's deleted. Another option would be to updated the events table so the release object on an `app_release` event has a `deleted_at` field when the release is deleted.

titanous:
Can we just hide deleted releases? An event change sounds good.

PR #3667, opened 2016-11-13
scheduler: Watch service events and mark formations as complete

### LINKED ISSUE #2043: scheduler: Create event once formation becomes fully active
The scheduler should create an event (e.g. `deploy_complete`) once a formation is finished scaling.

### LINKED ISSUE #2705: scheduler: Panic when stopping surplus omni jobs.
```
t=2016-04-16T13:29:59+0000 lvl=info msg="handling formation change" component=scheduler fn=HandleFormationChange app.id=288a2c03-585c-45bf-a42b-72aea39a660a release.id=05336755-107d-484a-93f9-95cd98e3eb80 processes=map[app:1]
t=2016-04-16T13:29:59+0000 lvl=info msg="handling formation change" component=scheduler fn=HandleFormationChange app.id=288a2c03-585c-45bf-a42b-72aea39a660a release.id=dbc6d65c-188f-4e87-a9ad-da5301a7358d processes=map[app:-1]
t=2016-04-16T13:29:59+0000 lvl=info msg="updating processes and tags of existing formation" component=scheduler fn=handleFormation app.id=288a2c03-585c-45bf-a42b-72aea39a660a release.id=dbc6d65c-188f-4e87-a9ad-da5301a7358d processes=map[app:-1] tags=map[]
t=2016-04-16T13:29:59+0000 lvl=info msg="scheduler loop exited" component=scheduler fn=Run
panic: runtime error: slice bounds out of range

goroutine 1 [running]:
main.(*Scheduler).stopSurplusOmniJobs(0xc208001200, 0xc208126d20)
    /home/jpg/src/flynn/controller/scheduler/scheduler.go:604 +0xbf4
main.(*Scheduler).RectifyFormation(0xc208001200, 0xc20812c6c0, 0x24, 0xc20812c720, 0x24)
    /home/jpg/src/flynn/controller/scheduler/scheduler.go:560 +0x3dc
main.(*Scheduler).HandleRectify(0xc208001200, 0x0, 0x0)
    /home/jpg/src/flynn/controller/scheduler/scheduler.go:541 +0xbf
main.(*Scheduler).Run(0xc208001200, 0x0, 0x0)
    /home/jpg/src/flynn/controller/scheduler/scheduler.go:366 +0xaef
main.main()
    /home/jpg/src/flynn/controller/scheduler/scheduler.go:158 +0xd33
```

### LINKED ISSUE #2895: App processes running old release after a successful update of environment variable

### LINKED ISSUE #3144: Fixing a broken initial push does not scale back up
When deploying broken code as an initial release, the app is scaled back down (as expected):

```
-----> Creating release...
=====> Scaling initial release to web=1
-----> Waiting for initial web job to start...
-----> WARN: scaling initial release down to web=0 due to error
remote: ERROR: Initial web job failed to start
```

Subsequent deploys the do not try to scale the app back up:

```
-----> Creating release...
=====> Application deployed
```

### LINKED ISSUE #3518: worker: Using job / service events during deployments is error prone
When deploying a release, the worker initiates scale requests and then waits for job / service events to determine when that scale request has been fulfilled.

This tends to cause issues like #3509 where the worker detects a release state which is different to the current scale (e.g. it sees `web=1` using discoverd but the current scale is `web=2`, perhaps the `web` job is in the process of starting). The worker then uses this incorrect state when calculating the expected scale and ends up waiting for an incorrect number of events, thus causing the deploy to fail.

As a concrete example:
- actual scale is `scheduler=3 web=2 worker=2`
- worker calculates scale as `scheduler=3 web=1 worker=2` (`web` process is still starting)
- worker wants to stop the scheduler so sets its scale to zero, leaving other types the same
- worker requests scale `scheduler=0 web=1 worker=2`
- the actual diff is now `scheduler=-3 web=-1 worker=0`
- whilst waiting for 3 scheduler down events, worker also receives web down event, which it ignores
- worker later times out waiting for the web down event it has already received and ignored

This is a bug in the worker which can be fixed, but I feel like the worker should not be waiting for job and service events to determine when a scale request has been fulfilled, it should just make scale requests and wait for subsequent events indicating the scale has happened.

This would involve updating the scheduler to emit such scale events (#2043), but it seems that is the right place to be doing this, and would simplify other workflows (e.g. `flynn scale`).

### LINKED ISSUE #3621: cli: 'flynn scale' should wait for service events
Running `flynn scale` just waits for job events, but when scaling jobs which expose services this doesn't indicate that the job is actually ready to serve requests, evident in the following [CI failure](https://ci.flynn.io/builds/20161104075439-b61baf77):

```
++ 08:24:16.901 /home/ubuntu/go/src/github.com/flynn/flynn/test/../cli/bin/flynn -a gitreceive scale app=1
scaling app: 0=>1

08:24:17.060 ==> app 8522eaef-1d1d-4500-acdb-65de44d472a8 pending
08:24:17.139 ==> app 998ef605-8522eaef-1d1d-4500-acdb-65de44d472a8 starting
08:24:17.478 ==> app 998ef605-8522eaef-1d1d-4500-acdb-65de44d472a8 up

scale completed in 492.380781ms

++ 08:24:17.480 /usr/bin/git commit -m bump --allow-empty
[master 3162126] bump

++ 08:24:17.487 /usr/bin/git push flynn master --progress
fatal: unable to access 'https://git.flynn-e42891ff817ff9ec.local/crows-revolt-barrie.git/': The requested URL returned error: 503
```

I am actually wondering if the scheduler should only mark a job as `up` once the relevant service has started (rather than just when the job is marked as running by the host).

This would mean the job would remain in the `starting` state until the service starts, which I think makes sense. This would also simplify deployments (just wait for job events rather than job and/or service events, see #3518).

@titanous @josephglanville @jvatic wdyt?

### LINKED ISSUE #3699: Host tags not working
I am setting up a cluster to have higher powered hardware that will run the flynn cluster databases. I set a host tag as such `flynn-host tags set flynnhost private=true`. Then ran a scale `flynn -a postgres scale postgres=3,private=true web=2`, that output `persisting tag change`. When I ran `flynn -a postgres limit set postgres memory=6GB` it started the postgres on a server not tagged with `private=true`.

lmars:
Summary:

* the scheduler now waits for service events for any process types with a configured service before marking jobs as running
* if the service is being routed to, the scheduler waits for all the routers to emit a backend up event before marking the job as running (with a 10s timeout after first seeing the service)
* formations now have a state which is either `pending` or `complete`, with any calls to `PutFormation` marking the formation as `pending`, and the scheduler marking as `complete` once it determines the correct number of jobs are running for that formation
* a scale event is emitted once a formation is marked as complete, meaning clients no longer need to calculate how many job / service events to wait for when scaling a formation, instead just waiting for the `complete` event (I have added a `Scale` method to the controller client which does exactly this)
* lots of code has been simplified by using scale events rather than job / service events (mainly the deploy worker)

Fixes #3621 
Fixes #3518
Fixes #2043
Fixes #3144
Fixes #2895
Fixes #2705
Fixes #3699

### DIFF
--- a/controller/client/v1/client.go
+++ b/controller/client/v1/client.go
@@ -284,6 +284,107 @@ func (c *Client) DeleteResource(providerID, resourceID string) (*ct.Resource, er
 	return res, err
 }
 
+func (c *Client) PutScaleRequest(req *ct.ScaleRequest) error {
+	if req.AppID == "" || req.ReleaseID == "" {
+		return errors.New("controller: missing app id and/or release id")
+	}
+	return c.Put(fmt.Sprintf("/apps/%s/scale/%s", req.AppID, req.ReleaseID), req, req)
+}
+
+func (c *Client) ScaleAppRelease(appID, releaseID string, opts ct.ScaleOptions) error {
+	if opts.Processes == nil && opts.Tags == nil {
+		return errors.New("controller: missing processes or tags")
+	}
+	if opts.Timeout == nil {
+		opts.Timeout = &ct.DefaultScaleTimeout
+	}
+
+	var events chan *ct.Event
+	var stream stream.Stream
+	if !opts.NoWait {
+		events = make(chan *ct.Event)
+		var err error
+		stream, err = c.StreamEvents(ct.StreamEventsOptions{
+			AppID: appID,
+			ObjectTypes: []ct.EventType{
+				ct.EventTypeJob,
+				ct.EventTypeScaleRequest,
+			},
+		}, events)
+		if err != nil {
+			return err
+		}
+		defer stream.Close()
+	}
+
+	scaleReq := &ct.ScaleRequest{
+		AppID:     appID,
+		ReleaseID: releaseID,
+		State:     ct.ScaleRequestStatePending,
+	}
+	if opts.Processes != nil {
+		scaleReq.NewProcesses = &opts.Processes
+	}
+	if opts.Tags != nil {
+		scaleReq.NewTags = &opts.Tags
+	}
+	if err := c.PutScaleRequest(scaleReq); err != nil {
+		return err
+	}
+
+	if opts.ScaleRequestCallback != nil {
+		opts.ScaleRequestCallback(scaleReq)
+	}
+
+	if opts.NoWait {
+		return nil
+	}
+
+	timeout := time.After(*opts.Timeout)
+	for {
+		select {
+		case event, ok := <-events:
+			if !ok {
+				return fmt.Errorf("event stream closed unexpectedly: %s", stream.Err())
+			}
+			switch event.ObjectType {
+			case ct.EventTypeJob:
+				if opts.JobEventCallback == nil {
+					continue
+				}
+				var job ct.Job
+				if err := json.Unmarshal(event.Data, &job); err != nil {
+					continue
+				}
+				if job.ReleaseID != releaseID {
+					continue
+				}
+				if err := opts.JobEventCallback(&job); err != nil {
+					return err
+				}
+			case ct.EventTypeScaleRequest:
+				var req ct.ScaleRequest
+				if err := json.Unmarshal(event.Data, &req); err != nil {
+					continue
+				}
+				if req.ID != scaleReq.ID {
+					continue
+				}
+				switch req.State {
+				case ct.ScaleRequestStateCancelled:
+					return errors.New("scale request cancelled")
+				case ct.ScaleRequestStateComplete:
+					return nil
+				}
+			}
+		case <-opts.Stop:
+			return ct.ErrScalingStopped
+		case <-timeout:
+			return fmt.Errorf("timed out waiting for scale to complete (waited %.f seconds)", opts.Timeout.Seconds())
+		}
+	}
+}
+
 // PutFormation updates an existing formation.
 func (c *Client) PutFormation(formation *ct.Formation) error {
 	if formation.AppID == "" || formation.ReleaseID == "" {
--- a/controller/scheduler/job.go
+++ b/controller/scheduler/job.go
@@ -99,6 +99,8 @@ type Job struct {
 
 	// hostError is the error from the host if the job fails to start
 	hostError *string
+
+	serviceFirstSeen *time.Time
 }
 
 // Tags returns the tags for the job's process type from the formation
@@ -125,6 +127,13 @@ func (j *Job) Volumes() []ct.VolumeReq {
 		return []ct.VolumeReq{{Path: "/data"}}
 	}
 	return nil
+}
+
+func (j *Job) Service() string {
+	if j.Formation == nil {
+		return ""
+	}
+	return j.Formation.Release.Processes[j.Type].Service
 }
 
 func (j *Job) IsRunning() bool {
--- a/appliance/redis/cmd/flynn-redis-api/main.go
+++ b/appliance/redis/cmd/flynn-redis-api/main.go
@@ -225,24 +225,18 @@ func (h *Handler) servePostCluster(w http.ResponseWriter, req *http.Request, _ h
 		return
 	}
 
-	h.Logger.Info("put formation", "release.id", release.ID)
-	formation := &ct.Formation{
-		AppID:     app.ID,
-		ReleaseID: release.ID,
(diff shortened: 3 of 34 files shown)


titanous:
I think we need an intermediate "formation change" object (maybe deployment?) to track the state of the change. Toggling the singleton formation object status back and forth is vulnerable to races and there is no way to correlate the events.

lmars:
@titanous I was trying to prevent the races by only setting the formation's state to complete if the `UpdatedAt` field hadn't changed (so the scheduler only marks a formation as complete when it knows it responded to the most recent update), but I admit it isn't ideal.

How about this:

* Each call to `PutFormation` creates an object, I'll call it `ScaleRequest` (although I'm not sold on that name)
* Scale requests have three states: `pending` `cancelled` and `complete`, defaulting to `pending` on creation
* If a new scale request is made, all currently pending scale requests for the formation are marked as cancelled (so any clients waiting on those scale requests to complete are notified that their scale request is no longer valid)
* The scheduler listens for scale requests rather than formation updates (doing a periodic sync of active formations for anti-entropy) and marks them as `complete` once the appropriate number of jobs are running (the controller ignores complete events for already cancelled requests)
* Clients get back a `ScaleRequest` from `PutFormation` and wait for it to either complete or be cancelled

titanous:
@lmars Sounds good to me.

lmars:
Unless it has already been fixed, we should add a test covering #3699.

lmars:
My proposed changes are backwards incompatible with the current API (I want to return a `ct.ScaleRequest` from `PutFormation`, but it currently returns `ct.Formation`).

What is the plan for making these kinds of changes in a backwards compatible way? Do it based on a header value sent by the client?

/cc @jvatic @titanous

titanous:
The plan is to handle this in GraphQL. In this case make a new endpoint and try it first then fall back to the old endpoint if you get a 404.

lmars:
Ready for another review.

controller/client/v1/client.go
```diff
@@ -283,6 +283,102 @@ func (c *Client) DeleteResource(providerID, resourceID string) (*ct.Resource, er
...
+
+func (c *Client) ScaleAppRelease(appID, releaseID string, opts ct.ScaleOptions) error {
+	if opts.Processes == nil && opts.Tags == nil {
+		return errors.New("controller: missing processes or tags")
+	}
+	if opts.Timeout == nil {
+		opts.Timeout = &ct.DefaultScaleTimeout
+	}
+
+	events := make(chan *ct.Event)
```
imjorge: should we create any events + stream if `opts.NoWait` is true?

controller/scheduler/job.go
```diff
@@ -99,6 +99,8 @@ type Job struct {
 
 	// hostError is the error from the host if the job fails to start
 	hostError *string
+
+	serviceFirstSeen *time.Time
```
titanous: This doesn't appear to be used?

lmars: This is set [here](https://github.com/flynn/flynn/blob/scheduler-service-events/controller/scheduler/scheduler.go#L1476-L1479) and used [here](https://github.com/flynn/flynn/blob/scheduler-service-events/controller/scheduler/scheduler.go#L1492-L1496) to enforce a timeout from when the service is first seen to marking the job as running.

controller/client/v1/client.go
lmars: Good spot, updated in df38798.

lmars:
@titanous @imjorge comments addressed, ready for another review.

lmars:
@titanous comments addressed

PR #3675, opened 2016-11-14
docs: Update mongodb.md

wasi:
the current directory has to be specified. Otherwise a empty tar would be created.


lmars:
Thanks!

PR #3678, opened 2016-11-14
scheduler: Don't exit at startup whilst waiting for events

lmars:
Exiting may leave the cluster without a scheduler, it is better to just wait for the events to happen.

### DIFF
--- a/controller/scheduler/scheduler.go
+++ b/controller/scheduler/scheduler.go
@@ -161,13 +161,11 @@ func main() {
 
 	go s.startHTTPServer(os.Getenv("PORT"))
 
-	if err := s.Run(); err != nil {
-		shutdown.Fatal(err)
-	}
+	s.Run()
 	shutdown.Exit()
 }
 
-func (s *Scheduler) streamFormationEvents() error {
+func (s *Scheduler) streamFormationEvents() {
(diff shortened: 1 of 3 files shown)


controller/scheduler/scheduler.go
```diff
@@ -272,20 +273,20 @@ func (s *Scheduler) streamHostEvents() error {
...
 
-	select {
-	case <-current:
-		return nil
-	case <-time.After(30 * time.Second):
-		return errors.New("timed out waiting for current host list")
+	// wait until we have the current list of hosts and their
+	// jobs before starting the main scheduler loop
+	start := time.Now()
+	tick := time.Tick(30 * time.Second)
```
imjorge: won't we be leaking a Ticker, here and on all other cases?

lmars: The Ticker is only in scope whilst waiting for the event, once the event happens the function returns and the Ticker is garbage collected.

titanous: This is not actually correct, since the ticker is on a timer heap internally, you need to call `Stop` to remove it from the heap so that it can be garbage collected.

lmars: Ah ok, didn't know that, a bit surprising! Good spot @imjorge

PR #3691, opened 2016-11-16
bootstrap: Support migrating between singleton & HA clusters

lmars:


host/cli/bootstrap.go
```diff
@@ -331,8 +336,96 @@ UPDATE releases SET processes = jsonb_set(processes, '{app,volumes}', '[{"path":
 WHERE release_id IN (SELECT release_id FROM apps WHERE name = 'gitreceive');
 `)
 
+	// update the SINGLETON environment variable for database appliances
+	// (which includes updating legacy appliances which had SINGLETON set
+	// on the database type rather than the release)
+	singleton := strconv.FormatBool(cfg.Singleton)
+	data.Postgres.Release.Env["SINGLETON"] = singleton
+	sqlBuf.WriteString(fmt.Sprintf(`
+UPDATE releases SET env = jsonb_set(env, '{SINGLETON}', '%q')
```
titanous: I think this can be `%t` instead of `%q`

lmars: It needs to be a quoted string because the JSON is deserialized into `map[string]string` (rather than `map[string]bool`).

PR #3702, opened 2016-11-17
router: Add support for inbound PROXY protocol v1

titanous:
This adds support for version 1 of the PROXY protocol to indicate the source IP for incoming HTTP and HTTPS connections.

http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt

The feature is off by default, and can be enabled by setting the environment variable `PROXY_PROTOCOL=true` while running a non-backup `flynn-host bootstrap` or configuring the router directly after bootstrap with the command:

    flynn -a router env set PROXY_PROTOCOL=true

This feature is useful to integrate external load balancers in TCP mode like the AWS Classic Load Balancer (previously called ELB) and HAProxy.

https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html
https://cbonte.github.io/haproxy-dconv/1.6/configuration.html#5.2-send-proxy

References #68

/cc @temujin9 @sleighter @blurrcat

### DIFF
--- a/router/proxyproto/protocol.go
+++ b/router/proxyproto/protocol.go
@@ -1,0 +1,154 @@
+// Derived from https://github.com/bradfitz/go-proxyproto
+
+// Package proxyproto implements a net.Listener supporting HAProxy PROXY protocol.
+//
+// See http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt for details.
+package proxyproto
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+// prefix is the string we look for at the start of a connection
+// to check if this connection is using the PROXY protocol.
+var prefix = []byte("PROXY ")
+
+// Listener wraps an underlying listener whose connections may be using the
+// HAProxy PROXY Protocol (version 1). If the connection is using the protocol,
+// RemoteAddr will return the correct client address.
+type Listener struct {
+	net.Listener
+}
+
+// conn is used to wrap an underlying connection which may be speaking the PROXY
+// Protocol. If it is, the RemoteAddr() will return the address of the client
+// instead of the proxy address.
+type conn struct {
+	net.Conn
+	connBuf  []byte
+	srcAddr  *net.TCPAddr
+	initOnce sync.Once
+}
+
+// Accept waits for and returns the next connection to the listener.
+func (p Listener) Accept() (net.Conn, error) {
+	rawc, err := p.Listener.Accept()
+	if err != nil {
+		return nil, err
+	}
+	return newConn(rawc), nil
+}
+
+func newConn(c net.Conn) *conn {
+	return &conn{Conn: c}
+}
+
+// Read checks for the PROXY protocol header when doing the initial scan. If
+// there is an error parsing the header, it is returned and the socket is
+// closed.
+func (p *conn) Read(b []byte) (int, error) {
+	var err error
+	p.initOnce.Do(func() { err = p.checkPrefix() })
+	if err != nil {
+		return 0, err
+	}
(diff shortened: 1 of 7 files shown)


router/proxyproto/protocol.go
```diff
@@ -0,0 +1,154 @@
...
+var prefix = []byte("PROXY ")
+
+// Listener wraps an underlying listener whose connections may be using the
+// HAProxy PROXY Protocol (version 1). If the connection is using the protocol,
+// RemoteAddr will return the correct client address.
+type Listener struct {
+	net.Listener
+}
+
+// conn is used to wrap and underlying connection which may be speaking the
```
lmars: s/and/an/

router/proxyproto/protocol.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil, err
+	}
+	return newConn(rawc), nil
+}
+
+func newConn(c net.Conn) *conn {
+	return &conn{Conn: c}
+}
+
+// Read is check for the proxy protocol header when doing
```
lmars: s/is check/checks/

router/proxyproto/protocol.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// RemoteAddr returns the address of the client if the proxy protocol is being
+// used, otherwise just returns the address of the socket peer. If there is an
+// error parsing the header, the address of the client is not returned, and the
+// socket is closed. One implication of this is that the call could block if the
+// client is slow. Using a Deadline is recommended if this is called before
+// Read().
+func (p *conn) RemoteAddr() net.Addr {
+	p.initOnce.Do(func() { p.checkPrefix() })
```
lmars: Should there be an error check on `checkPrefix()`?

router/proxyproto/protocol.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// Read the header line
+	header, err := buf.ReadString('\n')
+	if err != nil {
+		p.Conn.Close()
+		return err
+	}
+
+	// Strip the carriage return and new line
+	header = header[:len(header)-2]
```
lmars: Should this check there is a carriage return to strip?

router/proxyproto/protocol.go
```diff
@@ -0,0 +1,154 @@
...
+	parts := strings.Split(header, " ")
+	if len(parts) != 6 {
+		p.Conn.Close()
+		return fmt.Errorf("proxyconn: invalid header line: %q", header)
+	}
+
+	// Verify the type is known
+	if parts[1] != "TCP4" && parts[1] != "TCP6" {
+		p.Conn.Close()
+		return fmt.Errorf("proxyconn: hnhandled address type: %q", parts[1])
```
lmars: s/hnhandled/unknown/

router/server.go
```diff
@@ -89,6 +89,11 @@ func main() {
 		shutdown.Fatal("Missing random 32 byte base64-encoded COOKIE_KEY")
 	}
 
+	var proxyProtocol bool
+	if pp := os.Getenv("PROXY_PROTOCOL"); pp != "" {
+		proxyProtocol = pp == "true"
+	}
```
lmars: Could this just be `proxyProtocol := os.Getenv("PROXY_PROTOCOL") == "true"`?

router/proxyproto/protocol.go
titanous: No, the carriage return will always be included for compliant implementations, anything else may break parsing but has no meaningful effect.

router/proxyproto/protocol.go
titanous: No, the method signature (defined in the `net.Conn` interface) does not have an error return parameter.

titanous:
Comments addressed.

PR #3718, opened 2016-11-21
controller: Associate releases with a single app

titanous:
Releases were previously related to apps indirectly via formations, but pretty much everywhere a release is created, there is an associated app that a formation is created for, so whilst associating a release with multiple apps via formations was possible, it was rarely done.

This commit adds an `app_id` column and makes it required if the release is not deleted. Old releases that can't reliably be associated with an app are marked as deleted.


titanous:
I refactored the handling of domain migrations, please take a quick look.

Issue #2970, opened 2016-06-21
Flynn CI Runner V2


lmars:
We currently have a CI runner which runs as a standalone service on a single bare metal VM. The runner builds and tests Flynn in Linux VMs running under KVM.

We should implement a new CI runner which runs in a Flynn cluster. It should be much more generic and support building and testing all types of software, not just Flynn.

What follows is a brain dump of some of my thoughts on this.

---
### Current runner implementation

Here is an overview of how the current CI runner builds and tests Flynn:

1) We have an [Upstart job](https://github.com/flynn/flynn/blob/v20160621.0/test/scripts/upstart.conf) which pulls the latest Flynn code, creates a base VM image with Flynn build dependencies then starts the runner. To create the image, we download [ubuntu-base](https://wiki.ubuntu.com/Core), chroot into it and run a setup script (see [build.sh](https://github.com/flynn/flynn/blob/v20160621.0/test/rootfs/build.sh)).

2) On startup, the runner creates a CoW image from the base image, starts a VM, then pulls the latest Flynn code and builds Flynn using [this script](https://github.com/flynn/flynn/blob/v20160621.0/test/cluster/cluster.go#L396-L438) via SSH. This CoW image serves as a cache so that triggered builds don't need to build Flynn from scratch.

3) When the runner receives a GitHub webhook, it creates a CoW image from the cached build image, fetches the commit referenced in the webhook and builds Flynn again. It then uses CoW images of this resulting image to boot VMs to form two Flynn clusters (`default` and `release`). It uses `start-stop-daemon` to start `flynn-host` on each VM (see [here](https://github.com/flynn/flynn/blob/v20160621.0/test/cluster/cluster.go#L474-L498)), then runs bootstrap on the first VM in the cluster (see [here](https://github.com/flynn/flynn/blob/v20160621.0/test/cluster/cluster.go#L511-L531)).

4) With a Flynn cluster up and running, the runner runs the `flynn-test` binary on the first VM in the `default` cluster using [this script](https://github.com/flynn/flynn/blob/v20160621.0/test/runner/runner.go#L301-L350).

5) As part of running the test suites, `flynn-test` interacts with the runner via HTTP to boot new VMs to tests things like scheduling jobs on new hosts and promoting hosts in service discovery.

---
### Proposed runner implementation

A sketchy outline of how I think a new runner could work:
- The runner runs in Flynn, has a database of builds (probably PostgreSQL) and exposes a simple UI for creating / viewing / editing builds
[...]

PR #3719, opened 2016-11-21
test: Support booting Flynn in Flynn!

lmars:
This adds support for booting Flynn clusters inside Flynn clusters :smile:.

Summary:

* the `test/cluster2` package (which will eventually replace `test/cluster`) boots Flynn clusters by creating a flynn-host app, scaling it up, then running the bootstrapper once the hosts have started
* flynn-test now has a `bootCluster` function which boots a cluster and returns an object which can be used to run commands against the cluster
* flynn-test now runs in a Flynn job, which means it can happily change `/etc/hosts` at will, and also use discoverd for DNS without [hacking /etc/resolv.conf](https://github.com/flynn/flynn/blob/v20161114.0p1/script/run-integration-tests#L92-L94)
* flynn-host needs extra Linux capabilities (`CAP_SYS_ADMIN` and `CAP_NET_ADMIN`), needs to access `/dev/zfs` and `/dev/zd*`, and needs write access to `cgroupfs`, so I've added support for customising these through `ct.ProcessType` and `host.ContainerConfig`
* I've added a xenial based `host` image which has the `zfsutils-linux` package and a custom udev rule to create ZFS device nodes in the host's mount namespace
* flynn-host jobs have `/var/lib/flynn` mounted from the host and create volumes in `/var/lib/flynn/JOB_ID/volumes` (`zfs` doesn't like the sparse vdev file being in the container's filesystem, it needs to be available at the same path both inside and outside the container)
* some disruptive tests have been updated to boot their own clusters

This is a first step in refactoring CI (see #2970). I have some grand plans, but one step at a time :wink:.

Lets make CI great again!

### DIFF
--- a/host/libcontainer_backend.go
+++ b/host/libcontainer_backend.go
@@ -269,8 +269,11 @@ func (l *LibcontainerBackend) ConfigureNetworking(config *host.NetworkConfig) er
 	}
 
 	// enable IP forwarding
-	if err := ioutil.WriteFile("/proc/sys/net/ipv4/ip_forward", []byte("1\n"), 0644); err != nil {
-		return err
+	ipFwd := "/proc/sys/net/ipv4/ip_forward"
+	if data, err := ioutil.ReadFile(ipFwd); err != nil && !bytes.HasPrefix(data, []byte("1")) {
+		if err := ioutil.WriteFile(ipFwd, []byte("1\n"), 0644); err != nil {
+			return err
+		}
 	}
 
 	// Set up iptables for outbound traffic masquerading from containers to the
--- a/host/libcontainer_backend.go
+++ b/host/libcontainer_backend.go
@@ -182,25 +182,6 @@ func writeContainerConfig(path string, c *containerinit.Config, envs ...map[stri
 	return json.NewEncoder(f).Encode(c)
 }
 
-func writeHostname(path, hostname string) error {
-	f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0644)
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-	pos, err := f.Seek(0, os.SEEK_END)
-	if err != nil {
-		return err
-	}
-	if pos > 0 {
-		if _, err := f.Write([]byte("\n")); err != nil {
-			return err
-		}
-	}
-	_, err = fmt.Fprintf(f, "127.0.0.1 localhost %s\n", hostname)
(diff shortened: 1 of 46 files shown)


host/libcontainer_backend.go
```diff
@@ -448,7 +453,7 @@ func (l *LibcontainerBackend) Run(job *host.Job, runConfig *RunConfig, rateLimit
 		Cgroups: &configs.Cgroup{
 			Path: filepath.Join("/flynn", job.Partition, job.ID),
 			Resources: &configs.Resources{
-				AllowedDevices: configs.DefaultAllowedDevices,
+				AllowedDevices: append(configs.DefaultAllowedDevices, job.Config.AllowedDevices...),
```
titanous: In order to allow restricting jobs further, it probably makes sense to use _only_ the provided devices and capabilities if specified so that the defaults can be overridden?

test/cluster2/cluster.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+func (c *Cluster) Destroy() error {
+	if err := c.config.Client.PutFormation(&ct.Formation{
+		AppID:     c.App.ID,
+		ReleaseID: c.Release.ID,
+		Processes: map[string]int{"host": 0},
+	}); err != nil {
+		return err
+	}
```
titanous: Doesn't the `DeleteApp` below do this?

lmars:
I have updated some more tests to boot their own clusters (1bae5e5) including ones which add hosts to existing clusters (f1a0227).

Booting multi-node clusters inside multi-node clusters doesn't seem to work (the Flannel overlay network packets don't seem to make it to the right place), but multi-node within single-node works fine. Given tests can boot multi-node clusters if needed, I have changed the default CI cluster to be single-node (see c305723).

The multi-node tests will fail here, but I have tested in custom CI so should pass once we merge / restart CI.

@titanous I have addressed your comments too.

Ready for a review.

lmars:
I changed the backups to use unique domain names in 9a18478 so we can run the backup tests in parallel but this has broken the tests on master which expect the domain to be `1.localflynn.com` for all the backups (see [here](https://github.com/flynn/flynn/blob/master/test/test_zz_backup.go#L137)).

The sooner we can merge this the better (the only failing tests are the ones which try and boot multi-node clusters, which will pass once we restart CI).

If there are any blockers for this though I'll revert the backups in CI.

PR #3722, opened 2016-11-23
docker-receive: Fix converting AUFS whiteouts

lmars:


titanous:
Is there any reasonable way around this? `CAP_SYS_ADMIN` is a massive sledgehammer.

lmars:
We could add the list of opaque directories to the layer configuration, and then insert a layer after the given layer at mount time which has the OverlayFS whiteout dir.

titanous:
Okay, let's do it this way for now and then we can sort it out when we have a Go package that can write out squashfs filesystems.

PR #3743, opened 2016-12-02
host/cli: Add --filename option to collect-debug-info

### LINKED ISSUE #3732: Define an output file for `flynn-host collect-debug-info`
I want to automate the process of collecting debug info from several hosts and uploading the resulting tarballs to an s3 bucket. I'd like to trigger this operation via an Ansible playbook. To do this, I need to know specifically where `collect-debug-info` writes its output. 

Currently the command uses a random temp file, making it difficult to automatically upload the file after creation. The location of the tarball would have to be parsed from the humanized output message from the command.

I have two ideas for providing this capability:
 1. Add a `--out-file` option to `flynn-host collect-debug-info` to define the output file. Could either overwrite an existing file or consider a `--overwrite` flag to force the operation.
 2. Add option to write the tarball to stdout (just like `flynn cluster backup`), so the output can be piped directly to an S3 file.

josephglanville:
Closes #3732

### DIFF
--- a/host/cli/collect-debug-info.go
+++ b/host/cli/collect-debug-info.go
@@ -47,6 +47,7 @@ Options:
   --tarball          Create a tarball instead of uploading to a gist
   --include-env      Include sensitive environment variables
   --log-dir=DIR      Path to the log directory [default: /var/log/flynn]
+  --filename=PATH    Path to write tarball, only valid if --tarball is specified
 
 Collect debug information into an anonymous gist or tarball`)
 }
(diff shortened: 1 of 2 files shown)


host/cli/collect-debug-info.go
```diff
@@ -108,7 +109,20 @@ func runCollectDebugInfo(args *docopt.Args) error {
...
+		path := args.String["--filename"]
+		if path == "" {
+			tmpDir, err := ioutil.TempDir("", "flynn-host-debug")
+			if err != nil {
+				return err
+			}
+			defer func() {
+				if err != nil {
+					os.RemoveAll(tmpDir)
+				}
```
lmars: Which `err` is this checking?

josephglanville:
Ready for review. I dropped removing the temp path on failures as we probably don't want that behavior.

PR #3690, opened 2016-11-15
Update manual installation docs for Ubuntu 16.04

michaelyork:
Signed-off-by: Michael York <michael@york.fm>

PR #3744, opened 2016-12-02
Missing "service" command on "service flynn-host st

aatishnn:
Missing "service" command on "start flynn-host" and "status flynn-host"


lmars:
These instructions are correct for Ubuntu 14.04 (see #3690 for updates for 16.04, which will be officially supported in the next stable release).

Issue #1865, opened 2015-09-13
Build System


benbjohnson:
## Overview ![status](https://img.shields.io/badge/status-draft-red.svg)

This is a proposal for a new build system in Flynn that will function as both the build tool for applications running in Flynn as well as the build tool for the Flynn project itself. There are other existing build tools but there are none (to my knowledge) that are written in Go and that can be easily integrated into the Flynn project's workflow.

There is currently no name for this project, however in the tradition of other `make`-like tools, it seems like `flake` (`flynn` + `make`) would be a reasonable choice.
## Goals

There are 3 goals of the build system:
1. Functional: Outputs from a package should only depend on its inputs. Given the same inputs, the output should always be the exact same.
2. Content Addressable: A file should be addressable in a particular state based on its contents and permissions. This means that output files can be cached in order to perform incremental builds. Changes to a file's contents should trigger a rebuild of dependencies.
3. Pluggable: Allow the build system to plug into a local or remote Flynn cluster to allow parallel builds and utilize a shared build cache.
### Dependency graph

Flynn's current build system is [`tup`](http://gittup.org/tup/). Tup creates a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph) of dependencies by using a [FUSE file system](http://fuse.sourceforge.net/) to monitor file access by subprocesses. This differs from traditional `make` in that input files do not need to be specified explicitly. The downside is that the FUSE filesystem can be finicky and requires set up. Another downside of tup is that its rule language is unfamiliar to developers coming from a build system like `make`.

More details of tup's dependency graph can be found in [Build System Rules & Algorithms](http://gittup.org/tup/build_system_rules_and_algorithms.pdf).

A similar approach to dependency graph management will be used in the new build system, however, we would like to avoid FUSE if possible.
#### FUSE Alternatives

Some alternatives to FUSE have been evaluated:
- [`seccomp(2)`](http://man7.org/linux/man-pages/man2/seccomp.2.html) - The Secure Computing state allows syscalls to be filtered and trapped. This would allow us to capture `read` calls from subprocesses within the build and rewrite them as needed. I spent some time trying to implement this but rewriting syscalls is a fairly obscure feature and I haven't had luck with it so far. I attempted to use [`libseccomp-golang`](https://github.com/seccomp/libseccomp-golang) but there's not much documentation so I didn't get far with it. `seccomp(2)` is also only available on later versions of Linux so this would not be compatible with Mac OS X or Windows.
[...]

PR #3753, opened 2016-12-09
builder: Build Flynn images in Flynn!

lmars:
This adds a builder component which runs a Flynn job and generates a SquashFS layer containing the diff, and uses this to build the Flynn component images.

This removes the dependency on Docker 1.9.1 (#2459) allowing us to upgrade more easily to Ubuntu 16.04 in dev / CI (Docker 1.9.1 is not in the xenial apt repository).

The builder has the following commands:

* `flynn-build run` - runs in a Flynn job, execs a command, creates a SquashFS layer of the diff (which is now mounted read-only at `/.container-diff`) and streams it to STDOUT
* `flynn-build layer` - runs a build job using a base artifact, mounting the current directory into the job at `/src` as a 9p filesystem, savng the resulting layer into `/var/lib/flynn/local/layers` and writing the layer JSON to STDOUT
* `flynn-build artifact` - appends a list of layers to an artifact and writes it to STDOUT

These commands are used by `tup` to build the component images:

* `builder/bootstrap` is used to build Ubuntu trusty / xenial base images as well as the builder image which contains `flynn-build` and `mksquashfs`
* once `flynn-host` has been compiled, it is started with IP `192.0.2.100` and is used to run build jobs
* each component has shell scripts like `img/${NAME}.sh` which are built with `flynn-build layer` and stored in `layer/${NAME}.json`
* `flynn-build artifact` is run after the layers have been built and the artifact is stored in the root `image` directory (reading the default entrypoint from `img/entrypoint.json`)

Some of the tup rules are a bit hairy (`dashboard` and `installer` in particular), but I feel like this is a stepping stone to a better suited build language which will build the binaries as well as the images (see #1865).

I've made the layer storage an interface so that it will be easy in future to run the build jobs in a remote Flynn cluster and upload the layers to a blobstore (I plan to do this for the new CI runner when building the VM rootfs).

There is some cleanup to do (e.g. delete pinkerton), but I'm putting this out there for review.


lmars:
I have updated `flynn-build layer` so that it mounts `/out` as a 9p filesystem so that the layer is written directly back to the builder (rather than previously creating it on a volume and then streaming it via STDOUT).

builder/layer.go
```diff
@@ -0,0 +1,272 @@
...
+func generateHash(name string, inputs ...string) (string, error) {
+	h := sha512.New512_256()
+	h.Write([]byte(name))
+	for _, input := range inputs {
+		f, err := os.Open(input)
+		if err != nil {
+			return "", err
+		}
+		defer f.Close()
+		if _, err := io.Copy(h, f); err != nil {
```
titanous: The name of the input and size should be written in the hash, to ensure that collisions can't happen if a piece of a file is removed from the end of one and added to the beginning of the next. This will probably not happen in practice, but it doesn't hurt to be correct.

titanous: The name of the input and size should be written in the hash, to ensure that collisions can't happen if a piece of a file is removed from the end of one and added to the beginning of the next. This will probably not happen in practice, but it doesn't hurt to be correct.

builder/layer.go
```diff
@@ -0,0 +1,272 @@
...
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Error("error getting current directory", "err", err)
+		return err
+	}
+
+	// check if we already created a layer for the given name + script + inputs
+	name := filepath.Base(cwd)
+	script := args.String["<script>"]
+	hash, err := generateHash(name, append([]string{script}, args.All["<inputs>"].([]string)...)...)
```
titanous: This doesn't appear to include the base artifact and builder artifact hashes, it looks like it should?

builder/layer.go
```diff
@@ -0,0 +1,272 @@
...
+		for typ, limit := range resources {
+			cmd.Resources[typ] = limit
+		}
+	}
+
+	// mount the current directory at /src and a temp dir at /out as 9p
+	// filesystems, listening on the local flynn-host's network
+	//
+	// TODO(lmars): only expose the explicit inputs at /src so we know
+	//              the generated hash is accurate
```
titanous: How hard is it to do this before merging? For correctness this is important.

builder/layer.go
lmars: I've added the base artifact but I'm not adding the builder artifact because then any changes in the builder code (e.g. changing `controller/types` which we do quite often in dev) leads to every image being rebuilt from scratch (this is the main reason I added the caching, otherwise tup would handle it).

Eventually we will use an out-of-tree builder (probably build using released flynn-host / flynn-build binaries), in which case we can use their hashes here.

lmars:
@titanous comments addressed

host/start-build-daemon.sh
```diff
@@ -0,0 +1,76 @@
...
+    daemon \
+    --force \
+    --id          "build" \
+    --external-ip "${IP}" \
+    --listen-ip   "${IP}" \
+    --state       "${dir}/state.bolt" \
+    --volpath     "${dir}/volumes" \
+    --log-dir     "${dir}/log" \
+    --flynn-init  "${dir}/bin/flynn-init" \
+    --bridge-name "buildbr0" \
```
imjorge: shouldn't this use `BRIDGE` var from above?

lmars:
I have made some major changes to this branch so I am going to close this PR and open a fresh one.

PR #3770, opened 2016-12-15
all: Switch to Ubuntu 16.04 base VM images

### LINKED ISSUE #2789: Support Ubuntu 16.04
Currently, upon running manual install of Flynn on Ubuntu 16.04, the installer fails noting that 14.04 is required. Given 16.04 is an LTS release, is Flynn planning on supporting it in the near term?

lmars:
Packer has been switched to use `virtualbox-iso` rather than `virtualbox-ovf` as I couldn't get the latter working with the xenial cloud image (the OVF file targets VMWare rather than VirtualBox).

Using an ISO is slower, but given we only release once a month I don't think it is worth the hassle trying to get the OVF to work.

I have also added a custom script to install Docker 1.9.1 as we still need it for building images (although that will soon change: #3753).

CI is still using Ubuntu 14.04 as I am planning a larger refactor.

Closes #2789.


lmars:
I have successfully released Ubuntu 16.04 AWS AMIs and a Vagrant box with version `	v20161215.2` and installed a 3 node cluster via the installer in AWS and DigitalOcean.

I think this is good to go.

PR #3780, opened 2016-12-21
cli: Improve usability and docs of release update --clean

josephglanville:
Mention how to use `flynn release update--clean` in usage and make it more usable by omitting the ID from the JSON file. This usability enhancement is especially useful when using `--clean` with JSON generated by `flynn release show --json`.

### DIFF
--- a/cli/release.go
+++ b/cli/release.go
@@ -59,6 +59,9 @@ Commands:
 		will override existing config with any values set thus. Omit the ID to
 		update the current release.
 
+		By default, the new release fields will be merged with the previous one,
+		specify --clean to not copy any fields from the previous release.
+
 	delete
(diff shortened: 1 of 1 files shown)


cli/release.go
```diff
@@ -59,6 +59,9 @@ Commands:
 		will override existing config with any values set thus. Omit the ID to
 		update the current release.
 
+		Use --clean if you wish to remove fields. Doing so requires you to specify
+		in full all release values however as any ommited values will be lost.
```
titanous: How about something like:

> By default, the new release fields will be merged with the previous one, specify --clean to not copy any fields from the previous release.

josephglanville: That does sound better, will update.

PR #3783, opened 2016-12-23
slugbuilder: Determine process types in Go

lmars:
This avoids potentially matching unrelated buildpack output as the configured process types.

### DIFF
--- a/slugbuilder/artifact/main.go
+++ b/slugbuilder/artifact/main.go
@@ -12,12 +12,15 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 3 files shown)


slugbuilder/artifact/main.go
```diff
@@ -12,6 +12,10 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"gopkg.in/yaml.v2"
```
lmars: Hmm `goimports` seems to want to put a new line in there on my machine. Life is too short to figure out why 😛

lmars:
@titanous comments addressed

PR #3788, opened 2016-12-27
Add Canada Central to AWSRegionPicker

beautifulcode:
Allows for new Canada region to be selected as an option via installer.
Need to confirm this region doesn't require any additional changes to the installer etc to fully function.

### DIFF
--- a/installer/app/src/views/aws-region-picker.js.jsx
+++ b/installer/app/src/views/aws-region-picker.js.jsx
@@ -15,6 +15,7 @@ var AWSRegionPicker = React.createClass({
 					<option value="ap-southeast-2">Asia Pacific (Sydney)</option>
 					<option value="ap-northeast-1">Asia Pacific (Tokyo)</option>
 					<option value="sa-east-1">South America (Sao Paulo)</option>
+					<option value="ca-central-1">Canada Central (Montreal)</option>
 				</PrettySelect>
 			</label>
 		);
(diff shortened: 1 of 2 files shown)


lmars:
This needs to be added to the Packer template [here](https://github.com/flynn/flynn/blob/master/util/packer/ubuntu.json#L100-L109) so that an AMI is available in the region.

lmars:
Please add a [DCO sign off](https://github.com/flynn/flynn/blob/master/CONTRIBUTING.md#developers-certificate-of-origin) to the commit message, and prefix the commit message with `installer: `.

beautifulcode:
👍 will do.

beautifulcode:
@lmars rebased to squash into single commit with `installer:` prefix and DCO sign off.

lmars:
We are going to hold off merging this just yet as it will break the installer as there will be no AMI in the new region, so we need to create that AMI first (which we will do soon).

beautifulcode:
Sure. Appreciate the insight and effort. Is that something that is quick to create/config? Assuming that is outside of this repo? Let me know if I can help. I'm looking forward to spinning up some some Canadian clusters eh? :) Thanks!

lmars:
@beautifulcode we build the AMIs using [Packer](https://packer.io), and we usually only run it when we do a stable release once a month, but I'll merge this now and get the AMIs released sooner.

PR #3804, opened 2017-01-05
host: Expose container PID

josephglanville:
This exposes the container PID via the host API.

Necessary for providing PID -> container mapping in tools such as Scope that need to join on the host namespace PID.

### DIFF
--- a/host/types/types.go
+++ b/host/types/types.go
@@ -216,6 +216,7 @@ type ActiveJob struct {
 	Job        *Job      `json:"job,omitempty"`
 	HostID     string    `json:"host_id,omitempty"`
 	InternalIP string    `json:"internal_ip,omitempty"`
+	PID        *int      `json:"pid,omitempty"`
 	ForceStop  bool      `json:"force_stop,omitempty"`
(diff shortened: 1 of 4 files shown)


josephglanville:
@lmars can you think of any edge cases that need to be handled here?
I have it setting the PID back to 0 in the state when transititioning to non-running states but I couldn't think of anything else off the top of my head.

host/types/types.go
```diff
@@ -214,6 +214,7 @@ type ActiveJob struct {
 	Job        *Job      `json:"job,omitempty"`
 	HostID     string    `json:"host_id,omitempty"`
 	InternalIP string    `json:"internal_ip,omitempty"`
+	PID        int       `json:"pid,omitempty"`
```
lmars: What do you think about making this a pointer like `ExitStatus`? Then it is set iff the job is running.

josephglanville: I like that, `nil` is much more appropriate than 0 for a non-running job.

josephglanville:
@lmars updated to use `*int` for PID.

PR #3813, opened 2017-01-09
host: Add 'flynn-host cli-add-command' to generate 'flynn cluster add' command

### LINKED ISSUE #2873: let flynn-host generate a proper `flynn cluster add` line
It would be useful if flynn-host had a command for printing a proper `flynn cluster add` line to the console, ready for copy and pastel

imjorge:
closes #2873

```
$ flynn-host cli-add-command
Install the Flynn CLI (see https://flynn.io/docs/cli for instructions) and paste the line below into a terminal window:

flynn cluster add -p iG8I30Ewxw5P76EF10ogjYKBTg3Gjfc9ulbE/Vm7IVo= default 1.localflynn.com f488bf68ea6628d1d2ec8c18907d8254

The built-in dashboard can be accessed at http://dashboard.1.localflynn.com and your login token is 6d94a7313568aaf6c0cec0e52c0d2162
```

### DIFF
--- a/host/cli/cli-add-command.go
+++ b/host/cli/cli-add-command.go
@@ -1,0 +1,129 @@
+package cli
+
+import (
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"time"
+
+	host "github.com/flynn/flynn/host/types"
+	"github.com/flynn/flynn/pkg/cluster"
+	"github.com/flynn/go-docopt"
+)
+
+func init() {
+	Register("cli-add-command", runCliAddCommand, `
+usage: flynn-host cli-add-command
+
+Get the 'flynn cluster add' command to manage this cluster.`)
+}
+
+func runCliAddCommand(args *docopt.Args, client *cluster.Client) error {
+	hosts, err := client.Hosts()
+	if err != nil {
+		return fmt.Errorf("could not list hosts: %v", err)
+	}
+	if len(hosts) == 0 {
+		return errors.New("no hosts found")
+	}
+	var (
+		domain           string
+		key              string
+		pin              string
+		dashboardAppName string
+		dashboardDomain  string
+		dashboardToken   string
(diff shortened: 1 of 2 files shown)


imjorge:
Probably we should put this somewhere on the docs. I would guess that https://flynn.io/docs/cli#adding-clusters seems ideal? Anyway, could not find the relevant source for that documentation.

host/cli/cli-add-command.go
```diff
@@ -0,0 +1,89 @@
...
+		pin              string
+		dashboardAppName string
+		dashboardDomain  string
+		dashboardToken   string
+	)
+	for _, job := range jobs {
+		app := job.Job.Metadata["flynn-controller.app_name"]
+		typ := job.Job.Metadata["flynn-controller.type"]
+		if app == "controller" && typ == "web" {
+			domain = job.Job.Config.Env["DEFAULT_ROUTE_DOMAIN"]
```
lmars: We need to make sure these are the most recent jobs in case the environment variables have been changed since bootstrapping the cluster.

imjorge: i.e. greatest `CreatedAt`?

lmars: Yep sounds good.

imjorge: @lmars was wondering since the type is named `host.ActiveJob` it there will be multiple instances of the job type?

lmars: The name is unfortunate, a job becomes "active" once it is accepted by a host, and then we continue to name it "active" even if it has stopped, so there could be multiple "active" controller web jobs from different releases and with different environment variables (i.e. some stopped, some running, some starting etc.).

We should just pick the most recent one.

imjorge:
@lmars comments addressed.

imjorge:
@lmars Anything blocking this PR on my end?

lmars:
Thanks!

PR #3817, opened 2017-01-11
builder: Build Flynn in Flynn!

lmars:
This PR extends the changes proposed in #3753 to cover building all of Flynn inside Flynn (i.e. both the Go programs and the container images).

This replaces our current build system of `tup` + `docker` with a custom builder which builds layers with explicit inputs, using both the local filesystem and a TUF repository as a cache to avoid rebuilding layers with identical inputs.

I initially had it building layers into the blobstore, but this then lead to a situation where you always needed a cluster running to bootstrap another cluster and I kept getting stuck in a catch 22, so for now I have stuck with using `file://` URIs and storing layers in `/var/lib/flynn/layer-cache`.

This isn't quite generic enough to support arbitrary builds (see #1865), but I found it easier to just focus on building Flynn with the hope to generalise in the future.

The diff is pretty big but it would be hard to do this in smaller chunks. I have commented in places that I feel need clarification, but please ask questions if anything doesn't make sense.


builder/build.go
```diff
@@ -0,0 +1,1197 @@
...
+			f.Close()
+			return "", err
+		}
+		fmt.Fprintf(h, "%s%d", input, stat.Size())
+		_, err = io.Copy(h, f)
+		f.Close()
+		if err != nil {
+			return "", err
+		}
+	}
```
titanous: I think all of the details that are written to the hash should be encoded to a JSON object which is subsequently hashed. This will completely avoid a few (probably extremely rare) edge cases that could be caused by for example joining command line arguments together without a delimiter.

builder/img/go.sh
```diff
@@ -0,0 +1,30 @@
...
+tar xzf /tmp/go.tar.gz -C "${dir}"
+rm /tmp/go.tar.gz
+
+export GOROOT="/usr/local/go"
+export GOPATH="/go"
+export PATH="${GOROOT}/bin:${PATH}"
+go get github.com/jteeuwen/go-bindata/go-bindata
+go get github.com/tools/godep
+go get github.com/flynn/go-tuf/cmd/tuf
+go get github.com/flynn/go-tuf/cmd/tuf-client
```
titanous: These packages should all be checked out by git SHA (or use a vendoring system to download them if there are many dependencies).

builder/manifest.json
```diff
@@ -0,0 +1,712 @@
...
+          "mkdir -p /etc/flynn-controller",
+          "cp -r schema /etc/flynn-controller/jsonschema"
+        ]
+      }],
+      "entrypoint": {
+        "args": [
+          "/bin/start-flynn-controller"
+        ]
+      }
+    },
```
titanous: Does it make more sense to put these manifests in their corresponding component directories similar to how the `Tupfile` used to work?

script/build-flynn
```diff
@@ -0,0 +1,226 @@
...
+    esac
+  done
+
+  if [[ $# -ne 0 ]]; then
+    usage
+    exit 1
+  fi
+
+  # TODO: remove
+  rm -f "${ROOT}/cli/tuf.go"
```
titanous: TODO?

script/build-flynn
```diff
@@ -0,0 +1,226 @@
...
+
+  local flynn_host="${ROOT}/build/bin/flynn-host"
+
+  # if building from clean, download binaries + images
+  if ! [[ -e "${flynn_host}" ]]; then
+    local sha="132903ae94d7c1b174261e738424b94022233d938ef091659727a30bddeb7a61d114b762f0fc2ffeecb967aef87d92bc71a79b9c3df31220c24ee9d8d83fa932"
+    local url="https://s3.amazonaws.com/flynn-temp/tuf/targets/${sha}.flynn-host.gz"
```
titanous: It seems like we should land the prerequisite changes first and then switch this to a released version? Also, should this just default to the latest nightly release?

builder/manifest.json
lmars: I was thinking eventually we would but this in a Flynnfile in each component, but right now I've found it easier to edit the rules in one place. Happy to split them out if you feel strongly about it though.

script/build-flynn
lmars: This is here so I can get things to build in CI where it still has files built from master which are interfering. I'll drop the WIP commit before merge, but I would like to see it pass in CI.

script/build-flynn
lmars: I was imagining releasing this as a branch when we feel it is ready to merge then updating the references just before merging.

This may also be a chance to do a clear up of the main TUF repo, and perhaps change the key structure, WDYT?

lmars:
@titanous comments addressed (I will drop the CI WIP commit before merging).

lmars:
@titanous I've dropped the WIP commit and this is ready for review (it will not pass CI as is, but I am confident once merged it will pass after CI is restarted).

builder/build.go
```diff
@@ -0,0 +1,1210 @@
...
+
+	// run the job in the host network to avoid a kernel bug which causes
+	// subsequent jobs to block waiting on the lo network device to become
+	// free (see https://github.com/docker/docker/issues/5618).
+	//
+	// NOTE: this leads to an impure build, jobs sometimes use the state of
+	//   the network to change the installation procedure (e.g. PostgreSQL
+	//   changes the default port to 5433 if something is already listening
+	//   on port 5432 at install time)
+	job.Config.HostNetwork = true
```
titanous: Removing this would be a good test of #4089.

lmars: Removed in 7d59d3d, we can revert that if we get issues.

lmars:
w00t!

PR #3826, opened 2017-01-12
test/runner: Boot multi node default cluster

lmars:
This updates `test/cluster2` to boot clusters on a single host using tags and reverts CI back to using multiple VMs which should hopefully make it more reliable.


lmars:
I have tested this in a custom CI runner and it works ok.

PR #3842, opened 2017-01-17
schema: Remove entries for non-existent sink examples

titanous:
These examples don't exist, so the metadata for them shouldn't either.


josephglanville:
Aren't the examples the 4 files you deleted in this PR under `schema/examples/controller/sink_*`?
I have never fully understood the JSONSchema stuff but that was my assumption when adding them.

titanous:
The actual examples [are generated by this](https://github.com/flynn/flynn/blob/master/controller/examples/examples.go), those files are just the metadata for the stored HTTP transactions (which don't exist in this case)>

PR #3875, opened 2017-01-25
host,cli: Add newline delimited option for syslog sink

### LINKED ISSUE #3858: Questions regarding logging and formats
I'm struggling with getting logs from Flynn and the applications we have running in it. I've been able to use log-sink to send logs to Papertrail, but using Logentries results in a bunch of run-together lines that it's unable to parse, and for some reason I can't get log-sink to work with LogDNA at all.

LogDNA's agent is able to monitor entire directories and does a great job of shipping logs from the files written in `/var/log`, but Flynn's log filenames and log entries have no application name, only process IDs.

The log-sink feature is new and poorly documented. What format is the `prefix` supposed to be in? How do we determine the state of a sink?

josephglanville:
Add option to support non-RFC6587 compliant sinks that expect syslog lines delimited by newlines.

Fixes #3858

### DIFF
--- a/cli/cluster.go
+++ b/cli/cluster.go
@@ -31,7 +31,7 @@ usage: flynn cluster
        flynn cluster migrate-domain <domain>
        flynn cluster backup [--file <file>]
        flynn cluster log-sink
-       flynn cluster log-sink add syslog [--use-ids] [--insecure] <url> [<prefix>]
+       flynn cluster log-sink add syslog [--use-ids] [--insecure] [--format <format>] <url> [<prefix>]
        flynn cluster log-sink remove <id>
 
 Manage Flynn clusters.
@@ -82,8 +82,9 @@ Commands:
         Supported schemes are syslog and syslog+tls
 
         options:
-            --use-ids   Use app IDs instead of app names in the syslog APP-NAME field
-            --insecure  Don't verify servers certificate chain or hostname. Should only be used for testing.
+            --use-ids          Use app IDs instead of app names in the syslog APP-NAME field.
+            --insecure         Don't verify servers certificate chain or hostname. Should only be used for testing.
+            --format=<format>  One of rfc6587 or newline, defaults to rfc6587.
 
         examples:
             $ flynn cluster log-sink add syslog syslog+tls://rsyslog.host:514/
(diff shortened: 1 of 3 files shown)


cli/cluster.go
```diff
@@ -82,8 +82,9 @@ Commands:
         Supported schemes are syslog and syslog+tls
 
         options:
-            --use-ids   Use app IDs instead of app names in the syslog APP-NAME field
+            --use-ids   Use app IDs instead of app names in the syslog APP-NAME field.
             --insecure  Don't verify servers certificate chain or hostname. Should only be used for testing.
+			--format	One of rfc6587 or newline, defaults to rfc6587.
```
lmars: This whitespace looks odd, I guess it needs to be spaces rather than tabs?

cli/cluster.go
```diff
@@ -482,11 +483,20 @@ func runLogSinkAddSyslog(args *docopt.Args, client controller.Client) error {
 		return fmt.Errorf("Invalid syslog protocol: %s", u.Scheme)
 	}
 
+	var format ct.SyslogFormat
+	switch args.String["--format"] {
+	case "newline":
+		format = ct.SyslogFormatNewline
+	default:
+		format = ct.SyslogFormatRFC6587
```
lmars: I think we should only use `ct.SyslogFormatRFC6587` if the flag is either `rfc6587` or empty, so if someone makes a typo like `--format newlin` this will error.

host/logmux/sink.go
```diff
@@ -611,7 +613,14 @@ func (s *SyslogSink) Write(m message) error {
 	}
 
 	// Write to the remote syslog
-	_, err := s.conn.Write(rfc6587.Bytes(m.Message))
+	var data []byte
+	switch s.format {
+	case ct.SyslogFormatRFC6587:
+		data = rfc6587.Bytes(m.Message)
+	case ct.SyslogFormatNewline:
+		data = append(m.Message.Bytes(), byte('\n'))
```
lmars: Could just be `append(m.Message.Bytes(), '\n')`

cli/cluster.go
josephglanville: Agreed, will change.

josephglanville:
@lmars comments addressed.

lmars:
@josephglanville looks like there is a genuine test failure in `HostSuite.TestLogSinks`.

josephglanville:
Yep, looking into it.

lmars:
@josephglanville I think `s.format` needs a default value in `host/logmux/sink.go`.

josephglanville:
Yeah, the integration tests add the sink via the controller API so it doesn't use the defaults that are setup in the CLI.

PR #3887, opened 2017-01-26
appliance/redis: Fix setting password in discoverd registration test

josephglanville:
Not having a password set was causing Redis tests to hang for quite some time locally (though eventually exiting with exit status 0).
I'm unsure if this effects CI, it probably does but is likely masked by the long run times of other unit tests.


josephglanville:
This doesn't seem to be correct. Tests fail locally for me but work correctly in the CI environment so it's likely a local state/version problem that I need to look into.

PR #3892, opened 2017-01-30
router: allow catch-all route

### LINKED ISSUE #3779: Allow top-level wildcard route
Currently routes like `*.com` and `*.net` work as expected, but `*` doesn't work. The router should support a catch-all wildcard route.

jbrady42:
Add support for top level wildcard route

Closes #3779

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -460,6 +460,10 @@ func (s *HTTPListener) findRoute(host string, path string) *httpRoute {
 			return tree.Lookup(path)
 		}
 	}
+	// use catch-all if available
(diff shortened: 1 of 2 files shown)


lmars:
Thanks!

hadifarnoud:
so I can use `flynn route add http *`? on dashboard * comes of as `%2A/` and using cli just prints usage info. however `*.domain.com` works fine

Flynn version: `v20170417.2` and cli version is `v20170505.0` which are the latest versions I guess

Issue #2438, opened 2016-02-10
Persistent Volumes


josephglanville:
This is a preliminary list of the work required to get persistent volumes working in a best-effort fashion.
That is if the volume is available it will be used again, else a job will be scheduled with a new volume.
Additional deployer and scheduler strategies/policies are out of the scope of this initial work.
- [x] Add volumes_cache table to schema
- [x] Add volumes API to controller
- [x] Teach scheduler to sync volumes with controller
- [x] Enhance the scheduler to use previous volume information to schedule jobs to re-use volumes.
- [x] Add volumes CLI to display volumes and their attachments

josephglanville:
Work is happening on this branch https://github.com/flynn/flynn/tree/persistent-volumes

holtgrewe:
Would this also add support for large volumes where data is only read after writing them once? I am planning to deploy some Bioinformatics tools with Flynn. Such programs often use large "database" files that are used for answering user queries.

Adjusting the programs to just put everything into a postgres database is usually not easy but no update is required to the database files after a volume is created.

titanous:
@holtgrewe I think our work on squashfs images in #3100 is relevant, it would allow you to put a dataset in a compressed squashfs file that is part of the container image definition.

lmars:
@josephglanville I was taking a look through the changes on the persistent-volumes branch to see what scheduler changes are needed, and although you mention additional deploy strategies are out of scope, I do think we need to take deployments into consideration.

This is my understanding of the current proposed approach when starting a job which requires a volume:

* if no volumes exist for the given app + process type, create a new one
* if volumes exist for the given app + process type but they are all attached, create a new one
* if volumes exist for the given app + process type and one is unattached, use it

This works well to maintain volume attachments when jobs restart (which is useful for db appliances when a host is rebooted for example), but then it leads to the following scenario during deployments:

* `one-by-one`: the first new job gets a new volume, then subsequent new jobs get the volume from the previously stopped job, then leaving a single unattached volume when the last job stops
* `all-at-once`: all new jobs come up with new volumes, old jobs stop leaving N unattached volumes

We now have a pool of unattached volumes which get stale over time, and then the next deployment will use these stale volumes.

This is probably ok for jobs which manage their own replication (they will just bring the stale data up-to-date), but for non replicated appliances like Redis where we just want a single job with the same data, the first deployment would effectively lose the data, and then subsequent deploys would rollback the data to when the last deploy happened.

So, I propose:

[...]

PR #3897, opened 2017-02-02
Implement Persistent Volumes

lmars:
This pull request is a continuation of the work already done on the [persistent-volumes branch](https://github.com/flynn/flynn/compare/persistent-volumes) to implement persistent volumes as per #2438.

Summary:

* the scheduler syncs volumes from all hosts and persists them to the `volumes` and `job_volumes` tables in the controller database
* when placing a job which requires volumes, the scheduler first tries to locate existing, unassigned volumes for the job's app / release / type and the volume's path, and if it finds one then places the job on the same host as the volume whilst also assigning the volume to the job (i.e. setting `Volume.JobID`)
* volumes can be "decommissioned" via `flynn volume decommission ID` which causes the volume to not be attached to any new jobs
* if a host which has unassigned volumes is down, jobs are still scheduled on that host but enter a new `blocked` state waiting for either the host to come back up or the volume to be decommissioned. This then means if a host is rebooted which has data for a process type, the job just stays down until the host comes back rather than being restarted on a different host with an empty volume, but if the host has really gone away then it is up to the operator to decommission the volume and unblock the job to be scheduled on a different host

Things which need to be considered but not included in this PR:

* volumes won't be persisted through deployments, but the design allows for that to be added later (see [this comment](https://github.com/flynn/flynn/issues/2438#issuecomment-274272589))
* we should add `flynn volume backup` and `flynn volume restore` so that if a host is lost and volumes need to be decommissioned in order to move jobs to other hosts, operators can first restore volumes so that the unblocked job doesn't have to start with a completely empty volume

### DIFF
--- a/controller/scheduler/scheduler.go
+++ b/controller/scheduler/scheduler.go
@@ -16,6 +16,7 @@ import (
 	"github.com/flynn/flynn/controller/utils"
 	discoverd "github.com/flynn/flynn/discoverd/client"
 	host "github.com/flynn/flynn/host/types"
+	"github.com/flynn/flynn/host/volume"
 	"github.com/flynn/flynn/pkg/attempt"
 	"github.com/flynn/flynn/pkg/cluster"
 	"github.com/flynn/flynn/pkg/httphelper"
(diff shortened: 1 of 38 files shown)


controller/scheduler/scheduler.go
```diff
@@ -1035,6 +1274,10 @@ func (s *Scheduler) HandleInternalStateRequest(req *InternalStateRequest) {
 		req.State.Formations[key.String()] = &f
 	}
 
+	for id, vol := range s.volumes {
+		req.State.Volumes[id] = &(*vol)
```
lmars: We need to copy it to create a "snapshot" of the scheduler's state to pass back to the caller.

lmars:
@titanous comments addressed.

PR #3907, opened 2017-02-08
controller: avoid fd leak if JSON schema parse fails

imjorge:


controller/schema/schema.go
```diff
@@ -39,6 +39,7 @@ func Load(schemaRoot string) error {
 		schema := &jsonschema.Schema{Cache: schemaCache}
 		err = schema.ParseWithoutRefs(file)
 		if err != nil {
+			file.Close()
```
titanous: I believe you can move this up one line outside of the `if` and then delete the close below.

imjorge: @titanous true, just changed. Edited through Github Web UI, can this be squashed on merge?

titanous:
Thanks!

PR #3947, opened 2017-02-27
router: Pass client-specified Request-ID header

### LINKED ISSUE #3939: Allow HTTP Clients to Define X-Request-ID header
Heroku supports client-defined X-Request-ID headers, but the flynn HTTP router always assigns a random UUID. The Flynn router should use an existing value for X-Request-ID, if one exists. 

Relevant code: https://github.com/flynn/flynn/blob/master/router/http.go#L598

titanous:
If a client specifies a Request-ID request header, it will be passed to a backend without replacement it meets some basic criteria: the value must be between 20 and 200 characters, and consist of ASCII letters, digits, or the characters +, /, =, and -. Invalid IDs will be ignored and replaced with a randomly generated UUID.

Closes #3939

### DIFF
--- a/router/http.go
+++ b/router/http.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


router/http.go
```diff
@@ -607,3 +608,12 @@ func mustPortFromAddr(addr string) string {
 	}
 	return port
 }
+
+var validRequestIDPattern = regexp.MustCompile("^[a-zA-z0-9+/=-]+$")
```
lmars: Should the regex contain `A-Z` rather than `A-z`?

titanous: Yes, fixed.

PR #3957, opened 2017-03-02
test: Add debug logging to TestUpdateLogs

lmars:
To help investigate why the test keeps failing in CI.


lmars:
The test failed again, and although these logs didn't lead me to a fix, the change is still worth merging.

The next step in figuring this out is configuring a log sink for the nested cluster so we get all the logs  from the hosts before they die.

PR #3963, opened 2017-03-03
appliance/mongodb: Use a data volume for restore job

### LINKED ISSUE #3962: Restore mongodb Dump > 97mb
Hi,

the volume for the restore Job is limited to 97mb. So when I try to restore a dump > 97 mb I get out of disk error:

```
tar: xxx.metadata.json: Cannot write: No space left on device
```

volume from df -h:
```
/dev/zd640                                                97M   97M  1.0K 100% /var/lib/flynn/volumes/zfs/mnt/ext2/flynn1-b4ae79d1-1e6b-45c5-9c70-d2233d65c6c4
```

Is there a possibility to increase the volume size? I would not want to open access to mongodb and use mongorestore, but that would be a possibility.

lmars:
Fixes #3962.

### DIFF
--- a/appliance/mongodb/restore.sh
+++ b/appliance/mongodb/restore.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 # Extract dump into a temporary directory.
-TMPFILE=`mktemp -d /tmp/dump.XXXXXXXXXX`
+TMPFILE=`mktemp -d /data/dump.XXXXXXXXXX`
 mkdir -p $TMPFILE
 trap "rm -rf ${TMPFILE}" EXIT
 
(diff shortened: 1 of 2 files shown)


josephglanville:
@lmars we may need to make use of the version check here that we did for pg parallel restore.
Older clusters don't support using data volumes for one-off jobs.

lmars:
@josephglanville I don't think so, if restoring to an old cluster the `data: true` in the JSON will just be ignored, and the old cluster will continue to use `/tmp` (the pg restore change was different as the client was referencing `/data` which would then not exist in older clusters).

josephglanville:
Ahh I see, I overlooked that the MongoDB image was what was being update to reference `/data` rather than the client.

PR #3968, opened 2017-03-04
Vagrantfile,test/rootfs: Install protoc as build dependency

josephglanville:
Install the protobuf compiler as a build dependency.
Required to test #3425 in CI.

### DIFF
--- a/test/rootfs/setup.sh
+++ b/test/rootfs/setup.sh
@@ -205,9 +205,13 @@ curl https://godeb.s3.amazonaws.com/godeb-amd64.tar.gz | tar xz
 ./godeb install 1.7.5
 rm godeb
 
+# setup tmpdir
+tmpdir=$(mktemp --directory)
+trap "rm -rf ${tmpdir}" EXIT
+
 # install go-tuf
-export GOPATH="$(mktemp --directory)"
-trap "rm -rf ${GOPATH}" EXIT
+export GOPATH="${tmpdir}/gopath"
+mkdir ${GOPATH}
 go get github.com/flynn/go-tuf/cmd/tuf
 mv "${GOPATH}/bin/tuf" /usr/bin/tuf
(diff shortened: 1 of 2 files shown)


test/rootfs/setup.sh
```diff
@@ -227,6 +227,13 @@ git clone https://github.com/sstephenson/bats.git "${tmpdir}/bats"
 curl -fsLo "/usr/local/bin/jq" "http://stedolan.github.io/jq/download/linux64/jq"
 chmod +x "/usr/local/bin/jq"
 
+# install protobuf compiler
+curl -sL https://github.com/google/protobuf/releases/download/v3.0.0/protoc-3.0.0-linux-x86_64.zip > ~/protoc.zip
+trap "rm ~/protoc.zip" EXIT
```
lmars: This trap will overwrite the earlier traps (so we are already leaking some temp files as we have multiple traps). We should just create a tmpdir at the top of the script with a single trap to remove it then use it throughout the script.

josephglanville: I forgot that traps overwrite handlers. We also have umount and netns cleanup stuff attempting to be run on EXIT too, probably need hack together a solution to clean that up too.
For now I have changed it to use a single tmpdir atleast so we shouldn't leak files.

lmars: Yeh so ideally we would run this in a job and allow flynn-host to do the cleanup :smile:.

I actually have a branch where I did that (I think that is when I realised the trap issue), so that can be the proper fix for this.

jvatic:
@josephglanville @lmars updated with latest protobuf version

titanous:
As of #3817 I don't think this is the right place to do it, this should go in the Go builder layer afaik. /cc @lmars

lmars:
Yes this should be moved to [builder/img/go.sh](https://github.com/flynn/flynn/blob/master/builder/img/go.sh).

PR #3976, opened 2017-03-08
cli: Support updating Linux caps and allowed devices

lmars:


onnimonni:
Could we have some examples in the docs how to use this?

For example I have a case where I would want to run a vpn client inside docker container with `NET_ADMIN` capability.

lmars:
@onnimonni the workflow is currently a bit unfriendly until we have better support with Flynnfile (#1084), for now you need to add the following to a local `config.json` file:

```
{
  "processes": {
    "web": {
      "linux_capabilities": [
        "CAP_NET_RAW",
        "CAP_NET_BIND_SERVICE",
        "CAP_DAC_OVERRIDE",
        "CAP_SETFCAP",
        "CAP_SETPCAP",
        "CAP_SETGID",
        "CAP_SETUID",
        "CAP_MKNOD",
        "CAP_CHOWN",
        "CAP_FOWNER",
        "CAP_FSETID",
        "CAP_KILL",
        "CAP_SYS_CHROOT",
        "CAP_NET_ADMIN"
      ]
    }
  }
}
```

And then use it to update your release with:

```
$ flynn -a APP release update config.json
```

Note that this list *overwrites* the set of capabilities, so it is best to include the defaults (see [here](https://github.com/flynn/flynn/blob/master/host/types/defaults.go#L7-L24)) plus the ones you want.

PR #3990, opened 2017-03-14
sirenia: Clone volumes to new release on deploy

lmars:
This is a proposal for improving sirenia deployments by creating snapshots of existing volumes and cloning them to the new release during a deployment.

The deployer makes a snapshot request to the process about to be replaced (giving it chance to do tasks pre / post snapshot), and then creates a new volume from that snapshot on the same host, tagging it with the new release so that the scheduler will use it when starting the replacement process. 

Volume IDs are now used as the sirenia process IDs, so during a deployment the new nodes come up with new IDs (as they have distinct, cloned volumes), although the cluster fixer will end up creating nodes with existing IDs (as the volume ID will be the same).

TODO:

- [x] decide whether each appliance needs pre / post snapshot tasks
- [x] check using the volume ID as the process ID works in all cases
- [ ] convince ourselves that this is reliable (either with an integration test or manually testing with real data)

### DIFF
--- a/appliance/mariadb/cmd/flynn-mariadb/main.go
+++ b/appliance/mariadb/cmd/flynn-mariadb/main.go
@@ -2,16 +2,15 @@ package main
 
 import (
 	"encoding/binary"
-	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
-	"path/filepath"
 
 	"github.com/flynn/flynn/appliance/mariadb"
 	"github.com/flynn/flynn/discoverd/client"
+	"github.com/flynn/flynn/host/volume"
+	"github.com/flynn/flynn/pkg/cluster"
 	"github.com/flynn/flynn/pkg/httphelper"
-	"github.com/flynn/flynn/pkg/random"
 	"github.com/flynn/flynn/pkg/shutdown"
 	sd "github.com/flynn/flynn/pkg/sirenia/discoverd"
 	"github.com/flynn/flynn/pkg/sirenia/state"
@@ -37,17 +36,15 @@ func main() {
 	serverId := ip2id(net.ParseIP(ip))
 
 	const dataDir = "/data"
-	idFile := filepath.Join(dataDir, "instance_id")
-	idBytes, err := ioutil.ReadFile(idFile)
-	if err != nil && !os.IsNotExist(err) {
-		shutdown.Fatalf("error reading instance ID: %s", err)
+	volID := os.Getenv("VOLUME_0")
+	if volID == "" {
+		shutdown.Fatalf("error getting primary volume ID, VOLUME_0 not set")
 	}
-	id := string(idBytes)
-	if len(id) == 0 {
-		id = random.UUID()
-		if err := ioutil.WriteFile(idFile, []byte(id), 0644); err != nil {
-			shutdown.Fatalf("error writing instance ID: %s", err)
-		}
+
+	hostID, _ := cluster.ExtractHostID(os.Getenv("FLYNN_JOB_ID"))
+	host, err := cluster.NewClient().Host(hostID)
+	if err != nil {
+		shutdown.Fatal(err)
 	}
 
 	err = discoverd.DefaultClient.AddService(serviceName, &discoverd.ServiceConfig{
@@ -58,7 +55,7 @@ func main() {
 	}
 	inst := &discoverd.Instance{
 		Addr: ":" + mariadb.DefaultPort,
-		Meta: map[string]string{mariaIdKey: id},
+		Meta: map[string]string{mariaIdKey: volID},
 	}
 	hb, err := discoverd.DefaultClient.RegisterInstance(serviceName, inst)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


josephglanville:
This is pretty cool.

I think there is a few things to be looked into though.

The first is that when cloning volumes we should probably be cloning the volume of the peer that will be the new immediate upstream of the new peer.

For instance when we are replacing the primary this will currently create a new peer by cloning the volume of the primary. If there is replication lag and more than 1 async configured this could result in the new peer having a greater xlog value than it's upstream peer which would cause issues (they would likely resolve due to crash/restart but would be nice to avoid that).
Instead we should clone the last async in all cases I think as the new peer will always be appended to the current list of asyncs.

VolumeIDs as sirenia IDs seems a good improvement to me. 👍 

I think for now if the change to clone the last async is made then no pre/post snapshot hooks should be required for the current appliances as we have all our databases configured to be crash safe on disk at all times, that said using the base backup facilities in both PostgreSQL and MySQL could ensure that the on-disk state will not required crash recovery on boot of the new peer which would be a reasonable enhancement.

Overall, awesome. 🎉

lmars:
@josephglanville I agree with your recommendation to always clone the tail async node, but the issue then is that, at least with the current code, we will end up with all jobs on the same host.

For example, let's say we have 3 hosts A, B and C with the primary, sync and async on hosts A, B and C respectively. The first replacement clones on C, and we kill the old async, then the second replacement also clones on C (that is where the new async is), and kills the old sync, and finally the third replacement clones on C and kills the primary, leaving all new nodes on C.

A few options:

1) always snapshot the tail async but transfer that snapshot to other hosts (it is then questionable if this is better than just allowing postgres to do the replication rather than ZFS)
2) snapshot all jobs up front, wait until the current tail async is caught up with the snapshotted primary, then proceed as currently (so the new jobs will never have later entries than any running jobs)
3) at each step, once the snapshot is created wait until the current tail async is caught up with what is in the snapshot before proceeding
4) accept that the tail async could be ahead of its upstream and allow the database to deal with it

Thoughts?

titanous:
Assume a distribution of three database instances across three hosts. d0 is the primary, d1 is the sync, d2 is the async.

I think the sequence should effectively reverse the chain, like this:

1. Snapshot d2, start d2', stop d2. d2' is the new async.
2. Snapshot d1, start d1', d1' ends up chaining off of d2' as a second async. Stop d1, d2' is now the sync and d1' is now the async.
3. Snapshot d0, start d0', this instance ends up chaining off of d1' as a second async. Stop d0, d2' is now the primary, d1' is now the sync, and d0' is now the async.

lmars:
@titanous that is what the code currently does, but then after this step:

> Snapshot d1, start d1', d1' ends up chaining off of d2' as a second async

d1' could have a greater xlog value than d2' depending on how far behind d1 the original async d2 was.

titanous:
> d1' could have a greater xlog value than d2' depending on how far behind d1 the original async d2 was.

In that case can we wait for d2' to meet or pass the d1' xlog position before starting d1'?

lmars:
Ok, so my option 3) above?

titanous:
Yep, sounds good to me.

lmars:
@titanous I've implemented the tail xlog sync in 46080ec.

titanous:
Looks like the unit tests are broken:

```text
----------------------------------------------------------------------
22:52:09.460 PANIC: persistence_test.go:137: PersistenceTests.TestVolumeDeletion

... Panic: runtime error: invalid memory address or nil pointer dereference (PC=0x482EA8)

/home/ubuntu/go/src/github.com/flynn/flynn/build/_go.94aa79164c357ebf735445f634381e3b738d6b9418da6a21ba5b7c65c7851b9d/src/runtime/asm_amd64.s:514
  in call32
/home/ubuntu/go/src/github.com/flynn/flynn/build/_go.94aa79164c357ebf735445f634381e3b738d6b9418da6a21ba5b7c65c7851b9d/src/runtime/panic.go:489
  in gopanic
/home/ubuntu/go/src/github.com/flynn/flynn/build/_go.94aa79164c357ebf735445f634381e3b738d6b9418da6a21ba5b7c65c7851b9d/src/runtime/panic.go:63
  in panicmem
/home/ubuntu/go/src/github.com/flynn/flynn/build/_go.94aa79164c357ebf735445f634381e3b738d6b9418da6a21ba5b7c65c7851b9d/src/runtime/signal_unix.go:290
  in sigpanic
github.com/flynn/flynn/host/volume/manager/_test/_obj_test/manager.go:242
  in Manager.NewVolumeFromProvider
github.com/flynn/flynn/host/volume/manager/_test/_obj_test/manager.go:218
  in Manager.NewVolume
persistence_test.go:177
  in PersistenceTests.TestVolumeDeletion
/home/ubuntu/go/src/github.com/flynn/flynn/build/_go.94aa79164c357ebf735445f634381e3b738d6b9418da6a21ba5b7c65c7851b9d/src/runtime/asm_amd64.s:514
  in call32
/home/ubuntu/go/src/github.com/flynn/flynn/build/_go.94aa79164c357ebf735445f634381e3b738d6b9418da6a21ba5b7c65c7851b9d/src/reflect/value.go:302
  in Value.Call
/home/ubuntu/go/src/github.com/flynn/flynn/vendor/github.com/flynn/go-check/check.go:809
  in suiteRunner.forkTest.func1
/home/ubuntu/go/src/github.com/flynn/flynn/vendor/github.com/flynn/go-check/check.go:703
  in suiteRunner.forkCall.func1
/home/ubuntu/go/src/github.com/flynn/flynn/build/_go.94aa79164c357ebf735445f634381e3b738d6b9418da6a21ba5b7c65c7851b9d/src/runtime/asm_amd64.s:2197
  in goexit
OOPS: 0 passed, 4 PANICKED
```

lmars:
@titanous tests fixed.

lmars:
This is outdated, closing.

PR #4018, opened 2017-03-28
controller: Make Sink.Config a *json.RawMessage

lmars:
So that it is encoded as raw JSON rather than a base64 string.


lmars:
@titanous test fixed.

PR #4033, opened 2017-04-13
host: Force memory overcommitment

titanous:
This patch sets `/proc/sys/vm/overcommit_memory` to 1, which enables virtual memory overcommit and disables heuristic based denial of memory allocations (using these heuristics is the default setting of 0).

Due to the large virtual memory allocations that Go uses, and the fact that flynn-host frequently forks, the default heuristics can cause errors like this:

    fork/exec /sbin/zfs: cannot allocate memory

Since we rely on the OOM killer anyway, there's no need to arbitrarily cause memory allocations to fail and we can turn off the heuristics entirely.

### DIFF
--- a/host/libcontainer_backend.go
+++ b/host/libcontainer_backend.go
@@ -85,6 +85,10 @@ func NewLibcontainerBackend(config *LibcontainerConfig) (Backend, error) {
 		return nil, err
 	}
 
+	if err := forceMemoryOvercommit(); err != nil {
+		return nil, err
+	}
+
 	defaultTmpfs, err := createTmpfs(resource.DefaultTempDiskSize)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


hadifarnoud:
I had this exact issue. didn't see this issue and thought doing `flynn-host volume gc` might fix things.

now it appears gc removed some volumes that was needed for some jobs to run: https://gist.github.com/hadifarnoud/139d672c3435d7a20e37f6caba78ee25

is there any way to fix this?

hadifarnoud:
the more I try to fix it, the more I screw it up. I did a restore from backup but as it turns out my backup was corrupt. found an old backup but it's from my old cluster (flynnhub). at first it restored but did not work. 

now it just times out at `check online-hosts`

PR #4037, opened 2017-04-16
appliance/redis: Set a password in TestMain_Discoverd

lmars:
CI runs an old version of Redis which seems to handle empty passwords differently than newer versions (e.g. the version used in vagrant).

We have an outstanding task to bring CI up to date which would reveal this, but we are waiting on #3817 before doing that (I have a local branch refactoring CI based off that work).


lmars:
Hmm so this is now incompatible with the older redis version in CI, I'm going to bring CI redis up to date then will re-run this.

Issue #4049, opened 2017-04-24
Backup corrupt on Windows


schroedingersZombie:
When creating a backup on windows with
`flynn cluster backup --file=backup.tar`
the resulting tar file is corrupt.

The error seems to be in line 34 of pkg/backup/tar_writer.go here the result of syscall.Getuid is set as uid for the tar file. But syscall.Getuid returns -1 on a Windows system [0]. So the padded user (and group) id is "00000-1" which is not a valid numerical value, which tar expects at this field.

Therefore it is not possible to restore a backup made with windows.

A fix would be to either check if running on a Windows machine or if Getuid returns -1 and replace the uid in this case with a default.

lmars:
@schroedingersZombie are you interested in patching that code with your suggestion? I don't have a Windows machine to test myself.

PR #4059, opened 2017-04-27
Fix UID in backups on Windows

schroedingersZombie:
Fix for #4049.

When `syscall.Getuid()` returns a negative value zero is instead used as userid in the tar file.
This prevents userids like `00000-1` which are invalid.

Sorry for that new pull request, I previously fucked up with the branches on my fork.


titanous:
Thanks!

Issue #3910, opened 2017-02-10
docker-receive: Images with ENV containing tabs cannot be pushed


lmars:
The artifact fails to be created with the following error in the `docker-receive` logs:

```
ERROR: json: error calling MarshalJSON for type json.RawMessage: invalid character '\\t' in string literal\n
```

To reproduce:

* build a Docker image with a tab in an env var

```
$ docker build -t tab-env - <<< "$(echo -e 'FROM scratch\nENV FOO bar\tbaz')"                                                                                                          
Sending build context to Docker daemon 2.048 kB
Step 1 : FROM scratch
 ---> 
Step 2 : ENV FOO bar    baz
 ---> Running in 81741bc7e676
 ---> 06703e5d8c2f
Removing intermediate container 81741bc7e676
Successfully built 06703e5d8c2f
```

* push it to Flynn

```
$ flynn -a tab-env docker push tab-env
flynn: getting image config with "docker inspect -f {{ json .Config }} tab-env"
flynn: tagging Docker image with "docker tag tab-env docker.1.localflynn.com/tab-env:latest"
flynn: pushing Docker image with "docker push docker.1.localflynn.com/tab-env:latest"
The push refers to a repository [docker.1.localflynn.com/tab-env] (len: 1)
06703e5d8c2f: Pushed 
latest: digest: sha256:1346da490a595ff0b0f147771b79e83fb122c7f8fad0b3db142655eabdfefcca size: 1322
Received unexpected HTTP status: 500 Internal Server Error
```

The error is specifically trying to encode `Artifact.RawManifest` which contains the tab character in `Entrypoint.Env`, we should probably quote the raw manifest (i.e. convert literal tabs to `\t`).

schroedingersZombie:
I also encounter that problem when restoring a backup containing an docker image with tab in environment. At least I think, that this has the same cause.
`17:29:17.545360 migrate-artifacts running Docker image migrator
17:29:17.557127 migrate-artifacts converting 3 active Docker images to Flynn images
17:29:17.557134 migrate-artifacts converting Docker image 8ba9251a-366a-48f7-b1ee-df8ea1bf81a9 (1/3)
17:30:04.618664 migrate-artifacts ERROR: json: error calling MarshalJSON for type json.RawMessage: invalid character '\t' in string literal
17:30:04.621416 migrate-artifacts error converting Docker image 8ba9251a-366a-48f7-b1ee-df8ea1bf81a9 (1/3): exit status 1
[...]

PR #4060, opened 2017-04-27
Fix pushing of docker images with tab characters in env

schroedingersZombie:
Replaces tab characters in Manifest with `\t` an prevents so the error mentioned in #3910

### DIFF
--- a/test/test_docker_receive.go
+++ b/test/test_docker_receive.go
@@ -201,6 +201,6 @@ func (s *DockerReceiveSuite) TestTabsInEnv(t *c.C) {
 	// flynn docker push image
 	t.Assert(flynn(t, "/", "-a", app.Name, "docker", "push", repo), Succeeds)
 
-	// check the whiteouts are effective
+	// check the environment variable has the correct value
 	t.Assert(flynn(t, "/", "-a", app.Name, "run", "sh", "-c", "[[ \"$TAB\" = \"test\ttest\" ]]"), Succeeds)
 }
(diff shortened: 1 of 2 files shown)


lmars:
@schroedingersZombie this needs a regression test which builds a docker image with a tab in the env and successfully pushes it to Flynn (as described in #3910).

See [DockerReceiveSuite.TestConvertWhiteouts](https://github.com/flynn/flynn/blob/master/test/test_docker_receive.go#L70-L93) for a similar test.

schroedingersZombie:
@lmars added regression test. I hope this is sufficent.

titanous:
Looks like this needs a gofmt:

```text
These files are not properly gofmt'd:
 - test/test_docker_receive.go

Please reformat the above files using "gofmt -s -w" and commit the result
```

test/test_docker_receive.go
```diff
@@ -183,3 +183,24 @@ func (s *DockerReceiveSuite) TestReleaseDeleteImageLayers(t *c.C) {
...
+
+	// create app
+	client := s.controllerClient(t)
+	app := &ct.App{Name: repo}
+	t.Assert(client.CreateApp(app), c.IsNil)
+
+	// flynn docker push image
+	t.Assert(flynn(t, "/", "-a", app.Name, "docker", "push", repo), Succeeds)
+
+	// check the whiteouts are effective
```
lmars: Update this comment to describe the action being taken here.

lmars:
Thanks!

PR #4061, opened 2017-04-28
script/lib: Don't use magic for S3 upload MIME types

### LINKED ISSUE #4051: Vagrant box metadata JSON wrong Content-Type
When trying to run `vagrant up` I get the following error:
```You specified a box version constraint with a direct box file
path. Box version constraints only work with boxes from Vagrant
Cloud or a custom box host. Please remove the version constraint
and try again.
```
I suspect that is caused because `https://dl.flynn.io/vagrant/flynn-base.json` has it's content-typer set to `text/plain` if I host the file myself with content-type `application/json` this step works like expected.
I also tested with self served `text/plain` - same error as with the official server.

titanous:
s3cmd uses python-magic by default to figure out the MIME type of a file. Some version combinations return `text/plain` for JSON files instead of `application/json`, which breaks the Vagrant box manifest download.

Setting `--no-mime-magic` falls back to using the file extension, which is what we want.

Closes #4051

### DIFF
--- a/script/lib/aws.sh
+++ b/script/lib/aws.sh
@@ -17,6 +17,7 @@ sync_cloudfront() {
     --cf-invalidate \
     --no-preserve \
     --follow-symlinks \
+    --no-mime-magic \
     "${src}" "${dst}" \
     | tee /dev/stderr \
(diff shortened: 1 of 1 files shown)


jvatic:
LGTM

PR #4073, opened 2017-05-01
docs: easy issues aren't easy, they're complexity/easy

simonamor:
https://flynn.io/docs/development links to github issues which are
labelled easy, but the label is now complexity/easy so the link
results in no entries.

Signed-off-by: Simon Amor <simon@leaky.org>


titanous:
Thanks!

PR #4076, opened 2017-05-02
dashboard: Reduce github/provision button padding

simonamor:
When the app-list is long enough to cause a scrollbar on the middle
panel, there's not quite enough width for the github and provision
database buttons and this causes them to appear one above the other
instead of side by side.

This commit has the side-effect of reducing the padding by 0.5rem in
right-hand panel too.

Signed-off-by: Simon Amor <simon@leaky.org>


jvatic:
Thanks!

PR #4087, opened 2017-05-04
appliance/postgresql: Update to Postgres 9.6

titanous:


lmars:
@titanous do the tests pass if run manually with 9.6?

titanous:
Haven't tested that yet, will do so and post an update.

PR #4120, opened 2017-05-18
Fix issue with ADDITIONAL_HTTPS_PORTS

temujin9:
I made an assumption that turned out to be incorrect, about the router.Route member which gets populated for an HTTPS route. This fixes the issue, and wraps it in an integration test.


temujin9:
@titanous I have pushed the same code three times; it has failed each time, but in other tests. Can I get a pass?
- https://ci.flynn.io/builds/20170518231116-c776a4ea
- https://ci.flynn.io/builds/20170518222007-8b7d0d70
- https://ci.flynn.io/builds/20170518223255-750bf31a

PR #4137, opened 2017-06-09
blobstore: add support for openstack swift backend

fab120:
Blobstore support for openstack swift


titanous:
This is awesome! Is there any lightweight installation that could be run in a container? We'd like to have integration tests for the backend as well: https://github.com/flynn/flynn/blob/master/test/test_blobstore.go

fab120:
I don't think there is a simple way to run openstack (and swift service in particular) in a container.

I think the simplest way is to have an account on a [openstack provider](https://github.com/fab120/flynn/blob/eb34dc4e6afdd5c5057731221bb708e2889c4b1c/docs/content/production.html.md#blobstore-backend) and running test with it.

How are you handling tests with s3, gcs and azure now?

titanous:
Okay, if you write an integration test, we can hook it up to a hosted provider in CI after this gets merged.

fab120:
I've already written tests by looking at other backends implementations.
https://github.com/fab120/flynn/blob/eb34dc4e6afdd5c5057731221bb708e2889c4b1c/blobstore/blobstore_test.go#L145-L162
https://github.com/fab120/flynn/blob/eb34dc4e6afdd5c5057731221bb708e2889c4b1c/test/test_blobstore.go#L78-L88

Do you think this pr needs more test coverage?

titanous:
Perfect, for some reason GitHub was hiding the diff I was looking for. Will follow up after testing and review.

PR #4184, opened 2017-07-17
vendor: Update go-p9p

titanous:
This fixes the `error serving conn: error reading fcall: EOF` logspam during builds.


lmars:
@titanous the commit needs a DCO sign off.

PR #4187, opened 2017-07-19
test: Fix race in RouterSuite.TestAdditionalHttpPorts

titanous:
Ensure that the configuration change is applied to the router before moving forward, by watching for the up/down job events from the deployment.

This fixes a test failure that looks like:

```
++ 17:30:41.803 /go/src/github.com/flynn/flynn/build/bin/flynn-linux-amd64.4698f835b92d4d66342307a6dfede31b9a7d8bfc6b98001b79e04842a3587658 -a dashboard route add http -s dashboard-web -p 8080 dashboard.063ab384872a5776cad013a88d0e1b71.local
unknown_error: Something went wrong
```

### DIFF
--- a/test/test_router.go
+++ b/test/test_router.go
@@ -34,7 +34,6 @@ func (s *RouterSuite) TestAdditionalHttpPorts(t *c.C) {
 	// check a non-routed HTTP request to an additional port fails
 	req, err := http.NewRequest("GET", "http://dashboard."+x.Domain+":8080", nil)
 	t.Assert(err, c.IsNil)
-	req.SetBasicAuth("", x.Key)
 	res, err := http.DefaultClient.Do(req)
 	t.Assert(err, c.IsNil)
 	t.Assert(res.StatusCode, c.Equals, http.StatusNotFound)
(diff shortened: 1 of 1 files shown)


test/test_router.go
```diff
@@ -26,13 +26,14 @@ func (s *RouterSuite) TestAdditionalHttpPorts(t *c.C) {
 	defer x.Destroy()
 
 	// Test that setting added HTTP and HTTPS ports succeeds
-	t.Assert(x.flynn("/", "-a", "router", "env", "set", "ADDITIONAL_HTTP_PORTS=8080"), Succeeds)
-	t.Assert(x.flynn("/", "-a", "router", "env", "set", "ADDITIONAL_HTTPS_PORTS=8081"), Succeeds)
+	watcher, err := x.controller.WatchJobEvents("router", "")
+	t.Assert(err, c.IsNil)
+	t.Assert(x.flynn("/", "-a", "router", "env", "set", "ADDITIONAL_HTTP_PORTS=8080", "ADDITIONAL_HTTPS_PORTS=8081"), Succeeds)
+	t.Assert(watcher.WaitFor(ct.JobEvents{"app": {ct.JobStateUp: 1, ct.JobStateDown: 1}}, 10*time.Second, nil), c.IsNil)
```
lmars: Doesn't `env set` already do this?

titanous: No, for two reasons:

1. It's a `all-at-once` deploy, which means that the old formation is just scaled to zero at the end without waiting.
2. Even if it was a `one-by-one` deploy, the deployer waits for the discoverd registration to be pulled, not for the job to exit, so requests can still end up at the wrong router instance.

lmars: Ah of course, no router draining for the router itself :)

Issue #4228, opened 2017-08-25
Python buildpack, failing with undescriptive error


lukaspj:
Hello!

I receive the following error when pushing my django application:

```
Counting objects: 130, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (125/125), done.
Writing objects: 100% (130/130), 38.91 KiB | 0 bytes/s, done.
Total 130 (delta 41), reused 0 (delta 0)
remote: -----> Building django-test-1...
remote: -----> Python app detected
remote: -----> Installing python-3.6
remote: -----> Installing pip
remote: -----> Installing requirements with pip
remote:        /builder/buildpacks/000006_heroku-buildpack-python/bin/steps/pip-install: line 7: /app/.heroku/python/bin/pip: No such file or directory
remote: ERROR: Build failed: exec: job exited with status 1
```

Now, I'm certain that I'm doing something wrong. However, the error message is not very descriptive, how can I figure out what "No such file or directory" refers to?

For reference, my Profile is the following:
```web: gunicorn mysite.wsgi:application --log-file -```

And my requirements:
```
django ~= 1.11
gunicorn ~= 19.7.1
```

**Comment by lukaspj** [comment]
So, based no the docs which says the following:

```
Supported Runtimes
The latest python-2.7 and python-3.6 are officially supported, but any runtime between 2.4.4–3.6.1 can be used, including PyPy runtimes. See the buildpack’s GitHub page for a full list.
```

I interpreted it as "write python-3.6 in your runtime.txt, that is the officially supported". Maybe it should be clarified as `python-3.6.x`?

Anyways, this was clearly an error on my end. I'm both new to Flynn and Django, so I wasn't aware that Python 3.6 wasn't a valid version number.

PR #4229, opened 2017-08-25
Clarification of Python version numbers

lukaspj:
This is an attempt to avoid new users making the same mistake as I did, explained in #4228 . It is a minor clarification of the Python version numbers.

### DIFF
--- a/docs/content/languages/python.md
+++ b/docs/content/languages/python.md
@@ -33,10 +33,7 @@ python-2.7.8
 
 ### Supported Runtimes
 
-The latest `python-2.7.x` and `python-3.6.x` are officially supported, but any
-runtime between 2.4.4–3.6.2 can be used, including PyPy runtimes. See
-[the buildpack's GitHub page](https://github.com/heroku/heroku-buildpack-python/tree/master/builds/runtimes)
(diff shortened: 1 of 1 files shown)


docs/content/languages/python.md
```diff
@@ -33,8 +33,8 @@ python-2.7.8
 
 ### Supported Runtimes
 
-The latest `python-2.7` and `python-3.6` are officially supported, but any
-runtime between 2.4.4–3.6.1 can be used, including PyPy runtimes. See
+The latest `python-2.7.x` and `python-3.6.x` are officially supported, but any
+runtime between 2.4.4–3.6.2 can be used, including PyPy runtimes. See
```
lmars: This will continue to become stale over time, we should just reference the buildpacks [own documentation](https://github.com/heroku/heroku-buildpack-python#specify-a-python-runtime).

lukaspj: I concur, how is the rewording in https://github.com/flynn/flynn/pull/4229/commits/fc1652b67267916b9ead2b7d2daed63716c59d43 ?

lmars:
Thanks!

PR #4230, opened 2017-08-27
Remove Google DNS server as default recursors

### LINKED ISSUE #4227: discoverd ignores current DNS recursors
We run into a problem with discoverd service overriding 
 [current DNS recursors](https://github.com/flynn/flynn/blob/master/discoverd/main.go#L133)  with [default ones](https://github.com/flynn/flynn/blob/master/discoverd/main.go#L571 ). "recursor" option is not passed from command line and `StringVar` will default it to "8.8.8.8,8.8.4.4".

Issue happens whenever a new discoverd release is created. For example running `flynn -a discoverd env set DEBUG=true`  will lead to discoverd restart and hitting this [path](https://github.com/flynn/flynn/blob/master/discoverd/main.go#L118) effectively overriding upstream dns resolvers returned by `waitHostDNSConfig` Looks like checking for existing resolvers should fix the problem

```golang
if len(opt.Recursors) > 0 && len(resolvers) == 0 {
	resolvers = opt.Recursors
}
```

crawlik:
The fix address the following [issue](https://github.com/flynn/flynn/issues/4227) .

Fixes #4227.

### DIFF
--- a/discoverd/main.go
+++ b/discoverd/main.go
@@ -568,7 +568,7 @@ func (m *Main) ParseFlags(args ...string) (Options, error) {
 	fs.StringVar(&opt.Host, "host", "", "advertised hostname")
 	fs.StringVar(&opt.Addr, "addr", ":1111", "address to serve http and raft from")
 	fs.StringVar(&opt.DNSAddr, "dns-addr", "", "address to service DNS from")
-	fs.StringVar(&recursors, "recursors", "8.8.8.8,8.8.4.4", "upstream recursive DNS servers")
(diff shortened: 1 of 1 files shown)


crawlik:
Not sure what is wrong with CI build. The test passes in dev VM 
```
script/run-integration-tests -f TestJavaBuildpack
...
Cluster "default" added and set as default.
17:04:29.541 START: test_git_deploy.go:93: GitDeploySuite.TestJavaBuildpack
17:05:40.961 PASS: test_git_deploy.go:93: GitDeploySuite.TestJavaBuildpack	71.419s
OK: 1 passed
```

lmars:
Thanks!

PR #4246, opened 2017-09-15
util/cloner: Fix app name rewrite order

titanous:
Ensure that `appname-web` is rewritten after` appname-extra-proc-web` to avoid improperly rewriting app names where the destination app name contains the source name as a prefix.

### DIFF
--- a/util/cloner/flynn-clone.sh
+++ b/util/cloner/flynn-clone.sh
@@ -87,12 +87,12 @@ main() {
         .processes |= (map_values(.env = (if(.env) then with_entries(select(.key | in($whitelist))) else null end)))
       ' \
       < "${tmp}/src.json" | \
-      sed "s|${src_app}-web|${dst_app}-web|g" | \
-      sed "s|${src_app}-\(.*\)-web|${dst_app}-\\1-web|g" \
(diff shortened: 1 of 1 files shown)


util/cloner/flynn-clone.sh
```diff
@@ -87,12 +87,12 @@ main() {
...
-      sed "s|${src_app}-web|${dst_app}-web|g" | \
-      sed "s|${src_app}-\(.*\)-web|${dst_app}-\\1-web|g" \
+      sed "s|${src_app}-\(.*\)-web|${dst_app}-\\1-web|g" | \
+      sed "s|${src_app}-web|${dst_app}-web|g" \
       > "${tmp}/dst.json"
   else
-    sed "s|${src_app}-web|${dst_app}-web|g" "${tmp}/src.json" | \
-    sed "s|${src_app}-\(.*\)-web|${dst_app}-\\1-web|g" \
+    sed "s|${src_app}-\(.*\)-web|${dst_app}-\\1-web|g" | \
+    sed "s|${src_app}-web|${dst_app}-web|g" "${tmp}/src.json" \
```
lmars: The path `"${tmp}/src.json"` needs to be on the line above.

titanous: Ah, yeah. Fixed.

PR #4293, opened 2017-11-29
dashboard: Update yajl-ruby

josephglanville:
Update from 1.2.1.


lmars:
The dashboard build was failing with:

```
/app/node_modules/node-sass/node_modules/request/node_modules/hawk/node_modules/boom/lib/index.js:5
const Hoek = require('hoek');
^^^^^
SyntaxError: Use of const in strict mode.
```

I made a minimal upgrade to Node.js 4.x and node-sass@3.3.3 which seems to have fixed it.

@josephglanville your commit is missing a DCO sign-off though.

PR #4305, opened 2017-12-19
CI updates

titanous:
Update to 16.04, scripts for GCE installation.


titanous:
@lmars I think this is ready to go.

PR #4347, opened 2018-02-28
cli: show env vars from job first

### LINKED ISSUE #4108: Per-process env variables not reported correctly
`flynn env -a APP -t JOB get VARIABLE` returns the value set for VARIABLE on APP, not the one set on JOB. This is probably a simple fix to the `env get` code.

imjorge:
Show env var from job if specified, instead from app.

Fixes #4108


lmars:
Thanks!

PR #4367, opened 2018-04-16
tarreceive: API to convert Docker tar layers

lmars:
This replaces the `docker-receive` app with a much simpler `tarreceive` app which just receives tar image layers from `docker save`, converts them to SquashFS and stores them in the blobstore, leaving the client to construct an image manifest and deploy it.

The main reason for this work is to remove the large dependency on an old version of docker so that we can vendor newer libraries like [cAdvisor](https://github.com/google/cadvisor) without descending into dependency hell 🙈

It also paves the way for building a tool to deploy [OCI images](https://github.com/opencontainers/image-spec) and integrating with [BuildKit](https://github.com/moby/buildkit).

PR #4373, opened 2018-04-27
metrics: Add Prometheus and Grafana

lmars:
This adds a `metrics` app which runs [Prometheus](https://prometheus.io/) and [Grafana](https://grafana.com/).

There are a wealth of metrics we'd like to collect, but for now I've just added the following:

* `host_running_jobs`
* `router_backend_http_conn_errors_total`
* `router_backend_http_status_total`
* `router_backend_http_latency_seconds`

There is a Prometheus wrapper which watches discoverd services (currently just `flynn-host` and `router-api`) and re-configures Prometheus by editing files in `/etc/prometheus`.

Grafana is available at `https://metrics.${CLUSTER_DOMAIN}` with an admin user `flynn` and password available in `flynn -a metrics env get -t grafana GF_SECURITY_ADMIN_PASSWORD`.

TODO:

- [x] add Grafana Host dashboard
- [ ] add Grafana Router dashboard
- [x] add tests for router metrics


philiplb:
Does this Grafana-setup use Flynns PostgreSQL for persistence? Then I could replace our https://github.com/philiplb/grafana-on-flynn with this. :)

host/http.go
```diff
@@ -28,8 +28,9 @@ import (
 	"github.com/flynn/flynn/pkg/tufutil"
 	"github.com/flynn/flynn/pkg/version"
 	tuf "github.com/flynn/go-tuf/client"
-	"github.com/julienschmidt/httprouter"
 	"github.com/inconshreveable/log15"
+	"github.com/julienschmidt/httprouter"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
```
titanous: It looks like this needs to be vendored?

titanous:
This should have a few tests.

metrics/img/grafana.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+URL="https://s3-us-west-2.amazonaws.com/grafana-releases/release/grafana-5.0.4.linux-x64.tar.gz"
```
titanous: v5.1 is now available.

metrics/main.go
```diff
@@ -0,0 +1,196 @@
...
+	log.Info("writing global config file", "path", globalConfigPath)
+	if err := os.MkdirAll(filepath.Dir(globalConfigPath), 0755); err != nil {
+		return fmt.Errorf("error creating global config directory: %s", err)
+	}
+	configFile, err := os.OpenFile(globalConfigPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return fmt.Errorf("error opening global config file: %s", err)
+	}
+	defer configFile.Close()
+	if err := globalConfig.Execute(io.MultiWriter(configFile, os.Stdout), &globalConfigData{jobs}); err != nil {
```
titanous: Is there a reason this is being written to stdout?

lmars:
@philiplb I have switched Grafana to use a PostgreSQL backend in a30b8ef.

@titanous comments addressed.

I've added the following host metrics (plus a rudimentary Host dashboard in Grafana):

```
host_memory_total_bytes
host_memory_avail_bytes
host_load_1m
host_load_5m
host_load_15m
host_jobs_active
host_jobs_starting
host_jobs_running
```

I've also added an integration test.

I'm going to leave the Router dashboard as a TODO for a further PR when we can add more metrics.

Ready for a review.

philiplb:
@lmars Awesome. :-)

host/metrics.go
```diff
@@ -0,0 +1,215 @@
...
+type metrics struct {
+	state *State
+	log   log15.Logger
+
+	memFreeDesc  *prometheus.Desc
+	memAvailDesc *prometheus.Desc
+
+	load1mDesc  *prometheus.Desc
+	load5mDesc  *prometheus.Desc
+	load15mDesc *prometheus.Desc
```
titanous: Would it make more sense to add the [node exporter](https://github.com/prometheus/node_exporter)?

titanous:
How does this handle fault tolerance/HA? What happens if the host that the prometheus container is running on is rebooted?

host/metrics.go
lmars: I did initially try and use it but I hit a couple of issues:

* there were vendoring issues due to our outdated docker fork (which will be resolved by #4367)
* it is tied quite tightly with [a flag parsing library](https://github.com/prometheus/node_exporter/blob/master/collector/collector.go#L71), so calling [`NewNodeCollector`](https://github.com/prometheus/node_exporter/blob/master/collector/collector.go#L83) doesn't actually give us any metrics (they're all disabled)

I decided to just do a few metrics manually until we get round to being able to use [cAdvisor](https://github.com/google/cadvisor).

lmars:
> How does this handle fault tolerance/HA? What happens if the host that the prometheus container is running on is rebooted?

Prometheus stores its data on a volume which will be re-attached by the scheduler, but beyond that it isn't very HA (much like the Redis appliance). We could try a HA storage backend, but I was focussing initially on just having it running.

router/proxy/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+		"app":     backend.App,
+		"service": backend.Service,
+		"job_id":  backend.JobID,
+	}).Inc()
+}
+
+var backendHTTPStatusMetric = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "router_backend_http_status_total",
+		Help: "Router backend HTTP status codes",
```
titanous: This should be `router_backend_http_responses_total`, the status code is just another label.

host/metrics.go
```diff
@@ -0,0 +1,215 @@
...
+	labels := prometheus.Labels{
+		"host": state.id,
+	}
+
+	return prometheus.Register(&metrics{
+		state: state,
+		log:   log,
+		memFreeDesc: prometheus.NewDesc(
+			"host_memory_total_bytes",
+			"Host total memory in bytes",
```
titanous: this is called `memFreeDesc`, but says "total memory", total memory doesn't change, so this should be called `host_memory_free_bytes`, right?

host/metrics.go
```diff
@@ -0,0 +1,215 @@
...
+			"Host active jobs",
+			nil, labels,
+		),
+		startingJobsDesc: prometheus.NewDesc(
+			"host_jobs_starting",
+			"Host starting jobs",
+			nil, labels,
+		),
+		runningJobsDesc: prometheus.NewDesc(
+			"host_jobs_running",
```
titanous: These three should end with `_total`.

titanous:
The [custom scripts](https://github.com/dimitri/pgextwlist#custom-scripts) functionality allows creating a script that will `GRANT` the appropriate permissions to access the data types/functions.

host/metrics.go
BrynjarM: The standard output of cAdvisor with small modification (join the name and type with the container id) provides great insights and might be worth aiming for in the first release (at least for container cpu and memory).

Recently configured it on my instance with a small replace script between cAdvisor and Prometheus
flynn-metrics

router/proxy/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+	},
+	[]string{"app", "service", "job_id", "status"},
+)
+
+func trackBackendHTTPStatus(backend *router.Backend, status int) {
+	backendHTTPStatusMetric.With(prometheus.Labels{
+		"app":     backend.App,
+		"service": backend.Service,
+		"job_id":  backend.JobID,
```
titanous: I think these should be `backend_app`, `backend_service`, and `backend_job_id`, and then we should also have `job_id` and maybe `host` for the router instance?

lmars:
@titanous 

> The custom scripts functionality allows creating a script that will GRANT the appropriate permissions to access the data types/functions.

I gave it a try, but I'm getting `@current_user@` being `postgres` and `@database_owner@` being `flynn`, neither of which are the user that actually needs the permissions on the schema.

I got around this by setting the database owner as the random user [here](https://github.com/flynn/flynn/blob/master/appliance/postgresql/cmd/flynn-postgres-api/main.go#L86) with `ALTER DATABASE name OWNER TO new_owner` but I don't really know if that is desirable?

Even with that, I sometimes get a syntax error as it seems `@database_owner@` is only quoted if it begins with a letter:

```
GRANT ALL ON SCHEMA prometheus TO @database_owner@;
```

For example, when the user is `ffbec53e3764df12d4f6827a415fa910` the script executes just fine, but when it is `17fd84f3d9b45cbf0af3e2c451d1657e` I get:

```
DEBUG:  Executing custom script "/usr/local/share/postgresql/extwlist/pg_prometheus/after-create.sql"
ERROR:  syntax error at or near "17" at character 35
```

titanous:
>I got around this by setting the database owner as the random user here with `ALTER DATABASE name OWNER TO new_owner` but I don't really know if that is desirable?

Yep, that's fine.

>Even with that, I sometimes get a syntax error as it seems @database_owner@ is only quoted if it begins with a letter:

Wrap it in quotes: `"@database_owner@"`

host/metrics.go
lmars: @BrynjarM nice! We'll definitely be integrating cAdvisor, there are just some blockers before we can do so that I don't want to get in the way of merging this PR.

host/metrics.go
lmars: I've renamed it `memTotalDesc` and it is supposed to be the total memory so that it can be used along with available memory (`host_memory_avail_bytes`) to calculate percentage used. Available memory is the most useful metric here though.

lmars:
@titanous comments addressed, in particular:

* host job metrics now end in `_total`
* router metric labels are prefixed with `backend_`
* all metrics get `job_id` and `host_id` labels if set in the discoverd instance metadata (see [here](https://github.com/flynn/flynn/blob/metrics/metrics/main.go#L181-L187))
* Prometheus now uses PostgreSQL to store metrics

host/metrics.go
```diff
@@ -0,0 +1,215 @@
...
+	"sync"
+
+	host "github.com/flynn/flynn/host/types"
+	"github.com/inconshreveable/log15"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func registerMetrics(state *State, log log15.Logger) error {
+	labels := prometheus.Labels{
+		"host_id": state.id,
```
titanous: Is this redundant?

metrics/main.go
```diff
@@ -0,0 +1,223 @@
...
+				}
+				switch event.Kind {
+				case discoverd.EventKindCurrent:
+					if !current {
+						current = true
+						firstErr <- writeJobConfig(job, instances)
+					}
+				case discoverd.EventKindUp:
+					instances[event.Instance.Addr] = event.Instance
+					if err := writeJobConfig(job, instances); err != nil {
```
titanous: Should this check `current` first?

hadifarnoud:
do we know how much more memory this addition will need?

philiplb:
Cautious ping? :)

philiplb:
Any chance maybe this could be picked up again and finished?

hadifarnoud:
is this PR stalled?

titanous:

Issue #1452, opened 2015-04-14
Document the dashboard development workflow


jvatic:
1. launch the vagrant development box (`vagrant up` from the project root)
1. `vagrant ssh`
1. `make`
1. `bootstrap-flynn`
1. `./script/dashboard-dev compile` (repeat this step after every change)
1. `CONTROLLER_KEY="<KEY>" ./script/dashboard-dev run` (see the output of `bootstrap-flynn` for the `CONTROLLER_KEY`)

jvatic:
Draft: https://gist.github.com/jvatic/df346cfdf1632ca0ccd3

philiplb:
On the vagrant box 20160624.1 I had to install some dependencies in order to make "go run compiler.go" work:
```bash
sudo apt-get update
sudo apt-get install nodejs
sudo apt-get install npm
apt-get install ruby-bundler
sudo gem install yajl-ruby
# this one was needed as well and feels hacky
sudo ln -s `which nodejs` /usr/bin/node
```

lmars:
FYI I've added a convenience script in #3024.

PR #4402, opened 2018-06-12
dashboard: Show processes diff in confirmation dialog when scaling

jvatic:
screen shot 2018-06-12 at 11 20 35

Also updated the dashboard-dev script (related: #1452)

### DIFF
--- a/dashboard/api_test.go
+++ b/dashboard/api_test.go
@@ -22,16 +22,19 @@ const (
 func Test(t *testing.T) { TestingT(t) }
 
 type S struct {
-	srv *httptest.Server
+	srv        *httptest.Server
+	cookiePath string
 }
 
 var _ = Suite(&S{})
 
 func (s *S) SetUpSuite(c *C) {
+	cookiePath := "/"
 	s.srv = httptest.NewServer(APIHandler(&Config{
 		SessionStore:  sessions.NewCookieStore([]byte("session-secret")),
 		LoginToken:    testLoginToken,
 		ControllerKey: testControllerKey,
+		CookiePath:    cookiePath,
 	}))
 }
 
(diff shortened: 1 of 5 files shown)


titanous:
typo in commit message: "dashbaord"

dashboard/bindata.go
```diff
@@ -1,11 +1,21 @@
...
 
 func Asset(path string) ([]byte, error) {
-	return nil, nil
+    file, err := os.Open(path)
+    if err != nil {
+        return nil, err
+    }
+    var buf bytes.Buffer
+    _, err = io.Copy(&buf, file)
+    return buf.Bytes(), err
```
titanous: Use `ioutil.ReadFile`

script/dashboard-dev
```diff
@@ -55,27 +60,20 @@ run() {
...
-    _, err = io.Copy(&buf, file)
-    return buf.Bytes(), err
+	return ioutil.ReadFile(path)
 }
 
 func AssetInfo(path string) (os.FileInfo, error) {
-    return os.Stat(path)
+	return os.Stat(path)
 }
 BINDATA
```
jvatic: We possibly don't need this. I'll double check and see if any of the compile steps overwrite the checked-in version of `bindata.go` (this was previously the case).

jvatic: Confirmed. This isn't needed.

dashboard/api_test.go
```diff
@@ -22,16 +22,19 @@ const (
 func Test(t *testing.T) { TestingT(t) }
 
 type S struct {
-	srv *httptest.Server
+	srv        *httptest.Server
+	cookiePath string
```
titanous: This looks unused?

jvatic: Oops, supposed to be used down in one the assertion for the Location header. I'll create another PR.

jvatic: https://github.com/flynn/flynn/pull/4403

PR #4425, opened 2018-08-29
test: Run the CI runner in Flynn

lmars:
This updates the CI runner so that it will run inside a Flynn cluster (rather than running directly on a host machine using systemd), which makes it much easier to cleanup leaking VMs (just restart the job) and to test changes to the CI runner.

Since the CI runner needs access to loop and KVM devices, and needs to configure networking, I've added the concept of "job profiles" which give the job the extra devices and capabilities it needs:

- the `loop` profile finds an available loop device and creates it at `/dev/loop0` inside the job
- the `kvm` profile gives access to the `/dev/kvm` and `/dev/net/tun` devices as well as adding the `CAP_NET_ADMIN` capability for creating bridge and TAP devices
- the `zfs` profile gives access to the `/dev/zfs` device and `/dev/zvol` symlinks

I've also updated `dep` to `v0.5.0` as I was having issues updating other vendored dependencies.


host/profiles.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil, fmt.Errorf("unexpected data in %s: %q", path, data)
+	}
+	dev := &device{}
+	dev.major, err = strconv.ParseInt(s[0], 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing device major number from %q: %s", data, err)
+	}
+	dev.minor, err = strconv.ParseInt(s[1], 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing device minor number from %q: %s", data, err)
```
titanous: These two errors should include the path that failed to parse as well.

titanous:
What happens to the database of test runs? Do we need to move that to Postgres?

host/profiles.go
lmars: Done in b0a97b9.

PR #4434, opened 2018-09-13
dep: Cleanup vendor, remove golang.org/x/crypto/acme

jvatic:


Gopkg.lock
```diff
@@ -1117,6 +1117,7 @@
     "github.com/tiborvass/uniline",
     "github.com/vanillahsu/go_reuseport",
     "github.com/vishvananda/netlink",
+    "golang.org/x/crypto/acme",
     "golang.org/x/crypto/acme/autocert",
```
lmars: Did you run this through `dep ensure`? I think it's the case that we need to remove `golang.org/x/crypto/acme/autocert` rather than add `golang.org/x/crypto/acme` (we removed the dependency on autocert when moving the CI runner to run in Flynn, see https://github.com/flynn/flynn/commit/f1d9d490a7dddc820af2b3398298da19ed1d483f#diff-e3b09e9ffe6a8fd563ad4e263b4eb343L46)

jvatic: Ah, I was operating under the assumption it should be there due to it being in `vendor`.

jvatic: No worries! I've updated this PR to remove it completely.

PR #4466, opened 2018-11-08
Allow specifying an alternative 'docker' binary via DOCKER_BIN

### LINKED ISSUE #4465: [feature] support alternative container cli in 'flynn docker ...' (podman)
It would be nice if `flynn docker ...` commands could support different container clis compatible with `docker`. Specifically, I am recently using [`podman`](https://www.projectatomic.io/blog/2018/02/reintroduction-podman/), which is almost fully compatible to the `docker` cli but can run entirely in user-namespaces without any daemons etc.

Looking at the [subcommand sources](https://github.com/flynn/flynn/blob/master/cli/docker.go), most - if not all - docker commands are implemented with `os/exec.Command`. If we could simply allow for replacing `"docker"` with `"podman"` in all those instances I believe this should be doable?

I am currently trying to find incompatabilities by symlinking `podman` as `docker` in my `PATH`:

- [x] login fails with a TLS error, I am not sure yet how to add the CA cert in `podman`
  - just add `docker.$CLUSTER` to `[registries.insecure]`? https://www.projectatomic.io/blog/2018/05/podman-tls/ --> this does not "magically" work
  - put the CA certificate in `/etc/containers/certs.d/docker.$CLUSTER/*.crt` --> success
- [x] podman uses `.ContainerConfig` instead of `.Config` in `inspect` output
  - ~this currently blocks further investigation because it requires patching either `podman` or `flynn`~ see patch in https://github.com/flynn/flynn/issues/4465#issuecomment-437087854
  - see discussion in https://github.com/containers/libpod/issues/305
  - Docker does use both `Config` and `ContainerConfig`. So I _think_ this patch wouldn't even break compatability with `docker`?

ansemjo:
The patch introduces the ability to use different `docker` - or compatible - binaries to the `flynn docker` subcommands by providing their path in the environment variable `DOCKER_BIN`. This is useful if you want to use a different Docker build or an alternative CLI like [`podman`](https://github.com/containers/libpod) temporarily.

fixes #4465

### DIFF
--- a/cli/docker.go
+++ b/cli/docker.go
@@ -44,7 +44,7 @@ Example:
 	Assuming you have a Docker image tagged "my-custom-image:v2":
 
 	$ flynn docker push my-custom-image:v2
-	flynn: getting image config with "docker inspect -f {{ json .Config }} my-custom-image:v2"
+	flynn: getting image config with "docker inspect -f {{ json .ContainerConfig }} my-custom-image:v2"
 	flynn: tagging Docker image with "docker tag my-custom-image:v2 docker.1.localflynn.com/my-app:latest"
 	flynn: pushing Docker image with "docker push docker.1.localflynn.com/my-app:latest"
 	The push refers to a repository [docker.1.localflynn.com/my-app] (len: 1)
@@ -201,7 +201,7 @@ func runDockerPush(args *docopt.Args, client controller.Client) error {
 	}
 
 	// get the image config to determine Cmd, Entrypoint and Env
-	cmd := exec.Command("docker", "inspect", "-f", "{{ json .Config }}", image)
+	cmd := exec.Command("docker", "inspect", "-f", "{{ json .ContainerConfig }}", image)
 	log.Printf("flynn: getting image config with %q", strings.Join(cmd.Args, " "))
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {


titanous:
Given how rare this use case is, can you just have a binary in your `PATH` called Docker that links to whatever other binary you want to run?

ansemjo:
Surely you could and that's essentially what I did for initial testing:

    ln -s $(which podman) ~/.local/bin/docker

However, the first commit 6ee9e96 is still required for this to work (see initial comment in #4465).

titanous:
Yeah, remove the other commit and I'll merge.

ansemjo:
Done, I force-pushed the `HEAD~1` commit ..

ansemjo:
Oh well, that's fun .. because `podman inspect -f '{{ json .ContainerConfig }}'` provides the same output as `docker inspect -f '{{ json .Config }}'` it seems. And `.Config` just doesn't exist in podman's output ...

```
$ podman inspect -f '{{ json .ContainerConfig }}' configtest | jq '{ Entrypoint: .Entrypoint, Cmd: .Cmd, Env: .Env }'
{
  "Entrypoint": [
    "/bin/echo"
  ],
  "Cmd": [
    "hello world"
  ],
  "Env": [
    "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
    "FOO=BAR"
  ]
}

$ docker inspect -f '{{ json .ContainerConfig }}' configtest | jq '{ Entrypoint: .Entrypoint, Cmd: .Cmd, Env: .Env }'
{
  "Entrypoint": null,
  "Cmd": [
    "/bin/sh",
    "-c",
    "/bin/echo image command"
  ],
  "Env": [
    "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
    "FOO=BAR"
  ]
}

$ docker inspect -f '{{ json .Config }}' configtest | jq '{ Entrypoint: .Entrypoint, Cmd: .Cmd, Env: .Env }'
{
  "Entrypoint": [
    "/bin/echo"
  ],
  "Cmd": [
    "hello world"
  ],
  "Env": [
    "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
    "FOO=BAR"
  ]
}
```

Would a fallback work in this case, first checking `Config` and then `ContainerConfig` if that does not exist?

ansemjo:
I had opened an issue upstream which was closed pretty quickly. I'll retest once it hits the next release of `podman`.

PR #4512, opened 2019-03-06
remove no-op append calls

quasilyte:
The `append(s)` call has no effect, it just returns `s`.

Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>


titanous:
Thanks!

PR #4532, opened 2019-07-08
Add support for Ubuntu 18, update dependencies

### LINKED ISSUE #4392: Ubuntu 18.04 Support
Hi everyone,
I already asked this question in flynn irc channel. Just wantet to know if there are any plans to support ubuntu 18.04?

regards

titanous:
Closes #4392

Note that there is an untriaged bug in Linux 4.15/ZoL that causes hangs when using a sparse file to back the `flynn-default` pool. The workaround is to use a physical device to back the pool instead.


Gopkg.toml
```diff
@@ -18,6 +18,10 @@ required = [
   branch = "master"
   name = "github.com/cupcake/jsonschema"
 
+[[override]]
+  name = "github.com/docker/docker"
+  branch = "master"
+
```
lmars: Is this override necessary? I thought we dropped the docker dependency?

gitreceive/receiver/flynn-receive.go
```diff
@@ -96,6 +86,36 @@ Options:
...
+	if stack := prevRelease.Env["FLYNN_STACK"]; stack != "" {
+		switch stack {
+		case "heroku-18":
+		case "cedar-14":
+			fmt.Printf("WARNING: The cedar-14 stack is deprecated and does not receive security updates.")
+			fmt.Printf("WARNING: Unset FLYNN_STACK to use the default stack.")
+			slugbuilderImageID = os.Getenv("SLUGBUILDER_14_IMAGE_ID")
+			slugrunnerImageID = os.Getenv("SLUGRUNNER_14_IMAGE_ID")
+			stackName = "cedar-14"
+			cfStackName = "cflinuxfs3"
```
lmars: Should this be the legacy version `cflinuxfs2`?

Gopkg.toml
titanous: Removed, thanks.

gitreceive/receiver/flynn-receive.go
titanous: Fixed, thanks.

kenaniah:
@titanous Could you please link us to the bug and provide (or link to) an example of how the workaround could be implemented? I've got a new v20190715.0 cluster that appears to be hanging on gitreceive / slugbuilder and it might be related to that.

PR #4543, opened 2019-07-31
v2 dashboard

jvatic:
This is the new v2 dashboard which uses the gRPC controller.

Features:

- [x] Scale
- [x] Edit environment variables
- [x] Edit metadata
- [x] See release history
- [x] Rollback to any point in release history


dashboardv2/.prittierrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
amtrack: Shouldn't this filename be `.prettierrc.js` (with an `e` instead of `i`)?

jvatic:
Rebased on master and squashed commits

builder/img/protocjs.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+set -eo pipefail
+
+# install googleapis common protos
+shasum="9584b7ac21de5b31832faf827f898671cdcb034bd557a36ea3e7fc07e6571dcb"
+curl -fSLo /tmp/common-protos.tar.gz "https://github.com/googleapis/googleapis/archive/common-protos-1_3_1.tar.gz"
```
lmars: Is this needed?

builder/manifest.json
```diff
@@ -64,6 +64,13 @@
         "script": "builder/img/protoc.sh"
       }]
     },
+    {
+      "id": "protocjs",
+      "base": "protoc",
+      "layers": [{
+        "script": "builder/img/protocjs.sh"
```
lmars: I think we should just put the JS stuff in the `protoc` image

builder/manifest.json
```diff
@@ -602,6 +609,76 @@
         ]
       }
     },
+    {
+      "id": "dashboardv2-base",
+      "base": "go",
+      "layers": [{
+        "script": "builder/img/dashboardv2-base.sh",
+        "inputs": [
+          "controller/api/*.proto"
```
lmars: Is this input required for this layer?

builder/manifest.json
```diff
@@ -602,6 +609,76 @@
...
+           "dashboardv2/src/generated/google/api/*.js",
+           "dashboardv2/src/KeyValueEditor/*.ts",
+           "dashboardv2/src/KeyValueEditor/*.tsx"
+         ],
+         "limits": { "memory": "3G", "temp_disk": "3G" }
+       }
+      ]
+    },
+    {
+      "id": "dashboardv2",
```
lmars: Assuming we're no longer going to work on the old dashboard, should we just call this one `dashboard` now and remove the old build (we can always reference the old dashboard layers from the TUF repo if we want to use them)

controller/worker/domain_migration/domain_migration.go
```diff
@@ -167,6 +167,15 @@ func (m *migration) Run() error {
 				"URL":                  fmt.Sprintf("https://dashboard.%s", m.dm.Domain),
 			},
 		},
+		{
+			AppName: "dashboardv2",
+			NewEnv: map[string]string{
+				"DEFAULT_ROUTE_DOMAIN": m.dm.Domain,
+				"CONTROLLER_DOMAIN":    fmt.Sprintf("controller-grpc.%s", m.dm.Domain),
+				"CONTROLLER_HOST":      fmt.Sprintf("https://controller-grpc.%s", m.dm.Domain),
```
lmars: Should these be `controller.<domain>`?

dashboardv2/api.go
```diff
@@ -0,0 +1,284 @@
...
+	api := &API{
+		conf:       conf,
+		assetCache: make(map[string][]byte),
+	}
+
+	router := httprouter.New()
+	router2 := httprouter.New()
+
+	prefixPath := func(p string) string {
+		return p // TODO(jvatic): handle path prefix?
```
lmars: Can this be dropped?

bootstrap/manifest_template.json
```diff
@@ -711,6 +711,52 @@
...
+    }
+  },
+  {
+    "id": "dashboardv2-route",
+    "action": "add-route",
+    "app_step": "dashboard",
+    "cert_step": "controller-cert",
+    "type": "http",
+    "service": "dashboardv2-web",
+    "domain": "dashboardv2.{{ getenv \"CLUSTER_DOMAIN\" }}",
```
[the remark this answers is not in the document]
titanous: Yeah, we can delete the old dashboard entirely.

builder/build.go
```diff
@@ -508,6 +515,37 @@ func (b *Builder) BuildImage(image *Image) error {
...
+					fmt.Sprintf("mkdir -p %s", outDir),
+					fmt.Sprintf(`protoc -I /usr/local/include -I %s \
+						--plugin="protoc-gen-ts=/usr/local/bin/protoc-gen-ts"\
+						--js_out="import_style=commonjs,binary:%s" \
+						--ts_out="service=grpc-web:%s" \
+						/usr/local/include/google/api/annotations.proto \
+						/usr/local/include/google/api/http.proto \
+						%s
+`, dir, outDir, outDir, strings.Join(paths, " ")),
+				)
```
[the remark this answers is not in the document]
jvatic: @lmars I think it would need to be more like

```
protobuild: [
  {
    src: "input/dir".
    dst: "output/dir",
    lang: "js"
  }
]
```

as the js/ts build (i.e. the dashboard) will have a different output directory than input (unlike the go build). I don't think we need a list of langs as it seems unlikely there would be multiple being build to the same output dir (and js/ts should always be part of the same build). WDYT?

builder/img/protocjs.sh
jvatic: Hmm, that's used for the `/usr/local/include/google/api/...` args to `protoc` above, though those don't appear required anywhere so I think this is just dead code (there would be an error somewhere in the build process if it is in fact required once it's removed, so I'll just remove it).

builder/build.go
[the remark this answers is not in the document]
lmars: SGTM 👍

builder/img/protocjs.sh
lmars: Yeh I think we removed the `google/api/...` imports

builder/manifest.json
jvatic: It shouldn't be. I think that's probably remanence of an earlier configuration.

dashboardv2/api.go
jvatic: Yeah, I don't think need the ability to mount the dashboard any deeper than to domain root.

PR #4544, opened 2019-08-01
builder/img: Add vim-tiny to heroku-18 image

titanous:


builder/img/heroku-18.sh
```diff
@@ -106,6 +106,7 @@ apt-get install -y --no-install-recommends \
     zip \
     pigz \
     daemontools \
+    vim-tiny \
```
lmars: I realise that the list was previously this way, but shouldn't this last item finish the line (i.e. _not_ have an escaped newline at the end)?

titanous: This is intentional, it means that any changes to the list only add/remove actual lines, the diff won't touch existing lines to add/remove a escaped newline.

lmars: Gotcha, I was just going mad thinking this would escape into the `cat` two lines beneath, but of course there is another newline in between.

PR #4563, opened 2019-11-16
builder: Downgrade host image to ubuntu-xenial

lmars:
The bionic ZFS packages are not compatible with the older xenial packages used in CI (due to a bug in recent versions of ZFS using backing files), and so the host image needs to be based off xenial so
that its ZFS package can be used to create zpools when starting test clusters.

This has no effect on the released flynn-host as the binary will continue to be built with the Go image (that is based on bionic) and the host image is only used for testing.

(This is a blind fix, I'll know if it's effective once this PR is opened and CI runs...)

### DIFF
--- a/builder/img/ubuntu-xenial.sh
+++ b/builder/img/ubuntu-xenial.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+TMP="$(mktemp --directory)"
+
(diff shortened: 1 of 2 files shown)


lmars:
VMs are now successfully booting in CI, so this change has worked (after a force push to update the base xenial URL to a 2019 image).

PR #4564, opened 2019-11-19
router: Implement "power of two random choices" algorithm

lmars:


router/proxy/transport.go
```diff
@@ -50,6 +50,106 @@ type transport struct {
...
+		}
+		l.Error("retriable dial error", "job.id", backend.JobID, "addr", backend.Addr, "err", err, "attempt", attempt)
+		// remove the backend now that we've tried it
+		backends = append(backends[:index], backends[index+1:]...)
+		attempt++
+		return err, true
+	}
+
+	// prioritise the sticky backend if it exists
+	for index, backend := range backends {
```
titanous: This should have a guard so that it doesn't run if `stickyBackend == ""`.

router/proxy/transport.go
```diff
@@ -50,6 +50,106 @@ type transport struct {
...
+			if err, shouldRetry := try(index); err == nil || !shouldRetry {
+				return err
+			}
+			break
+		}
+	}
+
+	// keep picking two random backends and trying the one with the least
+	// number of in flight requests
+	for len(backends) > 0 {
```
titanous: We should cap this at a few retries, maybe 3 or 4?

router/proxy/transport.go
```diff
@@ -50,6 +50,106 @@ type transport struct {
...
+		if len(backends) == 1 {
+			err, _ := try(0)
+			return err
+		}
+
+		// pick two distinct random backends
+		n1 := random.Math.Intn(len(backends))
+		n2 := random.Math.Intn(len(backends))
+		for n2 == n1 {
+			n2 = random.Math.Intn(len(backends))
```
titanous: We don't need to be perfect here, just do `n2+1 % len(backends)`.

lmars:
@titanous comments addressed in d47a3a7a (I force pushed the commit like we used to, but let me know if you'd prefer commits on top once the PR is opened, I know some people prefer that to help to see what got resolved)

titanous:
Yeah, going forward let's push changes as new commits, and don't forget to resolve the relevant conversations.

lmars:
> Yeah, going forward let's push changes as new commits, and don't forget to resolve the relevant conversations.

Will do (I've resolved all the current comments).

router/proxy/transport.go
```diff
@@ -50,6 +54,109 @@ type transport struct {
...
+	for len(backends) > 0 {
+		// if there is only one backend, try it and return
+		if len(backends) == 1 {
+			err, _ := try(0)
+			return err
+		}
+
+		// pick two distinct random backends
+		n1 := random.Math.Intn(len(backends))
+		n2 := (n1 + 1) % len(backends)
```
titanous:
```suggestion
		n2 := random.Math.Intn(len(backends))
		if n2 == n1 {
			n2 = (n2 + 1) % len(backends)
		}
```

router/proxy/transport.go
```diff
@@ -50,6 +54,109 @@ type transport struct {
...
 	stickyCookieKey   *[32]byte
 	useStickySessions bool
+
+	inFlightMtx      sync.Mutex
+	inFlightRequests map[string]int64
+}
+
+func (t *transport) trackRequestStart(backend *router.Backend) {
+	t.inFlightMtx.Lock()
+	defer t.inFlightMtx.Unlock()
```
titanous: `defer` has some overhead and this is a hot path, just unlock it after incrementing (make the same change in `trackRequestEnd`.

titanous:
Can you also run a one-off benchmark of request throughput against this branch and master to ensure this doesn't have a major perf impact?

router/proxy/transport.go
lmars: Updated in 62a723a.

router/proxy/transport.go
lmars: Updated in d71e611.

lmars:
@titanous here is the benchmark:

https://gist.github.com/lmars/a37a8472a39404a26be97d58ff6d2390

So no discernible difference really.

Issue #3010, opened 2016-06-28
Allow configuration of the deploy strategy from the CLI


quentez:
At the moment, it's possible to change the deploy strategy from `all-at-once` to `one-by-one` by directly changing the value in the controller DB:

```
flynn -a controller pg psql -- -c "UPDATE apps SET strategy = 'all-at-once' WHERE name = '<app_name>'"
```

It would be useful to have this as a configuration setting on the CLI, at least before we end up having other ways to set and version an app's configuration.

I would also like to suggest another strategy that would be `one-by-one-per-server`. Our reason for wanting a `one-by-one` deployment is to limit the impact on the app's performance, and in that regard it makes sense at a server level but not really at a cluster level (and the current `one-by-one` can take a really long time if the app is deployed across many nodes).

lmars:
I think we should move the deployment strategy (and perhaps the deployment timeout) to app metadata so we don't have to keep adding explicit support for new variables, so something like:

```
$ flynn meta set deploy.strategy=one-by-one
$ flynn meta set deploy.timeout=120
```

@titanous @josephglanville @jvatic WDYT?

> I would also like to suggest another strategy that would be `one-by-one-per-server`

I'll be doing some work on moving deployments to jobs once #2744 is merged so I will take this into consideration.

josephglanville:
I think the deployment strategy and other variables affecting how a job is deployed/scheduled should be per process rather than per app. If we could come up with a way to support it in the process type structure I think that would be ideal.

titanous:
> the deployment strategy and other variables affecting how a job is deployed/scheduled should be per process

I agree, this would be ideal.

lmars:
FYI the request for a "one per host" deployment strategy has been added in https://github.com/flynn/flynn/pull/4567.

titanous:
Flynn is unmaintained and our infrastructure will shut down on June 1, 2021. See [the README](https://github.com/flynn/flynn#flynn-is-unmaintained) for details.

PR #4567, opened 2019-11-30
controller: Add in-batches deploy strategy

lmars:
This adds a `one-per-host` deployment strategy that is similar to a `one-by-one` deployment but scales one job per host at each step of the deploy.

The actual strategy just scales in batches equal to the host count and relies on the scheduler distributing the jobs evenly across the hosts in the cluster to effectively scale one per host.

Ref #3010.


test/test_deployer.go
```diff
@@ -191,8 +192,40 @@ func (s *DeployerSuite) TestOneByOneStrategy(t *c.C) {
 	d.waitForDeploymentStatus("complete")
 }
 
+func (s *DeployerSuite) TestOnePerHostStrategy(t *c.C) {
```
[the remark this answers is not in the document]
lmars: The function name is pretty self explanatory...

lmars:
The original changeset missed the requirement that this should support a configurable batch size rather than just the number of hosts.

Given that, in fa56f82 I have:

- Renamed the new strategy to `in-batches`
- Added `flynn deployment batch-size <size>` to set the batch size
- Added `flynn deployment batch-size` to get the current batch size

If the batch size is not set explicitly, then the deployment will behave as originally implemented (i.e. use the host count as the batch size).

@titanous ready for review.

schema/controller/deployment.json
```diff
@@ -45,6 +45,10 @@
     "deploy_timeout": {
       "$ref": "/schema/controller/common#/definitions/deploy_timeout"
     },
+    "batch_size": {
```
titanous: I think this should be called `deploy_batch_size` here and in the database schema.

lmars: Updated in d9abd6ab

PR #4569, opened 2019-12-12
controller: Merge the router db and API into the controller

lmars:
This moves the router db tables and data into the controller database, and also merges the router REST API into the controller (which already handles route related requests, but previously acted as a proxy to the internal router API).

This paves the way for a more robust router API that supports atomic route changes across apps and is more tightly integrated with other controller data like deployments and releases.

The external facing functionality is the same apart from the following:

- there is no longer support for adding HTTP routes on ports other than the defaults (i.e. 80 and 443), which was previously controlled via the router `ADDITIONAL_HTTP(S)_PORT` environment variables (the plan is to reintroduce this in a later change with the addition of listeners as first class API objects)

- there is no longer an `AddCert` router API method that switches the certificate for all its dependent routes (whilst useful, this was not exposed in the controller and not called from any other code, this functionality will be added to the controller API in due course)

---

A note on the db migration:

The existing router db migrations (i.e. 1 to 9) have been added as-is to the controller migrations as 35-43, which means new clusters run all those migrations like they used to for the router, but we can check when restoring backups which of the migrations had previously run on the router, and just apply those when we do a router -> controller db import, leaving the controller to run the rest of the migrations as normal.

A further migration has also been added that introduces the `set_tcp_route_port` trigger that sets an unused port for TCP routes that don't explicitly set a port (this was previously done in the router API by consulting the current list of TCP listeners and picking an unused port, which isn't possible in the controller API).

---

Further planned changes not included in this PR:

- ~~switch the router to get route change events from the controller API rather than the database~~ **[UPDATE: 2019-12-18] this has been added to the PR**
- move static router configuration (e.g. default TLS certs, additional listeners etc.) into the controller API
- add certificate management to the controller API

---

/cc @titanous @jvatic


lmars:
I've updated the PR to include switching the router to use the controller API for route change events rather than the database (see 86016f2).

lmars:
FYI I rebased on master to fix a conflict in `controller/data/schema.go` (I needed to bump the migration numbers)

PR #4574, opened 2020-01-07
CI Updates

lmars:
This fixes an image reference in `TestReleaseImages` and also updates test VMs to log the systemd journal to help investigate why some VMs don't start in CI.

Opening as a draft as I suspect this will lead to a fix for the failing VMs once run in CI.


lmars:
Closing for now, will re-open once I've tracked down the issue (this PR didn't really help).

PR #4577, opened 2020-01-28
builder: add flags to speed up building specific images

jvatic:
This speeds up the build process when you know which image(s) need to be rebuilt.

Example usage:

```sh
$ vagrant up && vagrant ssh
# build "controller" and "status" images if not already cached
# and use layer cache for all other images
vagrant@flynn:~/go/src/github.com/flynn/flynn$ ./script/build-flynn --filter controller,status
# force building of the "controller" image and use layer cache for all other images
vagrant@flynn:~/go/src/github.com/flynn/flynn$ ./script/build-flynn --filter controller --force
```

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -49,6 +49,7 @@ options:
   -t, --tuf-db=<path>       path to TUF database [default: build/tuf.db]
   -v, --verbose             be verbose
   --filter=<imageIDs>       imageIDs to build (defaults to all)
+  --force                   ignore layer cache
 
 Build Flynn images using builder/manifest.json.
 `[1:],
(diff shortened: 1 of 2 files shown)


lmars:
@jvatic how much of a speed up does this give?

I'm curious because one of the goals of the builder is to only build what has changed, and if it is slow at doing that then I'm wondering if there is some underlying optimisation we can make to speed that up and avoid adding these flags.

jvatic:
@lmars it's significant. I don't have specific numbers off the top of my head, but it's somewhere around ten minutes. It appears to be the layer ID generation that's causing some of the slowness.

jvatic:
As per slack conversation, the reason it's so slow is the fs driver in the VM is broken.

PR #4579, opened 2020-01-29
router: Support disabling keep alives

lmars:
This allows keep alives between the router and backends to be disabled with the `--disable-keep-alives` flag, or re-enabled with the `--enable-keep-alives` flag:

```bash
$ flynn route add http -s app-web --disable-keep-alives $domain

$ flynn route update $id --disable-keep-alives

$ flynn route update $id --enable-keep-alives
```

### DIFF
--- a/cli/route.go
+++ b/cli/route.go
@@ -36,8 +36,8 @@ Options:
 	--no-leader                disable leader-only routing mode (update only)
 	-p, --port=<port>          port to accept traffic on
 	--no-drain-backends        don't wait for in-flight requests to complete before stopping backends
-	--disable-keep-alives      disable keep alives between the router and backends for the given route
-	--enable-keep-alives       enable keep alives between the router and backends for the given route
+	--disable-keep-alives      disable keep-alives between the router and backends for the given route
+	--enable-keep-alives       enable keep-alives between the router and backends for the given route
 
 Commands:
 	With no arguments, shows a list of routes.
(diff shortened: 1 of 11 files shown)


cli/route.go
```diff
@@ -36,6 +36,8 @@ Options:
 	--no-leader                disable leader-only routing mode (update only)
 	-p, --port=<port>          port to accept traffic on
 	--no-drain-backends        don't wait for in-flight requests to complete before stopping backends
+	--disable-keep-alives      disable keep alives between the router and backends for the given route
+	--enable-keep-alives       enable keep alives between the router and backends for the given route
```
titanous:
```suggestion
	--enable-keep-alives       enable keep-alives between the router and backends for the given route (default for new routes)
```

router/http.go
```diff
@@ -249,7 +249,7 @@ func (h *httpSyncHandler) Set(data *router.Route) error {
 	} else {
 		bf = backendFunc(r.Service, service.sc.Instances)
 	}
-	r.rp = proxy.NewReverseProxy(bf, h.l.cookieKey, r.Sticky, service, logger.New("service", r.Service))
+	r.rp = proxy.NewReverseProxy(bf, h.l.cookieKey, r.Sticky, r.DisableKeepAlives, service, logger.New("service", r.Service))
```
titanous: This should probably be an options struct given the number of arguments building up.

cli/route.go
lmars: See 0a8622d3 and 53d2a4ec

router/http.go
lmars: See 4299f290

PR #4584, opened 2020-02-24
Add deployment events stream

jvatic:
This allows streaming scale requests along with deployments and is the first step towards allowing a scale request to be included in a create deployment request. I'm opening this now to get feedback before I write any tests.


lmars:
@jvatic 

> is the first step towards allowing a scale request to be included in a create deployment request

Looking at the change here, it feels like it would be best to take the next step now, and flip it so that creating a new deployment can also include a scale request (rather than a scale request creating a dummy deployment), because it seems we'll just end up reverting this change when we do that anyway.

WDYT?

jvatic:
@lmars It does feel a bit weird as is. I am wanting to merge the two constructs so that

1. A release may be deployed and scaled in a single request,
2. It is easy to filter out the previous release being scaled down on the client (i.e. in the dashboard), and
3. Scale requests and deployments may be retrieved in a single list

In order to solve #2, the scale request for the old release of a deployment would also need to be associated with the deployment (and be returned as part of an `ExpandedDeployment`): https://github.com/flynn/flynn/blob/beaecc379bae37b30debd0c373919f9b927123dc/controller/worker/deployment/job.go#L99-L112

Perhaps there needs to be a third API construct to house a deployment and/or a scale.

WDYT?

lmars:
So a deployment currently gets the desired process count for the new release from the current formation [here](https://github.com/flynn/flynn/blob/master/controller/data/deployment.go#L128), so I think it should be as simple as just allowing `Deployment.Processes` to be specified when creating the deployment? The deployment will then create the appropriate scale requests to realise those processes in the worker job.

jvatic:
As per slack discussion, I'll change this to 

1. add optional `deployment_id` to scale requests (and set for deployment scales)
1. add additional deployment events for more visibility into the deployment (so all events will be: `running`, `job_started`, `job_stopped`, `success`, `failure` plus two `scale_request` events with `deployment_id` set)
1. add a `StreamDeploymentEvents` gRPC endpoint that returns all `Event`s associated with the specified deployment (i.e. deployment (expanded), scale request, and job events)
1. change `CreateScaleRequest` and `CreateDeploymentRequest` to each have a `config` field:

```protobuf
message ScaleConfig {
  map<string, int32> processes = 1;
  map<string, DeploymentProcessTags> tags = 2;
}

message CreateScaleRequest {
  // parent = "apps/APP_ID/releases/RELEASE_ID"
  string parent = 1;

  ScaleConfig config = 2;
}

message DeploymentConfig {
  int32 timeout = 1;
  int32 batch_size = 2;
  ScaleConfig scale_config = 3;
}

message CreateDeploymentRequest {
  // parent = "apps/APP_ID/releases/RELEASE_ID"
  string parent = 1;

  DeploymentConfig config = 2;
}
```

gitreceive/server.go
```diff
@@ -317,9 +317,56 @@ git-archive-all() {
...
 while read oldrev newrev refname; do
 	if [[ $refname = "refs/heads/master" ]]; then
-		git-archive-all $newrev | /bin/flynn-receiver "$RECEIVE_APP" "$newrev" --meta git=true --meta "git.commit=$newrev"| sed -u "s/^/"$'\e[1G\e[K'"/"
+		git-archive-all $newrev | /bin/flynn-receiver "$RECEIVE_APP" "$newrev" \
+			--meta git=true \
+			--meta "git.commit=$newrev" \
+			--meta "git.subject=$(git-subject $newrev)" \
+			--meta "git.body=$(git-body $newrev)" \
+			--meta "git.signature-status=$(git-signature-status $newrev)" \
+			--meta "git.signature-fingerprint=$(git-signature-fingerprint $newrev)" \
```
titanous, line 364: Remove the signature-related fields, we won't be using them.

titanous:

PR #4586, opened 2020-03-09
Issue TLS certificates using the ACME protocol

### LINKED ISSUE #1995: Automatic HTTPS through Let's Encrypt
(This is an enhancement)

It would be really cool if it were possible to automatically enable https through the new Let's Encrypt service, since it's now trusted by all major browsers. It's also part of the Linux Foundation so it's rooted in the Open Source community.

lmars:
This PR adds the concept of a Managed Certificate which controls the issuance of TLS certificates using the ACME protocol.

The router has a new ACME service that watches for managed certificates being added to the controller, then places orders, responds to challenges and fetches issued certificates for those managed certificates.

Route config now has a `config.managed_certificates` function that configures a route with a managed certificate, which triggers the ACME service to automatically provision a certificate for the route.

Example:

```starlark
load("flynn.routeconfig.v1", "config")

def main(ctx):
  return config.app_routes({
    "app": [
      config.http_route(
        domain = "app.flynn.lmars.net",
        target = config.service("app-web"),
        certificate = config.managed_certificate("app.flynn.lmars.net"),
      ),
    ],
  })
```

```bash
$ route config apply routes.star 

!!! You are about to make the following 2 route changes: !!!

+--------+----------------------------+------------------+-------------------------------+
| ACTION |           ROUTE            |      BEFORE      |             AFTER             |
+--------+----------------------------+------------------+-------------------------------+
| DELETE | http:app.1.localflynn.com/ | service: app-web | -                             |
+--------+----------------------------+------------------+-------------------------------+
| CREATE | http:app.flynn.lmars.net/  | -                | service: app-web              |
|        |                            |                  | auto-tls: app.flynn.lmars.net |
+--------+----------------------------+------------------+-------------------------------+

Are you sure you want to proceed? (yes/no): yes
The confirmed route changes were made successfully.

$ flynn -a app route
ROUTE                      SERVICE  ID                                         STICKY  LEADER  PATH
https:app.flynn.lmars.net  app-web  http/e65ef5ac-dc86-4ee5-93ae-4cf1d6c3b6ab  false   false   /

$ openssl s_client -connect app.flynn.lmars.net:443 -servername app.flynn.lmars.net </dev/null 2>/dev/null | grep -A4 -F 'Certificate chain' 
Certificate chain
 0 s:CN = app.flynn.lmars.net
   i:CN = Fake LE Intermediate X1
 1 s:CN = Fake LE Intermediate X1
   i:CN = Fake LE Root X1
```

This is a WIP to get feedback on the approach, I plan to tackle the following in upcoming changes:

- [ ] support re-issuance (both for soon-to-expire certificates and manually using the CLI)
- [ ] support DNS and ALPN challenges
- [ ] support OCSP
- [ ] ensure  the ACME service handler follows the spec (i.e. checks all relevant status fields on orders / authorizations / challenges)
- [ ] run the ACME service in HA mode (i.e. like the scheduler with only the leader being active)
- [ ] switch to production Let's Encrypt by default

Closes #1995.


router/acme/acme.go
```diff
@@ -0,0 +1,480 @@
...
+func (s *Service) satisfyHTTP01Challenge(cert *ct.ManagedCertificate, challenge *acme.Challenge) error {
+	log := s.log.New("domain", cert.Domain, "challenge.token", challenge.Token)
+
+	log.Info("configuring http-01 challenge response")
+	return s.responder.RespondHTTP01(cert, challenge, func() error {
+		// make multiple attempts to complete the challenge
+		attempts := attempt.Strategy{
+			Total: 60 * time.Second,
+			Delay: 1 * time.Second,
+		}
```
titanous: We should only retry like this for 5xx.

Validations should retry every 13 minutes, as there is a limit of 5 failed validations per domain per hour and we need to handle the case where DNS is set up later, potentially several hours later.

For both issuance, renewals, and validation retries it may make sense to have a "task" data structure each with a scheduled run timestamp. There are two ways to model this, one involves locking in the database (we could use que-go for this) so we can have multiple `acme` workers with no leader. Alternatively we can do a scheduler-style leader lock, but we have to be even more careful about tasks being idempotent because we can't get a guarantee that there is only one leader due to the way consensus works.

router/acme/responder.go
```diff
@@ -0,0 +1,164 @@
...
+// completing the challenge and then removing the route
+func (r *Responder) RespondHTTP01(cert *ct.ManagedCertificate, challenge *acme.Challenge, complete func() error) error {
+	route, err := r.createChallengeRoute(cert, challenge)
+	if err != nil {
+		return err
+	}
+	if err := complete(); err != nil {
+		return err
+	}
+	return r.deleteChallengeRoute(route, challenge)
```
titanous: Should this be deferred so it gets cleaned up if `complete()` errors? We may want some kind of GC function for these.

titanous:
Prioritization suggestion:

This PR:
- support re-issuance (both for soon-to-expire certificates and manually using the CLI)
- ensure the ACME service handler follows the spec (i.e. checks all relevant status fields on orders / authorizations / challenges)
- ensure internal data structures/tests have support for multiple domains so we don't have to migrate later
- run the ACME service in HA mode

Separate PRs:
- CLI to provision managed certificates with multiple domains/options, support for certificate options like RSA keys and must staple
- support DNS (w/wildcard certificate support) and ALPN challenges
- use ACME with wildcard cert instead of self-signed cert for built-in apps
- OCSP stapling

router/acme/acme.go
titanous: Worth noting that OCSP stapling will be a task on a schedule too, so I think we do need an abstraction to handle it.

router/acme/responder.go
titanous: We could have a GC function as a recurring task.

rugwirobaker:
Is this PR ready for merging? This the only thing I'm waiting for to try flynn out. You probably have other commitments but you folks have done an amazing job so far. Hope this project doesn't die.

titanous:

PR #4603, opened 2020-07-09
controller: OAuth

jvatic:
@titanous this is the controller oauth stuff pulled out of the dashboard v2 branch.


controller/controller.go
```diff
@@ -64,6 +70,31 @@ func main() {
 		name.SetSeed(s)
 	}
 
+	var tokenKey *ecdsa.PublicKey
+	if tk := os.Getenv("ACCESS_TOKEN_KEY"); tk != "" {
```
jvatic, line 74: When `ACCESS_TOKEN_KEY` is not set, `tokenKey` remains `nil` which causes a panic if `authorizer.AuthorizeToken` is called. What should the behaviour be in this case? Should the controller crash at boot when this env is not set, or should a key be generated?

PR #4604, opened 2020-07-09
Remove dashboard

jvatic:
This removes the dashboard and all references to it.


kipparker:
So no more web dashboard in future versions of Flynn?

